PR #58, opened 2016-01-31
Add constant propagation pass

ducky64:
It compiles, but I don't actually know how to test it.
I also needed to add a TypedExpression trait since Scala can detect (and will error on) pure abstract 
classes but Stanza doesn't.

Looks like my IDE was also set to remove trailing whitespaces, which makes things conformant with Chisel3's style checker.


azidar:
It was easier to just rewrite the pass myself, but thanks for looking up the the right shifts etc. for BigInteger!

PR #75, opened 2016-02-25
Fixed printf bugs in scala and stanza versions

azidar:
Required special casing prints in SplitExp, and emitting expressions instead of their toString counterparts


jackkoenig:
So the current check for scala FIRRTL is terrible, it just compares rocket.fir compiled to Verilog to the a "golden" Verilog result. Tomorrow I will try to get either (or perhaps both) rocket-chip running through Chisel3 + FIRRTL or at least run the Chisel3 tests that use FIRRTL so that we can dump this abomination of a regression test.

PR #78, opened 2016-02-26
Travis add chisel3

jackkoenig:
Added Verilator installation to the Travis cache
Added Chisel 3 checkout and "sbt test"
Removed brittle string comparison of compiled rocket.fir and rocket-golden.v
Actual FIRRTL tests will be added in another pull request.


jackkoenig:
I'm not 100% sure how the caching works, but I think each branch has it's own cache, but all Pull Requests share a cache. This means that Verilator will be rebuilt every time someone pushes to a new branch (which is kind of annoying).

Also, the chisel tests take a little while to run because I don't think Travis gives us a lot of parallel stuff, so perhaps we should figure out how to tag tests in Scala and only run the subset of Chisel3 tests that actually touch FIRRTL

PR #79, opened 2016-03-01
Move mapper functions to implicit methods on IR vertices.

jackkoenig:
This moves all mappers to new implicit methods so they can be called with .map(func)
It also moves mappers from Util.scala to new Mappers.scala (and new object)
It does NOT implement all possible/desired mappers, just the ones currently in use
It does NOT migrate any recursive mappers


azidar:
Looks good to me, assuming everything passes

PR #80, opened 2016-03-02
Remove implicits from serialize

jackkoenig:
implicits should be avoided, rather, a single function on AST can call serialize functions for all FIRRTL graph nodes. 

These serialize functions could be migrated to the actual traits and case classes in IR.scala. I originally didn't put them there in order to keep the IR "self documenting", but perhaps that's just better handled by the FIRRTL Spec and ScalaDocs.


azidar:
Looks good to me

jackkoenig:
Merged in 0aa246385d1d2eabafce0e659d6438a38c3b6519

PR #81, opened 2016-03-03
Add some integration tests: successful compilation and execution

jackkoenig:
Add a few simple tests so we aren't 100% relying on Chisel3 testing.
This commit borrows some code from Chisel3 to make it easier to run tests, ideally this code won't be replicated, but since we're keeping chisel3 and firrtl from depending on each other, I think it's okay for now. Suggestions welcome.


aswaterman:
looks ok to me, provided we eventually avoid duplication of this code -- perhaps by removing it from chisel3, since the latter inherently depends on firrtl.

PR #84, opened 2016-03-07
Fix inferred width on SInts

azidar:
SInts were given a width one larger than desired. This fixes it.

### DIFF
--- a/src/main/scala/firrtl/Visitor.scala
+++ b/src/main/scala/firrtl/Visitor.scala
@@ -245,7 +245,7 @@ class Visitor(val fullFilename: String) extends FIRRTLBaseVisitor[AST]
               (IntWidth(string2BigInt(ctx.IntLit(0).getText)), string2BigInt(ctx.IntLit(1).getText))
             else {
                val bigint = string2BigInt(ctx.IntLit(0).getText)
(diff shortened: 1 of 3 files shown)


src/main/scala/firrtl/Visitor.scala
```diff
@@ -245,7 +245,7 @@ class Visitor(val fullFilename: String) extends FIRRTLBaseVisitor[AST]
               (IntWidth(string2BigInt(ctx.IntLit(0).getText)), string2BigInt(ctx.IntLit(1).getText))
             else {
                val bigint = string2BigInt(ctx.IntLit(0).getText)
-               (IntWidth(BigInt(bigint.bitLength + 1)),bigint)
+               (IntWidth(BigInt(scala.math.max(bigint.bitLength,1))),bigint)
```
aswaterman: I don't think this is right.  bitLength gives the number of bits needed excluding the sign bit, i.e., BigInt(-1).bitLength == 0, not 1.  The old version looks right to me.

The question is really about what SInt("hF") means.  Do you intend for it to mean +15 or -1?  If you want it to be -1, that suggests to me that you need different width inference rules for string-based literals than for the other kinds.

jackkoenig:
I agree with Andrew, closed for now.

Issue #69, opened 2016-02-24
Shift Right not handling shifts larger than width of argument


jackkoenig:
This problem manifests in both Scala and Stanza FIRRTL

Consider the FIRRTL:
```
circuit Bug :                                                                                                                                                                                                                                 
  module Bug :
    input in : UInt<1>
    output out : UInt<1>
    
    out <= shr(in, 1)
```
This results in the following invalid Verilog:
```Verilog
module Bug(
   input   in,                                                                                                                                                                                                                                
   output  out 
);  
   assign out = in[0:1];                                                                                               
endmodule
```
Shifting by more than the width of the operand gives similar results:
```
circuit Bug : 
  module Bug : 
    input in : UInt<5>
    output out : UInt<5>
    
    out <= shr(in, 6)     
```
```Verilog
module Bug(
   input  [4:0] in,
   output [4:0] out                                                                                                                                                                                                                           
);  
   assign out = in[4:6];                                                                                               
endmodule
```

First of all, the spec requires that the shift amount be strictly less than the width of the operand, so neither implementation is correctly implementing the spec here.

Secondly, and more importantly, since Chisel relies on FIRRTL for width-inference, it cannot prevent itself from emitting invalid FIRRTL here. Moreover, dynamic shift right does not have any constraint like this (as it obviously cannot) and just seems to punt to Verilog as in this example:
```
circuit Bug : 
  module Bug : 
    input in : UInt<5>
    output out : UInt<5>
    
    out <= dshr(in, UInt(6))                         
```
```Verilog
module Bug(                                                                                                                                                                                                                                   
   input  [4:0] in,
   output [4:0] out
);
   assign out = in >> 3'h6;
endmodule
```
Thirdly, allowing for shift amounts greater than or equal to the width of the operand makes it easier to write generators in that you don't have add special cases in your code.

I think FIRRTL should define the semantics of right shifting a number by an amount equal to or greater than the number's width to be (both are width = 1):

* 0 when unsigned
* sign bit when signed

These semantics strike me as both intuitive and necessary since they can be consistent between static and dynamic right shift.

aswaterman:
I concur with Jack's proposal for defining right shifts by amounts N >= W.  These remain consistent with the usual mathematical definitions of right shifts by N.  For unsigned, that's division by 2^N, rounding towards 0 (hence a result of 0), and for signed, that's division by 2^N, rounding towards -infinity (hence the sign bit).

palmer-dabbelt:
[...]

PR #85, opened 2016-03-07
Right shifts with const size > width size is now allowed. If a UInt, …

azidar:
…it returns UInt(0). If a SInt, returns sign bit

### DIFF
--- a/src/main/scala/firrtl/passes/Passes.scala
+++ b/src/main/scala/firrtl/passes/Passes.scala
@@ -86,9 +86,12 @@ object PassUtils extends LazyLogging {
        val p = passes.head
        val name = p.name
        logger.debug(s"Starting ${name}")
+       println(s"Starting ${name}")
        val x = p.run(c)
        logger.debug(x.serialize)
        logger.debug(s"Finished ${name}")
+       println(x.serialize)
+       println(s"Finished ${name}")
        executePasses(x, passes.tail)
     }
   }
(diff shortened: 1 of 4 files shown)


azidar:
Fix for #69

src/main/scala/firrtl/passes/Passes.scala
```diff
@@ -86,9 +86,12 @@ object PassUtils extends LazyLogging {
        val p = passes.head
        val name = p.name
        logger.debug(s"Starting ${name}")
+       println(s"Starting ${name}")
        val x = p.run(c)
        logger.debug(x.serialize)
        logger.debug(s"Finished ${name}")
+       println(x.serialize)
```
jackkoenig: Please remove these printlns, I'm assuming they were just for debugging anyway.

They make the Travis log impossible to read (and since it's failing, we need to figure out why). Also, those logger.debugs will print what you need, you can locally change how much is printed by the logger in src/main/resource/logback.xml. Simply change root level to be "debug" instead of "warn"

jackkoenig:
I think instead of this being handled in the Verilog Emitter, it could perhaps instead be handled in a Pass? I'm not sure which one, but since static right shifts are just bit selections, I think we can just replace them with a bits_select. Any idea where this should be done @azidar ?

jackkoenig:
I think this really should be handled in an earlier pass because it could be relevant to ConstProp too.

PR #88, opened 2016-03-10
Fix StringLit to ignore escaped double quotes

davidbiancolin:
String literal tokens were erroneously being terminated at double quotes that have been escaped. This change prevents that from happening.


aswaterman:
lgtm

Issue #65, opened 2016-02-23
No Support for Zero-width wires


jackkoenig:
See bug fix in b487ab24c0e298e7781f74811b8bfcd24384d1b5
This was due to rocket regression having memory of depth = 1, thus address width = 0. This makes sense because the address port is unnecessary in this case.

If I recall correctly, support for zero-width wires was mentioned as future work? But if they are supported in Chisel, I think it should be now work.

colinschmidt:
I would also like to see this feature implemented sooner rather than later!

azidar:
Yeah, the tricky thing is handling every corner case correctly. Off the top of my head - 

    wire w0 : UInt<0>
    wire w1 : UInt<1>
    wire w2 : UInt<1>

    add(w1,w0) // is the result width=1 or width=2?
    mul(w1,w0) // is the result width=1 or width=0?
    mux(w0,w1,w2) // => w2? or is it w1?
    mux(w1,w0,w2) // no clue

I think in general, a good rule of thumb is for arithmetic operations, its value should be treated as zero. I think the trickier part is determining the width inference rules.

ccelio:
Didn't Jim already handle zero-width wires in Chisel2? I don't think a complete reinvent of the wheel is needed on how different operations are handled.

A zero-width wire basically doesn't exist. So most things should just be an identity property, I would imagine. It's also fair to throw errors on some operations and allow others.

aswaterman:
The proposal isn't to change the semantics, but to get it working in
Chisel3, where the feature is currently missing.  This should require
nothing of Chisel3, but needs support in FIRRTL.

ucbjrl:
On Feb 23, 2016, at 11:39 AM, Adam Izraelevitz <notifications@github.com> wrote:

> Yeah, the tricky thing is handling every corner case correctly. Off the top of my head -
> 
> wire w0 : UInt<0>
> wire w1 : UInt<1>
> wire w2 : UInt<1>
> 
> add(w1,w0) // is the result width=1 or width=2?
> mul(w1,w0) // is the result width=1 or width=0?
> mux(w0,w1,w2) // => w2? or is it w1?
> mux(w1,w0,w2) // no clue
> 
> I think in general, a good rule of thumb is for arithmetic operations, its value should be treated as zero. I think the trickier part is determining the width inference rules.
> 

[...]

PR #90, opened 2016-03-10
Add support for right shift by amount larger than argument width

jackkoenig:
Solution to #69
Replacement for #85 

Right shifting UInt by amount greater than or equal to width gives value 0 with width of 1, once zero-width wires are supported #65, it should probably return zero-width wire instead.

I decided to implement this change as a new pass because I couldn't see where else it would fit in. I'm not sure I like the pass name "Legalization", but I couldn't think of anything better and I can see there possibly being other similar transformations that could be added to the pass. I left the actual process of turning a standard right shift (ie. amount < width of argument) in the Verilog Emitter because this really is a Verilog backend specific change. In a C++ emitter, bit extraction would map to a shift and mask. 

The test is added to the "integration" tests list. Ultimately the tests should be refactored, but this is a nice Chisel-generated self-testing circuit to check that this feature is supported.

The changes to the Width trait have some overlap with Width in WorkingIR, and I think refactoring of the InferWidths pass will clean up the overlap.

### DIFF
--- a/src/main/scala/firrtl/passes/Passes.scala
+++ b/src/main/scala/firrtl/passes/Passes.scala
@@ -1128,24 +1128,63 @@ object ExpandWhens extends Pass {
    }
 }
 
+// Replace shr by amount >= arg width with 0 for UInts and MSB for SInts
+// TODO replace UInt with zero-width wire instead
+object Legalize extends Pass {
+  def name = "Legalize"
+  def legalizeShiftRight (e: DoPrim): Expression = e.op match {
+    case SHIFT_RIGHT_OP => {
+      val amount = e.consts(0).toInt
+      val width = long_BANG(tpe(e.args(0)))
+      lazy val msb = width - 1
+      if (amount >= width) {
+        e.tpe match {
+          case t: UIntType => UIntValue(0, IntWidth(1))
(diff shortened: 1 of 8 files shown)


jackkoenig:
Before merging, need to check that this fixes https://github.com/ucb-bar/uncore/commit/417565332484b2094e1518449b3cb2d80eafb0bf

src/main/scala/firrtl/passes/Passes.scala
```diff
@@ -1128,24 +1128,66 @@ object ExpandWhens extends Pass {
...
+                  e.args(0) match {
+                     case x: UIntValue => {
+                        val v = x.value >> amount
+                        val w = (x.width - IntWidth(amount)) max IntWidth(1)
+                        UIntValue(v, w)
                      }
-                     case (x:SIntValue) => {
-                        val b = x.value >> e.consts(0).toInt
-                        SIntValue(b,tpe(e).as[SIntType].get.width)
+                     case x: SIntValue => { // take sign bit if shift amount is larger than arg width
```
jackkoenig: It turns out that >> in Scala does the right thing here so UInt and SInt can be treated the same and no special casing is necessary

PR #92, opened 2016-03-18
Add guard to emission of simulation constructs

jackkoenig:
Fix to https://github.com/ucb-bar/chisel3/issues/121 and replacement to https://github.com/ucb-bar/chisel3/pull/122.

This adds a new preprocessor macro to guard the simulation constructs stop and printf. The user can now define PRINTF_COND to prevent stop and printf from firing when not desired.

Main use case: we generally don't want these firing before the circuit comes out of reset, and pipelined reset signals can lead to this happening.

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -55,7 +55,7 @@ object FIRRTLEmitter extends Emitter {
 case class VIndent()
 case class VRandom()
 class VerilogEmitter extends Emitter {
-   val tab = "   "
+   val tab = "  "
    val ran = VRandom()
    var w:Option[Writer] = None
    var mname = ""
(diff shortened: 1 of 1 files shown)


src/main/scala/firrtl/Emitter.scala
```diff
@@ -344,10 +344,16 @@ class VerilogEmitter extends Emitter {
       def simulate (clk:Expression,en:Expression,s:Seq[Any]) = {
          if (!at_clock.contains(clk)) at_clock(clk) = ArrayBuffer[Seq[Any]]()
```
aswaterman: could also do ArrayBuffer[String], then use ++= for Seqs vs. += for Strings

aswaterman: ah, I see, \n is folded into this

PR #109, opened 2016-04-07
Make ConstProp pass more concise

aswaterman:
I was going to augment it, but thought it best to clean it up first.

### DIFF
--- a/src/main/scala/firrtl/passes/Passes.scala
+++ b/src/main/scala/firrtl/passes/Passes.scala
@@ -1164,68 +1164,53 @@ object Legalize extends Pass {
 }
 
 object ConstProp extends Pass {
-   def name = "Constant Propogation"
-   var mname = ""
+  def name = "Constant Propogation"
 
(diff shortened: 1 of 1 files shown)


src/main/scala/firrtl/passes/Passes.scala
```diff
@@ -1164,68 +1164,53 @@ object Legalize extends Pass {
...
+      }
+      case _ => e
+    }
+    case _ => e
+  }
+
+  private def constPropExpression(e: Expression): Expression = {
+    e map constPropExpression match {
+      case e: DoPrim => constPropPrim(e)
+      case _ => e
```
jackkoenig: should be `case e => e` so that we still return the result of the map constPropExpression

aswaterman: good catch: great reason not to alias names between scopes.

jackkoenig:
Merged ecc5c3d0934b11a9b727390853f84996c13dbb42

Issue #97, opened 2016-03-29
Makefile Issues


jackkoenig:
`make build-scala` doesn't rerun if you change a source file. It needs to depend on .scala files.

At this point, the scala implementation should be default, so `make build` should probably be equivalent to `make build-scala` (the current `make build` could become `make build-stanza` and a new command could do the actual building of stanza)

jackkoenig:
Fixed 084e7ca9f9ecb76b91e143931d08195d45672ab7

PR #110, opened 2016-04-07
Update Makefile to use sbt to determine sources; make build-scala the default build - #97

ucbjrl:
also remove redundant make targets and rules.


Makefile
```diff
@@ -4,12 +4,15 @@ regress_dir ?= $(root_dir)/regress
...
 
+sbt ?= sbt
 stanza ?= $(install_dir)/stanza
 stanza_bin ?= $(install_dir)/firrtl-stanza
 scala_jar ?= $(install_dir)/firrtl.jar
-scala_src=$(shell ls src/main/scala/firrtl/*.scala)
+# The following is slow but accurate.
+# We use 'unmanaged' so we don't include source files generated by the
+#  build itself (otherwise, we'd need to include the rules to build them).
+scala_src := $(shell $(sbt) "show compile:unmanaged-sources compile:unmanaged-resources" | ./sourcesFromSbt.sh )
```
aswaterman: I'm glad we're fixing this.  Can we just do

    find . -type f -name "*.scala".

to avoid invoking sbt twice?

ucbjrl: If you've built once,

    find . -type f -name "*.scala"

will pick up the managed (generated sources) and it won't pick up resources.

    find src -type f \( -name "*.scala" -o -path "*/resources/*" \)

should pick up almost everything.

I'd be willing to sacrifice accuracy for speed in this case. You can always force a complete build by doing a

    make clean build

jackkoenig:
We should also change `make test-scala` to be `sbt test` since we're using sbt testing now

jackkoenig:
It also seems like every time you run `make build-scala` it reruns `sbt assembly` even if no Scala files have changed. Maybe that's okay, but I think the dependency on the scala_src essentially isn't working.

ucbjrl:
`sbt assembly` is pretty clever. If you edit one of the scala files, but the change doesn't impact the output jar, `sbt assembly` won't update it. The source file will be newer than the jar so make will think it needs to rebuild the jar. Try the following:

    % make clean build
    % make build

The second (and subsequent) build(s) should do nothing. If that's not the case, something is broken.

jackkoenig:
Fair point, I guess my main complaint was sbt startup time since it's non-trivial

jackkoenig:
merged in 168392889d42506fae5a1aa637ebe1e61d799e62

PR #111, opened 2016-04-07
Add basic constant propagation for logical operators

aswaterman:
This is deliberately incomplete because I wanted to get feedback before
plowing ahead.  These passes handle constant propagation for bitwise
and equality operators on UInt only, usually only when the widths match.


jackkoenig:
I really like the use of OOP, I think it makes the code super clear.

This and other things is making me think UIntValue and SIntValue should be refactored to extend some common Literal parent type. This would make supporting SInts in this code easier.

jackkoenig:
Looks good to me. I think we should generally put new and refactored passes into new files so I'm glad you did that.

src/main/scala/firrtl/passes/ConstProp.scala
```diff
@@ -0,0 +1,198 @@
...
+      }
+      if (old ne propagated)
+        nPropagated += 1
+      propagated
+    }
+
+    def constPropStmt(s: Stmt): Stmt = {
+      s match {
+        case x: DefNode => nodeMap(x.name) = x.value
+        case _ =>
```
azidar: This is incorrect, constPropStmt should be recursive:
`case _ => s map constPropStmt`

azidar:
Though I myself am guilty of this, we should really start to build up the number of Scala tests, can we write some for this pass that shows the constant propagation working?

aswaterman:
What's the right testing strategy?  Functional testing can show the pass doesn't screw things up.  Is there a good way to show propagation happening that isn't unduly brittle?

azidar:
I think the only way is string comparison (or a slightly fancier string comparison) of the FIRRTL output. If each test is written with the minimal amount of FIRRTL to produce the specific result, it should be pretty robust to other changes in the compiler (other than the parser/emitter, which hopefully shouldn't change that much any more).

aswaterman:
Yeah, makes sense.  Are there any examples of tests written in that manner?

jackkoenig:
The original Stanza tests are written in this manner, for example see: https://github.com/ucb-bar/firrtl/blob/master/test/passes/const-prop/bits.fir

It is definitely much easier to write tests in this way when you're looking for specific changes to the FIRRTL. The one place where it gets kind of hairy is when wanting to look at information that isn't printed by default like Types of Expressions. Adam solved this in the Stanza implementation by adding special debug printing that would print this information so you could check that way. We could essentially create a similar thing in Scala. 

I've thought about how to do this in a more "Scala-y" way but it's obviously really hard. I've floated around the idea of creating some "equivalence function generator" so that you could essentially create a list of things you expect to find in the Firrtl graph and search for them. Unfortunately, the internals of Firrtl are more likely to change than the text form, so this is probably more difficult to write and more brittle than Adam's proposed solution. 

Generally, I think executable tests that can "test themselves" are the best way to write tests where we can, but obviously for things like constant propagation and dead code elimination that approach won't really work that well.

azidar:
One option is to write the firrtl test in a Scala string, then call the parser, then execute the passes you want, and then emit firrtl again, and do a string compare. This keeps it all in Scala and thus easily runnable by scalatest. If you want I can come up with a short example?

jackkoenig:
I think a short example would be great. It seems like doing this would involve kind of writing our own version of FileCheck or perhaps there is a library out there we could use.

aswaterman:
Yeah, I'd be happy to write tests once there's an example in the repo I can
go off of.

azidar:
See #116

PR #112, opened 2016-04-07
Add Expression.tpe accessor

aswaterman:
Almost all of the code was already there.  This is cleaner (and faster)
than calling tpe(Expression).

### DIFF
--- a/src/main/scala/firrtl/IR.scala
+++ b/src/main/scala/firrtl/IR.scala
@@ -89,15 +89,21 @@ case object BITS_SELECT_OP extends PrimOp
 case object HEAD_OP extends PrimOp
 case object TAIL_OP extends PrimOp
 
-trait Expression extends AST
+trait Expression extends AST {
+  def tpe: Type
+}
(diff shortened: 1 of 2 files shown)


jackkoenig:
Agreed, I think a lot of things should be moved from Utils to be methods on the actual IR classes.

jackkoenig:
Should removing `def tpe` from Utils be part of this PR? And thus changing all the code to use the new method?

azidar:
I disagree - don't merge this yet? I'll write up my thoughts in a bit.

aswaterman:
sure -- I just wanted to force the discussion of the matter

jackkoenig:
@azidar Any chance you could expand a bit? I think the public API will need to be cleaned up soon so we should talk about how exactly to do it.

azidar:
Whoops sorry. I chatted with @aswaterman offline and decided adding tpe is an ok thing for Expression. However, we should be very very careful adding methods to any of our AST nodes (the bar should be extremely high).

PR #114, opened 2016-04-08
Rewrite lower types

jackkoenig:
DO NOT MERGE.

This should fail Travis, there is a slight bug that I need to fix.
I will rebase this deleting most of the commits before it should actually be merged.

I'm PRing this to request code and algorithm review.


jackkoenig:
This rewrite of LowerTypes fixes a few minor bugs in the pass, but most importantly it adds a technique for resolving name collisions during lowering of types. I will first describe the algorithm then talk about why I chose this particular approach and compare it to others.

**The Algorithm:**
* The general idea remains that bundles lower by adding _ while vectors lower by adding _#
* The pass first lowers the ports of every single module, then lowers the body of each module. The ports must be done first for detection of name collisions with the ports of instances within a module.
* Name collisions are detected and resolved by: 
    * First creating a namespace of everything at a given level of scope
    * For each aggregate:
        * Enumerate all lowered names of aggregate elements
        * If there are any collisions with the elements, add an _ to the end of the name of the aggregate
        * Iterate on the above until no collisions are detected
        * Add all lowered elements to the namespace
* The above is probably best illustrated by an example:
    * Given pseudo-Firrtl:  `a : { b, c : { d, e }[2], c_0_d, c__1_e }` and `a_b`
    * We would first recursively lower `a`, starting with the namespace `b, c, c_0_d, c__1_e`
        * Trying to lower `c` would give `c_0_d` which collides, thus we would rename c to `c_`
        * We try to lower `c_` giving `c__1_e` which collides, thus we rename c to `c__`
        * We now lower `c__` mapping `a` to     
`a : { b, c___0_d, c___0_e, c___1_d, c___1_e, c_0_d, c__1_e}`
    * We now consider the outter namespace `a, a_b`
    * We try to lower `a` but that gives `a_b` which collides, thus we rename `a` to `a_`
    * We then successfully lower `a_` giving the final lowered names:    
`a__b, a__c___0_d, a__c___0_e, a__c___1_d, a__c___1_e, a__c_0_d, a__c__1_e, a_b`
    * This example is pretty brutal but it shows the overall idea of dealing with name collisions in such a way that it keeps fields of bundles together in the naming

src/main/scala/firrtl/Mappers.scala
```diff
@@ -79,7 +79,10 @@ object Mappers {
           case s: DefWire => DefWire(s.info,f(s.name),s.tpe)
           case s: DefPoison => DefPoison(s.info,f(s.name),s.tpe)
           case s: DefRegister => DefRegister(s.info,f(s.name), s.tpe, s.clock, s.reset, s.init)
-          case s: DefMemory => DefMemory(s.info,f(s.name), s.data_type, s.depth, s.write_latency, s.read_latency, s.readers, s.writers, s.readwriters)
+          case s: DefMemory =>
+            DefMemory(s.info,f(s.name), s.data_type, s.depth, s.write_latency,
```
azidar: This is tricky - I'm not sure you should map onto the names of the ports, because generally, mapping String=>String on a statement should only change the name of the statement, but not to any types in the design (the type of a reference to this memory would change). I suppose another way to do this would be to do something in forType's map(Stmt):Stmt, where the data type of the DefMemory is created from its readers/writers/readwriters and data_type, passed into the mapper function, and then map any changes back.

jackkoenig:
**Discussion/Comparison**

1. The original approach as defined in the Firrtl Spec (lower with $, disallow prefixes in naming)
    1. Pros
        1. Names are trivial to lower
        1. The names of other things can never change the lowered name of something
        1. Elegant in it's simplicity
    1. Cons
        1. $ is allowed in Verilog identifiers, but its not allowed in C
            * Verilator maps $ to __024, thus `a.b.c` -> `a__024b__024c`
            * I think this issue is actually really important because people will write Chisel3 code and the Verilated interface they see will be covered in these escape sequences and be very hard to read. They will wonder why we did things this way and made it hard to use when we could have just not inserted $
        1. $ is difficult to use in VIM: * for searching quits on $
        1. $ is already weird in some waveform viewers, and it's not clear that it is actually supported by all tools (although perhaps it is)
        1. This puts the the effort on the user and transformation writers to ensure no prefix collisions happen. I can expand on this point if requested.
1. Map all single underscores to double underscores and use underscore as delimiter
  1. Pros
    1. Names are trivial to lower
    1. The names of other things can never change the lowered name of something else
  1. Cons
    1. It's not idempotent, if it's rerun things really blow up
       * This could perhaps be mitigated, but perhaps my favorite quality of Firrtl is the ability to lower, transform, then relower, and that requires idempotence*
        * We could mitigate but running this pass early and not running it again upon relowering, but now we have the potential to insert aliases upon relowering. I suspect this would require transformation writers that insert high-firrtl types to essentially roll their own solution to alias detection and mitigation, rather than just solving it ourselves.
1. The proposal in this PR
  1. Pros:
    1. Keeps fields of bundles and vectors in consistent naming
    1. Relatively simple to understand, and thus predictable
  1. Cons:
    1. It still is name mangling, adding an additional value can change the resulting name of something else
    1. Technically can create things with lots of underscores, but collisions are rare, and examples like above are pretty pathological. They can only occur when users themselves write code with multiple underscores
    1. Because instances and memories are aggregate types, they can get underscores appended to their names. 

I got kind of tired of writing this, might add more later or in response to feedback.

src/main/scala/firrtl/Mappers.scala
jackkoenig: Yeah that's a good point. 

I originally did that to make mapping _ to __ easier. Regardless, I think you're right that String -> String mapping does not imply changing the readers and writers. Also, I no longer believe _ -> __ is the right approach to dealing with name collisions anyway.

azidar:
Overall, I like the algorithm, because it does gracefully keep the names, is good in the common case, and resolves collisions. The existing algorithm with the prefix definitions is certainly simpler, but I feel that it might be too restrictive. Good work.

Have you thought about splitting this into two passes? The first pass is simply a renaming pass that changes the names of bundle fields to avoid eventual collisions but does not actually expand nodes (and can be done very early in the compiler), and then later on running the existing lowering types pass ( or the rest of your pass, though I'm not sure if there was anything wrong with the original one).

The advantage of doing it this way is we can remap names very early on, which allows names to more easily be traced, versus renaming AND lowering all the names. They really are two separate concerns.

azidar:
In fact, the renaming pass can occur (and probably should occur) before type inference (though you will need to calculate the type of the instances after the first pass of renaming port names and types). After you rename, you can just run type inference for the first time after the pass.

jackkoenig:
Yeah I think splitting it into two passes makes a lot of sense. We should probably do a check during lowering that no aliasing names have been introduced (just for safety between the running of the rename part vs the actual lowering), but that's trivial.

The part that does the actual lowering is very similar to the original pass, I actually copy pasted lots of code in writing it, pretty much all of this work was in doing the actual name collisions and resolutions.

PR #115, opened 2016-04-08
Fix remove subaccess

azidar:


jackkoenig:
I think one way to do this is to just iteratively run RemoveAccesses till they're all gone, something similar to what Andrew did in DCE: https://github.com/ucb-bar/firrtl/blob/master/src/main/scala/firrtl/passes/DeadCodeElimination.scala

azidar:
The problem wasn't nested subaccesses. What happened was the pass wasn't visiting the index of a WSubAccess to look for accesses to remove - it was a pretty simple fix.

src/test/scala/firrtlTests/FeatureSpec.scala
```diff
@@ -7,7 +7,7 @@ import org.scalatest._
 class FeatureSpec extends FirrtlPropSpec {
 
   property("Nested SubAccesses should be supported!") {
-    compileFirrtlTest("NestedSubAccessTester", "/features")
+    runFirrtlTest("NestedSubAccessTester", "/features")
```
jackkoenig: I agree the test should run rather than just compile, but this commit message seems to say the opposite?

jackkoenig:
Okay github is hiding my comment, but I don't think the diff is outdated?

Anyway the commit message on 6afe7f02822f4218239279ab5cefe6cf85a0b675 seems to be doing the opposite of what the message says (I think running the test is better than just compiling it).

jackkoenig:
Anyway this looks good, I'm going to squash some commits and merge

jackkoenig:
I'm trying to merge with https://github.com/ucb-bar/firrtl/tree/fix-remove-subaccess-rebase but it looks like there is a race condition causing the tests to fail when run in parallel

jackkoenig:
Merged 4ebc6e2d874b2a936efd262aedb87acb2abcd854

PR #116, opened 2016-04-08
Added Example test

azidar:
An example test that contains the FIRRTL text, parses it, emits it, and checks it.

### DIFF
--- a/src/test/scala/firrtlTests/Example.scala
+++ b/src/test/scala/firrtlTests/Example.scala
@@ -1,0 +1,35 @@
+package firrtlTests
+
+import java.io._
+import org.scalatest._
+import org.scalatest.prop._
+import firrtl.Parser
+import firrtl.FIRRTLEmitter
+import firrtl.Circuit
+import firrtl.passes.Pass
+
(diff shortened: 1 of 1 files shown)


jackkoenig:
I think this is a reasonable way to write some small tests, but it might be nice to not have to have 100% matching, like just check for certain lines or whatever. Perhaps we could have some utility functions that let you search for certain lines, ignoring whitespace or whatever. I think this can be done pretty easily though.

azidar:
Definitely. We should encourage lots of very small, very tailored tests, so I like the fact that this is all contained in one file, is simple to understand, and is relatively lightweight. We should certainly start writing more complicated string comparison operators to make it slightly more robust.

jackkoenig:
I'll try something like this for my name expansion test for #114. I can perhaps make some short tests there that check the different types of collisions.

colinschmidt:
I think we should merge this as is and if we want it to be easier via utilities that it should be another PR

jackkoenig:
We have tests of this sort in master, this was just an example

Issue #89, opened 2016-03-10
Register lack of initialization has bad error


jackkoenig:
In the following FIRRTL

```
circuit Bug :                                                                                                                                                                                                                                                                                                              
  module Bug : 
    input clk : Clock
    input reset : UInt<1>

    wire buggyReset : {valid : UInt<1>}
    buggyReset is invalid
    reg buggyReg : {valid : UInt<1>, bits : UInt<3>}, clk with : (reset => (reset, buggyReset))
```

buggyReg is initialized, but because it is of BundleType, it is not fully initialized by buggyReset. 

This leads to an IndexOutOfBounds Exception during LowerTypes when the pass tries to connect each subfield of buggyReset to buggyReg (similar error in Stanza FIRRTL). 

It seems to me that this should be detected and an error thrown during CheckInitialization. I don't entirely understand Working IR Void, but they seem to be added in ExpandWhens. I'll look into this later but @azidar , maybe the bug is obvious to you.

aswaterman:
Unclear whether or not this is related, but the following causes an IndexOutOfBoundsException in InferWidths:

    circuit RocketTile : 
      module ClientTileLinkIOArbiter : 
[...]

PR #117, opened 2016-04-09
Fixed bundle type equality. 

azidar:
Was not checking for length of bundles, so if the bundles matched but one had more fields, it was not caught. #89


jackkoenig:
Merged in f93dc6831e282338bcda1db04dbf272a36b4a895

PR #118, opened 2016-04-11
Fix keywords connect lhs

jackkoenig:
Turns out that just assuming any Stmt that begins with "reg" is a register declaration is a bad idea.

Added a bit of safety to visitStmt so that we actually check that the first child of a stmt is a TerminalNode before matching on a register or wire, or the like. 

Also added some Parser tests I had been sitting on a while, as well as a check for the specific problem with keywords on lefthand side of connects. 

I also noticed the tests were missing licenses, so might as well add that too.

Fixes ucb-bar/chisel3#87


aswaterman:
lgtm

PR #119, opened 2016-04-13
Add CSE; improve DCE and ConstProp

aswaterman:
Unit tests are forthcoming, pending a test template that's appropriate
for this kind of optimization pass.

### DIFF
--- a/src/main/scala/firrtl/passes/CommonSubexpressionElimination.scala
+++ b/src/main/scala/firrtl/passes/CommonSubexpressionElimination.scala
@@ -38,7 +38,7 @@ object CommonSubexpressionElimination extends Pass {
 
   private def cseOnce(s: Stmt): (Stmt, Long) = {
     var nEliminated = 0L
-    val expressions = collection.mutable.HashMap[Expression, String]()
+    val expressions = collection.mutable.HashMap[MemoizedHash[Expression], String]()
     val nodes = collection.mutable.HashMap[String, Expression]()
 
     def recordNodes(s: Stmt): Stmt = s match {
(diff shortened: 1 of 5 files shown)


jackkoenig:
retest this please

jackkoenig:
It seems that 6fe81557f893081a7365d66c1711c6f21ef16284 makes "Legalize" as it's currently written redundant. I guess we don't necessarily want Constant Prop to be required, but it does make Legalize pointless if it is run

src/main/scala/firrtl/passes/CommonSubexpressionElimination.scala
```diff
@@ -0,0 +1,84 @@
...
+
+    def recordNodes(s: Stmt): Stmt = s match {
+      case x: DefNode =>
+        nodes(x.name) = x.value
+        expressions.getOrElseUpdate(x.value, x.name)
+        x
+      case _ => s map recordNodes
+    }
+
+    def eliminateNodeRef(s: Expression): Expression = s match {
```
jackkoenig: We generally use s for Stmt, e for Expression. Maybe it doesn't matter though

aswaterman:
My inclination is also against requiring ConstProp for correctness.

src/main/scala/firrtl/passes/CommonSubexpressionElimination.scala
aswaterman: that's accidental, I'll fix it

jackkoenig:
looks good to me, Travis might fail because of the concurrency issues but it passed before

jackkoenig:
Travis is broken at the moment, this passed all FIRRTL and Chisel tests locally so I'm merging

PR #120, opened 2016-04-14
Add Uniquify

jackkoenig:
Replaces #114 

Refactor LowerTypes a bit, very similar to old implementation.
Add some tests.
Add Uniquify Pass and several tests.
Some utility function changes to make things make more sense.


jackkoenig:
It looks like Travis fails fairly frequently because of "fork := true", but unfortunately this is required for tests to work because the compiler currently is not threadsafe...

aswaterman:
Time to bite the bullet and make it threadsafe? I can help...

jackkoenig:
Yeah now that I'm done with this (finally), I'll take a look

src/main/scala/firrtl/Utils.scala
```diff
@@ -691,10 +701,21 @@ object Utils {
...
+     // TODO Should these be more safe? (ie. error on UnknownType)
+     def isGround: Boolean = t match {
+       case (_: UIntType | _: SIntType | _: ClockType) => true
+       case _ => false
+     }
+     def isAggregate: Boolean = t match {
+       case (_: BundleType | _: VectorType) => true
+       case _ => false
+     }
+
```
aswaterman: I think in all the contexts these are used, (isGround ^ isAggregate) should be true.  So you might be able to get away with just one, which is resistant to unexpected inputs:

    def isGround: Boolean = t match {
      case (_: UIntType | _: SIntType | _: ClockType) => true
      case (_: BundleType | _: VectorType) => false
    }

src/main/scala/firrtl/Utils.scala
```diff
@@ -428,6 +437,14 @@ object Utils {
          case REVERSE => swap(d)
       }
    }
+   def times (g: Gender, d: Direction): Direction = times(d, g)
+   // TODO Does this make sense?
+   def times (d: Direction, g: Gender): Direction = g match {
+     case FEMALE => d
+     case MALE => swap(d) // MALE == INPUT == REVERSE
+     case _ => error(s"times called on invalid Gender: $g")
```
aswaterman: Not a big deal, but since this is just an internal compile error that is completely unhelpful to the user, I'd just leave off the case _ entirely and let it throw a match error... unless this has frequently proved useful for debugging.

src/main/scala/firrtl/passes/LowerTypes.scala
```diff
@@ -0,0 +1,274 @@
+/*
```
aswaterman: In future, it would be better to not move the pass to a separate file in the same commit as you modify it -- it's hard to see what changed.  (I guess in this case, it was a near-total rewrite, so it doesn't matter much anyway.)

aswaterman:
the code looks fine to me, but I don't have a great idea what's going on in LowerTypes, so it's hard to comment on that.

src/main/scala/firrtl/Utils.scala
jackkoenig: Yeah I thought about doing it that way, I originally split it because I wasn't sure if ClockType was ground, but it actually is so I think you're right. 

Do you think it would make sense to still have isAggregate as `def isAggregate = !t.isGround`?

aswaterman: if it makes the intent clearer when it's used, then yeah

src/main/scala/firrtl/passes/LowerTypes.scala
jackkoenig: I think this is definitely better, even in this case, so I've added it!

azidar:
Have you thought about how this uniquifying pass would work if I wanted to safely inline a module?

jackkoenig:
I hadn't thought about it directly but it should extend quite nicely.

1. Choose a prefix for the module being inlined.
1. Map the supermodule to a type (see stmtToType)
1. Map the inlining module to a type
1. Create a field in the supermodule with it's name as the prefix above and type as the type of the inlining module
1. Create the same style of name mapping structure as used in the pass

The only issue I see is how to insert the stmts from the inlining module, expressions are easy since you can just wrap them in SubFields with the prefix. I think we could introduce some private "SubStmt" that the Uniquify pass could handle similarly and thus properly insert prefixes to the names of DefRegisters, DefWires, etc.

PR #123, opened 2016-04-15
Fix Verilog emission for Modelsim compliation

aswaterman:
Statements like:

    if (foo)
      x <= y;
    end else begin
      ;
    end

are now emitted without the else clause:

    if (foo)
      x <= y;
    end

and statements like:

    if (foo)
      ;
    end else begin
      x <= y;
    end

are now emitted without the empty statement:

    if (foo)
    end else begin
      x <= y;
    end


azidar:
lgtm.

jackkoenig:
Merged 798dda219da440cbbb2e49fed959384c37ca6bb6

PR #124, opened 2016-04-15
add sbt-launch.jar

yunsup:
Commit sbt-launch.jar so we don't rely on the installed sbt versions on systems.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ regress_dir ?= $(root_dir)/regress
 firrtl_dir ?= $(root_dir)/src/main/stanza
 install_dir ?= $(root_dir)/utils/bin
 
-sbt ?= sbt
+sbt ?= java -Xmx8G -Xss512M -jar sbt-launch.jar
 stanza ?= $(install_dir)/stanza
 stanza_bin ?= $(install_dir)/firrtl-stanza
 scala_jar ?= $(install_dir)/firrtl.jar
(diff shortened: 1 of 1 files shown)


ucbjrl:
I don't think this is a good idea. We want all our projects to use the same version of sbt.

If the installed one is incorrect, you should install a "good" one in your private bin. Keeping per-project versions of sbt dooms them to be obsolete.

And where does it end? Should we also package a version of clang? Our own copy of verilator? Linux?

yunsup:
Seems like committing sbt-launch.jar isn't the popular choice.  The original reason was to expose java as the dependency, not the sbt tool itself.  I guess I'm indifferent either way.  However we should streamline all projects that use sbt to use the same variable to point to the sbt build tool.  I suggest to use variable SBT with a ?= so that people can override locally for all projects (rocket-chip, firrtl, and torture)?

aswaterman:
@yunsup's commit message incorrectly describes the problem he was trying to solve: that Java 1.8's default stack size is insufficient to run the Scala compiler on the FIRRTL source code.  Committing sbt-launch.jar isn't the right fix, but providing a way to specify how sbt is invoked is quite reasonable.  Actually, it seems this feature is already in the Makefile ("sbt ?= sbt"), but with the less-than-standard lowercasing.  All caps is most common for this kind of make option.

yunsup:
As @aswaterman points out, I have conflated two changes into this commit.  I'll streamline other repositories with SBT and submit a separate pull request.

PR #126, opened 2016-04-16
Add namespace

jackkoenig:
Replace gensym with Namespace

Most replacing was trivial, InferTypes required a little bit of massaging because of its use of public utility function

IsDeclaration trait is a useful thing to match on as shown in Namespace.scala, but are Fields declarations? They are if you want to create the namespace for a bundle.

I put Namespace in its own file as opposed to Utils because the next step is to replace all names with Identifier wrappers and that will need to be in a separate file with Namespace


aswaterman:
lgtm

PR #127, opened 2016-04-19
add better type mismatch error message

colinschmidt:
also check for it int unittest

It was kind of annoying to only see "Type mismatch" so now we print out the left and right hand sides of the connect as well.

### DIFF
--- a/src/test/scala/firrtlTests/UnitTests.scala
+++ b/src/test/scala/firrtlTests/UnitTests.scala
@@ -60,12 +60,11 @@ class UnitTests extends FirrtlFlatSpec {
         |    input y: {a : UInt<1>}
         |    output x: {a : UInt<1>, b : UInt<1>}
         |    x <= y""".stripMargin
-    val thrown = intercept[PassExceptions] {
+    intercept[CheckTypes.InvalidConnect] {
       passes.foldLeft(parse(input)) {
         (c: Circuit, p: Pass) => p.run(c)
       }
     }
-    assert(thrown.getMessage contains "Type mismatch. Cannot connect")
   }
 
   "Initializing a register with a different type" should "throw an exception" in {
@@ -83,7 +82,7 @@ class UnitTests extends FirrtlFlatSpec {
        |    wire x : { valid : UInt<1> }
        |    reg y : { valid : UInt<1>, bits : UInt<3> }, clk with :
        |      reset => (reset, x)""".stripMargin
-    intercept[PassExceptions] {
+    intercept[CheckTypes.InvalidRegInit] {
       passes.foldLeft(parse(input)) {
         (c: Circuit, p: Pass) => p.run(c)
       }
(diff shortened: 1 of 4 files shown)


src/test/scala/firrtlTests/UnitTests.scala
```diff
@@ -47,11 +47,12 @@ class UnitTests extends FlatSpec with Matchers {
         |    input y: {a : UInt<1>}
         |    output x: {a : UInt<1>, b : UInt<1>}
         |    x <= y""".stripMargin
-    intercept[PassExceptions] {
+    val thrown = intercept[PassExceptions] {
       passes.foldLeft(Parser.parse("",input.split("\n").toIterator)) {
         (c: Circuit, p: Pass) => p.run(c)
       }
     }
+    assert(thrown.getMessage contains "Type mismatch. Cannot connect")
```
aswaterman: Rather than looking for the string, why not do something like

    assert(throw.isInstanceOf[InvalidConnect])

aswaterman: otherwise, lgtm

colinschmidt: I guess I wanted to make sure my updated message gets in, but I'll change this, now that I've manually checked

colinschmidt:
Well actually this causes an issue in the current implementation, changin as you suggest leads to this warning and a failure

```
fruitless type test: a value of type firrtl.passes.PassExceptions cannot also be a firrtl.passes.CheckTypes.InvalidConnect
```

Telling it to intercept the InvalidConnect leads to this failure

```
Expected exception firrtl.passes.CheckTypes$InvalidConnect to be thrown, but firrtl.passes.PassExceptions was thrown. (UnitTests.scala:50)
```

jackkoenig:
I actually had written this up, but got dragged into a meeting before hitting submit:

The main issue is that since we want to get all the errors then throw them all at the same time, we actually just throw one big exception (ie. PassExceptions). The best way would be to actually detect that an InvalidConnect is thrown.

One way we could solve this is by providing some flag to "Except on first error"

jackkoenig:
So there is kind of a fundamental problem in fixing this.

Currently, passes only accept and return a Circuit, there is no global option information available. I've been floating around the idea of creating a "CompilerContext" class that gets instantiated with the Circuit as one of it's children. Then this Context can get passed around rather than just the Circuit. The Context could be used for passing around debug flags (like this one), as well as Analyses and Annotations that add information to the Circuit without being a part of it. @azidar has been doing some revamping of the compiler that he will hopefully upstream soon, how can/should we pass around this kind of information?

jackkoenig:
An alternative that doesn't bring the compiler organization issue to the forefront would be to create some type of debug *compilation* option, such that when it is enabled, the code is compiled to throw the first exception. Then the tests could compile FIRRTL with this option enabled

jackkoenig:
I still want to check for the right exception thrown, but we still need some way to pass debug flags through the compiler. Waiting for @azidar's compiler reorganization, then I'll add that feature.

PR #129, opened 2016-04-19
streamline sbt

yunsup:
following PR#124, this commit streamlines the invocation of sbt throughout all projects so that developers can overwrite the default sbt call by defining environment variable SBT.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ regress_dir ?= $(root_dir)/regress
 firrtl_dir ?= $(root_dir)/src/main/stanza
 install_dir ?= $(root_dir)/utils/bin
 
-sbt ?= sbt
+SBT ?= sbt
 stanza ?= $(install_dir)/stanza
 stanza_bin ?= $(install_dir)/firrtl-stanza
 scala_jar ?= $(install_dir)/firrtl.jar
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -84,7 +84,7 @@ clean:
 	rm -f $(install_dir)/firrtl.jar
 	rm -f $(install_dir)/firrtl
 	rm -f $(install_dir)/firrtl-stanza
-	"$(sbt)" "clean"
+	$(SBT) "clean"
```
ucbjrl: @yunsup, you may want to include the quotes around $(SBT) so it will work with paths with spaces. However, that would prevent you from including some leading sbt options as part of the $(SBT) definition.

yunsup: @ucbjrl, if you need to point to paths with spaces, can't you include the quotes within the definition of environment variable SBT, such as SBT = "/path/with spaces"?

yunsup: I just remembered why I got rid of the quotes, as @ucbjrl points out, I couldn't do:

SBT = java -Xmx8G -Xss512M -jar /path/to/sbt-launch.jar

yunsup:
@jackkoenig @azidar can you review this PR and merge?

aswaterman:
Agreed with Yunsup; it is best to escape the spaces externally.

azidar:
lgtm.

PR #136, opened 2016-04-20
Use a global namespace for VarWidth names

### LINKED ISSUE #135: Strange width inference bug crosses unrelated modules
https://gist.github.com/aswaterman/75f8016abf83c0df569273862c4c1bc6

A wire in module A (which is never instantiated) is given the width of a wire in module C (which is also never instantiated).

aswaterman:
This matches the unstated assumption in InferWidths.

Closes #135


jackkoenig:
lgtm

PR #140, opened 2016-04-21
Remove line in Verilog Emitter erroneously printing ); before module def

### LINKED ISSUE #133: No Ports gives invalid Verilog
```
circuit Bojack :                                                                                                       
  module Bojack :                                                                                                      
    wire x : UInt<32>                                                                                                  
    x <= UInt("hdeadbeef")   
```
compiles to:
```Verilog
);                                                                                                                                                                                                                                            
module Bojack(
);
  wire [31:0] x;
  assign x = 32'hdeadbeef;
endmodule
```

jackkoenig:
Fixes #133 

Not sure why the line was there, perhaps I'm missing something

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -396,8 +396,6 @@ class VerilogEmitter extends Emitter {
       }
       def build_ports () = {
          (m.ports,0 until m.ports.size).zipped.foreach{(p,i) => {
-            var end = ",\n"
-            if (m.ports.size - 1 == i) end = "\n);\n"
             p.direction match {
                case INPUT => portdefs += Seq(p.direction,"  ",p.tpe," ",p.name)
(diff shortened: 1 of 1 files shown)


aswaterman:
I think build_ports is actually correct.  Instead, I'm wondering why it's the only thing being printed for an InModule...

jackkoenig:
So wait, is this the right thing to do or not?

aswaterman:
I believe it is incorrect.  Actually, build_ports is totally fubar.  Note that the result of this code is not even used:

            var end = ",\n"
            if (m.ports.size - 1 == i) end = "\n);\n"

aswaterman:
My bad.  The dead code with "var end" is what confused me.  I think your fix is right, and you should delete the dead code, too.

jackkoenig:
Deleted dead code and rebased on top of master

PR #141, opened 2016-04-21
Improve chisel assert

jackkoenig:
I got sick of the printf and stop resulting from Chisel asserts being guarded separately, so by moving SplitExp to before ConstProp, CSE, and DCE, and by making sure expressions are split out of Stops, we get my desired result:

```
circuit Assert :
  module Assert :
    input clk : Clock
    input reset : UInt<1>

    reg foo : UInt<16>, clk with : (reset => (reset, UInt(0)))
    foo <= add(foo, UInt(1))

    when not(reset) :
      when eq(foo, UInt(13)) :
        printf(clk, UInt(1), "Assertion failed!\n")
        stop(clk, UInt(1), 1)
```

Will now lower to:

```
 circuit Assert : 
  module Assert : 
    input clk : Clock
    input reset : UInt<1>
    reg foo : UInt<16>, clk with : 
      reset => (reset, UInt<1>("h0"))
    node GEN_0 = not(reset)
    node GEN_1 = eq(foo, UInt<4>("hd"))
    node GEN_2 = and(GEN_0, GEN_1)
    printf(clk, GEN_2, "Assertion failed!\n")
    stop(clk, GEN_2, 1)
    foo <= add(foo, UInt<1>("h1"))
```

### DIFF
--- a/src/main/scala/firrtl/Compiler.scala
+++ b/src/main/scala/firrtl/Compiler.scala
@@ -83,11 +83,11 @@ object VerilogCompiler extends Compiler {
     InferTypes,
     ResolveGenders,
     InferWidths,
+    SplitExp,
     ConstProp,
     CommonSubexpressionElimination,
     DeadCodeElimination,
     VerilogWrap,
-    SplitExp,
     VerilogRename
   )
   def run(c: Circuit, w: Writer)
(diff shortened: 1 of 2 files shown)


aswaterman:
lgtm

PR #142, opened 2016-04-21
Fix VCS lint errors

aswaterman:
Rocket-chip gets through cleanly now.


azidar:
Can you write a Unit test for SplitExpressions?

azidar:
Also, why is emitting nested muxes not ok for VCS?

jackkoenig:
I think nested muxes are okay but if you split them you can do some CSE (I think)

aswaterman:
There's no problem with nested muxes...

The problem was that Legalize was emitting bits(mux(...)), but that was never getting split.  The bits() was just being silently dropped in the Emitter.  Adding Mux to SplitExpressions fixed that.  Is there a better approach?

jackkoenig:
Comparing this branch to one that also doesn't include the recent commit moving SplitExp to before CSE, rocket.v resulting from regress/rocket.fir was 124,671 lines, in this branch it is 100,989

aswaterman:
I didn't do this to improve CSE, but I guess it has that effect.

azidar:
So I think there should be a "MergeExpressions" pass that combines split muxes to make the verilog less verbose, and we keep the behavior of SplitExpressions to split muxes. So the sequence of passes is:
SplitExpressions
CSE
MergeExpressions.

Insofar as "Legalize" is concerned, verilator emits warnings for any mismatch of widths, including assignments. I'll hopefully push a pass upstream soon that will eliminate any mismatched widths, and that can replace Legalize.

For now, though, I'm fine with going ahead and using this fix.

aswaterman:
Makes sense to me.

azidar:
I would like to see some Unit tests for SplitExp though :stuck_out_tongue:

aswaterman:
we can do that after @jackkoenig merges his uniquify pass, because he is making it easier to write this kind of test.

jackkoenig:
Agreed, Merged via ec32c852b57a42c7741f8ae27d59c21fcdb86a82

PR #146, opened 2016-04-26
Fix expand whens

jackkoenig:
There's a lot going on in here, I will probably split out some of the minor changes to a different PR, I also need to add in tests for Expand Whens. I would also like to clean up the code a little more, but it seems to work.


jackkoenig:
@azidar a couple of questions about checking initialization:

WVoids are used to see when something is not initialized, that is still used in my changes. However, in the case of expanding a when and there not being a default value to use, I currently just throw the not initialized exception right away in ExpandWhens. I did this because values are memoized as the when is expanded, so the currently CheckInitialization pass won't work. 

I'm not a huge fan of this because it splits where lack of initialization can be detected so I have a couple of ways it might be fixed:

1. Move all initialization checking to Expand Whens
1. Make Check Initialization more complicated in that when a WVoid() is seen being assigned to a node, we trace that node until we hit a DefRegister, DefWire, Port, DefMemory field, or DefInstance field and then error that it isn't fully initialized. Would this work correctly?

jackkoenig:
For the record, I ran rocket-chip in vsim with this branch and successfully ran the asm tests for DefaultL2VLSIConfig

colinschmidt:
This allows Hwacha to get through the compiler. It currently does not pass any tests but it hasn't done that before in chisel3 so this isn't necessarily related. I'm investigating and will open issues as I can.

src/main/scala/firrtl/Parser.scala
```diff
@@ -44,20 +44,31 @@ case class InvalidEscapeCharException(message: String) extends ParserException(m
 
 object Parser extends LazyLogging
 {
+  private def time[R](name: String)(block: => R): R = {
```
azidar: Should this go in Parser? Maybe move it to Utils?

src/main/scala/firrtl/Parser.scala
```diff
@@ -44,20 +44,31 @@ case class InvalidEscapeCharException(message: String) extends ParserException(m
...
   def parse(filename: String, lines: Iterator[String], useInfo: Boolean = true): Circuit = {
-    val fixedInput = Translator.addBrackets(lines)
+    val fixedInput = time("Translator") { Translator.addBrackets(lines) }
     //logger.debug("Preprocessed Input:\n" + fixedInput.result) 
     val antlrStream = new ANTLRInputStream(fixedInput.result)
     val lexer = new FIRRTLLexer(antlrStream)
     val tokens = new CommonTokenStream(lexer)
     val parser = new FIRRTLParser(tokens)
 
     // FIXME Dangerous (TODO)
```
azidar: Why is this dangerous?

src/main/scala/firrtl/Utils.scala
```diff
@@ -569,6 +569,81 @@ object Utils {
...
+    * @example
+    *   Given:   SubField(SubIndex(SubField(Ref("a", UIntType(IntWidth(32))), "b"), 2), "c")
+    *   Returns: (Ref("a"), SubField(SubIndex(Ref("b"), 2), "c"))
+    *   a.b[2].c -> (a, b[2].c)
+    * @example
+    *   Given:   SubField(SubIndex(Ref("b"), 2), "c")
+    *   Returns: (Ref("b"), SubField(SubIndex(EmptyExpression, 2), "c"))
+    *   b[2].c -> (b, EMPTY[2].c)
+    * @note This function only supports WRef, WSubField, and WSubIndex
+    */
```
azidar: Beautiful comment

src/main/scala/firrtl/Utils.scala
```diff
@@ -569,6 +569,81 @@ object Utils {
...
+  case class DeclarationNotFoundException(msg: String) extends FIRRTLException(msg)
+
+  /** Gets the root declaration of an expression
+    *
+    * @param m the [[firrtl.InModule]] to search
+    * @param expr the [[firrtl.Expression]] that refers to some declaration
+    * @return the [[firrtl.IsDeclaration]] of `expr`
+    * @throws DeclarationNotFoundException if no declaration of `expr` is found
+    */
+  def getDeclaration(m: InModule, expr: Expression): IsDeclaration = {
```
azidar: We really need to restructure IR vs WIR, because its starting to bleed everywhere - e.g. which Util functions are for WIR vs IR?

src/main/scala/firrtl/passes/ExpandWhens.scala
```diff
@@ -0,0 +1,189 @@
...
+            val c_netlist = getEntries(netlist, exps)
+            val conseqStmt = expandWhens(c_netlist, AND(p, s.pred))(s.conseq)
+            val altStmt = expandWhens(netlist, AND(p, NOT(s.pred)))(s.alt)
+
+            for (lvalue <- c_netlist.keys) {
+              val value = netlist.get(lvalue) match {
+                case v: Some[Expression] => v.get match {
+                  case WVoid() =>
+                    val decl = getDeclaration(m, lvalue.e1)
+                    throw new RefNotInitializedException(decl.info, m.name, decl.name)
```
azidar: This check isn't correct - I'll commit a test case that shows it. You need to check both the consequence and the alternative branches for the WVoid(), instead of just the consequent.

azidar:
Looks good, other than the check is incomplete. However, I'd recommend Option 2 anyways, and we can make the error message look something like a stack trace, where:

```
node GEN_0 = mux(p,x,VOID)
node GEN_1 = mux(q,GEN_0,r)
w <= GEN_1
```

would trigger an error like:
`w is not fully initialized (w <= GEN_1 <= GEN_0 <= VOID)`

src/main/scala/firrtl/Parser.scala
jackkoenig: I agree it should be in Utils over the Parser.

I would prefer we eventually have a way this data can be collected but not necessarily printed out so it could be used in performance regressions, etc.

src/main/scala/firrtl/Parser.scala
jackkoenig: I should remove this comment. It was due to my misreading the ANTLR4 documentation. Using PredictionMode.LL was too slow, so I switched it to SLL, but I thought SLL could return incorrect results. It turns out SLL _can_ error on correct inputs, but still will always error on incorrect inputs. Erroring on correct inputs is a property of the grammar, and since we have yet to run into this, I think we're probably okay.

See:  http://www.antlr.org/api/Java/org/antlr/v4/runtime/atn/PredictionMode.html#enum_constant_detail

jackkoenig:
I suspected I was missing something on the Check. I agree that providing stack-like trace error messages through Option 2 would be pretty cool, I'll look in to it

jackkoenig:
So I've fixed some of the stuff up, including updating Check Initialization to trace VOIDs and provide [what I think] is a decent error message. Other than moving the time method in the parser to Utils, I think this PR is ready.

jackkoenig:
Okay I've removed the Parser timing stuff and will put it in a different PR.

Confusingly, the order of commits shown here in Github is different than what I have locally, the order should be such that the CheckInitialization tests are added after the pass is refactored, weird.

Other than that strange inconsistency (is it a problem?), this branch is ready for CR and merging.

src/main/scala/firrtl/passes/CheckInitialization.scala
```diff
@@ -0,0 +1,129 @@
...
+            }
+            node
+          case s => s map checkInitS
+        }
+      }
+      checkInitS(m.body)
+
+      // Build Up Errors
+      if (voidExprs.nonEmpty) {
+        voidExprs.keys foreach { expr =>
```
aswaterman: shoudn't need this if-construct; it should suffice just to do something like

    for ((expr, _) <- voidExprs) {

which won't do anything if voidExprs is empty.

src/main/scala/firrtl/passes/ExpandWhens.scala
```diff
@@ -0,0 +1,201 @@
...
+      exps: Seq[Expression]): LinkedHashMap[WrappedExpression, Expression] = {
+    val hashx = LinkedHashMap[WrappedExpression, Expression]()
+    exps foreach (e => if (hash.contains(e)) hashx(e) = hash(e))
+    hashx
+  }
+  private def getFemaleRefs(n: String, t: Type, g: Gender): Seq[Expression] = {
+    def getGender(t: Type, i: Int, g: Gender): Gender = times(g, get_flip(t, i, DEFAULT))
+    val exps = create_exps(WRef(n, t, ExpKind(), g))
+    val expsx = ArrayBuffer[Expression]()
+    for (i <- 0 until exps.size) {
```
aswaterman: suggest not reusing variable name _i_ here -- the indices floating around are confusing enough as it is

jackkoenig:
Okay I've addressed Andrew's two comments from yesterday, and rebased upon master

aswaterman:
I'm fine w/merging

jackkoenig:
I'll merge after travis runs

azidar:
lgtm

Issue #147, opened 2016-04-27
Readwrite ports don't work


jackkoenig:
In writing a test to check that Mems work, I found that readwrite ports are not supported. Specifically, in the Emitter, [this line](https://github.com/ucb-bar/firrtl/blob/54184c2a08aea4c8682d2fe899718c369e00a240/src/main/scala/firrtl/Emitter.scala#L546) creates a nested AND PrimOp, but nested primops are [specifically disallowed](https://github.com/ucb-bar/firrtl/blob/54184c2a08aea4c8682d2fe899718c369e00a240/src/main/scala/firrtl/Emitter.scala#L168-L176)

PR #148, opened 2016-04-27
Fix rwmem

jackkoenig:
This fixes a bug where LowerTypes was using the obsolete field rmode instead of wmode for readwrite ports of memories.

It also fixes a bug where readwrite ports weren't properly supported and adds an executable test to check that they are.

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -554,7 +554,13 @@ class VerilogEmitter extends Emitter {
                   val rmem_port = WSubAccess(mem,raddrx,s.data_type,UNKNOWNGENDER)
                   assign(rdata,rmem_port)
                   val wmem_port = WSubAccess(mem,waddrx,s.data_type,UNKNOWNGENDER)
-                  update(wmem_port,datax,clk,AND(AND(enx,maskx),wmode))
+
+                  val tempName = namespace.newTemp
+                  val tempExp = AND(enx,maskx)
+                  declare("wire", tempName, tpe(tempExp))
+                  val tempWRef = wref(tempName, tpe(tempExp))
+                  assign(tempWRef, tempExp)
+                  update(wmem_port,datax,clk,AND(tempWRef,wmode))
                }
             }
             case (s:Begin) => s map (build_streams)
--- a/src/main/scala/firrtl/passes/LowerTypes.scala
+++ b/src/main/scala/firrtl/passes/LowerTypes.scala
@@ -100,7 +100,7 @@ object LowerTypes extends Pass {
       def lowerTypesMemExp(e: Expression): Seq[Expression] = {
         val (mem, port, field, tail) = splitMemRef(e)
         // Fields that need to be replicated for each resulting mem
-        if (Seq("addr", "en", "clk", "rmode").contains(field.name)) {
+        if (Seq("addr", "en", "clk", "wmode").contains(field.name)) {
           require(tail.isEmpty) // there can't be a tail for these
           val memType = memDataTypeMap(mem.name)
 
--- a/src/test/scala/firrtlTests/IntegrationSpec.scala
+++ b/src/test/scala/firrtlTests/IntegrationSpec.scala
@@ -35,7 +35,8 @@ class IntegrationSpec extends FirrtlPropSpec {
   case class Test(name: String, dir: String)
 
   val runTests = Seq(Test("GCDTester", "/integration"),
-                     Test("RightShiftTester", "/integration"))
+                     Test("RightShiftTester", "/integration"),
+                     Test("MemTester", "/integration"))
       
 
   runTests foreach { test =>
--- a/test/integration/MemTester.fir
+++ b/test/integration/MemTester.fir
@@ -1,0 +1,55 @@
+
+circuit MemTester :
+  module ReadWrite :
+    input clk : Clock
+    input reset : UInt<1>
+
+    reg n : UInt<32>, clk with :
+      reset => (reset, UInt(0))
+
+    reg wmode : UInt<1>, clk with : 
+      reset => (reset, UInt(1))
+    wmode <= not(wmode)
+
+    reg addr : UInt<5>, clk with :
+      reset => (reset, UInt(0))
+
+    when eq(wmode, UInt(0)) :
+      n <= add(n, UInt(1))
+      addr <= add(addr, UInt(1))
+
+    mem m :
+      data-type => UInt<32>
+      depth => 32
+      read-latency => 0
+      write-latency => 1
+      readwriter => rw
+      read-under-write => undefined
+    m.rw.clk <= clk
+    m.rw.addr <= addr
+    m.rw.wmode <= wmode
+    m.rw.data <= n
+    m.rw.mask <= UInt(1)
+    m.rw.en <= UInt(1)
+
+    when not(reset) :
+      when eq(wmode, UInt(0)) :
+        when neq(m.rw.rdata, n) :
+          printf(clk, UInt(1), "Assertion failed! m.rw.data has the wrong value!\n")
+          stop(clk, UInt(1), 1)
+
+  module MemTester :
+    input clk : Clock
+    input reset : UInt<1>
+
+    reg count : UInt<32>, clk with :
+      reset => (reset, UInt(100))
+    count <= tail(sub(count, UInt(1)), 1)
+
+    inst rwMod of ReadWrite
+    rwMod.clk <= clk
+    rwMod.reset <= reset
+
+    when eq(count, UInt(0)) :
+      stop(clk, UInt(1), 0)
+


src/main/scala/firrtl/passes/LowerTypes.scala
```diff
@@ -139,7 +139,7 @@ object LowerTypes extends Pass {
       def lowerTypesMemExp(e: Expression): Seq[Expression] = {
         val (mem, port, field, tail) = splitMemRef(e)
         // Fields that need to be replicated for each resulting mem
-        if (Seq("addr", "en", "clk", "rmode").contains(field.name)) {
+        if (Seq("addr", "en", "clk", "wmode").contains(field.name)) {
```
aswaterman: this bug wouldn't have come up if these were checked by the parser, and turned into fields in IR nodes that can be statically checked by the scala compiler

src/main/scala/firrtl/Emitter.scala
```diff
@@ -543,7 +543,13 @@ class VerilogEmitter extends Emitter {
...
                   assign(rdata,rmem_port)
                   val wmem_port = WSubAccess(mem,waddrx,s.data_type,UNKNOWNGENDER)
-                  update(wmem_port,datax,clk,AND(AND(enx,maskx),wmode))
+
+                  val tempName = namespace.newTemp
+                  val tempExp = AND(enx,maskx)
+                  declare("wire", tempName, tpe(tempExp))
+                  val tempWRef = wref(tempName, tpe(tempExp))
+                  assign(tempWRef, tempExp)
+                  update(wmem_port,datax,clk,AND(tempWRef,wmode))
```
aswaterman: The emitter really shouldn't be creating nodes (which is why this looks ugly).  This should probably be moved into some other pass.

jackkoenig: I'm curious how this could be moved to another pass. This essentially creates logic that guards the writing of a rw port to be the AND of write enable, write mask, and write mode. According to FIRRTL semantics, these are 3 separate fields that are only being merged here to simulate in Verilog. 

I guess there could be a pass on LoFirrtl that replaces FIRRTL mems with some VerilogSimMem or something that merges these three signals into a single one?

A reasonable alternative is to go back to `AND(AND(enx, maskx), wmode)` and instead just make the emitter handle that since it's a simple logical condition for writing to the port. The main issue with how it is currently implemented is that this update function is using the same emission function used to emit all primops (which disallows nesting) when instead it could use a special one that allows for these 3 things to be ANDed together.

aswaterman:
Yeah, I don't see how this can be done in a way that could be serialized and re-parsed.  Given that the warts are localized, maybe it's OK

jackkoenig:
I'd like to get this fix merged in the near future so here are my thoughts:

On replacing "rmode" with "wmode,":  Redoing how memory ports are handled in the compiler would have prevented this from showing up, but I think that change is much larger than this bug fix. I'm planning to look into it, but it's going to require substantial changes all over the compiler.

On guarding writing to a readwrite port with en, mask, and wmode:  I agree that the emitter should not create a node, instead I think the emitter should just support the original nested AND. It's really not that complicated and while we could introduce some transformation that would not be serializable and reparseable, I think that's complete overkill for just trying to and 3 fields together. Additionally, if we support aggregate typed memories in the future as requested in #165, then I don't see how we can possibly handle this in another pass, those 3 signals have to be ANDed together as the condition for writing to the memory.

azidar:
So, I've been thinking about this, and I'm considering a different approach - what if we removed memories entirely from the compiler, and instead had "intrinsic" memories that were library Firrtl modules that would contain the necessary Firrtl logic to implement cmem and smem behaviors. We tag them as special using the annotation system, so they will get preserved throughout the compiler (and can be substituted with black boxes, for example). The specified behavior would be identical to the current mems (e.g. write ordering is unspecified), but the Firrtl intrinsic memory would have the same behavior as a vector-typed register.

I think this proposal goes in tandem with trying to emit bundles/vectors directly. If we do this, then the emitted special memories wouldn't need any special casing in Firrtl - we would just emit the intrinsic memory module as normal, and the vector-typed register would get emitted in its entirety.

Thoughts?

jackkoenig:
I'm liking this idea a lot.

I'm not even sure it needs to be a _FIRRTL_ library in that FIRRTL could just provide the appropriate first class annotations, and Chisel could just emit the appropriate FIRRTL to emulate its smem and cmem. I do think providing them as a FIRRTL library could standardize how these memories are emitted by other frontends; however, it may be desirable that the frontends are required to emit the proper FIRRTL code to emulate the semantics of their memories. 

This also would result in another very desirable quality in the FIRRTL compiler:  black boxes that have a provided implementation that is used for simulation but not synthesis.

aswaterman:
I think it is a technically reasonable idea.  But so far, "annotate it" has
been a euphemism for "maybe we'll get around to figuring it out some other
day."  So I'm a actually quite skeptical of it from a project management
perspective.

As a meta-issue, this kind of change represents significant instability in
FIRRTL.  I recognize that the project is still in its early stages, but
this is no minor change, and it does undermine my confidence in it being a
stable platform against which to develop.

CuppoJava:
I don't support this idea.

We've talked about this before. The same criticisms continue to apply.

1) The mem construct is parameterized by the data type that it holds.

   mem m :
      data-type => {a:SInt, b:SInt}

   The module system is not powerful enough (and I don't think it
   should be) to be parameterized by a data type.

   If you choose to instantiate a different black box for each kind of
   memory, then what relation do they have to each other?

   instance mem_a of SInt_Mem_with_a_b_fields
   instance mem_b of SInt_Mem_with_a_b_c_fields
   etc...

2) The semantics of the mem construct is clearly defined. Why is it that

      mem mymem :
         mask => mk
         enable => UInt(1)

   is equivalent to

      mem mymem :
         mask => UInt(1)
         enable => mk

   ? What rules are we following to do this reasoning? These rules
   exist, whether or not we choose to write them down explicitly.

   Currently, those rules are written in the specification. But we
   could instead have an annotation called *mask signal* coupled with
   a programmer comment that says masks can be implemented as enables
   in certain situations. etc. Would that make the system any clearer?

3) From our reasoning, we know that a memory containing multiple fields

      mem m :
         data-type => {a:UInt, b:UInt}

   can be split up into multiple smaller ones.

      mem m_a :
         data-type => UInt

      mem m_b :
         data-type => UInt

   How do we know we can do that? If we take the blackbox philosophy,
   and we see the following:

      instance m of BlackBox137 (annotation = memory_containing_a_b_fields)

   Would you be able to deduce that the following is equivalent?

      instance m_a of BlackBox211 (annotation =
memory_containing_UInt_field)
      instance m_b of BlackBox211 (annotation =
memory_containing_UInt_field)

4) The semantics of the memory come with clear error conditions. Given
   the following mem:

      mem m :
         depth => 10
         addr => address

   We can conceivably write a pass that issues a warning if address is
   ever greater than 10. How do we know that's possible? Taking the
   blackbox example again.

      instance m of BlackBox11 (annotation = memory)

   How rich would you have to make the annotation system in order to
   tell the pass that the addr line has to be less than 10? And
   supposing you did this, would the resulting system have any less
   special casing than the current compiler?

The above are my reasonable criticisms. The following is my
philosophical stance.

If the goal is simply to reduce special case constructs, we can take
this philosophy to absurdity and propose that even primitive
operations be black boxes. (By the way, this *was* proposed at some
point.)

So instead of

   val z = x + y

we can write

   instance z of Adder
   z.input_a <= x
   z.input_b <= y

and we can annotate Adder with a special flag that says that the
output is one bit greater than the maximum width of either input.

This annotation system is becoming a proverbial rug under which we
sweep any immediate difficulties we encounter.

The same general arguments against black boxing primitive operations
apply, and for the same reasons, to memories.

   - Patrick

azidar:
Lots to comment on, but before I do, here is a different proposal:

Forget the intrinsic memories. What if we had a pass right before emission that replaces all mem constructs with a module implementing the mem? This would remove a lot of the complexity from the emitter (my concern), wouldn't depart from the spec (Andrew's concern), and preserves the semantic information for the majority of the compiler (Patrick's concern).

CuppoJava:
Completely on board with that.

aswaterman:
I think that's quite sensible.  It preserves the spec, and it provides a clear path to supporting technology-specific memories.  We should, of course, still package with FIRRTL a pass, or script, or whatever to provide behavioral Verilog implementations of these memories, which will also serve as a template for mapping to technology-specific RAMs.

jackkoenig:
Alright so it looks like we have broad agreement, what's the migration strategy?

We have two outstanding issues:

* #147 - with the fix proposed here
* #155 - with a fix in fbc396e2cf42505c76188666153373471f98b00c

Obviously these fixes do not fit the new proposal, but they could be used temporarily if the new proposal is going to take a while.

azidar:
I think we should fix the immediate concern, and also write tests so that when we implement the new solution, we can ensure it works.

jackkoenig:
Another thought:

Also, as discussed in #155, we have an implicit validif on the read data from non-power-of-two memories. We discussed emitting ifdef guards for the randomization logic since it comes at a significant simulation performance penalty. In the case of memories, however, this randomization logic is not  optional (but obviously not synthesizable). 

We have two proposals on how to deal with this:

* Adam proposed emitting the randomization logic for validifs based on a compiler option.
  * This requires some (very necessary) work to pass these options through the compiler. I want to get this work done anyway for other things as well.
*  Andrew proposed guarding validifs in the Verilog with RANDOMIZE_DISCONNECT_NETS or something so the user could specify if they want to use the randomization in a particular simulation (without rerunning firrtl)
  * This requires some way of distinguishing between required and non-required validifs (read: annotations)
  * #154 (which I am trying to merge today) has a proposal for annotations that could make this work

In the meantime, we can continue to remove validifs (ie. there is no randomization support), and then transform the circuit to replace memories *after* the remove validifs pass has been run. Then by just emitting the randomization logic we will have correct operation equivalent to what we already have (except with the mentioned bugs fixed) without having to worry about the above issues.

PR #149, opened 2016-04-27
Remove Verilator Warnings

azidar:
The following changes were necessary to emit warning-free Verilator Verilog:

- Remove ValidIfs. This pass replaces ValidIfs with their value, with the hopes of future simplification of Emitter.scala
- Pad Widths. This pass pads/truncates assignments and expressions (e.g. mux inputs are the same width)
- ConstProp case. This addition simplifies two cases: comparisons against zero, and comparisons against a value outside the range of a wire.
- Memory ports are under same clock
- Randomized values are bit-selected from (to eliminate corresponding Verilator warning)

### DIFF
--- a/src/main/scala/firrtl/passes/ConstProp.scala
+++ b/src/main/scala/firrtl/passes/ConstProp.scala
@@ -125,6 +125,82 @@ object ConstProp extends Pass {
     }
   }
 
+  private def foldComparison(e: DoPrim) = {
+    def foldIfZeroedArg(x: Expression): Expression = {
+      def isUInt(e: Expression): Boolean = tpe(e) match {
+        case UIntType(_) => true
+        case _ => false
+      }
+      def isZero(e: Expression) = e match {
+          case UIntValue(value,_) => value == BigInt(0)
+          case SIntValue(value,_) => value == BigInt(0)
+          case _ => false
+        }
+      x match {
+        case DoPrim(LESS_OP,      Seq(a,b),_,_) if(isUInt(a) && isZero(b)) => zero
+        case DoPrim(LESS_EQ_OP,   Seq(a,b),_,_) if(isZero(a) && isUInt(b)) => one
+        case DoPrim(GREATER_OP,   Seq(a,b),_,_) if(isZero(a) && isUInt(b)) => zero
+        case DoPrim(GREATER_EQ_OP,Seq(a,b),_,_) if(isUInt(a) && isZero(b)) => one
+        case e => e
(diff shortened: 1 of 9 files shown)


src/main/scala/firrtl/passes/ConstProp.scala
```diff
@@ -125,6 +125,82 @@ object ConstProp extends Pass {
...
+            case GREATER_EQ_OP if (r0 >= r1) => one
+            //case EQUAL_OP      if (r0 === r1) => one
+            //case NEQUAL_OP     if ((r0 < r1) || (r0 > r1)) => one
+            // Always false
+            case LESS_OP       if (r0 >= r1) => zero
+            case LESS_EQ_OP    if (r0 > r1) => zero
+            case GREATER_OP    if (r0 <= r1) => zero
+            case GREATER_EQ_OP if (r0 < r1) => zero
+            //case EQUAL_OP      if ((r0 < r1) || (r0 > r1)) => zero
+            //case NEQUAL_OP     if (r0 === r1) => zero
```
aswaterman: any particular reason these are commented out?

aswaterman: I see that they were removed in a later commit, nevermind

aswaterman:
This generates incorrect Verilog for the randomization code by behaviorally assigning to wires.  To generate a 1-bit random signal, it's emitting

    wire [31:0] GEN_36;
    reg [1:0] GEN_17;
    ...
    GEN_36 = {1{$random}}; 
    GEN_16 = GEN_36[0:0];

aswaterman:
Haven't had time to do a CR, but I did verify rocket-chip passes.

src/main/scala/firrtl/passes/RemoveValidIf.scala
```diff
@@ -0,0 +1,26 @@
+package firrtl
+package passes
+import firrtl.Mappers.{ExpMap, StmtMap}
+
+// Removes ValidIf as an optimization
+object RemoveValidIf extends Pass {
```
jackkoenig: Legalize was intended to be a pass for miscellaneous things that are required for Verilog Legalization. Perhaps this pass should be merged with that?

azidar: Agreed - we will need to add the shr transform to PadWidths.

jackkoenig:
Replaced by #164, I'm looking over the code but should merge soon.

PR #152, opened 2016-04-29
Update timing and parser

jackkoenig:
Minor cleanup, new "time" function makes code that prints timing information cleaner


azidar:
lgtm

PR #153, opened 2016-04-29
Update check high form

jackkoenig:
Minor unused code deletion and useful error message for those porting code from Chisel2 to Chisel3 dealing with innumerable Check High Form errors

### DIFF
--- a/src/main/scala/firrtl/passes/Checks.scala
+++ b/src/main/scala/firrtl/passes/Checks.scala
@@ -45,7 +45,6 @@ object CheckHighForm extends Pass with LazyLogging {
 
   // Custom Exceptions
   class NotUniqueException(name: String) extends PassException(s"${sinfo}: [module ${mname}] Reference ${name} does not have a unique name.")
-  class IsPrefixException(prefix: String) extends PassException(s"${sinfo}: [module ${mname}] Symbol ${prefix} is a prefix.")
   class InvalidLOCException extends PassException(s"${sinfo}: [module ${mname}] Invalid connect to an expression that is not a reference or a WritePort.")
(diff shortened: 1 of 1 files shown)


src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -289,7 +256,10 @@ object CheckHighForm extends Pass with LazyLogging {
     }
     sinfo = c.info
     if (numTopM != 1) errors.append(new NoTopModuleException(c.main))
-    if (errors.nonEmpty) throw new PassExceptions(errors)
+    if (errors.nonEmpty) {
+      errors.append(new PassException(s"${errors.length} high form error(s) detected"))
+      throw new PassExceptions(errors)
```
aswaterman: The error reporting idiom

    if (errors.nonEmpty)
      throw new PassExceptions(errors)

and perhaps

    if (errors.nonEmpty) {
      errors.append(...)
      throw new PassExceptions(errors)
    }

are probably common enough to justify utility functions (e.g., throwPassExceptions(errors), which does nothing if errors is empty).  Or perhaps make errors a class rather than ArrayBuffer[PassException]

jackkoenig: I think making errors a class makes the most sense, I'll look into it.

I'd also like an elegant (and threadsafe) way to provide commandline options that could disable "grouped" error reporting, instead giving the more specific exception so we can check it in testing. I'm thinking maybe we pass around a CircuitContext object along with the Circuit that contains the errors object. At instantiation time the errors object could be told to either relay the exception or save it to be thrown as a group later.

azidar:
lgtm.

PR #154, opened 2016-05-02
Restructured Compiler to contain Transforms. Added an InlineInstance pass.

azidar:
New structure:

- A Compiler contains a sequence of Transforms.
- Each Transform consumes a Circuit and a list of CircuitAnnotations.
- Each Transform returns a Circuit, an optional CircuitAnnotation, and an optional  RenameMap.
- After executing each transform, the compiler updates the list of CircuitAnnotations with the RenameMap.

An example Transform, InlineInstances,  uses this system.


aswaterman:
Ah, good stuff.

src/main/scala/firrtl/Compiler.scala
```diff
@@ -33,71 +33,75 @@ import java.io.Writer
...
+// ===========================================
+//                 Compilers
+// -------------------------------------------
+
+case class CompilerResult (circuit: Circuit, annotations: Seq[CircuitAnnotation])
+
+// - A sequence of transformations
+// - Call compile to executes each transformation in sequence onto
+//    a given circuit.
+trait Compiler extends LazyLogging {
```
jackkoenig: I've used trait all over the place, but I'm thinking we should start using abstract class for things like this that are meant to be the primary parent class of other things, and traits as mixins. Perhaps I am wrong though. Thoughts?

src/main/scala/firrtl/Driver.scala
```diff
@@ -24,76 +24,93 @@ ANY, PROVIDED HEREUNDER IS PROVIDED "AS IS". REGENTS HAS NO OBLIGATION
...
+  -X <compiler>    Specify the target compiler
+                   Currently supported: high low verilog
   """
-  private val defaultOptions = Map[Symbol, Any]().withDefaultValue(false)
 
-  def compile(input: String, output: String, compiler: Compiler)
+  // Compiles circuit. First parses a circuit from an input file,
+  //  executes all compiler passes, and writes result to an output
+  //  file.
+  def compile(input: String, output: String, compiler: Compiler, annotations: Seq[CircuitAnnotation] = Seq.empty)
```
jackkoenig: We should do these comments as ScalaDoc style eg.
```scala
 /** Compiles a circuit
   *
   *  First parses a circuit from an input file, executes all compiler passes, and writes result to an output file
   */
 def compile(input: String, output: String, compiler: Compiler, annotations: Seq[CircuitAnnotation] = Seq.empty)
```

src/main/scala/firrtl/Driver.scala
```diff
@@ -24,76 +24,93 @@ ANY, PROVIDED HEREUNDER IS PROVIDED "AS IS". REGENTS HAS NO OBLIGATION
...
   }
 
-  def main(args: Array[String])
-  {
+  // Arguments specify the compiler, input file, and output file
+  def main(args: Array[String]) = {
     val arglist = args.toList
-    type OptionMap = Map[Symbol, Any]
 
+    sealed trait CompilerOption
```
jackkoenig: Good work sealing, I need to do this more.

src/main/scala/firrtl/Driver.scala
```diff
@@ -24,76 +24,93 @@ ANY, PROVIDED HEREUNDER IS PROVIDED "AS IS". REGENTS HAS NO OBLIGATION
...
+       }
+
+    type OptionMap = Map[CompilerOption, String]
     def nextOption(map: OptionMap, list: List[String]): OptionMap = {
       list match {
         case Nil => map
+        case "-inline" :: value :: tail => {
+          handleInlineOption(value)
+          nextOption(map, tail)
+        }
```
jackkoenig: Perhaps it's splitting hairs, but I think we should stick to the standard one dash for flaglike options, two dashes for words

src/main/scala/firrtl/LoweringCompilers.scala
```diff
@@ -0,0 +1,202 @@
...
+trait SimpleRun extends LazyLogging {
+   def run (circuit: Circuit, passes: Seq[Pass]): TransformResult = {
+      val result = passes.foldLeft(circuit) {
+         (c: Circuit, pass: Pass) => {
+            val name = pass.name
+            logger.debug(s"Starting ${name}")
+            val x = pass.run(c)
+            logger.debug(x.serialize)
+            logger.debug(s"Finished ${name}")
+            x
```
jackkoenig: Should we put the new Utils.time function around this?

src/main/scala/firrtl/passes/Inline.scala
```diff
@@ -0,0 +1,171 @@
...
+// Tags an annotation to be consumed by this pass
+case object InlineCAKind extends CircuitAnnotationKind
+
+// Only use on legal Firrtl. Specifically, the restriction of
+//  instance loops must have been checked, or else this pass can
+//  infinitely recurse
+object InlineInstances extends Transform {
+   val inlineDelim = "$"
+   def name = "Inline Instances"
+   def execute(circuit: Circuit, annotations: Seq[CircuitAnnotation]): TransformResult = {
```
jackkoenig: This isn't going to work just using "$" as a delim, as with the previous lowering issues, this can lead to collisions (try adding output i$a to any of the InlineInstancetests). I can look in to modifying this pass to use the (currently private) utility functions in Uniquify to achieve the same affect. If I do that, we can also use _ as the delimiter instead.

src/test/scala/firrtlTests/AnnotationTests.scala
```diff
@@ -0,0 +1,57 @@
...
+ */
+abstract class AnnotationSpec extends FlatSpec {
+   def parse (s: String): Circuit = Parser.parse("",s.split("\n").toIterator)
+   def compiler: Compiler
+   def input: String
+   def getResult (annotation: CircuitAnnotation): CompilerResult = {
+      val writer = new StringWriter()
+      compiler.compile(parse(input), Seq(annotation), writer)
+   }
+}
```
jackkoenig: FirrtlFlatSpec exists so that things can extend it and get common functions for free. We could consider putting parse as a function in FirrtlRunners or something.

jackkoenig:
On the whole this looks great! 

If you agree @azidar, I'm going to look in to tweaking the InlineInstances pass to use Uniquify's stuff to help deal with name collisions. @davidbiancolin and I have a transformation that I'll try out with this branch just to make sure I understand what all is going on and how to use it. Good work!

src/main/scala/firrtl/Compiler.scala
```diff
@@ -33,71 +33,75 @@ import java.io.Writer
...
-  def run(c: Circuit, w: Writer)
+
+// ===========================================
+//                 Annotations
+// -------------------------------------------
+case class AnnotationException(message: String) extends Exception(message)
+trait Named { def name: String }
+case class ModuleName(name: String) extends Named
+case class ComponentName(name: String, module: ModuleName) extends Named
+
```
jackkoenig: I wonder if this `Named` can be unified with `HasName` and `IsDeclaration` in the IR.scala. Using references to the actual Named IR node would actually have a smaller memory footprint (since we don't have to allocate an additional string, although I plan to fix this later through unique `Identifier` objects) and it would provide more power. Similar with the ModuleName.

This different approach would require creating new annotation objects when IR nodes change even if their name does not change though, this may be an undesirable overhead.

src/main/scala/firrtl/Compiler.scala
azidar: I personally always prefer traits over abstract classes, but I'm open to hear other reasons.

src/main/scala/firrtl/passes/Inline.scala
azidar: Yup, that would be great if you looked into it.

src/test/scala/firrtlTests/AnnotationTests.scala
azidar: Yeah, I'm having a hard time figuring out the best parent class for all the different types of tests we want to run. I'm content having multiple parent classes, and once we settle more on our testing strategy, we can look for what is in common and merge them later.

src/main/scala/firrtl/Compiler.scala
azidar: > Using references to the actual Named IR node would actually have a smaller memory footprint

Not a good idea. All of the passes are operating under the assumption that the Firrtl graph is a tree. If there are cycles, we can introduce a whole number of errors (as well as having to deal with the HUGE headache of stale references). The cost of allocating a String is _totally_ acceptable.

src/main/scala/firrtl/Compiler.scala
jackkoenig: So there isn't a huge difference between traits and abstract classes, but the one big one is that a class can only extend a single abstract class while it can obviously mix in multiple traits. I think this is a pattern to take advantage of for various things. Not necessarily here, but I think in the IR for example, things like Expression and Statement should be abstract classes because it's not appropriate for something to extend both. Most things should definitely be traits, but for things that have a clear inheritance tree, using abstract classes makes this more explicit and thus clearer in my opinion

PR #158, opened 2016-05-03
Prototype supporting file info

jackkoenig:
Uses same StringLit as Print

Do not merge yet, this is just a prototype.

This PR provides fileinfo as double quoted strings before circuits, modules, ports, and statements. I think we should discuss how we want file info to actually look. In this current implementation, file info goes before the firrtl text and it looks something like this:
```
"TestFileInfo.scala@7.0" circuit Test :
  "TestFileInfo.scala@7.0" module Test :
    "TestFileInfo.scala@9.2" input clk : Clock
    "TestFileInfo.scala@9.2" input reset : UInt<1>

    "TestFileInfo.scala@11.2" reg x : UInt<32>, clk with :
      reset => (reset, UInt(0))

    "TestFileInfo.scala@13.2" x <= add(x, UInt(1))

    "TestFileInfo.scala@15.2" when eq(x, UInt(20)) :
      "TestFileInfo.scala@16.4" stop(clk, UInt(1), 0)
```
Personally I find that unreadable. An alternative would be to put the fileinfo on the previous line:
```
"TestFileInfo.scala@7.0"
circuit Test :  
  "TestFileInfo.scala@7.0"
  module Test : 
    "TestFileInfo.scala@9.2"
    input clk : Clock 
    "TestFileInfo.scala@9.2"
    input reset : UInt<1>
    
    "TestFileInfo.scala@11.2"
    reg x : UInt<32>, clk with : 
      reset => (reset, UInt(0))

    "TestFileInfo.scala@13.2"
    x <= add(x, UInt(1)) 

    "TestFileInfo.scala@15.2"
    when eq(x, UInt(20)) : 
      "TestFileInfo.scala@16.4"
      stop(clk, UInt(1), 0)
```
Still difficult to read, but with syntax highlighting it might be okay. I think putting the file info after the firrtl text is better: 
```
circuit Test : "TestFileInfo.scala@7.0"  
  module Test : "TestFileInfo.scala@7.0" 
    input clk : Clock "TestFileInfo.scala@9.2"  
    input reset : UInt<1> "TestFileInfo.scala@9.2" 
    
    reg x : UInt<32>, clk with : "TestFileInfo.scala@11.2"
      reset => (reset, UInt(0))
    
    x <= add(x, UInt(1)) "TestFileInfo.scala@13.2" 
    
    when eq(x, UInt(20)) : "TestFileInfo.scala@15.2"
      stop(clk, UInt(1), 0) "TestFileInfo.scala@16.4"
```
With proper syntax highlighting I think this is the most readable.

### DIFF
--- a/src/main/antlr4/FIRRTL.g4
+++ b/src/main/antlr4/FIRRTL.g4
@@ -37,16 +37,16 @@ grammar FIRRTL;
 
 // Does there have to be at least one module?
 circuit
-  : 'circuit' id ':' '{' module* '}'
+  : info? 'circuit' id ':' '{' module* '}'
   ;
 
 module
-  : 'module' id ':' '{' port* block '}'
-  | 'extmodule' id ':' '{' port* '}'
+  : info? 'module' id ':' '{' port* block '}'
+  | info? 'extmodule' id ':' '{' port* '}'
   ;
 
 port
-  : dir id ':' type
+  : info? dir id ':' type
   ;
 
 dir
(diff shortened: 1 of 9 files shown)


azidar:
Yeah, I think after is probably the best

jackkoenig:
We should also discuss how we want the file info to look. I used string literals because it was the easy to implement using the existing StringLit implementation, but perhaps this is not the right thing to do. We could use brackets: `[TestFileInfo.scala@11.2]` and colon as the conventional "append" `[TestFileInfo.scala@11.2:test.fir@7.4]`. 

Also, what characters should be allowed in file info? The same characters as StringLit? Should people be able to escape tabs, and eventually escape any arbitrary character? Currently string lit just supports a very restricted subset of ASCII since I thought it would be better to be strict now and add support later rather than the other way around.

jackkoenig:
Replaced by #166

Issue #94, opened 2016-03-25
Scala IR revamp


jackkoenig:
So I think there are some things that could/should change in the Scala IR definition that would clean it up a little. Some ideas are to match scala style, some are more my own stylistic opinions. 

https://github.com/ucb-bar/firrtl/blob/master/src/main/scala/firrtl/IR.scala

1. Change all case class Name() to case object Name
  1. This would match scala style
1. Abbreviate/Rename various vertices 
  1. Rename Expression to Exp (we use Stmt for Statements so why not have Exp to match?)
  1. Rename DefRegister to DefReg (to match abbreviation in concrete syntax)
  1. Rename DefInstance to DefInst (to match abbreviation in concrete syntax)
  1. Rename DefMemory to DefMem (to match abbreviation in concrete syntax)
  1. Rename Conditionally to When (less important but Conditionally is kind of long)
  1. Rename Print to Printf (to match name in spec)
1. Change Primops to camel-case 
  1. ADD_OP -> AddOp (perhaps just Add?)
1. Make primops match concrete syntax name
  1. GREATER_EQ_OP -> GeqOp 
  1. Do we gain anything by making the name long?  This makes it confusing to know what names in a .fir file match the various PrimOps
  1. I'm sympathetic to the self-documenting names idea, but DYN_SHIFT_RIGHT_OP just looks ridiculous when that information can easily be put in a scala-doc
1. Use camelcase within case classes
  1. read_latency -> readLatency (or readLat)
1. Rename Direction to PortDir
1. Rename Flip to FieldDir
  1. DEFAULT is just confusing since the flip keyword really means REVERSE
  1. I think Default and Flip should be the case objects, to match "flip" keyword
1. Rename BulkConnect to PartialConnect (because spec)
1. Rename top-level trait from AST to ???
  - Something about the name bothers me. While FIRRTL circuits are ASTs, hardware is really a DAG.
  - I'd like it to be like FirrtlNode, but Node is something else, what about like FirrtlVertex?
1. Standardize use of FIRRTL vs Firrtl vs firrtl
  - Formerly, I put FIRRTL everywhere, but that doesn't really work well with camelcase
  - FIRRTLException is awkwad, I'd like FirrtlException better.

This issue may sound stupid but I'd like feedback *before* I change things so I won't change anything I shouldn't

@azidar I'd appreciate your feedback in particular

ccelio:
[...]

PR #160, opened 2016-05-05
Proposal for cleaning up the IR

jackkoenig:
Originally proposed some of these changes in #94 here is a proposed cleanup.

I lengthened the names of most things, but I'm still partial to reasonable abbreviations. The problem is that some things start to get long...See stuff in MemoryPortField area, Orientation, and the fact that I did not extend PrimOp to PrimitiveOperation. I think spelling out Expression (as opposed to Expr) and Statement (as opposed to Stmt) suggests we should spell out PrimOp as well. 

Most of the changes here are minor and not controversial:

1. Flip -> Orientation
1. case class Name() -> case object Name
1. Stmt -> Statement
1. `exp` abbreviation -> `expr`
1. Add GroundType and AggregateType
1. [US]IntValue -> [US]IntLiteral and creation of superclass Literal
1. CamelCase and match spec for PrimOps
  1. That being said, with reduction operations it could be confusing eg. `OrOp` vs `OrrOp`
  1. The above *should* be fixed with ScalaDoc comments
1. AST -> FIRRTLNode

### DIFF
--- a/src/main/scala/firrtl/IR.scala
+++ b/src/main/scala/firrtl/IR.scala
@@ -109,6 +109,52 @@ case class SIntLiteral(value: BigInt, width: Width) extends Literal {
 }
 case class DoPrim(op: PrimOp, args: Seq[Expression], consts: Seq[BigInt], tpe: Type) extends Expression
 
+// Mem Ports/Fields
+case class MemoryPortField(name: String, orientation: Orientation, tpe: Type) extends FIRRTLNode
+case class MemoryPortData(tpe: Type, orientation: Orientation) extends MemoryPortField("data", orientation, tpe)
+case class MemoryPortAddress(tpe: Type) extends MemoryPortField("addr", Default, tpe)
+case object MemoryPortEnable extends MemoryField("en", Default, UIntType(IntWidth(1)))
+case object MemoryPortClock extends MemoryField("clk", Default, ClockType)
+case class MemoryPortWriteMask(tpe: Type) extends MemoryField("mask", Default, tpe)
+case object MemoryPortWriteMode extends MemoryField("wmode", Default, UIntType(IntWidth(1)))
+
+// ********** Memory Port 1 **********
+abstract class MemoryPort extends FIRRTLNode {
+  val name: String
+  val data: MemoryPortData
+  val addr: MemoryPortAddress
+  val en: MemoryPortEnable
+  val clk: MemoryPortClock
+}
+case class Reader(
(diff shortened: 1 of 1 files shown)


jackkoenig:
The second commit in this PR address some of the discussion in #148 

I'm not sure that it addresses @aswaterman's concern https://github.com/ucb-bar/firrtl/pull/148#discussion_r61216334, but I think this is in the general direction of more explicit representation of memory ports in the IR.

Note that there are two proposals for Memory Ports as indicated by the comments.

jackkoenig:
The proposed Memory Port changes *could* address Andrew's concern if we changed the way references work. We could change references to contain an actual reference to the node being referred rather than the String name of it. We could also instead provide some sort of Circuit Analysis that provides this feature, and then use that instead of checking string names as in the issue Andrew was pointing out.

chick:
@jackkoenig I don't have any good ideas on the name expansion.  Personally I'd be ok with PrimitiveOperation, but I use an IDE that means I never type stuff like that.  I don't think that any rule you make regarding naming conventions will be without problems, your are living this code right now, just do what you think is right.

azidar:
> The proposed Memory Port changes could address Andrew's concern if we changed the way references work. We could change references to contain an actual reference to the node rather than its String name. We could also instead provide some sort of "Analysis" that provides this actual reference feature.

I'm not sure I understand - you mean WRef would replace its name field with a Stmt field?

jackkoenig:
> >The proposed Memory Port changes could address Andrew's concern if we changed the way references work. We could change references to contain an actual reference to the node rather than its String name. We could also instead provide some sort of "Analysis" that provides this actual reference feature.

> I'm not sure I understand - you mean WRef would replace its name field with a Stmt field?

Perhaps not a Stmt field since WRef could refer to Ports as well, could be an IsDeclaration or something like that.

azidar:
Yeah, I understand it is tempting to put the actual node there instead of a string, but this is very intentional in the design of the IR. Not only do we want the Firrtl graph to be a tree, we really want to prevent having stale data in the graph, and this would be a huge cause of headaches because we would essentially need to update these after every since pass.

jackkoenig:
Oh yeah that makes sense, I was more just referring to how we could better handle the issue Andrew brought up in #148 . I think providing a utility function (or analysis) that creates a mapping from WRef and the like to their declarations would work equally well

jackkoenig:
Replaced by #163

PR #163, opened 2016-05-10
Ir cleanup real

jackkoenig:
Replaces #160 

Few functionality changes, just major reorganization of the public API of the IR

I think this should wait to be merged until after #149 and #154, then I will clean things up again and rebase this. 

My previous proposal was more of a sketch but this is actual changes I think are reasonable. Please consider and comment so we can get this merged before Beta.

See commit messages for comments on the changes


azidar:
Why the move from trait to abstract class?

azidar:
Overall I'm happy with the changes. Lgtm.  @jackkoenig can you rebase and merge?

jackkoenig:
My thinking on the use of trait vs. abstract class revolves around the intent. I think abstract classes are useful for things where single inheritance is desired, while traits are useful for mixins. 

In the case of the IR, a FIRRTL vertex cannot be both an Expression and a Statement, therefore I think both of those should be abstract classes to make the code more self-documenting. Alternatively, a FIRRTL vertex can be both a Statement and a Declaration, or it can not be a Statement and yet still be a Declaration.

I don't suggest abstract class be used all over the place, just where it makes the intent more clear. What do you (and others) think?

jackkoenig:
Also if anyone has a better name for the parent of all FIRRTL vertices/nodes, I would greatly appreciate it. In the past it was AST, currently this PR has it as FIRRTLNode, so any better suggestion would be great. We could also standardize on Firrtl for naming so FirrtlNode might be less painful to look at.

jackkoenig:
Also, I put IR into it's own object, but I'm thinking that should actually either be a separate package or package object called `ir`, thoughts?

azidar:
I've been using Firrtl instead of FIRRTL, so I'd prefer to standardize on Firrtl.

Maybe Component? FirrtlComponent? Element? FirrtlElement? ASTComponent? ASTNode? I think FirrtlNode is actually ok, though.

I think all IR (not working IR) stuff should go into its own package, so people can reuse the IR parser and IR serializer. We should then include the IR package in our WIR, and be able to add our own serializing methods. I think this would be a good forcing function for clarifying the difference between IR and WIR.

azidar:
And I like your explanation of using abstract class over trait, I'm convinced.

jackkoenig:
I'm thinking FirrtlNode will be good, I'm working on updating the branch now, unfortunately (although not unexpected) there are a ton of conflicts

jackkoenig:
See #191

PR #164, opened 2016-05-10
Pad widths rebase

jackkoenig:
Slightly cleaned up git history for #149


jackkoenig:
Rocket chip currently is not passing, it fails on master as well though so I'm pretty sure it has nothing to do with these commits. I'm going to try to fix that before merging this.

PR #166, opened 2016-05-12
Implement file info

jackkoenig:
Replacement for #158 and related to ucb-bar/chisel3#186

This is a mostly fully functional implementation of file info/source locators in the FIRRTL compiler.
Because of difficulty in implementation, single line scopes are disallowed for everything except register resets (no single line scoping is used except register reset). This implementation requires that the source locators (if available) come at the end of the line, except registers where it must come after the reset definition (if there is one).

The Parser now supports 4 FileInfo options:    

1. IgnoreInfo - populate all nodes with NoInfo (even if info is provided in the firrtl code)    
1. UseInfo - populate nodes with any source locators provided in the firrtl code    
1. GenInfo - generate source locators from the firrtl code file itself    
1. AppendInfo - append GenInfo results to the end of parsed UseInfo    

The one big thing that needs some work is that the file info text itself is handled similarly to how StringLits are handled for Printf. This means that it doesn't really support escaped close bracket the way it should (instead supporting escaped double quotes because it thinks it's a double quoted string). Fixing this requires some refactoring of the StringLiteral handling that seems less important than getting this PR out asap.

### DIFF
--- a/src/main/antlr4/FIRRTL.g4
+++ b/src/main/antlr4/FIRRTL.g4
@@ -37,16 +37,16 @@ grammar FIRRTL;
 
 // Does there have to be at least one module?
 circuit
-  : 'circuit' id ':' '{' module* '}'
+  : 'circuit' id ':' info? '{' module* '}'
   ;
 
 module
-  : 'module' id ':' '{' port* block '}'
-  | 'extmodule' id ':' '{' port* '}'
+  : 'module' id ':' info? '{' port* block '}'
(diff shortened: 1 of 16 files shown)


jackkoenig:
I also need to write some tests :)

jackkoenig:
This is required for ucb-bar/chisel3#186 to pass tests, and we'd really like source locators asap so we should prioritize merging this

azidar:
lgtm.

jackkoenig:
Thanks! merging

PR #167, opened 2016-05-13
Restructured Compiler to use Transforms. Added an InlineInstance pass.

jackkoenig:
Transforms are new unit of modularity within the compiler.

Replacement for #154 

I created a new PR to rebase the old one on master and resolve any conflicts.
I also wanted to fixup the inline instances pass.

It turns out there is kind of a problem, it inlines a module, but then leaves it there so that in the emitted Verilog there is a module with no instantiations. This is problematic in that Verilator cannot tell what the top-level module is (and I imagine there is a similar problem in VCS). Users can, of course, specify the top-level module at the command line, maybe this is the right way to do it. We could also consider removing any modules that are no longer instantiated from the circuit, perhaps in the InlineInstances pass, perhaps in a later pass (like move Legalize to after it InlineInstances and do it there?), or perhaps in the Verilog Emitter. 

Thoughts?


jackkoenig:
I ran this inlining a random module in rocket-chip (Reorder Queue), and it worked, it passed all the assembly tests!

jackkoenig:
Added a commit that changes the testing to use --top-module when invoking Verilator. It seems that most people will easily be able to specify their top module (and can get it from the circuit if they want), so perhaps this is the right way to do it.

aswaterman:
cool stuff

Issue #155, opened 2016-05-03
Random memory initialization generates out of bounds address 


colinschmidt:
A read from a non-power of two memory can be initialized with an address that is out of the range of the memory. In vcs this results in an X output, potentially propagating throughout the rest of your design.

This showed up in one of the queues in the outer memory system of rocket-chip.

relevant firrtl looking like:
```
cmem ram : {client_xact_id : UInt<7>, addr_beat : UInt<2>, client_id : UInt<3>, is_builtin_type : UInt<1>, a_type : UInt<3>}[5]
...
infer mport T_326 = ram[T_274], clk
```
output verilog
```
assign ram_addr_beat_T_326_addr = T_274;
...
assign ram_addr_beat_T_326_data = ram_addr_beat[ram_addr_beat_T_326_addr];
...
initial begin
...
T_274 = {1{$random}};
```

aswaterman:
I guess it could be emitted as something like

    `ifdef SYNTHESIS
      assign rdata = mem[raddr];
    `else
      assign rdata = raddr >= n ? {1{$random}} : mem[raddr];
    `endif

jackkoenig:
Andrew and I talked a bit about this, we think the right way to deal with it is to create a temporary node that is assigned the read data if the address is within range, garbage otherwise (through a validIf construct). Then all references to the read data are replaced with this node. eg:
```
    node readOutData = mem.reader.data
```
->
```
    node GEN_N = validif(lt(mem.reader.addr, UInt(MEM_SIZE)), mem.reader.data)
    node readOutData = GEN_N
```

azidar:
If we do this strategy, then note that it is no longer legal to replace ValidIf(p, e) with e. I think this is ok, and the RemoveValidIf pass should replace ValidIf(p,e) with Mux(p,e,UInt(0)) (or something like that).

aswaterman:
Hmm, Mux(p,e,0) will add a whole bunch of hardware in some cases. Let's not
do that.

azidar:
Then perhaps we should not remove ValidIfs, and just emit them in Verilog like:

`node GEN_N = validif(lt(mem.reader.addr, UInt(MEM_SIZE)), mem.reader.data)`

becomes

```
`ifdef SYNTHESIS
  assign GEN_N = mem_reader_data;
`else
  assign GEN_N = mem.reader.addr < MEM_SIZE ? {1{$random}} : mem_reader_data;
`endif
```

[...]

PR #174, opened 2016-05-18
Fix bits type

jackkoenig:
This commit fixes a host of minor bugs and adds the appropriate tests to find them. These bug fixes are required in order for the compiler to be rerun after lowering (eg. for transformations that want to operate on low FIRRTL then rerun the compiler). 

The one thing I'm not super sure about is whether or not the constant propogation of pad widths I added to Legalize should be there. This is a linting warning so it's not clear that it is a required "legalization" but the warning it was exercising used to be an error in Verilator so it's probably a good idea to not exercise it. 

I also added the concept of "Debug Compilers" which extend the compilers defined in main by inserting check passes in between the passes already run. I think this concept could be extended in various ways to have the testing infrastructure run more checks than the regular compiler does (as we've talked about in the past).


jackkoenig:
This PR is required for my next PR (adding the substitute memories pass) which should fix #147 and #155

azidar:
I think the Legalization pass should be subsumed by the PadWidths pass and Constant propagation. The legalizeBits/legalizePad should be done during constant propagation, and legalize connect should be done during PadWidths.

The changes to account for SInt's are very good.

The Debug compilers is a good idea, I think there is more to flesh out there, but I'm ok accepting the change for now.

aswaterman:
I'm a little nervous about mandating constant propagation for correct code generation, but OTOH, having a separate legalization pass results in code duplication.  Maybe the former is the pragmatic thing to do.

jackkoenig:
I definitely think there is more to flesh out with the Debug Compilers, for instance it would be great if each pass had a `checkInvariants` function or something just to give us a little more confidence they're doing the right thing.

jackkoenig:
I think we might be able to do both. We could essentially combine the two into a single pass with "optimization" and "required" propagation functions. This pass accepts an argument to select which we're doing as it's constructor. Then the ConstantPropogation and Legalize passes are subclasses of this class, and the compiler chooses which to run based on command line options

azidar:
Another alternative is to add the required prop functions to PadWidths? PadWidths is essentially a legalization pass, in that it also removes Verilator warnings. Then again, some verilator warnings are also removed during constant propagation.....

jackkoenig:
Replaced by #296

PR #175, opened 2016-05-19
Update verilogToCpp in FirrtlSpec to match version in Chisel 3

jackkoenig:
In particular, PRINTF_COND is needed to guard assertions from firing early.

Honestly, I'm not sure how the tests were working before. A new test I'm writing doesn't get out of reset without this fix.


jackkoenig:
I have a better idea

PR #176, opened 2016-05-20
Add "test" library dependency on Chisel. Delete code copied from Chisel.

jackkoenig:
This will fail until ucb-bar/chisel3#196 is merged and published.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -18,6 +18,8 @@ libraryDependencies += "org.scalatest" % "scalatest_2.11" % "2.2.6" % "test"
 
 libraryDependencies += "org.scalacheck" %% "scalacheck" % "1.12.5" % "test"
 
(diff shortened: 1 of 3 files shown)


ucbjrl:
This introduces a dependency on chisel3. As we haven't yet published chisel3 3.0, this would require developers build and publish-local a version of chisel3. I realize this is only required for testing, so the circular dependency can be managed, but it does brings up a larger topic.

Chisel, firrtl, testers, and yet-to-be-developed tools, are really libraries. We should have a separate driver repo that "glues" them together into an application. This "driver" would be responsible for command line argument parsing, choosing backends, and configuring the various library components. Neither chisel nor firrtl should care about (or know) the consumer of their output.

azidar:
I'm going to close this issue until the "Driver" solution has been implemented.

PR #179, opened 2016-05-24
Fix bug in FIRRTL width inference, refactor associated functions

sdtwigg:
When folding over lists for MinWidth and MaxWidth, would assume 0 as a start
value. 0 persists through MinWidth resulting in under-constraining.

The functions were also refactored to be more readable and aligned with scala
style/best practices.


azidar:
lgtm.

PR #180, opened 2016-06-02
Suppress "match may not be exhaustive" warning

aswaterman:


aswaterman:
Seems like there might still be some nondeterminism in the tests.

jackkoenig:
I believe that the compiler is actually not thread safe. One of the tests will randomly give up and fail but that's much more rare than these threading issues 👎

PR #181, opened 2016-06-02
Ensure testsuites run in series.

ucbjrl:
The Parser is a singleton object with internal state.


jackkoenig:
I have a patch that I think solves the fundamental issue, trying to submit asap

PR #183, opened 2016-06-07
Fix non-thread safe Serialize by splitting it into class and object

### LINKED ISSUE #159: Threading issues
As the number of tests grows, thread unsafe behavior creeps out of the woodwork. I have no idea what is causing it but something is breaking parallel running of tests. I have to babysit Travis to restart it if it fails randomly, and locally I usually have to run `sbt test` followed by `sbt test-quick` to re-run anything that fails to make sure it isn't actually failing.

jackkoenig:
As far as I can tell this fixes the non-thread safe behavior we've been seeing.

It is kind of awkward that the BigInt serialize function is part of Serialize since this function is really for nodes of the FIRRTL AST. You can see in Checks.scala that calling it requires (new Serialize).serialize which is kind of awkward, perhaps this function should be replaced with a Utils function called prettyPrint or something?

Resolves #181 
Fixes #159 ?

### DIFF
--- a/src/main/scala/firrtl/Serialize.scala
+++ b/src/main/scala/firrtl/Serialize.scala
@@ -31,25 +31,29 @@ import firrtl.PrimOps._
 import firrtl.Utils._
 
 private object Serialize {
-
   def serialize(root: AST): String = {
+    lazy val ser = new Serialize
     root match {
-      case r: PrimOp => serialize(r)
-      case r: Expression => serialize(r)
-      case r: Stmt => serialize(r)
-      case r: Width => serialize(r)
-      case r: Flip => serialize(r)
-      case r: Field => serialize(r)
-      case r: Type => serialize(r)
-      case r: Direction => serialize(r)
-      case r: Port => serialize(r)
-      case r: Module => serialize(r)
-      case r: Circuit => serialize(r)
-      case r: StringLit => serialize(r)
+      case r: PrimOp => ser.serialize(r)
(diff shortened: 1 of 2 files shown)


azidar:
You could make an apply method of the Serialize object create a new Serialize, so its (new Serialize).serialize turns into Serialize().serialize?

I don't see how this fixes the thread-safe issue, maybe you can explain what was going wrong before?

jackkoenig:
Yeah I think adding an apply method is better. Having a companion object and a class makes me think we should rename Serialize to Serializer, what do you think?

So I wasn't sure what the problem was before, it was just clear that something wasn't thread safe. I basically blasted threads at the Translator thinking it was that but it wasnt. Then I tried the Parser, wasn't that either, then I added serialize to my test and it failed. If you look at the code for Serialize, it has a private var indentLevel. This is global so if two threads call in to Serialize at the same time, boom problem. This could possibly be solved by adding indentLevel as an explicit or implicit argument to the serialize functions, but this solution required the least changes.

I'm trying to write a concurrency test which could help prevent this in the future.

azidar:
Serializer works for me. Thanks for the explanation. I'd like to merge this, so if the concurrency test is going to take a while (>1 day) to write, I'd be ok adding it in a separate pull request.

jackkoenig:
That's fine, let me push a minor change to add the apply method, Serializer can wait since that might break stuff

jackkoenig:
K you can see the minor change, go ahead and merge after Travis is done

Issue #184, opened 2016-06-08
Width check regression after #179 merge


kammoh:
after `commit 1362c56c96e387f4657f639b03ba27e5160342d7` firrtl breaks during Width check:
```...
[INFO] Starting Infer Widths
[INFO] Finished Infer Widths
[INFO] Infer Widths took 21.8 ms

[INFO] Starting Width Check
[error] (run-main-0) firrtl.passes.PassExceptions: 
[error] firrtl.passes.CheckWidths$UninferredWidth:  : [module GCD]  Uninferred width.
[error] firrtl.passes.CheckWidths$UninferredWidth:  : [module GCD]  Uninferred width.
[error] firrtl.passes.PassException: 2 errors detected!
firrtl.passes.PassExceptions: 
firrtl.passes.CheckWidths$UninferredWidth:  : [module GCD]  Uninferred width.
firrtl.passes.CheckWidths$UninferredWidth:  : [module GCD]  Uninferred width.
firrtl.passes.PassException: 2 errors detected!
	at firrtl.passes.Errors.trigger(Passes.scala:61)
```
using GCD.ir from chisel-tutorial here.
everything seams to be ok in `e7ca89642342a10fc1780cc13b88c9bba91ae875`

PR #185, opened 2016-06-08
Fix for bug introduced in #174

azidar:
Addresses #184

Problem was that the width inferencer must assume a minimal width
for self-referencing widths contained in a MaxWidth. Otherwise, it
cannot solve the constraint.


aswaterman:
sounds reasonable to me

Issue #132, opened 2016-04-20
Recursive Instantiation of Modules allowed


jackkoenig:
Recursive instantiation of modules is allowed, this obviously doesn't make sense.
```
circuit Foo : 
  module Foo : 
    input a : UInt<32>
    output b : UInt<32>
    inst bar of Bar
    bar.a <= a
    b <= bar.b
  
  module Bar : 
    input a : UInt<32>
    output b : UInt<32>
    inst foo of Foo
    foo.a <= a
    b <= foo.b
```

Results in the Verilog:
```
module Foo(                                                                                                            
  input  [31:0] a,                                                                                                     
  output [31:0] b                                                                                                      
);                                                                                                                     
  wire [31:0] bar_a;                                                                                                   
  wire [31:0] bar_b;                                                                                                   
  Bar bar (                                                                                                            
    .a(bar_a),                                                                                                                                                                                                                                
    .b(bar_b)                                                                                                          
  );                                                                                                                   
  assign b = bar_b;                                                                                                    
  assign bar_a = a;                                                                                                    
endmodule                                                                                                              
module Bar(                                                                                                            
  input  [31:0] a,                                                                                                     
  output [31:0] b                                                                                                      
);                                                                                                                     
  wire [31:0] foo_a;                                                                                                   
  wire [31:0] foo_b;                                                                                                   
  Foo foo (                                                                                                            
    .a(foo_a),                                                                                                         
    .b(foo_b)                                                                                                          
  );                                                                                                                   
  assign b = foo_b;                                                                                                    
  assign foo_a = a;                                                                                                    
endmodule             
```

azidar:
Yup. This should get checked in CheckHighForm.

chick:
Working on it

jackkoenig:
Fixed by #201

PR #188, opened 2016-06-09
Recursive modules now detected

chick:
Fixes issue #132 , catches direct and indirect recursion
Recursive modules tests now work
Add a println for multiple errors accumulated during a test


.gitignore
```diff
@@ -51,3 +51,4 @@ lib_managed/
 src_managed/
 project/boot/
 project/plugins/project/
+.idea
```
azidar: What is this?

src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -61,6 +61,35 @@ object CheckHighForm extends Pass with LazyLogging {
   class BadPrintfException(x: Char) extends PassException(s"${sinfo}: [module ${mname}] Bad printf format: " + "\"%" + x + "\"")
   class BadPrintfTrailingException extends PassException(s"${sinfo}: [module ${mname}] Bad printf format: trailing " + "\"%\"")
   class BadPrintfIncorrectNumException extends PassException(s"${sinfo}: [module ${mname}] Bad printf format: incorrect number of arguments")
+  class InstanceLoop(loop: String) extends PassException(s"${sinfo}: [module ${mname}] Has instance loop $loop")
+
+  class ModuleGraph {
```
azidar: Can you comment this class and its functions, explaining what they do?

src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -61,6 +61,35 @@ object CheckHighForm extends Pass with LazyLogging {
...
+        case Some(children) =>
+          if(children.contains(parent)) {
+            parent :: path
+          }
+          else {
+            children.foreach { grandchild =>
+              val newPath = pathExists(grandchild, parent, grandchild :: path)
+              if(newPath.nonEmpty)
+                return newPath
+              }
```
azidar: fix indentation

src/main/scala/firrtl/passes/Passes.scala
```diff
@@ -57,6 +57,7 @@ class Errors {
     case 0 =>
     case 1 => throw errors.head
     case _ =>
+      println(errors.map { e => e.getMessage}.mkString("multiple errors", "\n", ""))
```
azidar: Remove this.

colinschmidt:
You should try to keep the first line of your commit messages under 50 characters or at least 80 so that github doesn't do this weird "..." stuff

.gitignore
chick: This suppresses IntelliJ's configuration file from git

src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -61,6 +61,60 @@ object CheckHighForm extends Pass with LazyLogging {
...
+  class InstanceLoop(loop: String) extends PassException(s"${sinfo}: [module ${mname}] Has instance loop $loop")
+
+  /**
+    * Maintains a one to many graph of each modules instantiated child module.
+    * This graph can be searched for a path from a child module back to one of
+    * it's parents.  If one is found a recursive loop has happened
+    * The graph is a map between the name of a node to set of names of that nodes children
+    */
+  class ModuleGraph {
+    val nodes = new HashMap[String, HashSet[String]]
```
jackkoenig: It is preferable to not import mutable datastructures by name, instead importing scala.collection.mutable and then instantiate with `new mutable.HashMap[...]`

We have not followed this guideline in the past but I strongly believe we should enforce it in the future.

chick:
I got an intermittent error during a test run and it occurred in code below in line with FIRRTLStringLitHandler, which does inherit from a object with an member
```
    val byteArray = raw.getBytes(java.nio.charset.StandardCharsets.UTF_8)
    StringLit(unescape(byteArray, Array()))
```
I wonder if there could be some race condition intializing this thing

```
  property(s"Invalid escape characters should fail") {
    forAll(validCharSeq, invalidEsc, validCharSeq) {
      (head: Seq[Char], badEsc: Char, tail: Seq[Char]) =>
        val str = (head ++ Seq('\\', badEsc) ++ tail).mkString
        intercept[InvalidEscapeCharException] {
          firrtl.FIRRTLStringLitHandler.unescape(str)
        }
    }
  }
```

jackkoenig:
I believe the problem is actually in

```scala
  // Generators for random testing
  val validChar = for (c <- Gen.choose(0x20.toChar, 0x7e.toChar) if c != '\\') yield c
  val validCharSeq = Gen.containerOf[Seq, Char](validChar)
  val invalidChar = Gen.oneOf(Gen.choose(0x00.toChar, 0x1f.toChar),
                             Gen.choose(0x7f.toChar, 0xff.toChar))
  val invalidEsc = for (
    c <- Gen.choose(0x00.toChar, 0xff.toChar
  ) if (!validEsc.contains(c))) yield c
```

Everywhere you see an `if` in a scalacheck `Gen`, this means that it will generate a value then if it fails the if condition it will try again. If it fails more than a set number of times, scalacheck gives up and marks the test as a failure. This problem could be fixed by either increasing the number of times it tries to get a valid random value, or by fixing these generators to not use conditions like that.

chick:
retest this please

jackkoenig:
For Travis you have to rerun it through the web interface

chick:
@azidar @jackkoenig ,This appears to be failing running Chisel3 tests here.  Any idea what is going on here.  I think it unlikely that my changes are causing this problem
```
- /tmp/FailingAssertTester3792273207556073012/FailingAssertTester.v:55: Second verilog $finish, exiting
Assertion failed
    at Assert.scala:11 assert(Bool(false))
1
Assertion failed
    at Assert.scala:11 assert(Bool(false))
1
```

azidar:
It looks like travis is complaining about "Instance loops a -> b -> a" is failing? I know that some of the Chisel tests are supposed to fail, so although verilator prints it fails, it actually succeeds.

chick:
Too far behind, resubmitting in another branch

PR #189, opened 2016-06-09
Initializes register addresses.

### LINKED ISSUE #187: SeqMem address register not initialized
We're running into an annoying Chisel issue that's breaking one of the ground tests in VCS. Basically, the problem is that the address register SeqMem (GEN_48 in the following screenshot) is not initialized. This causes one of the memory cells (seq_plru_state[7]) to get overwritten with Xs during reset.

seqmem-bug-report

azidar:
Fixes #187.


aswaterman:
lgtm thanks

PR #190, opened 2016-06-10
Add test to check compiler is thread safe

jackkoenig:
Here's a test that just spams multiple instances of each of the main compilers to check for threadsafety. As far as I can tell, it always fails without the fix to Serialize that was merged earlier this week.

For any Scala experts, I tried to only have one Await at the end, but for the life of me I could not figure out how to turn Seq[Future[Seq[Future]]] to Future[Seq]. Is this okay?


azidar:
It looks good to me, do you want to wait to merge once you get comments on the Future[Seq] question?

jackkoenig:
I don't think it really matters, it would only be marginally more elegant.

PR #191, opened 2016-06-10
Ir cleanup fix

jackkoenig:
Replaces #163 

I tried very hard to put this in the same previous PR but Github was just not having it. I pushed the exact same history to that branch and this one and yet they look completely different, I have no clue why.

This is a big redo of the Firrtl IR. I could merge it all into one commit if y'all think that's the right thing to do.


jackkoenig:
Please look over before merging but it should be good

azidar:
Looked it over and seems good. One minor thing, can we change BoolType() -> BoolType? After that, I'll merge.

jackkoenig:
I tried to rebase it into the Type commit but this whole thing has been a nightmare so I think I'm done fighting with it 💯

Issue #157, opened 2016-05-03
Write Check Chirrtl pass


azidar:
To fix #121, we need to check CHIRRTL code is legitimate (similar structure to CheckHighForm). Things to check:
- No duplicate references
- Invalid connection to non-assignable expression (e.g. add(a,b) <= c)
- Negative unsigned integer
- Undeclared references
- Memories with flips
- Incorrect access/index to a non-referenceable expression
- All modules are defined
- Correct arguments/constants for primops
- No negative/zero widths (zero width restriction can be relaxed)
- No negative/zero sized vectors/memories
- Bad printf form

I think this can wait until after the restructuring of chisel3 which would use CHIRRTL as its AST instead of its pseudo-FIRRTL IR.

PR #198, opened 2016-07-21
Added a Chirrtl check for undeclared wires, etc.

azidar:
As Chisel3 is prone to emitting invalid Firrtl, this does a quick pre-check of basic correctness to ensure the Chirrtl passes do not screw up badly.


azidar:
Fixes #157

PR #208, opened 2016-07-28
remove ConstProp in HighFirrtlToMiddleFirrtl

donggyukim:
ConstProp before width padding causes errors for SIntLiteral. Any reason to add it here?

### DIFF
--- a/src/main/scala/firrtl/LoweringCompilers.scala
+++ b/src/main/scala/firrtl/LoweringCompilers.scala
@@ -108,7 +108,6 @@ class HighFirrtlToMiddleFirrtl () extends Transform with SimpleRun {
       passes.RemoveAccesses,
       passes.ExpandWhens,
       passes.CheckInitialization,
-      passes.ConstProp,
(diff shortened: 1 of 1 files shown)


azidar:
I think its fine to remove, but I'd also ping @aswaterman

aswaterman:
I have no idea what the implications of this change are, but I had expected this to be a low-FIRRTL pass, anyway.

Is there actually a bug in ConstProp that I need to fix, though?

donggyukim:
I'm not even sure whether there's a bug in ConstProp. I thought it's an issue of pass ordering.

aswaterman:
I see. We should merge this after you figure out why the build failed.

donggyukim:
I think this is due to #159

jackkoenig:
Looking at the Travis run, it seems that StringSpec was the only thing failing. That fails kind of randomly but #212 should fix that. Until #212 is merged the general strategy is to rerun Travis until it passes :)

PR #212, opened 2016-08-01
Fix StringSpec generators to only choose from valid values.

jackkoenig:
The old almost equivalent syntax gives the same result but can cause the
test to fail if too many invalid values are thrown away.

Fixes the issue where ScalaCheck sometimes gives up on random generation and causes tests to fail


aswaterman:
This looks fine to me... the failure seems to be unrelated to the test, though?

jackkoenig:
Travis won't load on my browser so I can't seem to see what the issue is (so I'm rerunning it).

jackkoenig:
Looks fine now

PR #217, opened 2016-08-03
Exposed the evaluate function in InferWidths

azidar:
This allows other passes to make use of the width constraint solver.


jackkoenig:
What other passes need to use it?

azidar:
I'm writing a fixed point library on top of Firrtl, and want to reuse the constraint solver. It won't go in mainline Firrtl.

azidar:
Turns out I don't need this anymore. I'm going to close this PR.

PR #218, opened 2016-08-03
Added RemoveEmpty.scala, which removes Empty and nested Blocks

azidar:
A utility pass other Firrtl compilers can use. Also makes testing easier because you can remove Empty statements that may be introduced via your compiler pass (but you don't want to have them in the check string).

### DIFF
--- a/src/main/scala/firrtl/passes/RemoveEmpty.scala
+++ b/src/main/scala/firrtl/passes/RemoveEmpty.scala
@@ -8,24 +8,8 @@ import firrtl.ir._
 object RemoveEmpty extends Pass {
   def name = "Remove Empty Statements"
   private def onModule(m: DefModule): DefModule = {
-    def onScope(s: Statement): Statement = {
-      val newStmts = mutable.ArrayBuffer[Statement]()
-      def onStmt(s: Statement): Unit = s match {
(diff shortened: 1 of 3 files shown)


jackkoenig:
I have a utility written for this in ExpandWhens. We should remove it from there and put some equivalent implementation in Utils that should then be used by this pass. I think mine is neat but whatever.

https://github.com/ucb-bar/firrtl/blob/master/src/main/scala/firrtl/passes/ExpandWhens.scala#L74

src/main/scala/firrtl/passes/RemoveEmpty.scala
```diff
@@ -0,0 +1,35 @@
...
+          Block(newStmts.toSeq)
+        case s => s map onScope
+      }
+    }
+    m match {
+      case m: Module => Module(m.info, m.name, m.ports, onScope(m.body))
+      case m: ExtModule => m
+    }
+  }
+  def run(c: Circuit): Circuit = Circuit(c.info, c.modules.map(onModule _), c.main)
```
jackkoenig: Prefer infix notation ie.
```scala
c.modules map onModule
```

azidar:
Ah ok, I'll move it to utils instead.

PR #223, opened 2016-08-08
Don't create output files until the compiler succeeds

aswaterman:
Creating the output file preemptively screws up make, as on
subsequent executions of make, it thinks the task succeeded.


palmer-dabbelt:
This fixes my problem.  Thanks!

PR #225, opened 2016-08-09
Move verilog_rename_n() into VerilogRename object …

ucbjrl:
… so it's available outside firrtl (chisel-testers).


aswaterman:
How is the tester going to depend on this?  If it's using it to attempt to mangle names the same way that FIRRTL does, it's not sufficient, because of name conflicts when flattening aggregates.

ucbjrl:
I believe this is applied after any name mangling. It's replacing existing duplicate code in PeekPokeTesterUtils.validName().

azidar:
Yeah, @aswaterman is correct, @ucbjrl you will also need to run the Uniquify pass as well, I believe. Perhaps this will be done later (as this is a simple first step to get 95% of the way there).

ucbjrl:
Sure, all this is well and good. My point is duplicating existing firrtl code in testers is bad. There is coupling and it's better to acknowledge it than hide it by duplicating code. We do need a better API, but until it's designed, let's at least acknowledge the dependency.

jackkoenig:
Tester is going to use annotations. No longer necessary.

PR #236, opened 2016-08-15
Change RW port names

shunshou:
* Updated FIRRTL spec + related code for readwrite ports.
    (write) data -> wdata & mask -> wmask for clarity

* Also removed simple.fir that snuck into master branch.

### DIFF
--- a/spec/spec.tex
+++ b/spec/spec.tex
@@ -827,10 +827,10 @@ If the \verb|en| field is high, then the non-masked portion of the \verb|data| f
 \subsubsection{Readwrite Ports}
 Finally, the readwrite ports have type:
 \begin{verbatim}
-{wmode:UInt<1>, flip rdata:T, data:T, mask:M, 
+{wmode:UInt<1>, flip rdata:T, wdata:T, wmask:M, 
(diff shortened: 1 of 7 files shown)


azidar:
data -> wdata and mask -> wmask does make more sense insofar as clarity is concerned. This is a backwards incompatible change, but since Chisel3 has only recently been able to generate read-write ports, perhaps this is ok? any thoughts @aswaterman @jackkoenig

aswaterman:
This is more consistent and IMO is worth the backwards incompatibility at this juncture.  I assume @shunshou is making the corresponding changes to Chisel.

shunshou:
From previous discussions (re: inferReadWrite), I think Chisel is agnostic to explicit readwrite ports. I just changed the InferReadWrite FIRRTL pass to be consistent with the spec change.

shunshou:
@jackkoenig any opinions? It'd be great if a decision could be made either way, as that affects how I'd recode my pass. I assume the sooner the better, as the backwards incompatibility becomes more of a problem the longer people have to use the InferReadWrite pass.

jackkoenig:
I agree with this change and agree that we should make backwards incompatible changes sooner rather than later, I will merge

Issue #222, opened 2016-08-08
Java OutOfMemoryErrors... during Infer Widths?


ccelio:
Something in the last 1-2 weeks seems to have broke FIRRTL with regards to java GCC memory limit in using BOOM - I believe this commit (76684cbeb8b5d46138185432bc323a5de2eb7f59).

It's the oldest commit along the mainline (76684cbeb8b5d46138185432bc323a5de2eb7f59) I can replicate the error with, and thankfully it also still prints out the compiler passes so I can see where it is.  Each pass was really really really slow: 38 seconds for expand-whens for example. It usually takes 2.5 minutes to get from "make" to a compiled emulator for BOOM.  But this one crashes out at 7-8 minutes during InferWidths.

Trying to go to older FIRRTLs just gives me gives lower types errors... thanks to trying to update to latest rocket-chip. But that's not why I'm here. 

**Here's the failing output:**

```
[info] [24.422] Done elaborating.
[success] Total time: 28 s, completed Aug 8, 2016 1:45:50 AM
mv /scratch/celio/boom-chip/vsim/generated-src/Top.fir /scratch/celio/boom-chip/vsim/generated-src/Top.BOOML1Config.fir
mkdir -p ./output
ln -fs /nscratch/celio/install/riscv/riscv64-unknown-elf/share/riscv-tests/isa/rv64ui-p-simple output/rv64ui-p-simple
mkdir -p /scratch/celio/boom-chip/vsim/generated-src/
echo "\`ifndef CONST_VH" > /scratch/celio/boom-chip/vsim/generated-src/consts.BOOML1Config.vh
[...]

PR #238, opened 2016-08-15
Changed Chirrtl pass for write mask

shunshou:
* If the memory datatype is not a ground type, write mask is initialized to 1.
    Chisel only allows write masks that are Vecs of Bools IF the Chisel memory
    type is a Vec of x. To reduce the # of extraneous nodes, the generated
    Firrtl/Verilog (from Chirrtl) should have a constant 1 connected to the mask.
* Changed ExpandWhens s.t. if a Mux is to be generated where the true value ==
    false value, skip generating the Mux and propagate the tv directly. (Simplifies
    the node graph earlier on when Chisel designs are parameterized).

### DIFF
--- a/src/main/scala/firrtl/passes/ExpandWhens.scala
+++ b/src/main/scala/firrtl/passes/ExpandWhens.scala
@@ -133,7 +133,11 @@ object ExpandWhens extends Pass {
                     case (WInvalid(), WInvalid()) => WInvalid()
                     case (WInvalid(), fv) => ValidIf(NOT(s.pred), fv, tpe(fv))
                     case (tv, WInvalid()) => ValidIf(s.pred, tv, tpe(tv))
-                    case (tv, fv) => Mux(s.pred, tv, fv, mux_type_and_widths(tv, fv))
+                    case (tv, fv) => {
+                      // parameterization might result in the same tv, fv -- get rid of unnecessary mux
+                      if (tv != fv) Mux(s.pred, tv, fv, mux_type_and_widths(tv, fv))
+                      else tv
+                    }
                   }
                 case None =>
                   // Since not in netlist, lvalue must be declared in EXACTLY one of conseq or alt
--- a/src/main/scala/firrtl/passes/Passes.scala
+++ b/src/main/scala/firrtl/passes/Passes.scala
@@ -1131,11 +1131,18 @@ object RemoveCHIRRTL extends Pass {
                      }}
                   def set_write (vec:Seq[MPort],data:String,mask:String) : Unit = {
                      val tmask = create_mask(s.tpe)
+                     // mem data type = ground type implies no masking used (per Chisel spec)
+                     val maskInitVal = tmask match {
+                        case BoolType => one
+                        case _ => zero
+                     }
                      for (r <- vec ) {
                         stmts += IsInvalid(s.info,SubField(SubField(Reference(s.name,ut),r.name,ut),data,tdata))
                         for (x <- create_exps(SubField(SubField(Reference(s.name,ut),r.name,ut),mask,tmask)) ) {
-                           stmts += Connect(s.info,x,zero)
-                        }}}
+                           stmts += Connect(s.info,x,maskInitVal)
+                        }
+                     }
+                  }
                   val rds = (hash.getOrElse(s.name,EMPs())).readers
                   set_poison(rds,"addr")
                   set_enable(rds,"en")


src/main/scala/firrtl/passes/ExpandWhens.scala
```diff
@@ -133,7 +133,11 @@ object ExpandWhens extends Pass {
                     case (WInvalid(), WInvalid()) => WInvalid()
                     case (WInvalid(), fv) => ValidIf(NOT(s.pred), fv, tpe(fv))
                     case (tv, WInvalid()) => ValidIf(s.pred, tv, tpe(tv))
-                    case (tv, fv) => Mux(s.pred, tv, fv, mux_type_and_widths(tv, fv))
+                    case (tv, fv) => {
+                      // parameterization might result in the same tv, fv -- get rid of unnecessary mux
+                      if (tv != fv) Mux(s.pred, tv, fv, mux_type_and_widths(tv, fv))
+                      else tv
```
jackkoenig: Since this is a constant propogation, I'm not sure that we should do this here. We generally want to avoid distributing similar/related functionality across passes.

shunshou: If basic constant propagation shouldn't be done here, then there definitely needs to be some utility mechanism to traverse the graph and figure out if 2 nodes have the same origin. I personally dislike the notion of always having to wait until the end to do constant propagation, because that makes intermediate FIRRTL code harder to decipher and potentially harder to do anything with. (If we want some conditional optimization). 

Also, by using the above + a ConstProp pass leads to:

```
assign mem_T_5_mask = 1'h1;

always @(posedge clk) begin
    GEN_1 <= mem_T_3_addr;
    GEN_3 <= mem_T_3_en;
    if(mem_T_5_en & mem_T_5_mask) begin
      mem[mem_T_5_addr] <= mem_T_5_data;
    end
  end
```

Which is considerably less redundant (given the fact that no mask is needed) than:

```
assign mem_T_5_mask = T_4;
assign mem_T_5_en = T_4;
assign T_4 = io_en & io_wen;

always @(posedge clk) begin
    GEN_1 <= mem_T_3_addr;
    GEN_3 <= mem_T_3_en;
    if(mem_T_5_en & mem_T_5_mask) begin
      mem[mem_T_5_addr] <= mem_T_5_data;
    end
  end
```

Which current FIRRTL generates. -- It seems like the ConstProp or other logic redunction pass didn't really do its job. While you could say that this optimization should be left for the synthesis tools, Verilog readibility + code line reduction is obviously still quite valuable to the hardware designer, so wherever possible, my opinion is you should try to sneak in optimizations that make sense. 

It's easier to catch areas for optimization tailored to specific points in the compilation process than it is to to create one big catch-all at the end...

jackkoenig: I think that or a similar utility would be useful no matter what.

There's no reason we have to wait until the end to do constant propagation. The pass could (and should) be expanded to work on mid firrtl and then could be run following Expand Whens. Better yet, it could be run in some restricted form on high firrtl (the proposed propagation in expand whens would be very difficult to do on high firrtl). However, I believe the crux of the reason why you're getting the bad QOR as in your example above is that Constant Propagation isn't doing its job right. And although I don't think it is causing this particular issue (although it might be), we should consider rerunning ConstProp, CSE, and DCE (perhaps in a loop) since they can create work for the other to do. I believe this is a common compiler pattern.

The reason why I am so hesitant to distribute constant propagation functionality throughout is that it violates separation of concerns. I think that Expand Whens is probably already doing too much work and I would love to reduce it's functionality if possible. Furthermore, Constant Propagation is not required for correctness, and thus should be easy to turn off. Distributing it throughout other passes makes that very difficult.

I totally agree with you about readability and not treating this as one big catch-all (although the one big catch-all should be able to do just that, and it's currently not!)

azidar: I agree with @jackkoenig that we should consider rerunning ConstProp, CSE, and DCE. We should also remember that these should never be required for correctness.

> If basic constant propagation shouldn't be done here, then there definitely needs to be some utility mechanism to traverse the graph and figure out if 2 nodes have the same origin.

Perhaps, but you shouldn't care whether the wmask/enable are equivalent, because your code should work regardless. Simple passes are better passes. We should really just take a look at improving ConstProp so all cases of the above are caught, not just for your use case of wmasks/enables.

> I personally dislike the notion of always having to wait until the end to do constant propagation, because that makes intermediate FIRRTL code harder to decipher and potentially harder to do anything with. (If we want some conditional optimization).

If your pass is too hard to look at, then you should run the ConstProp after your pass when debugging. Don't make ConstProp a necessary thing to run because other users may not want to have any ConstProp done on their code, but still want to use your pass (like @jackkoenig  said, separation of concerns)

> Verilog readibility + code line reduction is obviously still quite valuable to the hardware designer...

Absolutely! And if it isn't readable, we should do a better job.

> so wherever possible, my opinion is you should try to sneak in optimizations that make sense.

I totally disagree. Imagine trying to debug some code where some random pass is doing constant propagation, and I have no way of knowing which one. Sneaking in optimizations is an anti-pattern. I want to be able to turn on/off ALL ConstProp etc. easily, and if these are snuck in everywhere, that is impossible.

W.r.t. this change, I can see this specific optimization in ExpandWhens to _maybe_ be ok because, in a different light, its essentially redefining the behavior of ExpandWhens to be 'insert a mux if the values are different, otherwise return the value'. However, I agree with Jack that we definitely want a separation of concerns, and in general we should NOT add constprop/other optimizations in compiler passes that aren't specifically dedicated to that purpose. My vote is it leaves a sour taste in my mouth but I'm ok merging it. However, I'll leave this decision up to @jackkoenig ;)

aswaterman:
Also, running the optimization passes earlier will reduce compiler runtime.

shunshou:
I'm not married to these changes, so I'm ok either way. 

> Perhaps, but you shouldn't care whether the wmask/enable are equivalent, because your code should work regardless. Simple passes are better passes. We should really just take a look at improving ConstProp so all cases of the above are caught, not just for your use case of wmasks/enables.

I do care that write mask and enable are equivalent. There's a big difference between having code that is functionally correct and code that produces enhanced QoR. If you can reduce logic levels knowing that mask and enable are equivalent, or in the case of SRAM macros, eliminate peripheral circuitry that is baked in, you should always do that to get faster clock periods, lower area, etc. 

One of the things that's irked me in the past re: Chisel is that, while it might produce functionally correct code, a lot of things are just suboptimal by the nature of their construction. Take, for example, having multi-clock domain support be a low priority up until this point, and not supporting separate read/write clocks. If you're pushing hardware boundaries, you really need to be able to customize all of these things. The problem is that I think the interesting compiler/programming language problem is different and slightly orthogonal to the task of designing *good* hardware. 

> If your pass is too hard to look at, then you should run the ConstProp after your pass when debugging. Don't make ConstProp a necessary thing to run because other users may not want to have any ConstProp done on their code, but still want to use your pass (like @jackkoenig said, separation of concerns)

It's not so much that my code is "too hard to look at" that I *need* constant propagation. Yes, I'd like to make my generated FIRRTL as simple as possible to read. Maybe constant propagation isn't the best thing to really look into for my pass. As Rimas pointed out for a very specific instantiation of Rocket-chip, conditioning whether to remove or keep write mask only on whether the memory data type is of type Vec is insufficient to optimize all cases (i.e. when Vec size = 1, you never expect *Chisel* to generate a write mask, although FIRRTL doesn't know the difference, and thinks write mask is needed). 

> I totally disagree. Imagine trying to debug some code where some random pass is doing constant propagation, and I have no way of knowing which one. Sneaking in optimizations is an anti-pattern. I want to be able to turn on/off ALL ConstProp etc. easily, and if these are snuck in everywhere, that is impossible.

I'm coming from a 99% hardware background. In that sense, I would hope the tools I use (Chisel,FIRRTL, what have you) have been thoroughly tested, so that, in the ideal world, I wouldn't need to debug anything at the FIRRTL level. It's already enough of a headache to make sure that my logic is correct (oops, I inserted an extra register here when I didn't need it kind of thing). To have to add debugging FIRRTL passes into the grand scheme of things just seems like a huge deterrent for adoption. I just want a tool that will transform my Chisel into functionally correct, but also *optimized* RTL as long as I provide the right initial expressions. I really hope to take for granted that constant propagation is doing the right thing, and just focus on my own hardware problems. 

The way Chisel has been and kind of currently still is is that you basically need to be a Chisel developer to be able to write more interesting RTL with it. That really shouldn't be the case. 

> Also, running the optimization passes earlier will reduce compiler runtime.

Also something I wanted to mention yesterday but didn't. Although... I guess even with rocket-chip the compiler runtime is short enough where it's less of a big deal.

kammoh:
Sorry to just jump in the middle of the discussion... as a chisel/firrtl user, I pretty much feel @shunshou's concerns regarding juicing performance out of designs and would totally prefer getting better synthesis results to "better" coding inside the running tools, as long as the tools keep running ok. But I understand that could turn out iffy in the long run :)
Well, I was thinking if it would be possible to solve this with providing access to some sort of meta-data (about nodes dependency, equivalency, etc) from earlier passes or from chisel. Something like augmenting the SourceInfo with a list/stack of meta information which keeps being updated along the passes. Of course devising a general (yet manageable) scheme for this may turn out non-trivial. my 2¢ :)

azidar:
I agree that improved QOR is the end-all goal. I see two paths going forward:

Option 1: We sneak in little optimizations everywhere. QOR is good. Writing new passes is hard because you never know when some other pass may drastically change your expected input.

Option 2: We improve the constant propagation pass. QOR is good. Writing new passes is easy because all code optimizations are contained, known, and easy to understand.

Option 3: We do neither. QOR is bad. Writing new passes is easy.

Did I summarize all the options properly? If so, which option do you prefer? To me, Option 2 is clearly superior.

azidar:
@kammoh Thanks for your 2cents :) 

> as long as the tools keep running ok. 

You hit the nail on the head. Chisel2 had many problems because it wasn't design systematically. It is possible to get great QOR as well as a well designed system, but to do that we need to hold ourselves accountable and do things the "right" way, even when it is more annoying to do so.
All in all, I hope we all agree that our goal is getting good QOR while employing good software engineering, so the tools always "keep running ok" :)

Your idea about modifying SourceInfo is a good one, and we've thought about this type of problem. Our current solution is to use the annotation system (see Annotations.scala) to pass information along side the Firrtl passes. This separation of systems was to address the problem of a deleted node that shouldn't have been deleted (and there is no way of checking this fact, as the corresponding annotation is deleted as well). By keep the annotations separate from the circuit, we can check their consistency and error if something went wrong. The annotation system is not complete yet and is in its early stages of usages, so feedback is more than welcome :)

donggyukim:
We can resolve this issue if we can call ConstProp right after ExpandWhens? I think some passes between ExpandWhens and ConstProp prevent proper optimization. There's some issues calling ConstProp early, but may be able to figure out why.

shunshou:
Re Option 1: I don't understand the problem. You should never be considering an "expected input" to your pass. i.e. if the condition in which your pass is supposed to do an optimization exists, then your pass should do its thing. If the condition doesn't exist b/c it's already been optimized by a previous pass, then don't do anything. Your passes should never make any assumptions on what has/has not been previously done. 

This is unfortunately not the way I've constructed the write enable/write mask optimization, but the difficulty there doesn't really lie with FIRRTL in and of itself. It's more because of the wonky-ness of Chisel -> FIRRTL translation. The Chisel output has more specificity (in theory) than the FIRRTL output, and then the pass generalizes this back to what FIRRTL expects... so you lose some opportunities for optimizing without having to go through the tedious process of re-extracting the specificity out down the line (when your pass is inserted). 

My personal feeling is that the Chirrtl emitted by Chisel should be changed in such a way that Firrtl can still interpret it correctly + it doesn't result in loss of specificity, but that requires quite a bit of rethinking, and I'm not sure if people like that option.

----

Re: Option 2, I have no qualms, under the assumption that ConstProp or similar be allowed to be included in intermediate stages (i.e. before my pass at Middle to Low Firrtl) or some utility function should be created to analyze the circuit without doing the actual propagation, at the very least. I personally think you should be allowed to throw ConstProp (everything you could possibly want in it) everywhere in that case. It should not affect the correctness of a circuit, and if your pass is a good pass, your pass shouldn't be broken by doing pre ConstProp (unless you have a more specific non-debug related example to suggest that doing ConstProp earlier is legitimately a bad idea).

azidar:
@donggyukim yeah I was going to look into making ConstProp, DCE and sub expression elimination work on high firrtl. then you can include this whenever you want. also, it should fix the problem that @ccelio is seeing in #222.

azidar:
@shunshou 

Re: Option 2 will improve the various optimization passes to work on any Firrtl (high, mid, low). In a 'good QOR compiler', we should do these optimizations early and often. In a 'debugging compiler', we should be able to have zero optimizations (and still a correct output). Sounds like we agree, can we move forward with Option 2?

donggyukim:
@azidar Ok, I'll take a look at it.

shunshou:
@azidar Sure. Works for me. Thanks!

jackkoenig:
Will be resolved with #246

PR #242, opened 2016-08-17
emit correct enable signals for memories

### LINKED ISSUE #178: Enable signal is not attached to memory read ports in verilog
The enable signals are not attached to the memory read ports in verilog. Registers are generated for the signals, but they are not used.

donggyukim:
This fixes #178.


azidar:
Can you clarify exactly what was wrong before, and how you fixed it (for posterity)?

donggyukim:
Read Port/RW Port enables are not properly handled in emitting verilog. There are pipes generated for enables, but not used at all, which shows unexpected behaviors in some cases(e.g. FAME1 transforms). This PR handles them properly at least for SeqMems mapped to black RAMs on FPGA.

azidar:
lgtm. update the branch and I'll merge.

Issue #165, opened 2016-05-11
smems of bundle types are split into multiple memories


donggyukim:
smems of bundle types are split into multiple memories from chirrtl to firrtl. I think this is not desirable. For example:
`smem metaMem : {dirty : UInt<1>, tag : UInt<20>}[256]`
becomes two memories:
`mem metaMem_dirty:`
`  data-type => UInt<1>`
`mem metaMem_tag:`
`  data-type => UInt<20>`

PR #244, opened 2016-08-17
Add MemUtils to aid in interfacing with alternate memory implementations

albert-magyar:
@shunshou @azidar 

These utilities were handy in writing passes for interfacing with alternate memory implementations.

I'm still wondering how much of the WIR information can be filled out without losing generality of these utilities for other purposes. This is the most basic and robust version that relies on resolving and checking happening downstream of the pass.

### DIFF
--- a/src/main/scala/firrtl/passes/MemUtils.scala
+++ b/src/main/scala/firrtl/passes/MemUtils.scala
@@ -1,0 +1,161 @@
+/*
+ Copyright (c) 2014 - 2016 The Regents of the University of
+ California (Regents). All Rights Reserved.  Redistribution and use in
+ source and binary forms, with or without modification, are permitted
+ provided that the following conditions are met:
+ * Redistributions of source code must retain the above
+ copyright notice, this list of conditions and the following
+ two paragraphs of disclaimer.
+ * Redistributions in binary form must reproduce the above
+ copyright notice, this list of conditions and the following
+ two paragraphs of disclaimer in the documentation and/or other materials
+ provided with the distribution.
+ * Neither the name of the Regents nor the names of its contributors
+ may be used to endorse or promote products derived from this
+ software without specific prior written permission.
+ IN NO EVENT SHALL REGENTS BE LIABLE TO ANY PARTY FOR DIRECT, INDIRECT,
+ SPECIAL, INCIDENTAL, OR CONSEQUENTIAL DAMAGES, INCLUDING LOST PROFITS,
+ ARISING OUT OF THE USE OF THIS SOFTWARE AND ITS DOCUMENTATION, EVEN IF
+ REGENTS HAS BEEN ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ REGENTS SPECIFICALLY DISCLAIMS ANY WARRANTIES, INCLUDING, BUT NOT
+ LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ A PARTICULAR PURPOSE. THE SOFTWARE AND ACCOMPANYING DOCUMENTATION, IF
+ ANY, PROVIDED HEREUNDER IS PROVIDED "AS IS". REGENTS HAS NO OBLIGATION
+ TO PROVIDE MAINTENANCE, SUPPORT, UPDATES, ENHANCEMENTS, OR
+ MODIFICATIONS.
+ */
+
+package firrtl.passes
+
+import com.typesafe.scalalogging.LazyLogging
+
+import firrtl._
+import firrtl.ir._
+import firrtl.Utils._
+import firrtl.PrimOps._
+
+object seqCat {
+  def apply(args: Seq[Expression]): Expression = args.length match {
+    case 0 => error("Empty Seq passed to seqcat")
+    case 1 => args(0)
+    case 2 => DoPrim(PrimOps.Cat, args, Seq.empty[BigInt], UIntType(UnknownWidth))
+    case _ => {
+      val seqs = args.splitAt(args.length/2)
+      DoPrim(PrimOps.Cat, Seq(seqCat(seqs._1),seqCat(seqs._2)), Seq.empty[BigInt], UIntType(UnknownWidth))
+    }
+  }
+}
+
+object toBits {
+  def apply(e: Expression): Expression = e match {
+    case ex: WRef => hiercat(ex,ex.tpe)
+    case ex: WSubField => hiercat(ex,ex.tpe)
+    case ex: WSubIndex => hiercat(ex,ex.tpe)
+    case t => error("Invalid operand expression for toBits!")
+  }
+  def hiercat(e: Expression, dt: Type): Expression = dt match {
+    case t:VectorType => seqCat((0 to t.size).map(i => hiercat(WSubIndex(e, i, t.tpe, UNKNOWNGENDER),t.tpe)))
+    case t:BundleType => seqCat(t.fields.map(f => hiercat(WSubField(e, f.name, f.tpe, UNKNOWNGENDER), f.tpe)))
+    case t:GroundType => e
+    case t => error("Unknown type encountered in toBits!")
+  }
+}
+
+object toBitMask {
+  def apply(e: Expression, dataType: Type): Expression = e match {
+    case ex: WRef => hiermask(ex,ex.tpe,dataType)
+    case ex: WSubField => hiermask(ex,ex.tpe,dataType)
+    case ex: WSubIndex => hiermask(ex,ex.tpe,dataType)
+    case t => error("Invalid operand expression for toBits!")
+  }
+  def hiermask(e: Expression, maskType: Type, dataType: Type): Expression = (maskType, dataType) match {
+    case (mt:VectorType, dt:VectorType) => seqCat((0 to mt.size).map(i => hiermask(WSubIndex(e, i, mt.tpe, UNKNOWNGENDER), mt.tpe, dt.tpe)))
+    case (mt:BundleType, dt:BundleType) => seqCat((mt.fields zip dt.fields).map{ case (mf,df) =>
+      hiermask(WSubField(e, mf.name, mf.tpe, UNKNOWNGENDER), mf.tpe, df.tpe) })
+    case (mt:UIntType, dt:GroundType) => seqCat(List.fill(bitWidth(dt).intValue)(e))
+    case (mt, dt) => error("Invalid type for mask component!")
+  }
+}
+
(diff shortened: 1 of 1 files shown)


shunshou:
Sweet! :) It'd be great if someone can merge this with the master branch ASAP, and then I can give this a whirl with a mem -> blackbox pass that can be upstreamed. And if it makes writing passes easier/more general by leaving resolving + checking to the respective passes, then I think, at this point, it's probably better? Thanks!! :)

donggyukim:
Is it useful to prevent #165?

src/main/scala/firrtl/passes/MemUtils.scala
```diff
@@ -0,0 +1,185 @@
...
+ ANY, PROVIDED HEREUNDER IS PROVIDED "AS IS". REGENTS HAS NO OBLIGATION
+ TO PROVIDE MAINTENANCE, SUPPORT, UPDATES, ENHANCEMENTS, OR
+ MODIFICATIONS.
+ */
+
+package firrtl.passes
+
+import com.typesafe.scalalogging.LazyLogging
+
+import scala.collection.mutable.LinkedHashSet
```
azidar: Get rid of unused imports

src/main/scala/firrtl/passes/MemUtils.scala
```diff
@@ -0,0 +1,185 @@
...
+      }
+      case t:UIntType => getPartGround(lhs, t, rhs, offset)
+      case t:SIntType => getPartGround(lhs, t, rhs, offset)
+      case t => error("Unknown type encountered in fromBits!")
+    }
+  }
+}
+
+object MemPortUtils {
+  def rPortToBundle(name: String, mem: DefMemory) =
```
donggyukim: What's `name` for? It's not used in this method.

donggyukim:
Also, why don't we put them on Utils.scala?

src/main/scala/firrtl/passes/MemUtils.scala
```diff
@@ -0,0 +1,185 @@
...
+import firrtl._
+import firrtl.ir._
+import firrtl.Utils._
+import firrtl.Mappers._
+import firrtl.PrimOps._
+import firrtl.Annotations._
+import firrtl.WrappedExpression._
+
+object toBits {
+  def apply(e: Expression): Expression = {
```
azidar: These types of functions can be more eloquently and compactly expressed as:
def apply(e: Expression): Expression = e match {
  case ex: WRef => hiercat(ex,ex.tpe)
  ...
}

azidar:
@donggyukim potentially :)

src/main/scala/firrtl/passes/MemUtils.scala
```diff
@@ -0,0 +1,185 @@
...
+      case t:VectorType => t.size * bitWidth(t.tpe)
+      case t:BundleType => t.fields.map(f => bitWidth(f.tpe)).foldLeft(BigInt(0))(_+_)
+      case t:UIntType => baseWidth(t.width)
+      case t:SIntType => baseWidth(t.width)
+      case t => error("Unknown type encountered in bitWidth!")
+    }
+  }
+  def baseWidth(w: Width): BigInt = {
+    w match {
+      case w:IntWidth => w.width
```
azidar: Another idiom of scala is the break out the case class, as it can sometimes result in cleaner-looking code:

```
w match {
   case IntWidth(width) => width
   ...
}
```

src/main/scala/firrtl/passes/MemUtils.scala
```diff
@@ -0,0 +1,185 @@
...
+      case ex: WRef => hiercat(ex,ex.tpe)
+      case ex: WSubField => hiercat(ex,ex.tpe)
+      case ex: WSubIndex => hiercat(ex,ex.tpe)
+      case t => error("Invalid operand expression for toBits!")
+    }
+  }
+  def hiercat(e: Expression, dt: Type): Expression = {
+    dt match {
+      case t:VectorType => DoPrim(PrimOps.Cat, (0 to t.size).map(i =>
+          hiercat(WSubIndex(e, i, t.tpe, UNKNOWNGENDER),t.tpe)), Seq.empty[BigInt], UIntType(UnknownWidth))
```
azidar: This will create a DoPrim(Cat, args, ..., ...) with args.size ==t.size. This isn't correct, as a Cat operator can only have 2 arguments. You need to build a tree of concatenations.

src/main/scala/firrtl/passes/MemUtils.scala
albert-magyar: Yeah I do that elsewhere but I think I used a multi-statement block at one point and just was sloppy when reducing the block to a single statement.

src/main/scala/firrtl/passes/MemUtils.scala
```diff
@@ -0,0 +1,185 @@
...
+      case t:BundleType => DoPrim(PrimOps.Cat, t.fields.map(f =>
+          hiercat(WSubField(e, f.name, f.tpe, UNKNOWNGENDER), f.tpe)), Seq.empty[BigInt], UIntType(UnknownWidth))
```
azidar: Same with this case, if the bundle has a number of fields != 2, this will generate a DoPrim(Cat, args, ...) with args.size != 2.

src/main/scala/firrtl/passes/MemUtils.scala
```diff
@@ -0,0 +1,185 @@
...
+    e match {
+      case ex: WRef => hiermask(ex,ex.tpe,dataType)
+      case ex: WSubField => hiermask(ex,ex.tpe,dataType)
+      case ex: WSubIndex => hiermask(ex,ex.tpe,dataType)
+      case t => error("Invalid operand expression for toBits!")
+    }
+  }
+  def hiermask(e: Expression, maskType: Type, dataType: Type): Expression = {
+    (maskType, dataType) match {
+      case (mt:VectorType, dt:VectorType) => DoPrim(PrimOps.Cat, (0 to mt.size).map(i =>
```
azidar: Same problem here, for each case class.

albert-magyar:
I believe I got all the mentioned issues, most importantly Adam's. It looks like I didn't do the best job cherry picking the functions from my commits on other branches the first time around.

src/main/scala/firrtl/passes/MemUtils.scala
jackkoenig: Even better, you can nest unapplies, so in the above

```scala
case UIntType(IntWidth(width)) => width
case SIntType(IntWidth(width)) => width
```

Or even better! We can define unapply on GroundType and replace both of those with

```scala
case GroundType(IntWidth(width)) => width
```

src/main/scala/firrtl/passes/MemUtils.scala
```diff
@@ -0,0 +1,174 @@
...
+        var stmts = Seq.empty[Statement]
+        for (f <- t.fields.reverse) {
+          val (tmpOffset, substmts) = getPart(WSubField(lhs, f.name, f.tpe, UNKNOWNGENDER), f.tpe, rhs, currentOffset)
+          stmts = stmts ++ substmts
+          currentOffset = tmpOffset
+        }
+        (currentOffset, stmts)
+      }
+      case t:UIntType => getPartGround(lhs, t, rhs, offset)
+      case t:SIntType => getPartGround(lhs, t, rhs, offset)
```
jackkoenig: You can merge these two lines into just matching on GroundType

src/main/scala/firrtl/passes/MemUtils.scala
```diff
@@ -0,0 +1,174 @@
...
+      case ex: WRef => hiercat(ex,ex.tpe)
+      case ex: WSubField => hiercat(ex,ex.tpe)
+      case ex: WSubIndex => hiercat(ex,ex.tpe)
+      case t => error("Invalid operand expression for toBits!")
+  }
+  def hiercat(e: Expression, dt: Type): Expression = dt match {
+      case t:VectorType => seqCat((0 to t.size).map(i => hiercat(WSubIndex(e, i, t.tpe, UNKNOWNGENDER),t.tpe)))
+      case t:BundleType => seqCat(t.fields.map(f => hiercat(WSubField(e, f.name, f.tpe, UNKNOWNGENDER), f.tpe)))
+      case t:UIntType => e
+      case t:SIntType => e
```
jackkoenig: Could replace both of these with:
```scala
case _: GroundType => e
```

src/main/scala/firrtl/passes/MemUtils.scala
jackkoenig: Although GroundType does also include ClockType so perhaps that's not what you want (you can always match on ClockType first and error or whatever)

albert-magyar: Yeah seems like it's the same number of lines and less clear to get the same strictness.

albert-magyar:
Props to Jack for being clever and helping me cut 10 lines

src/main/scala/firrtl/passes/MemUtils.scala
```diff
@@ -0,0 +1,161 @@
...
+    case _ => {
+      val seqs = args.splitAt(args.length/2)
+      DoPrim(PrimOps.Cat, Seq(seqCat(seqs._1),seqCat(seqs._2)), Seq.empty[BigInt], UIntType(UnknownWidth))
+    }
+  }
+}
+
+object toBits {
+  def apply(e: Expression): Expression = e match {
+      case ex: WRef => hiercat(ex,ex.tpe)
```
azidar: fix indentation

src/main/scala/firrtl/passes/MemUtils.scala
```diff
@@ -0,0 +1,161 @@
...
+      case t:VectorType => seqCat((0 to t.size).map(i => hiercat(WSubIndex(e, i, t.tpe, UNKNOWNGENDER),t.tpe)))
+      case t:BundleType => seqCat(t.fields.map(f => hiercat(WSubField(e, f.name, f.tpe, UNKNOWNGENDER), f.tpe)))
+      case t:GroundType => e
+      case t => error("Unknown type encountered in toBits!")
+  }
+}
+
+object toBitMask {
+  def apply(e: Expression, dataType: Type): Expression = e match {
+      case ex: WRef => hiermask(ex,ex.tpe,dataType)
```
azidar: fix indentation

azidar:
lgtm, fix the indentation and i'll merge :)

albert-magyar:
Cool, I got rid of those spaces

jackkoenig:
Good work!

PR #245, opened 2016-08-18
Sealed types for IR

grebe:
When going through firrtl problems with the recent fixed point work, I noticed that there were match statements that didn't include the new cases introduced by adding a new type. I thought that sealed types and the exhaustiveness checking that comes for free might be a handy way to solve this problem. This is my quick attempt to seal some types that look like they could benefit from sealing.

This is quick and dirty- I just dumped a bunch of stuff in IR.scala without pondering organization too much. If you want to merge I don't mind cleaning it up a bit more.

If you don't want to merge, I recommend at least checking this out and compiling it to track down all the warnings. There are a bunch!


jackkoenig:
While sealing types like this is generally advisable, it's not clear that we actually want to seal everything (or anything).

There is a larger discussion to be had about the Working IR nodes (like WRef and stuff) and how we can better handle that, as well as how to publicly acknowledge Chirrtl.

Ignoring those issues, things like VRandom show exactly why we probably don't want to seal Expression at least. It can be useful to introduce custom IR nodes for use within a pass or between passes (but within a Transformation such that they should not leak out). We don't want to pollute the IR with such things and we want outside users to be able to do similar things in their own passes without having to modify Firrtl. That being said, there could be a better way to handle this use case.

Perhaps we could still seal the IR but introduce a preferred way to do things like VRandom. There could be special nodes in the IR like `case class CustomExpr(contents: Any)`. This is obviously more or less the exact same thing but it might make such uses a bit more self-documenting.

grebe:
I agree with everything you said.

I'm not sure how exhaustiveness checking works with a mixture of non-sealed types. If you match an object with a sealed type and match on non-sealed types, does it still try to do it's best with exhaustiveness checking? Is sealing some, but not all, of the IR types useful?

azidar:
>  Is sealing some, but not all, of the IR types useful?

I don't think so, I can see use cases for introducing custom nodes for all of the IR types.

> special nodes in the IR like ```case class CustomExpr(contents: Any)```

This is a neat idea and potentially worth exploring.

PR #247, opened 2016-08-19
Bugfix: ExpandWhen was emitting WInvalid() 

azidar:
In the case that a bundle-typed wire declared and partially assigned within a when statement, the following generated node could be created:
```
node GEN = INVALID
```

Fix entails looking at the expression before memoizing - if its a referencing expression, we don't memoize. Otherwise, we do.

### DIFF
--- a/src/main/scala/firrtl/passes/ExpandWhens.scala
+++ b/src/main/scala/firrtl/passes/ExpandWhens.scala
@@ -137,15 +137,20 @@ object ExpandWhens extends Pass {
                 conseqNetlist getOrElse (lvalue, altNetlist(lvalue))
             }
 
-            nodes get res match {
-              case Some(name) =>
-                netlist(lvalue) = WRef(name, res.tpe, NodeKind(), MALE)
+            res match {
+              case _: ValidIf | _: Mux | _: DoPrim => nodes get res match {
+                case Some(name) =>
+                  netlist(lvalue) = WRef(name, res.tpe, NodeKind(), MALE)
+                  EmptyStmt
+                case None =>
+                  val name = namespace.newTemp
+                  nodes(res) = name
+                  netlist(lvalue) = WRef(name, res.tpe, NodeKind(), MALE)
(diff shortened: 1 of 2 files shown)


src/test/scala/firrtlTests/ExpandWhensSpec.scala
```diff
@@ -32,9 +32,51 @@ import org.scalatest._
 import org.scalatest.prop._
 import firrtl._
 import firrtl.passes._
+import firrtl.ir._
+import firrtl.Parser.IgnoreInfo
 
 class ExpandWhensSpec extends FirrtlFlatSpec {
+  def parse (input:String) = Parser.parse(input.split("\n").toIterator, IgnoreInfo)
```
jackkoenig: Style nit, should be 
```scala
private def parse(input: String) = ...
```
(Interestingly, I just noticed that Github syntax highlighting doesn't highlight the parameter name unless you put a space after the colon)

src/main/scala/firrtl/passes/ExpandWhens.scala
```diff
@@ -140,10 +140,15 @@ object ExpandWhens extends Pass {
...
-              netlist(lvalue) = memoExpr
+              res match {
+                case _: WInvalid|_: WRef|_: WSubField|_: WSubIndex|_: Literal =>
+                  netlist(lvalue) = res
+                case _ =>
+                  val memoNode = DefNode(s.info, namespace.newTemp, res)
+                  val memoExpr = WRef(memoNode.name, res.tpe, NodeKind(), MALE)
+                  memos += memoNode
+                  netlist(lvalue) = memoExpr
+              }
```
jackkoenig: A comment at the top of this match statement saying something like "Only memoize non-referential expressions" would be nice.

I also think it might be better to match on the expressions we *are* memoizing (ie. Muxes and Primops) rather than the ones we are not.

donggyukim:
LGTM

PR #249, opened 2016-08-22
Transform for blackboxing memories

shunshou:
Initial ReplSeqMem transform. Would appreciate code review :) @azidar @jackkoenig 

Needs to be run with the following arguments: --replSeqMem -c:circuit:-i:inputConf:-o:outputConf
inputConf isn't actually used right now. It's just a placeholder. 
Default rocket "make run-asm-tests" seems to pass (still in the middle of running through tests). 

Some notes:
* I just commented out some code that might be useful later (i.e. filling the mask bits to the full datawidth) to be consistent w/ the newest vlsi_mem_gen template. 
* The transform has "sub arguments" delimited by ":" b/c I wanted to be able to pass in conf file locations.
* The transform consists of 3 separate passes (sort of separation of concern-y???).

 -  Annotate sequential memories that are candidates for macro replacement + determine if write mask is needed. The need for write mask is determined by checking to see if wmode/en and wmask have the same origin in a module. 
 - Look for unique mem modules that need to be created (memories tagged with "prototype" references) and change references to use more meaningful port names (i.e. RW0 instead T_408)
 - Replace DefMemory with memory blackbox + wrapper + conf file. This will not generate wmask ports if not needed. Creates the minimum # of black boxes needed by the design. 

Throughout the 3 passes, I "annotate" information like maskGran, etc. via modding the DefMemory "Info". I know that isn't the intention of "Info", but it was easier to inline the annotations that way, at least for the time being. When the Annotation system becomes more mature, I'd like to, at the very least, tag the generated Module/ExtModule's with more information. 

Future:
* Add more unit tests (rw vs. r,w; mask vs. no mask; properly takes care of redundant memories)
* Split memory based off of input conf information
* Eventually get rid of the need for vlsi_mem_gen

I know I'm not conforming to Scala coding best practices. Was thinking of running this through some auto formatter later.

### DIFF
--- a/src/main/scala/firrtl/Namespace.scala
+++ b/src/main/scala/firrtl/Namespace.scala
@@ -85,7 +85,7 @@ object Namespace {
     namespace
   }
 
-  /** Initializes a [[Namespace]] for [[Module]] names in a [[Circuit]] */
+  /** Initializes a [[Namespace]] for [[ir.Module]] names in a [[ir.Circuit]] */
   def apply(c: Circuit): Namespace = {
     val namespace = new Namespace
     c.modules foreach { m =>
--- a/src/main/scala/firrtl/Namespace.scala
+++ b/src/main/scala/firrtl/Namespace.scala
@@ -84,5 +84,14 @@ object Namespace {
 
     namespace
   }
+
+  /** Initializes a [[Namespace]] for [[Module]] names in a [[Circuit]] */
+  def apply(c: Circuit): Namespace = {
+    val namespace = new Namespace
+    c.modules foreach { m =>
+      namespace.namespace += m.name
+    }
+    namespace
+  }
 }
 
--- a/src/main/scala/firrtl/passes/AnnotateMemMacros.scala
+++ b/src/main/scala/firrtl/passes/AnnotateMemMacros.scala
@@ -1,0 +1,130 @@
+package firrtl.passes
+
+import scala.collection.mutable.{ArrayBuffer,HashMap}
+import AnalysisUtils._
+import firrtl.WrappedExpression._
+import firrtl.ir._
+import firrtl._
+import firrtl.Utils._
+
+case class AppendableInfo(fields: Map[String,Any]) extends Info {
+  def append(a: Map[String,Any]) = this.copy(fields = fields ++ a)
+  def append(a: Tuple2[String,Any]): AppendableInfo = append(Map(a))
+  def get(f: String) = fields.get(f)
+  override def equals(b: Any) = b match {
+    case i: AppendableInfo => fields - "info" == i.fields - "info"
+    case _ => false
+  }
+}
+
+object AnalysisUtils {
+
+  def getConnects(m: Module) = {
+    val connects = HashMap[String, Expression]()
+    def getConnects(s: Statement): Unit = s match {
+      case s: Connect  =>
+        connects(s.loc.serialize) = s.expr
+      case s: PartialConnect =>
+        connects(s.loc.serialize) = s.expr
+      case s: DefNode =>
+        connects(s.name) = s.value
+      case s: Block =>
+        s.stmts foreach getConnects
+      case _ =>
+    }
+    getConnects(m.body)
+    connects.toMap
+  }  
+
+  // only works in a module
+  def getConnectOrigin(connects: Map[String, Expression], node: String): Expression = 
+    if (connects contains node) getConnectOrigin(connects,connects(node))
+    else EmptyExpression
+  
+  // backward searches until PrimOp, Lit or non-trivial Mux appears
+  private def getConnectOrigin(connects: Map[String, Expression], e: Expression): Expression = e match {    
+    case Mux(cond, tv, fv, _) if we(tv) == we(one) && we(fv) == we(zero) =>
+      getConnectOrigin(connects,cond.serialize)
+    case _: WRef | _: SubField | _: SubIndex | _: SubAccess if connects contains e.serialize =>
+      getConnectOrigin(connects,e.serialize) 
+    case _ => e
+  }
+
+  def appendInfo[T <: Info](info: T, add: Map[String,Any]) = info match {
+    case i: AppendableInfo => i.append(add)
+    case _ => AppendableInfo(fields = add + ("info" -> info))
+  }
+  def appendInfo[T <: Info](info: T, add: Tuple2[String,Any]): AppendableInfo = appendInfo(info,Map(add))
+  def getInfo[T <: Info](info: T, k: String) = info match{
+    case i: AppendableInfo => i.get(k)
+    case _ => None
+  }
+  def containsInfo[T <: Info](info: T, k: String) = info match {
+    case i: AppendableInfo => i.fields.contains(k)
+    case _ => false
+  }
+
+  def eqMems(a: DefMemory, b: DefMemory) = {
+    a.info == b.info &&
+    a.dataType == b.dataType &&
+    a.depth == b.depth &&
+    a.writeLatency == b.writeLatency &&
+    a.readLatency == b.readLatency &&
+    a.readers == b.readers &&
+    a.writers == b.writers &&
+    a.readwriters == b.readwriters &&
+    a.readUnderWrite == b.readUnderWrite
+  }  
+
+}
+
+object AnnotateMemMacros extends Pass {
+
+  def name = "Analyze sequential memories and tag with info for future passes (useMacro,maskGran)"
+
+  def run(c: Circuit) = {
+
+    def annotateModMems(m: Module) = {
+      val connects = getConnects(m)
+
+      // returns # of mask bits if used
+      def getMaskBits(wen: String, wmask: String): Option[Int] = {
+        val wenOrigin = we(getConnectOrigin(connects,wen))
+        val one1 = we(one)
+        val wmaskOrigin = connects.keys.toSeq.filter(_.startsWith(wmask)).map(x => we(getConnectOrigin(connects,x)))
+        // all wmask bits are equal to wmode/wen or all wmask bits = 1(for redundancy checking)
+        val redundantMask = wmaskOrigin.map( x => (x == wenOrigin) || (x == one1) ).foldLeft(true)(_ && _)
+        if (redundantMask) None else Some(wmaskOrigin.length)
+      }
+
+      def updateStmts(s: Statement): Statement = s match {
+        // only annotate memories that are candidates for memory macro replacements
+        // i.e. rw, w + r (read,write 1 cycle delay)
+        case m: DefMemory if m.readLatency == 1 && m.writeLatency == 1 && 
+            (m.writers.length + m.readwriters.length) == 1 && m.readers.length <= 1 =>
+          val dataBits = bitWidth(m.dataType)
+          val rwMasks = m.readwriters map (w => getMaskBits(s"${m.name}.$w.wmode",s"${m.name}.$w.wmask"))
+          val wMasks = m.writers map (w => getMaskBits(s"${m.name}.$w.en",s"${m.name}.$w.mask"))
+          val maskBits = (rwMasks ++ wMasks).head
+          val memAnnotations = Map("useMacro" -> true)
+          val tempInfo = appendInfo(m.info,memAnnotations)
+          if (maskBits == None) m.copy(info = tempInfo)
+          else m.copy(info = tempInfo.append("maskGran" -> dataBits/maskBits.get))       
+        case b: Block => Block(b.stmts map updateStmts)
+        case s => s
+      }
+
+      m.copy(body=updateStmts(m.body))
+    } 
+
+    val updatedMods = c.modules map {
+      case m: Module => annotateModMems(m)
+      case m: ExtModule => m
+    }
+    c.copy(modules = updatedMods)
+
+  }  
+
+}
+
+// TODO: Add floorplan info?
--- a/src/main/scala/firrtl/passes/AnnotateMemMacros.scala
+++ b/src/main/scala/firrtl/passes/AnnotateMemMacros.scala
@@ -22,7 +22,7 @@ object AnalysisUtils {
   def getConnects(m: Module) = {
     val connects = HashMap[String, Expression]()
     def getConnects(s: Statement): Unit = s match {
-      case s: Connect  =>
+      case s: Connect =>
         connects(s.loc.serialize) = s.expr
       case s: PartialConnect =>
         connects(s.loc.serialize) = s.expr
@@ -37,14 +37,49 @@ object AnalysisUtils {
   }  
 
   // only works in a module
-  def getConnectOrigin(connects: Map[String, Expression], node: String): Expression = 
+  def getConnectOrigin(connects: Map[String, Expression], node: String): Expression = {
     if (connects contains node) getConnectOrigin(connects,connects(node))
     else EmptyExpression
-  
+  }
+
+  def checkLit(e: Expression) = e match {
+    case l : Literal => true
+    case _ => false
+  }
+
+  def getOrigin(connects: Map[String, Expression], e: Expression) = e match {
+    case DoPrim(_,_,_,_) => getConnectOrigin(connects,e)
+    case l if (checkLit(l)) => e
+    case _ => getConnectOrigin(connects,e.serialize)  
+  }
+
   // backward searches until PrimOp, Lit or non-trivial Mux appears
-  private def getConnectOrigin(connects: Map[String, Expression], e: Expression): Expression = e match {    
-    case Mux(cond, tv, fv, _) if we(tv) == we(one) && we(fv) == we(zero) =>
-      getConnectOrigin(connects,cond.serialize)
+  // technically, you should keep searching through PrimOp, because a node + 0 is still itself, 
+  // a node shifted by 0 is still itself, etc. 
+  // TODO: handle validif???, more thorough
+  private def getConnectOrigin(connects: Map[String, Expression], e: Expression): Expression = e match { 
+    case Mux(cond, tv, fv, _) =>
+      val fvOrigin = getOrigin(connects,fv)
+      val tvOrigin = getOrigin(connects,tv)
+      val condOrigin = getOrigin(connects,cond)
+      if (we(tvOrigin) == we(one) && we(fvOrigin) == we(zero)) condOrigin
+      else if (we(condOrigin) == we(one)) tvOrigin
+      else if (we(condOrigin) == we(zero)) fvOrigin
+      else if (we(tvOrigin) == we(fvOrigin)) tvOrigin
+      else if (we(fvOrigin) == we(zero) && we(condOrigin) == we(tvOrigin)) condOrigin
+      else e
+    case DoPrim(op, args, consts, tpe) if op == PrimOps.Or && args.contains(one) => one
+    case DoPrim(op, args, consts, tpe) if op == PrimOps.And && args.contains(zero) => zero
+    case DoPrim(op, args, consts, tpe) if op == PrimOps.Bits =>  
+      val msb = consts(0)
+      val lsb = consts(1) 
+      val extractionWidth = (msb-lsb)+1
+      val nodeWidth = bitWidth(args.head.tpe)
+      // if you're extracting the full bitwidth, then keep searching for origin
+      if (nodeWidth == extractionWidth) getOrigin(connects,args.head)
+      else e
+    case DoPrim(op, args, _, _) if (op == PrimOps.AsUInt || op == PrimOps.AsSInt || op == PrimOps.AsClock) =>
+      getOrigin(connects,args.head)
     case _: WRef | _: SubField | _: SubIndex | _: SubAccess if connects contains e.serialize =>
       getConnectOrigin(connects,e.serialize) 
     case _ => e
@@ -113,7 +148,6 @@ object AnnotateMemMacros extends Pass {
         case b: Block => Block(b.stmts map updateStmts)
         case s => s
       }
-
       m.copy(body=updateStmts(m.body))
     } 
 
--- a/src/main/scala/firrtl/passes/AnnotateMemMacros.scala
+++ b/src/main/scala/firrtl/passes/AnnotateMemMacros.scala
@@ -9,7 +9,7 @@ import firrtl.Utils._
 
 case class AppendableInfo(fields: Map[String,Any]) extends Info {
   def append(a: Map[String,Any]) = this.copy(fields = fields ++ a)
-  def append(a: Tuple2[String,Any]): AppendableInfo = append(Map(a))
+  def append(a: (String,Any)): AppendableInfo = append(Map(a))
   def get(f: String) = fields.get(f)
   override def equals(b: Any) = b match {
     case i: AppendableInfo => fields - "info" == i.fields - "info"
@@ -89,7 +89,7 @@ object AnalysisUtils {
     case i: AppendableInfo => i.append(add)
     case _ => AppendableInfo(fields = add + ("info" -> info))
   }
-  def appendInfo[T <: Info](info: T, add: Tuple2[String,Any]): AppendableInfo = appendInfo(info,Map(add))
+  def appendInfo[T <: Info](info: T, add: (String,Any)): AppendableInfo = appendInfo(info,Map(add))
   def getInfo[T <: Info](info: T, k: String) = info match{
     case i: AppendableInfo => i.get(k)
     case _ => None
--- a/src/main/scala/firrtl/passes/AnnotateMemMacros.scala
+++ b/src/main/scala/firrtl/passes/AnnotateMemMacros.scala
@@ -6,6 +6,7 @@ import firrtl.WrappedExpression._
 import firrtl.ir._
 import firrtl._
 import firrtl.Utils._
+import firrtl.Mappers._
 
 case class AppendableInfo(fields: Map[String,Any]) extends Info {
   def append(a: Map[String,Any]) = this.copy(fields = fields ++ a)
@@ -23,8 +24,6 @@ object AnalysisUtils {
     val connects = HashMap[String, Expression]()
     def getConnects(s: Statement): Unit = s match {
       case s: Connect =>
-        connects(s.loc.serialize) = s.expr
-      case s: PartialConnect =>
         connects(s.loc.serialize) = s.expr
       case s: DefNode =>
         connects(s.name) = s.value
@@ -53,6 +52,10 @@ object AnalysisUtils {
     case _ => getConnectOrigin(connects,e.serialize)  
   }
 
+  // takes in a list of connections in a given module and looks to find the origin of some node
+  // where origin is the source of a connect. note that if the source is a trivial primop that
+  // has yet to be optimized via constant propagation, the function will try to search backwards
+  // past the primop. this allows you to compare if two nodes have the same origin, for example. 
   // backward searches until PrimOp, Lit or non-trivial Mux appears
   // technically, you should keep searching through PrimOp, because a node + 0 is still itself, 
   // a node shifted by 0 is still itself, etc. 
@@ -68,9 +71,9 @@ object AnalysisUtils {
       else if (we(tvOrigin) == we(fvOrigin)) tvOrigin
       else if (we(fvOrigin) == we(zero) && we(condOrigin) == we(tvOrigin)) condOrigin
       else e
-    case DoPrim(op, args, consts, tpe) if op == PrimOps.Or && args.contains(one) => one
-    case DoPrim(op, args, consts, tpe) if op == PrimOps.And && args.contains(zero) => zero
-    case DoPrim(op, args, consts, tpe) if op == PrimOps.Bits =>  
+    case DoPrim(PrimOps.Or, args, consts, tpe) if args.contains(one) => one
+    case DoPrim(PrimOps.And, args, consts, tpe) if args.contains(zero) => zero
+    case DoPrim(PrimOps.Bits, args, consts, tpe) =>  
       val msb = consts(0)
       val lsb = consts(1) 
       val extractionWidth = (msb-lsb)+1
@@ -145,7 +148,7 @@ object AnnotateMemMacros extends Pass {
           val tempInfo = appendInfo(m.info,memAnnotations)
           if (maskBits == None) m.copy(info = tempInfo)
           else m.copy(info = tempInfo.append("maskGran" -> dataBits/maskBits.get))       
-        case b: Block => Block(b.stmts map updateStmts)
+        case b: Block => b map updateStmts
         case s => s
       }
       m.copy(body=updateStmts(m.body))
--- a/src/main/scala/firrtl/passes/AnnotateMemMacros.scala
+++ b/src/main/scala/firrtl/passes/AnnotateMemMacros.scala
@@ -1,6 +1,8 @@
+// See LICENSE for license details.
+
 package firrtl.passes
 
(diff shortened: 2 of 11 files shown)


azidar:
Overall, looks good but I need to spend more time looking over it. One comment though - as you said, I also don't think "info" is the right place to store the annotation information. What I would recommend instead is create your own IR.Expression node, call it DefAnnotatedMemory, and have it contain the additional fields (similar to how WRef takes the place of Reference). Do you think that would work with your given use case?

src/main/scala/firrtl/passes/ReplSeqMem.scala
```diff
@@ -0,0 +1,142 @@
...
+              (c, pass) =>
+                val x = Utils.time(pass.name)(pass run c)
+                logger debug x.serialize
+                x
+            }, 
+            None, 
+            Some(map)
+          )
+        }  
+        case _ => TransformResult(circuit, None, Some(map))
```
azidar: Error instead of doing nothing? If you get an annotation with the correct TransID but it isn't the annotation you were expecting, I believe this should be an error.

azidar:
I think its probably time to split the codebase into a few more packages. I think each "Transform" should have its own package, Utils, etc. that is shared among its internal passes. However, that shouldn't be part of this PR, so its fine for now :)

src/main/scala/firrtl/passes/ReplSeqMem.scala
```diff
@@ -0,0 +1,142 @@
...
+import java.io.Writer
+import AnalysisUtils._
+
+sealed trait PassOption
+case object InputConfigFileName extends PassOption
+case object OutputConfigFileName extends PassOption
+case object PassCircuitName extends PassOption
+
+object Error {
+  def apply[T <: Any](msg: String) = throw new Exception(msg)
```
azidar: Consider using https://github.com/ucb-bar/firrtl/blob/master/src/main/scala/firrtl/passes/Passes.scala#L54 instead? It can collect many errors and throw them together. If this isn't desired, then don't worry about it.

shunshou: I just wanted a shorthand for not having to type throw new Exception(msg) everytime. I'm ok with whatever/will look into that.

src/main/scala/firrtl/passes/ReplSeqMem.scala
shunshou: Ok. Maybe that should also be fixed in InferReadWrite? (stole this bit of code from it)

shunshou:
Re: your info comment -- DefAnnotatedMemory is a little weird, unless you mean I can have a separate "annotation" field that more or less looks like the "info" field I've made -- i.e. a HashMap[String,Any]. 

My issue is: between different sub-passes (which in the future, should be allowed to be disabled), I may or may not want to add additional annotations. I'm not sure how you handle it with the AnnotationMap right now. Each subsequent pass should look to see if an annotation exists, and I don't want to have to keep changing the DefAnnotatedMemory case class to add in optionable annotations as the Transform gets (optionally) more complicated. 

i.e. currently, I annotate stuff like
useMacro
maskGran

In the future, if an input conf file is fed in, I might want to add annotations like
splitWidth
splitDepth
splitNum 

(i.e. how to split the memory, which the last pass will then use -- not just simple black boxing anymore)

Regarding HashMap[String,Any], it does get a little annoying having to match to stuff, each time... i.e.
> x match {case i: Info => i}

I'm still really bad with Scala and types. Does it work to have an convertTo[Info](x) function? 

Also, thanks for the feedback!

src/main/scala/firrtl/passes/ReplSeqMem.scala
azidar: @donggyukim

shunshou:
Oh and re: Transform utils -- I assume there's some universally applicable set of Utils (i.e. find origin), so some thought needs to be put into whether Utils are X-Transform specific or more universal... 

Although separating analysis vs. manipulation (i.e. remove connection) would be useful

azidar:
> DefAnnotatedMemory is a little weird, unless you mean I can have a separate "annotation" field that more or less looks like the "info" field I've made -- i.e. a HashMap[String,Any].

Yeah that is exactly what I mean. Then it is all contained within one node, and your code isn't reappropriating the info field. In fact, I think out of all the IR nodes, Info is the one that makes the most sense to seal.

As far as a general strategy to pass annotations within a transform, you're right it is definitely clunky right now. I'm currently experimenting with having Pass extend Transform (so Transforms can merge with other transforms and reuse the annotation passing API). This is something we should talk more about at the Firrtl dev meeting.

shunshou:
"info" = source locators right now is an annotation in and of itself, though... so I don't understand why it should be its own separate thing. If you leave it out and seal it, I feel like you're just special casing something that doesn't need to be special cased...

azidar:
> you're just special casing something that doesn't need to be special cased...

This is the difference between a spec and an implementation. The spec defines the existence and propagation behavior of info. Any pass can look at info and know exactly what it means (without any knowledge of any other pass).

I agree that you could remove info and pass it via the annotation system as a proof of concept. You could also pass types the same way. For simplicity (and for the sake of sanity) we don't pass types this way, and similarly, we don't pass info the same way.

azidar:
Re: using HashMap[String,Any] - you can just add all the values you want in the DefAnnotatedMemory case class, and because you create and remove it within your Transform, at a later time you can always add more options without breaking any code.

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
```diff
@@ -0,0 +1,164 @@
...
+}
+
+object AnalysisUtils {
+
+  def getConnects(m: Module) = {
+    val connects = HashMap[String, Expression]()
+    def getConnects(s: Statement): Unit = s match {
+      case s: Connect =>
+        connects(s.loc.serialize) = s.expr
+      case s: PartialConnect =>
```
azidar: unnecessary, all partial connects will be gone in middle/low firrtl

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
```diff
@@ -0,0 +1,164 @@
...
+      case s: Block =>
+        s.stmts foreach getConnects
+      case _ =>
+    }
+    getConnects(m.body)
+    connects.toMap
+  }  
+
+  // only works in a module
+  def getConnectOrigin(connects: Map[String, Expression], node: String): Expression = {
```
azidar: comment why you need this. might be worthwhile putting this in MemUtils.scala

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
```diff
@@ -0,0 +1,164 @@
...
+      val fvOrigin = getOrigin(connects,fv)
+      val tvOrigin = getOrigin(connects,tv)
+      val condOrigin = getOrigin(connects,cond)
+      if (we(tvOrigin) == we(one) && we(fvOrigin) == we(zero)) condOrigin
+      else if (we(condOrigin) == we(one)) tvOrigin
+      else if (we(condOrigin) == we(zero)) fvOrigin
+      else if (we(tvOrigin) == we(fvOrigin)) tvOrigin
+      else if (we(fvOrigin) == we(zero) && we(condOrigin) == we(tvOrigin)) condOrigin
+      else e
+    case DoPrim(op, args, consts, tpe) if op == PrimOps.Or && args.contains(one) => one
```
azidar: case DoPrim(PrimOps.Or, args, consts, tpe) if args.contains(one) => one

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
azidar: I'd like to see multiple tests for this function to make sure its correct.

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
```diff
@@ -0,0 +1,164 @@
...
+            (m.writers.length + m.readwriters.length) == 1 && m.readers.length <= 1 =>
+          val dataBits = bitWidth(m.dataType)
+          val rwMasks = m.readwriters map (w => getMaskBits(s"${m.name}.$w.wmode",s"${m.name}.$w.wmask"))
+          val wMasks = m.writers map (w => getMaskBits(s"${m.name}.$w.en",s"${m.name}.$w.mask"))
+          val maskBits = (rwMasks ++ wMasks).head
+          val memAnnotations = Map("useMacro" -> true)
+          val tempInfo = appendInfo(m.info,memAnnotations)
+          if (maskBits == None) m.copy(info = tempInfo)
+          else m.copy(info = tempInfo.append("maskGran" -> dataBits/maskBits.get))       
+        case b: Block => Block(b.stmts map updateStmts)
```
azidar: can do:
case b: Block => b map updateStmts

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
shunshou: FYI -- this was also taken more or less from InferReadWrite.scala. I think these are useful functions that should be pulled out into some generic AnalysisUtils-like Util thing

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
shunshou: Also, this whole notion of going through all of the modules in a circuit and all of the statements in a module is so common... that should probably be templated up in the firrtl spec readme or... idk if there's a better way to package it in code...

src/main/scala/firrtl/Namespace.scala
```diff
@@ -84,5 +84,14 @@ object Namespace {
 
     namespace
   }
+
+  /** Initializes a [[Namespace]] for [[Module]] names in a [[Circuit]] */
```
jackkoenig: I'll need to check the Scaladoc (which is a mess right now anyway), but I believe that since Module and Circuit are in package firrtl.ir, but Namespace is just in package firrtl, this needs to be [[ir.Module]] and [[ir.Circuit]]

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
```diff
@@ -0,0 +1,167 @@
...
+    a.info == b.info &&
+    a.dataType == b.dataType &&
+    a.depth == b.depth &&
+    a.writeLatency == b.writeLatency &&
+    a.readLatency == b.readLatency &&
+    a.readers == b.readers &&
+    a.writers == b.writers &&
+    a.readwriters == b.readwriters &&
+    a.readUnderWrite == b.readUnderWrite
+  }
```
jackkoenig: Is this not the default equality for DefMemory since it's a case class?

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
```diff
@@ -0,0 +1,167 @@
...
+  override def equals(b: Any) = b match {
+    case i: AppendableInfo => fields - "info" == i.fields - "info"
+    case _ => false
+  }
+}
+
+object AnalysisUtils {
+
+  def getConnects(m: Module) = {
+    val connects = HashMap[String, Expression]()
```
jackkoenig: Generally prefer mutable.Blah when using mutable data structures.

src/main/scala/firrtl/passes/AnnotateValidMemConfigs.scala
```diff
@@ -0,0 +1,298 @@
...
+    // (output) config file path
+    configFile: Option[String],
+    // config pattern
+    configPattern: Option[String],
+    // read documentation for details 
+    defaultArgs: Option[String],
+    // default behavior (if not used) is to have wmask port width = datawidth/maskgran
+    // if true: wmask port width pre-filled to datawidth
+    fillWMask: Boolean
+){
```
jackkoenig: This should go on the previous line.
ie. 
```scala
    fillWMask: Boolean){
```
See http://docs.scala-lang.org/style/declarations.html#classes

src/main/scala/firrtl/passes/AnnotateValidMemConfigs.scala
```diff
@@ -0,0 +1,298 @@
...
+  import CustomYAMLProtocol._
+
+  def name = "Annotate memories with valid split depths, widths, #\'s"
+
+  // TODO: Consider splitting InferRW to analysis + actual optimization pass, in case sp doesn't exist
+  // TODO: Don't get first available? 
+  case class SRAMCompilerSet(
+    sp: Option[SRAMCompiler] = None, 
+    dp: Option[SRAMCompiler] = None 
+  ){
```
jackkoenig: This should be 
```scala
case class SRAMCompilerSet(
    sp: Option[SRAMCompiler] = None, 
    dp: Option[SRAMCompiler] = None){
```
ie. when you do multiline class definition, tab twice (ie. 4 spaces) and put the final ){ on the same line as the last field.

src/main/scala/firrtl/passes/AnnotateValidMemConfigs.scala
```diff
@@ -0,0 +1,298 @@
...
+    def serialize() = {
```
jackkoenig: For functions with no arguments, Scala-style specifies that invoking with () generally suggests changing mutable state. Thus if serialize is returning a String (as I suspect it is), there should not be parentheses here.

src/main/scala/firrtl/passes/AnnotateValidMemConfigs.scala
```diff
@@ -0,0 +1,298 @@
...
+  val outputBuffer = new java.io.CharArrayWriter
+  val separator = "--- \n"
+  def append(in: YamlValue) = {
+    outputBuffer.append(separator + in.prettyPrint)
+  }
+  def serialize = {
+    val outputFile = new java.io.PrintWriter(file)
+    outputFile.write(outputBuffer.toString)
+    outputFile.close()
+  }
```
jackkoenig: Throughout Firrtl, Serialize returns a String and is purely functional. For consistency I think this name should be different, something like "dump"

src/main/scala/firrtl/passes/AnnotateValidMemConfigs.scala
```diff
@@ -0,0 +1,298 @@
...
+  inc: Int,
+  max: Int,
+  // these values should not be used, regardless of min,inc,max
+  illegal: Option[List[Int]]
+){
+  def getValid = {
+    val range = (min to max by inc).toList
+    range.filterNot(illegal.getOrElse(List[Int]()).toSet)
+  }
+}
```
jackkoenig: Similar to other comments, double tab the class parameters (4 spaces), and put the ){ on the same line as the last parameter

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
```diff
@@ -0,0 +1,167 @@
...
+    def getConnects(s: Statement): Unit = s match {
+      case s: Connect =>
+        connects(s.loc.serialize) = s.expr
+      case s: DefNode =>
+        connects(s.name) = s.value
+      case s: Block =>
+        s.stmts foreach getConnects
+      case _ =>
+    }
```
jackkoenig: So this code is not going to be finding connects in Conditionals, if that's intentional, no worries, if it's not intentional though we should fix it. It general how I would write this is (still using connects from above and unapply matching which I generally prefer but is just a stylistic opinion):
```scala
    def getConnects(s: Statement): Statement = {
      s map getConnects match {
        case Connect(_, loc, expr) =>
          connects(loc.serialize) = expr
        case DefNode(_, name, value) =>
          connects(name) = value
        case _ => // do nothing
      }
      s // return because we only have map and not foreach right now
    }
```

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
```diff
@@ -0,0 +1,167 @@
...
+  // only works in a module
+  def getConnectOrigin(connects: Map[String, Expression], node: String): Expression = {
+    if (connects contains node) getConnectOrigin(connects,connects(node))
+    else EmptyExpression
+  }
+
+  def checkLit(e: Expression) = e match {
+    case l : Literal => true
+    case _ => false
+  }
```
jackkoenig: alternatively this *could* just be
```scala
def checkLit(e: Expression) = e.isInstanceOf[Literal]
```
nbd though

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
```diff
@@ -0,0 +1,167 @@
...
+
+  def getOrigin(connects: Map[String, Expression], e: Expression) = e match {
+    case DoPrim(_,_,_,_) => getConnectOrigin(connects,e)
+    case l if (checkLit(l)) => e
+    case _ => getConnectOrigin(connects,e.serialize)  
+  }
```
jackkoenig: per the above, if this is the only use of checkLit, then you can just match on the type here?
```scala
case _: Literal => e
```
or
```scala
case l: Literal => l
```

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
```diff
@@ -0,0 +1,167 @@
...
+    else EmptyExpression
+  }
+
+  def checkLit(e: Expression) = e match {
+    case l : Literal => true
+    case _ => false
+  }
+
+  def getOrigin(connects: Map[String, Expression], e: Expression) = e match {
+    case DoPrim(_,_,_,_) => getConnectOrigin(connects,e)
```
jackkoenig: If you don't care about any of the fields you can always just do `case _: DoPrim => ...`

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
```diff
@@ -0,0 +1,167 @@
...
+      else if (we(condOrigin) == we(one)) tvOrigin
+      else if (we(condOrigin) == we(zero)) fvOrigin
+      else if (we(tvOrigin) == we(fvOrigin)) tvOrigin
+      else if (we(fvOrigin) == we(zero) && we(condOrigin) == we(tvOrigin)) condOrigin
+      else e
+    case DoPrim(PrimOps.Or, args, consts, tpe) if args.contains(one) => one
+    case DoPrim(PrimOps.And, args, consts, tpe) if args.contains(zero) => zero
+    case DoPrim(PrimOps.Bits, args, consts, tpe) =>  
+      val msb = consts(0)
+      val lsb = consts(1)
```
jackkoenig: If you want you could be even more clever here and unapply consts as well. ie.
```scala
case DoPrim(PrimOps.Bits, args, Seq(msb, lsb), tpe) =>
```
You could do something similar with args.head since arg should only hvae 1 argument

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
```diff
@@ -0,0 +1,167 @@
...
+    case DoPrim(PrimOps.Bits, args, consts, tpe) =>  
+      val msb = consts(0)
+      val lsb = consts(1) 
+      val extractionWidth = (msb-lsb)+1
+      val nodeWidth = bitWidth(args.head.tpe)
+      // if you're extracting the full bitwidth, then keep searching for origin
+      if (nodeWidth == extractionWidth) getOrigin(connects,args.head)
+      else e
+    case DoPrim(op, args, _, _) if (op == PrimOps.AsUInt || op == PrimOps.AsSInt || op == PrimOps.AsClock) =>
+      getOrigin(connects,args.head)
```
jackkoenig: in general, prefer spacing after commas, ie.
```scala
getOrigin(connects, args.head)
```

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
shunshou: What do you mean by conditionals? = Whens or Muxes? 
I forget, but I was hoping at least for my pass, that by the mid to low firrtl transform, When's should've already been converted to Mux's? This makes this utility function not super meaningful at high firrtl, but I wasn't sure how to deal with that (and this particular function was pulled from InferReadWrite). 

Also, can you explain what "ValidIf" is? I'm not clear about it from what code I've read.

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
shunshou: This was mostly just me trying to pull out a utility function that might be helpful elsewhere. I have no implementation preference, but is it meaningful for future passes?

shunshou:
Not sure if you're going to add this separately, but note to self: Change the YamlReader map into flatmap

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
shunshou: Also, can you explain "unapply matching"? I'm having a hard time understanding how s map getConnects actually works... It just recursively calls getConnects on each element of s? But that only neds to happen for Blocks? So I personally feel it's more clear to do the Block stuff rather than s map getConnects, unless I'm missing something...

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
jackkoenig: As you pointed out to me, you want to check everything but name. As I mentioned, you *could* (although perhaps shouldn't) create new memories with the same name and compare using the builtin ==, thus keeping this somewhat more robust to hopefully rare changes to the IR. I'm not sure
ie.
```scala
def eqMems(a: DefMemory, b: DefMemory) = a.copy(name = "blah") == b.copy(name = "blah")
```

src/main/scala/firrtl/passes/AnnotateValidMemConfigs.scala
```diff
@@ -0,0 +1,298 @@
...
+  implicit val sc = yamlFormat11(SRAMCompiler)
+}
+
+case class DimensionRules(
+  min: Int,
+  // step size
+  inc: Int,
+  max: Int,
+  // these values should not be used, regardless of min,inc,max
+  illegal: Option[List[Int]]
```
jackkoenig: Does Option[List[_]] make sense? Can't it just be a list that is sometimes empty?

src/main/scala/firrtl/passes/AnnotateValidMemConfigs.scala
```diff
@@ -0,0 +1,298 @@
...
+class YamlFileReader(file: String){
+  import CustomYAMLProtocol._
+  def parse[A](implicit reader: YamlReader[A]) : Seq[A] = {
+    if (new java.io.File(file).exists) {
+      val yamlString = scala.io.Source.fromFile(file).getLines.mkString("\n")
+      val optionOut = yamlString.parseYamls.map(x => 
+        try Some(reader.read(x))
+        catch {case e: Exception => None} 
+      )
+      optionOut.filter(_ != None).map(_.get)
```
jackkoenig: yeah just delete the last line and make the parseYamls.map a flatMap

src/main/scala/firrtl/passes/ReplaceMemMacros.scala
```diff
@@ -0,0 +1,167 @@
...
+
+class ReplaceMemMacros(writer: ConfWriter) extends Pass {
+
+  def name = "Replace memories with black box wrappers (optimizes when write mask isn't needed) + configuration file"
+
+  def run(c: Circuit) = {
+
+    lazy val moduleNamespace = Namespace(c)
+    val memMods = ArrayBuffer[DefModule]()
+    val uniqueMems = ArrayBuffer[DefMemory]()
```
jackkoenig: mutable.ArrayBuffer

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
jackkoenig: Right okay. Yeah if you're on midfirrtl you don't need to worry about conditionals (meaning whens, not muxes). Regardless the code I wrote will do the right thing. And you make a good point, it's not entirely clear how to deal with this if you have multi assignment to the same thing, it seems like this really needs to be run on midfirrtl.

ValidIf is the thing that IsInvalid inside whens maps to in the same way that connections in whens map to muxes. It kind of means "ConnectIf" in that if the predicate is false just put garbage.

By unapply matching I simply mean you can match on Connect or DefNode above in two equivalent ways:
```scala
case c: Connect =>
  connects(c.loc.serialize) = c.expr
case Connect(_, loc, expr) =>
  connects(loc.serialize) = expr
```
those matches are equivalent

Finally, for s map getConnects, while it is true that on midfirrtl the only place it matters is on Blocks, if we had conditionals then it would matter there as well. You can do it on just Block if you want. Similarly, there is map for mapping on the Expression children of Statements. In that case you could do it manually (which requires a lot more than just Block), or you could use map.

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
jackkoenig: I don't think so since you can match on Literal as below. You can also call .isInstanceOf[Literal] on anything which is only marginally more characters than checkLit

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
```diff
@@ -0,0 +1,167 @@
...
+    case DoPrim(PrimOps.And, args, consts, tpe) if args.contains(zero) => zero
+    case DoPrim(PrimOps.Bits, args, consts, tpe) =>  
+      val msb = consts(0)
+      val lsb = consts(1) 
+      val extractionWidth = (msb-lsb)+1
+      val nodeWidth = bitWidth(args.head.tpe)
+      // if you're extracting the full bitwidth, then keep searching for origin
+      if (nodeWidth == extractionWidth) getOrigin(connects,args.head)
+      else e
+    case DoPrim(op, args, _, _) if (op == PrimOps.AsUInt || op == PrimOps.AsSInt || op == PrimOps.AsClock) =>
```
jackkoenig: A neat thing you could do here would be to unapply the PrimOps as:
```scala
```

src/main/scala/firrtl/passes/ReplaceMemMacros.scala
```diff
@@ -0,0 +1,167 @@
...
+    val stmts = ArrayBuffer[Statement]()
+    val wrapperioPorts = MemPortUtils.memToBundle(m).fields.map(f => Port(NoInfo, f.name, Input, f.tpe)) 
+    val bbProto = m.copy(dataType = flattenType(m.dataType))
+    val bbioPorts = MemPortUtils.memToFlattenBundle(m).fields.map(f => Port(NoInfo, f.name, Input, f.tpe)) 
+
+    stmts += WDefInstance(NoInfo,m.name,m.name,UnknownType)
+    val bbRef = createRef(m.name)
+    stmts ++= (m.readers zip bbProto.readers).map{ 
+      case (x,y) => adaptReader(createRef(x),m,createSubField(bbRef,y),bbProto)
+    }.flatten
```
jackkoenig: map followed by flatten is just a flatMap

jackkoenig: Also spacing around commas is appreciated

src/main/scala/firrtl/passes/ReplaceMemMacros.scala
```diff
@@ -0,0 +1,167 @@
...
+    val uniqueMems = ArrayBuffer[DefMemory]()
+
+    def updateMemMods(m: Module) = {
+      val memPortMap = HashMap[String,Expression]()
+
+      def updateMemStmts(s: Statement): Statement = s match {
+        case m: DefMemory if containsInfo(m.info,"useMacro") => 
+          if(!containsInfo(m.info,"maskGran")) {
+            m.writers foreach {w => memPortMap(s"${m.name}.${w}.mask") = EmptyExpression}
+            m.readwriters foreach {w => memPortMap(s"${m.name}.${w}.wmask") = EmptyExpression}
```
jackkoenig: when using curly braces there should be spaces before and after

src/main/scala/firrtl/passes/UpdateDuplicateMemMacros.scala
```diff
@@ -0,0 +1,95 @@
...
+    val rFields = defaultFields :+ "data"
+    val wFields = rFields :+ "mask"
+    val rwFields = defaultFields ++ Seq("wmode","wdata","rdata","wmask")
+
+    def updateMemPortMap(ports: Seq[String], fields: Seq[String], portType: String) = 
+      for (p <- ports.zipWithIndex; f <- fields) {
+        val newPort = createSubField(createRef(m.name),portType+p._2)        
+        val field = createSubField(newPort,f)
+        memPortMap(s"${m.name}.${p._1}.${f}") = field
+      }
```
jackkoenig: I generally prefer unapplying tuples to the the whole ._1 and ._2. ie.
```scala
for ((p, i) <- ports.zipWithIndex; f <- fields) {
```
then you can use i for the index and p for the port!

src/main/scala/firrtl/passes/UpdateDuplicateMemMacros.scala
```diff
@@ -0,0 +1,95 @@
...
+    m.copy(readers = rports, writers = wports, readwriters = rwports)
+  }
+
+  def updateStmtRefs(s: Statement, repl: Map[String,Expression]): Statement = {
+    def updateRef(e: Expression): Expression = e map updateRef match {
+      case e: WSubField => repl getOrElse (e.serialize,e)
+      case e => e
+    }
+    def updateStmtRefs(s: Statement): Statement = s map updateStmtRefs map updateRef match {
+      case Connect(info, loc, exp) if loc == EmptyExpression => EmptyStmt
```
jackkoenig:
```scala
case Connect(info, EmptyExpression, exp) => EmptyStmt
```

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
jackkoenig: Although this is really a judgement call if it makes it more clear. I personally think so but others may disagree.

src/main/scala/firrtl/passes/AnnotateValidMemConfigs.scala
```diff
@@ -206,24 +202,23 @@ case class SRAMCompiler(
     bestRule.getValidConfig(width = memWidth, depth = memDepth).get.copy(xsplit = validConfig.xsplit, ysplit = validConfig.ysplit)
   }
 
-  def serialize() = {
+  def serialize = {
     // TODO
   }
```
jackkoenig: if this is TODO in the sense that it shouldn't be called, you can use ??? and it'll error if someone calls it

ie.
```scala
def serialize = ???
```

shunshou: That is weirdly convenient. lol

shunshou:
I think I've made all the requisite changes. Note: This needs to be merged with current master to actually get rocket-chip to simulate.

donggyukim:
@azidar @jackkoenig Yeah, let's merge this prior to my dumb PRs :)

jackkoenig:
I'm planning to merge it tonight, I squashed a couple of the commits
together

shunshou:
Sweet. Super thanks!!! :)

jackkoenig:
Replaced by #265, I'll merge it in the morning, want to run the full jenkins build before I merge and I can't seem to get the tunnel working from home. @shunshou please take a look to make sure I didn't mess anything up, All I did was merge a few commits and move a couple of things around (mainly to make sure there were no broken commits)

PR #253, opened 2016-08-26
Emit wires instead of registers for invalid randomization

zhemao:
This fixes an issue with my previous PR that I overlooked.

Before, the verilog emitter would connect registers to the invalid ports
and use random initialization on the generated registers.

This makes it impossible to separately configure randomization of the invalid ports from randomization of the initial register contents.

It also means there may be a bunch of useless registers in the taped out design. Not sure how well the synthesis tools optimize out registers that never get initialized.

It is better to just generate wires instead and use random assignment on the
wires. This makes it possible to simply leave the invalid ports floating. If they were actually being used to drive other nets, you would see an issue in simulation. If not, the synthesis tools should optimize them out.

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -316,11 +316,16 @@ class VerilogEmitter extends Emitter {
          assigns += Seq("assign ",e," = ",value,";")
       // In simulation, assign garbage under a predicate
       def garbageAssign(e: Expression, syn: Expression, garbageCond: Expression) = {
-         assigns += Seq("`ifndef RANDOMIZE_ASSIGN")
+         assigns += Seq("`ifndef RANDOMIZE_GARBAGE_ASSIGN")
          assigns += Seq("assign ", e, " = ", syn, ";")
          assigns += Seq("`else")
          assigns += Seq("assign ", e, " = ", garbageCond, " ? ", rand_string(tpe(syn)), " : ", syn, ";")
          assigns += Seq("`endif")
+      }
+      def invalidAssign(e: Expression) = {
+        assigns += Seq("`ifdef RANDOMIZE_INVALID_ASSIGN")
+        assigns += Seq("assign ", e, " = ", rand_string(tpe(e)), ";")
+        assigns += Seq("`endif")
       }
       def update_and_reset(r: Expression, clk: Expression, reset: Expression, init: Expression) = {
         // We want to flatten Mux trees for reg updates into if-trees for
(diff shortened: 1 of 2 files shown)


ben-k:
The synthesis tools generally refuse to optimize away registers for any reason, so this change is welcome.

azidar:
lgtm, I'm not sure why it was a reg to begin with, can @aswaterman weigh in?

aswaterman:
I'm missing the context; could I see a code example?

zhemao:
It's for ports that are unconnected. So for the following chisel

    class Unconnected extends Module {
        val io = new Bundle {
            val no_connect = Bool(OUTPUT)
        }
    }

This will turn into the following FIRRTL

    module Unconnected :
        input clk : Clock
        input reset : UInt<1>
        output io : { no_connect : UInt<1> }
        io.no_connect is invalid

Previously, the generated verilog would be

    module Unconnected (
        input clk,
        input reset,
        output io_no_connect
    );
        reg GEN_0;
        assign io_no_connect = GEN_0;

        initial begin
            GEN_0 <= {1{$random}};
        end
    endmodule

With this change, the verilog will be

    module Unconnected :
        input clk : Clock
        input reset : UInt<1>
        output io : { no_connect : UInt<1> }
        io.no_connect is invalid

Previously, the generated verilog would be

    module Unconnected (
        input clk,
        input reset,
        output io_no_connect
    );
        wire GEN_0;
        assign io_no_connect = GEN_0;

        assign GEN_0 = {1{$random}};
    endmodule

aswaterman:
Gotcha.  My main question is whether this affects simulation performance (particularly in VCS--in Verilator, this should all be turned off, anyway).  I think this should be merged unless it has a deleterious effect in that respect.

The reg isn't a register in the circuit sense (there's no clock); there's no issue either way with synthesis pruning.

zhemao:
So I tried running pk on simv executables compiled with both the old and new version and with invalid randomization turned off.

Old (reg) randomization: 1m 30s

New (wire) randomization: 1m 32s

No randomization: 1m 27s

So if there is an effect on performance it seems pretty negligible. Also, I think in RocketChip we will want to have the randomization turned off by default so that we can more easily see which wires are disconnected.

aswaterman:
Agreed about the performance.

I don't agree that not randomizing wires will make it easier to detect disconnected wires in rocket-chip, however.  You'll just get a bunch of z's, which quickly become x's, and x-pessimism will kill the simulation, even though it would otherwise be correct.

zhemao:
I guess I'll merge this commit in then.

As for simulation. I think we *want* disconnected wires to show up as z's. It's hard for me to see a case in which doing logic on undriven nets would not be a bug.

aswaterman:
I missed the high-level intent of the commit.  It just doesn't make sense.  Randomization _helps_ with handling unknown values.  x's just don't work the way you'd like, at least without redesigning chisel and/or firrtl.

aswaterman:
The sequence of commits starting with 87ad4220c2eba50fb1ccc854b20021776c6d34de breaks rocket-chip.

zhemao:
You mean it breaks because the randomization flags need to be changed? I am aware, that's why I had the other PR for RocketChip.

PR #248, opened 2016-08-20
Fix perf bug

donggyukim:
This fixes #222. No more stack overflows for BOOM compilation with printfs. `InferWidths` is much faster, but `ExpandWhens` is still slow. We also need to optimize `CommonSubexpressionElimination` but it's not clear how to do for now.

@azidar 
* f925ea66e7f58ac1b60383eed863de0600298e3a is helpful for `RemoveAccesses`, but dramatically increate the latency of `LowerType`. I decide to use this technique only in `RemoveAccess`.(54d47b0)
* 8a4e13d7cf253470fd5c8a54b63ee2cf6f72f155 was not very helpful to speed up `ExpandWhens`, so I didn't cherry-pick it.

@ccelio 
Please confirm this works well with BOOM. `ExpandWhens` and `CommonSubexpressionElimination` are still slow though.

EDIT: Memoization was bad for `ExpandWhens`. Now CSE is the bottleneck.

PR #256, opened 2016-08-30
Fix performance bug

donggyukim:
@jackkoenig This is the first patch for #222 from #248. Utils will be clean with the following patches.

### DIFF
--- a/src/main/scala/firrtl/Namespace.scala
+++ b/src/main/scala/firrtl/Namespace.scala
@@ -63,22 +63,16 @@ object Namespace {
   def apply(m: DefModule): Namespace = {
     val namespace = new Namespace
 
-    def buildNamespaceStmt(s: Statement): Statement =
-      s map buildNamespaceStmt match {
-        case dec: IsDeclaration =>
-          namespace.namespace += dec.name
-          dec
-        case x => x
-      }
-    def buildNamespacePort(p: Port): Port = p match {
-      case dec: IsDeclaration =>
-        namespace.namespace += dec.name
-        dec
-      case x => x
+    def buildNamespaceStmt(s: Statement): Seq[String] = s match {
+      case s: IsDeclaration => Seq(s.name)
+      case s: Conditionally => buildNamespaceStmt(s.conseq) ++ buildNamespaceStmt(s.alt)
+      case s: Block => s.stmts flatMap buildNamespaceStmt
+      case _ => Nil
     }
-    m.ports map buildNamespacePort
+    namespace.namespace ++= (m.ports collect { case dec: IsDeclaration => dec.name })
     m match {
-      case in: Module => buildNamespaceStmt(in.body)
+      case in: Module =>
+        namespace.namespace ++= buildNamespaceStmt(in.body)
       case _ => // Do nothing
     }
 
--- a/src/main/scala/firrtl/Utils.scala
+++ b/src/main/scala/firrtl/Utils.scala
@@ -151,27 +151,18 @@ object Utils extends LazyLogging {
    }
    def create_exps (n:String, t:Type) : Seq[Expression] =
       create_exps(WRef(n,t,ExpKind(),UNKNOWNGENDER))
-   def create_exps (e:Expression) : Seq[Expression] = {
-      e match {
-         case (e:Mux) => {
-            val e1s = create_exps(e.tval)
-            val e2s = create_exps(e.fval)
-            (e1s, e2s).zipped.map { (e1,e2) => Mux(e.cond,e1,e2,mux_type_and_widths(e1,e2)) }
-         }
-         case (e:ValidIf) => create_exps(e.value).map { e1 => ValidIf(e.cond,e1,tpe(e1)) }
-         case (e) => {
-            tpe(e) match {
-               case (t:UIntType) => Seq(e)
-               case (t:SIntType) => Seq(e)
-               case ClockType => Seq(e)
-               case (t:BundleType) => {
-                  t.fields.flatMap { f => create_exps(WSubField(e,f.name,f.tpe,times(gender(e), f.flip))) }
-               }
-               case (t:VectorType) => {
-                  (0 until t.size).flatMap { i => create_exps(WSubIndex(e,i,t.tpe,gender(e))) }
-               }
-            }
-         }
+   def create_exps (e:Expression) : Seq[Expression] = e match {
+      case (e:Mux) =>
+         val e1s = create_exps(e.tval)
+         val e2s = create_exps(e.fval)
+         (e1s,e2s).zipped map ((e1,e2) => Mux(e.cond,e1,e2,mux_type_and_widths(e1,e2)))
+      case (e:ValidIf) => create_exps(e.value) map (e1 => ValidIf(e.cond,e1,tpe(e1)))
+      case (e) => tpe(e) match {
+         case (_:GroundType) => Seq(e)
+         case (t:BundleType) => (t.fields foldLeft Seq[Expression]())((exps, f) =>
+            exps ++ create_exps(WSubField(e,f.name,f.tpe,times(gender(e), f.flip))))
+         case (t:VectorType) => ((0 until t.size) foldLeft Seq[Expression]())((exps, i) =>
+            exps ++ create_exps(WSubIndex(e,i,t.tpe,gender(e))))
       }
    }
    def get_flip (t:Type, i:Int, f:Orientation) : Orientation = {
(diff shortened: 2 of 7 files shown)


src/main/scala/firrtl/Utils.scala
```diff
@@ -151,27 +151,18 @@ object Utils extends LazyLogging {
...
-               }
-            }
-         }
+   def create_exps (e:Expression) : Seq[Expression] = e match {
+      case (e:Mux) =>
+         val e1s = create_exps(e.tval)
+         val e2s = create_exps(e.fval)
+         (e1s,e2s).zipped map ((e1,e2) => Mux(e.cond,e1,e2,mux_type_and_widths(e1,e2)))
+      case (e:ValidIf) => create_exps(e.value) map (e1 => ValidIf(e.cond,e1,tpe(e1)))
+      case (e) => tpe(e) match {
```
azidar: e.tpe

src/main/scala/firrtl/Utils.scala
```diff
@@ -151,27 +151,18 @@ object Utils extends LazyLogging {
...
-                  (0 until t.size).flatMap { i => create_exps(WSubIndex(e,i,t.tpe,gender(e))) }
-               }
-            }
-         }
+   def create_exps (e:Expression) : Seq[Expression] = e match {
+      case (e:Mux) =>
+         val e1s = create_exps(e.tval)
+         val e2s = create_exps(e.fval)
+         (e1s,e2s).zipped map ((e1,e2) => Mux(e.cond,e1,e2,mux_type_and_widths(e1,e2)))
+      case (e:ValidIf) => create_exps(e.value) map (e1 => ValidIf(e.cond,e1,tpe(e1)))
```
azidar: e.tpe

src/main/scala/firrtl/passes/Passes.scala
```diff
@@ -1061,24 +1061,20 @@ case class DataRef( val exp : Expression, val male : String, val female : String
...
-               (0 until t.size).flatMap(i => create_exps(SubIndex(e,i,t.tpe)))
-            case UnknownType => Seq(e)
-         }
+   def create_exps (e:Expression) : Seq[Expression] = e match {
+      case (e:Mux) =>
+         val e1s = create_exps(e.tval)
+         val e2s = create_exps(e.fval)
+         (e1s,e2s).zipped map ((e1,e2) => Mux(e.cond,e1,e2,mux_type(e1,e2)))
+      case (e:ValidIf) =>
+         create_exps(e.value) map (e1 => ValidIf(e.cond,e1,tpe(e1)))
```
azidar: e1.tpe

src/main/scala/firrtl/passes/Passes.scala
```diff
@@ -1061,24 +1061,20 @@ case class DataRef( val exp : Expression, val male : String, val female : String
...
+      case (e) => (tpe(e)) match {
```
azidar: e.tpe

donggyukim: #257 did this :)

src/main/scala/firrtl/WIR.scala
```diff
@@ -105,12 +105,9 @@ class WrappedExpression (val e1:Expression) {
...
                case (e1:WInvalid,e2:WInvalid) => true
-               case (e1:DoPrim,e2:DoPrim) => {
-                  var are_equal = e1.op == e2.op
-                  (e1.args,e2.args).zipped.foreach{ (x,y) => { if (!weq(x,y)) are_equal = false }}
-                  (e1.consts,e2.consts).zipped.foreach{ (x,y) => { if (x != y) are_equal = false }}
-                  are_equal
-               }
+               case (e1:DoPrim,e2:DoPrim) => e1.op == e2.op &&
+                  ((e1.consts zip e2.consts) forall {case (x, y) => x == y}) &&
+                  ((e1.args zip e2.args) forall {case (x, y) => weq(x, y)})
```
jackkoenig: Scala style generally case for a space around curly braces. eg.
```scala
((e1.consts zip e2.consts) forall { case (x, y) => x == y }) &&
((e1.args zip e2.args) forall { case (x, y) => weq(x, y) })
```

src/main/scala/firrtl/passes/RemoveAccesses.scala
```diff
@@ -16,90 +16,81 @@ object RemoveAccesses extends Pass {
...
+  private def hasAccess(e: Expression): Boolean = {
+    def rec_has_access(e: Expression): Boolean = e match {
+      case (e:WSubAccess) => true
+      case (e:WSubField) => rec_has_access(e.exp)
+      case (e:WSubIndex) => rec_has_access(e.exp)
+      case (e:Mux) => rec_has_access(e.cond) || rec_has_access(e.tval) || rec_has_access(e.fval)
+      case (e:ValidIf) => rec_has_access(e.cond) || rec_has_access(e.value)
+      case (e:DoPrim) => e.args exists rec_has_access
+      case _ => false
     }
```
jackkoenig: I know you're trying to removes maps to improve performance, but at what cost?

I think this makes the code less readable and more difficult to maintain. When I moved maps from a more direct function call to the current implicit implementation which clearly has some overhead, the compiler took <5% hit in performance. It could be this overhead is much larger now (since we have fixed several performance bugs) but I'm skeptical of making all of our code way more verbose for debatable improvements. 

Per an issue faced by @shunshou, @azidar and I have discussed how to make Mappers more flexible (ie. not get a match error for custom IR nodes). The current plan is to add the mapper functions as functions defined on FirrtlNode so that custom nodes can extend them themselves. The reason why the separate mappers exists is because of type erasure, so we could add such functions as mapStmt, mapExpr, mapType, etc. to each IR node. Then, the nice "map" syntax could be supported by the implicits in Mappers just calling the right virtual function. I believe that this will fix what you're trying to do here because then you can just call the right mapper and the overhead of current Mappers is gone (and actually it would be even faster because the virtual function call should in general be superior to a match statement).

src/main/scala/firrtl/WIR.scala
donggyukim: Do you want me to fix it in this PR?

src/main/scala/firrtl/passes/RemoveAccesses.scala
donggyukim: I saw some terrible stack overflows w.r.t mappers before, so I just removed it if I can do easily. I'm totally fine to revert this change if this doesn't cause any problems.

src/main/scala/firrtl/Namespace.scala
```diff
@@ -63,22 +63,16 @@ object Namespace {
...
+      case _ => Nil
     }
-    m.ports map buildNamespacePort
+    namespace.namespace ++= (m.ports collect { case dec: IsDeclaration => dec.name })
     m match {
-      case in: Module => buildNamespaceStmt(in.body)
+      case in: Module =>
+        namespace.namespace ++= buildNamespaceStmt(in.body)
       case _ => // Do nothing
     }
```
jackkoenig: While the original maps should certainly be replaced with foreach (which needs to be written), it's not actually even clear to me that this is more performant than the original implementation. Per prior conversations, I really don't think that the mappers are the cause of the stack overflows and I think removing them in general hurts readability and maintainability (although not so much in this case). 

I like the collect on the ports, although technically all ports are IsDeclarations so we could probably just do m.ports map (_.name). The way I originally wrote this I wasn't sure if we were going to need to treat the fields of Aggregate ports as declarations but that was clearly wrong.

As for the Stmt visiting, I'm a bit more divided, I like how functional it is, but I don't like manually traversing. 

A bit more on the stack overflow stuff.  Manually traversing removes a couple of small function calls from the stack but doesn't fix the fundamental problem that we are recursively walking a tree. The weird thing is, Firrtl graphs *shouldn't* be that deep of trees. It really surprises me that our standard traversals are causing stack overflows even on a design as large as BOOM. I think there is another cause here (that might be fixed in another one of your commits), but that we should make sure we should focus on that rather than on manual traversal that is still recursive and thus subject to the same stack overflows on larger designs.

jackkoenig: I will say, it would be super useful if we could do what buildNamespaceStmt is doing here and really map from Statement => Seq[String]. I'm not sure exactly how to do this but I'm wondering if we shouldn't replace the current map with "flatMap" and change "map" to mean apply the function to each member of the FirrtlNode and return a Seq of that result?

azidar:
lgtm

jackkoenig:
lgtm

PR #257, opened 2016-08-31
Refactor

donggyukim:
@jackkoenig 
* code refactoring from #248 
* depends on #256


donggyukim:
Split into #282, #278, #279, #283

PR #260, opened 2016-09-01
Mem latency pipes

donggyukim:
@azidar @jackkoenig @aswaterman 
Previously latency pipes for memory ports are generated during verilog translation, which is not desirable because they're not visible as firrtl nodes. Instead, I added an additional pass to generate latency pipes and turn all memories into combinational. Thus, in verilog translation, all memories should be combinational or blackboxes. This PR also includes better verilog generation for register mux trees and some code cleanups.

EDIT: NO DEPENDENCY

PR #263, opened 2016-09-03
Refactor2

donggyukim:
@jackkoenig @azidar This is another code refactoring based on #260.
* Remove unnecessary functions in Utils
* Rewrite `LowerTypes` 
* `ports` in MemKind is not used, so Kinds can be case object
* fix MemUtils


donggyukim:
Split into #284, #285

PR #265, opened 2016-09-06
Transform for blackboxing memories - rebased

jackkoenig:
Replaces #249 

I merged a few commits together and changed messages to leave a blank line after the first line.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -18,6 +18,8 @@ libraryDependencies += "org.scalatest" % "scalatest_2.11" % "2.2.6" % "test"
 
 libraryDependencies += "org.scalacheck" %% "scalacheck" % "1.12.5" % "test"
 
+libraryDependencies += "net.jcazevedo" %% "moultingyaml" % "0.2"
+
 // Assembly
 
 assemblyJarName in assembly := "firrtl.jar"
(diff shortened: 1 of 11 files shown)


shunshou:
Looks like it's ok? At least +1,137 −44 is the same? 

Re: Jenkins -- I'm not super sure how it works. I assume it just checks that default rocket FIRRTL passes? 

I can separately run assembly tests on this version now that master's been merged (rather than merging master locally).

jackkoenig:
I think Jenkins runs Chisel3, Firrtl, and rocket chips tests (at least for default configurations). I hope we can add other configurations and/or BOOM, etc. in the future.

If you can run the rocket-chip tests and confirm it works I can go ahead and merge (which will actually trigger jenkins anyway because it always runs when master changes)

shunshou:
I'm running with my transform turned on, and at least it's already gotten through the first couple of tests. It'll take some time to get through all of them... Jenkins won't catch when my transform is on, but at least it'll prove that my transform didn't break anything when it's not used...

shunshou:
@ben-k @mwachs5 -- this should be more or less the same as versions you guys have tried in the past, save for lots of code style + restructuring tweaks. Can you guys double check that the versions of Rocket you guys are using compile + simulate? 

I've run with the generic replseqmem branch, but if you have anything fancier to try this on before this initial version gets merged, that'd be a good sanity check.

mwachs5:
I've been keeping up to date with your branch and tried with the latest as of Sunday, but I can try with this exact version in the morning. Can you update the replseqmem rocket -chip branch (or I can if it's simple)

shunshou:
@jackkoenig the assembly tests I ran looked OK. 
@mwachs5 I bumped rocket-chip replseqmem

PR #266, opened 2016-09-06
Revamp mappers

jackkoenig:
This PR allows custom IR nodes to implement the mapper functions and thus not match error when map is called.

It also changes the implicits to value classes which saves an object creation upon calling map. Not a huge deal but yeah.


donggyukim:
LGTM

PR #267, opened 2016-09-07
use case object for Kind

donggyukim:
* `ports` in `MemKind` are not used, so `Kind`s need not be case class
*  includes #284, #285, #260


azidar:
lgtm

PR #268, opened 2016-09-07
Code Cleanup

donggyukim:
Split some commits from #257


donggyukim:
Split into #271, #273 , #275, and #276.

PR #271, opened 2016-09-07
Clean up Utils

donggyukim:


aswaterman:
An advantage of Utils.ONE is that it is a singleton and so doesn't construct a new object every time.

Rather than reverting, you could do something like Scala's BigInt, which memoizes the smallest 1K values so it doesn't have to keep constructing them.  (IntWidth would have to stop being a case class for this to work, but that's probably OK.)  See https://github.com/scala/scala/blob/v2.11.8/src/library/scala/math/BigInt.scala#L39

src/main/scala/firrtl/Utils.scala
```diff
@@ -165,72 +165,54 @@ object Utils extends LazyLogging {
...
+       case (t: BundleType) => ((t.fields foldLeft (i, None: Option[Orientation])){
+          case ((n, ret), x) if n < get_size(x.tpe) => ret match {
+            case None => (n, Some(get_flip(x.tpe,n,times(x.flip,f))))
+            case Some(_) => (n, ret)
+          }
+          case ((n, ret), x) => (n - get_size(x.tpe), ret)
+       }) match {
+          case (_, Some(p)) => p
+          case (_, None) => error("Shouldn't be here")
+       }
```
jackkoenig: Rather than matching and checking for some, you could just assign the result of the fold to some value and then check that it is a some via .get or getOrElse error("shouldn't be here") should also probably work.

jackkoenig:
So @aswaterman  do you think we should 

1. Drop the commit for now until memoization of small widths comes with it
2. Keep the commit but file an issue to memoize later

donggyukim:
@aswaterman @jackkoenig I made an issue about this.

jackkoenig:
I'm kind of leaning toward 2 at the moment

aswaterman:
yeah

PR #273, opened 2016-09-07
Clean up passes

donggyukim:
Depends on #271

### DIFF
--- a/src/main/scala/firrtl/passes/LowerTypes.scala
+++ b/src/main/scala/firrtl/passes/LowerTypes.scala
@@ -61,10 +61,10 @@ object LowerTypes extends Pass {
     */
   def loweredName(e: Expression): String = e match {
     case e: WRef => e.name
-    case e: WSubField => loweredName(e.exp) + delim + e.name
-    case e: WSubIndex => loweredName(e.exp) + delim + e.value
+    case e: WSubField => s"${loweredName(e.exp)}$delim${e.name}"
+    case e: WSubIndex => s"${loweredName(e.exp)}$delim${e.value}"
   }
-  def loweredName(s: Seq[String]): String = s.mkString(delim)
+  def loweredName(s: Seq[String]): String = s mkString delim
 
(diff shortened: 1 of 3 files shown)


src/main/scala/firrtl/passes/LowerTypes.scala
```diff
@@ -100,157 +100,139 @@ object LowerTypes extends Pass {
...
+              case _ => error("Error! lowerTypesExp called on MemKind " + 
+                              "SubField that needs to be expanded!")
+            }
+          case _ => WRef(loweredName(e), e.tpe, kind(e), gender(e))
         }
         case e: Mux => e map (lowerTypesExp)
         case e: ValidIf => e map (lowerTypesExp)
-        case (_: UIntLiteral | _: SIntLiteral) => e
         case e: DoPrim => e map (lowerTypesExp)
+        case _ => e
```
azidar: Can you not do `case _ => e`? instead, leave the match on Literal, and add `case _ => throw PassException("Shouldn't be here")`

azidar:
Once comment is addressed, it lgtm

src/main/scala/firrtl/passes/LowerTypes.scala
```diff
@@ -100,157 +100,139 @@ object LowerTypes extends Pass {
...
+            exps.size match {
+              case 1 => exps.head
+              case _ => error("Error! lowerTypesExp called on MemKind " + 
+                              "SubField that needs to be expanded!")
+            }
+          case _ => WRef(loweredName(e), e.tpe, kind(e), gender(e))
         }
         case e: Mux => e map (lowerTypesExp)
         case e: ValidIf => e map (lowerTypesExp)
-        case (_: UIntLiteral | _: SIntLiteral) => e
```
donggyukim: @azidar Was this intentional? I'll revive it.

azidar: Yeah, it will error if a WSubAccess leaks into lower types (a compiler bug). Just make the error check explicit

PR #277, opened 2016-09-07
Refactor checks

donggyukim:
Depends on #276

### DIFF
--- a/src/main/scala/firrtl/passes/Checks.scala
+++ b/src/main/scala/firrtl/passes/Checks.scala
@@ -29,746 +29,576 @@ package firrtl.passes
 
 import com.typesafe.scalalogging.LazyLogging
 
-// Datastructures
-import scala.collection.mutable.{HashMap,HashSet}
-import scala.collection.mutable.ArrayBuffer
-
 import firrtl._
 import firrtl.ir._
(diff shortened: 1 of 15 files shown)


src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -29,746 +29,576 @@ package firrtl.passes
...
+             Eq | Neq | Dshl | Dshr | And | Or | Xor | Cat =>
+          correctNum(Option(2), 0)
+        case AsUInt | AsSInt | AsClock | Cvt | Neq | Not =>
+          correctNum(Option(1), 0)
+        case Pad | Shl | Shr | Head | Tail =>
+          correctNum(Option(1), 1)
+        case Bits =>
+          correctNum(Option(1), 2)
+        case Andr | Orr | Xorr =>
+          correctNum(None,0)
```
azidar: this is a bug in the implementation we should fix, instead of `correctNum(None,0)` it should be `correctNum(Option(1), 0)` - actually, this function shouldn't take Option[Int] at all, it should just take an Int.

azidar:
Can you make a bugfix as a separate PR as per the comment? This PR lgtm

PR #279, opened 2016-09-07
Refactor Passes.scala

donggyukim:
* SHOULD BE MERGED BY #283
* JUST NEEDS LGTM


jackkoenig:
It's generally better if you have separate commits for moving a pass to a new file then actually changing the code of the pass so we can see what changed. I'm not sure if I feel strongly enough in this case to make you change it though

azidar:
lgtm

donggyukim:
Merged by #283

PR #278, opened 2016-09-07
Refactor InferWidths

donggyukim:
* SHOULD BE MERGED BY #283 
* JUST NEEDS LGTM

PR #283, opened 2016-09-08
Refactor ExpandWhens

donggyukim:
* CHANGE BASE TO MASTER BEFORE MERGE
* Includes #278, #279


src/main/scala/firrtl/passes/ExpandWhens.scala
```diff
@@ -35,9 +35,8 @@ import firrtl.PrimOps._
 import firrtl.WrappedExpression._
 
 // Datastructures
-import scala.collection.mutable.HashMap
-import scala.collection.mutable.LinkedHashMap
-import scala.collection.mutable.ArrayBuffer
+import scala.collection.mutable
+import scala.collection.mutable.{HashMap, LinkedHashMap, ArrayBuffer}
```
jackkoenig: In general, I'd rather we just only import scala.collection.mutable and then explicitly put mutable.DataStructure whenever using mutable data structures. In this case it is just a clean up so I don't see a need to change, but just for future reference

src/main/scala/firrtl/passes/ExpandWhens.scala
```diff
@@ -98,90 +89,74 @@ object ExpandWhens extends Pass {
...
-          case c: Connect =>
-            netlist(c.loc) = c.expr
-            EmptyStmt
-          case c: IsInvalid =>
-            netlist(c.expr) = WInvalid()
-            EmptyStmt
-          case s: Conditionally =>
-            val memos = ArrayBuffer[Statement]()
+          defaults: Seq[mutable.Map[WrappedExpression, Expression]],
+          p: Expression)(s: Statement): Statement = s match {
```
jackkoenig: I generally ascribe to the style the parameters make the line to long, you switch to 1 parameter per line (see: http://docs.scala-lang.org/style/declarations#classes). I've only ever seen this written in regards to class definitions (not functions) but I think following it for functions is a reasonable thing to do.
eg.

```scala
       def expandWhens(
            netlist: LinkedHashMap[WrappedExpression, Expression],
            defaults: Seq[mutable.Map[WrappedExpression, Expression]],
            p: Expression)
            (s: Statement): Statement = s match {
```

An alternative might be to draw inspiration from multiple parameter list style (http://docs.scala-lang.org/style/declarations#multiple-parameter-lists) and do something like
```scala
       def expandWhens(netlist: LinkedHashMap[WrappedExpression, Expression],
                       defaults: Seq[mutable.Map[WrappedExpression, Expression]],
                       p: Expression)
                      (s: Statement): Statement = s match {
```
Note that defaults and p are spaced to line up with the start of netlist while the parentheses of the parameter lists are also lined up

src/main/scala/firrtl/passes/ExpandWhens.scala
```diff
@@ -98,90 +89,74 @@ object ExpandWhens extends Pass {
...
       val netlist = LinkedHashMap[WrappedExpression, Expression]()
-
       // Add ports to netlist
-      m.ports foreach { port =>
-        getFemaleRefs(port.name, port.tpe, to_gender(port.direction)) foreach (ref => netlist(ref) = WVoid())
-      }
-      val bodyx = expandWhens(netlist, Seq(netlist), one)(m.body)
-
-      (netlist, simlist, bodyx)
+      netlist ++= (m.ports flatMap {case Port(_, name, dir, tpe) =>
```
jackkoenig: There should be spaces on both sides of curly braces, ie.
```scala
netlist ++= (m.ports flatMap { case Port(_, name, dir, tpe) =>
```

src/main/scala/firrtl/Utils.scala
```diff
@@ -98,28 +98,6 @@ object Utils extends LazyLogging {
...
-     else if ((e1 == we(one)) | (e2 == we(one))) one
-     else if (e1 == we(zero)) e2.e1
-     else if (e2 == we(zero)) e1.e1
-     else DoPrim(Or,Seq(e1.e1,e2.e1),Seq(),UIntType(IntWidth(1)))
-  }
-  def NOT (e1:WrappedExpression) : Expression = {
-     if (e1 == we(one)) zero
-     else if (e1 == we(zero)) one
-     else DoPrim(Eq,Seq(e1.e1,zero),Seq(),UIntType(IntWidth(1)))
-  }
```
jackkoenig: Why do these need to be moved out of Utils? Is it just that they aren't currently be used anywhere but the Emitter?

donggyukim: The main reason is this is ConstProp. Also, I think this is not general enough to put into Utils

jackkoenig: I agree with the constant propagation reasoning. I actually think shorthand for constructing Firrtl nodes is something we need more of in general though.

jackkoenig:
Other than the tabbing of the expandWhens function definition (and preferably adding a space after the curly brace), lgtm

PR #284, opened 2016-09-08
More utils cleanups

donggyukim:
* ONLY NEEDS LGTM
* SHOULD BE MERGED BY #267

### DIFF
--- a/src/main/scala/firrtl/Utils.scala
+++ b/src/main/scala/firrtl/Utils.scala
@@ -106,14 +106,14 @@ object Utils extends LazyLogging {
       val e1s = create_exps(e.tval)
       val e2s = create_exps(e.fval)
       e1s zip e2s map {case (e1, e2) =>
-        Mux(e.cond, e1, e2, mux_type_and_widths(e1,e2))
+        Mux(e.cond, e1, e2, mux_type_and_widths(e1, e2))
       }
     case (e: ValidIf) => create_exps(e.value) map (e1 => ValidIf(e.cond, e1, e1.tpe))
     case (e) => e.tpe match {
       case (_: GroundType) => Seq(e)
       case (t: BundleType) => (t.fields foldLeft Seq[Expression]())((exps, f) =>
         exps ++ create_exps(WSubField(e, f.name, f.tpe,times(gender(e), f.flip))))
-      case (t: VectorType) => ((0 until t.size) foldLeft Seq[Expression]())((exps, i) =>
+      case (t: VectorType) => (0 until t.size foldLeft Seq[Expression]())((exps, i) =>
         exps ++ create_exps(WSubIndex(e, i, t.tpe,gender(e))))
     }
   }
@@ -165,30 +165,30 @@ object Utils extends LazyLogging {
   }
 
 //============== TYPES ================
-  def mux_type (e1:Expression, e2:Expression) : Type = mux_type(e1.tpe, e2.tpe)
-  def mux_type (t1:Type, t2:Type) : Type = (t1,t2) match { 
-    case (t1:UIntType, t2:UIntType) => UIntType(UnknownWidth)
-    case (t1:SIntType, t2:SIntType) => SIntType(UnknownWidth)
-    case (t1:VectorType, t2:VectorType) => VectorType(mux_type(t1.tpe, t2.tpe), t1.size)
-    case (t1:BundleType, t2:BundleType) => BundleType((t1.fields zip t2.fields) map {
+  def mux_type(e1: Expression, e2: Expression): Type = mux_type(e1.tpe, e2.tpe)
+  def mux_type(t1: Type, t2: Type): Type = (t1, t2) match {
+    case (t1: UIntType, t2: UIntType) => UIntType(UnknownWidth)
+    case (t1: SIntType, t2: SIntType) => SIntType(UnknownWidth)
+    case (t1: VectorType, t2: VectorType) => VectorType(mux_type(t1.tpe, t2.tpe), t1.size)
+    case (t1: BundleType, t2: BundleType) => BundleType(t1.fields zip t2.fields map {
       case (f1, f2) => Field(f1.name, f1.flip, mux_type(f1.tpe, f2.tpe))
     })
     case _ => UnknownType
   }
-  def mux_type_and_widths (e1:Expression,e2:Expression) : Type =
+  def mux_type_and_widths(e1: Expression,e2: Expression): Type =
     mux_type_and_widths(e1.tpe, e2.tpe)
-  def mux_type_and_widths (t1:Type, t2:Type) : Type = {
-    def wmax (w1:Width, w2:Width) : Width = (w1,w2) match {
-      case (w1:IntWidth, w2:IntWidth) => IntWidth(w1.width max w2.width)
+  def mux_type_and_widths(t1: Type, t2: Type): Type = {
+    def wmax(w1: Width, w2: Width): Width = (w1, w2) match {
+      case (w1: IntWidth, w2: IntWidth) => IntWidth(w1.width max w2.width)
       case (w1, w2) => MaxWidth(Seq(w1, w2))
     }
     (t1, t2) match {
-      case (t1:UIntType, t2:UIntType) => UIntType(wmax(t1.width, t2.width))
-      case (t1:SIntType, t2:SIntType) => SIntType(wmax(t1.width, t2.width))
-      case (t1:VectorType, t2:VectorType) => VectorType(
+      case (t1: UIntType, t2: UIntType) => UIntType(wmax(t1.width, t2.width))
+      case (t1: SIntType, t2: SIntType) => SIntType(wmax(t1.width, t2.width))
+      case (t1: VectorType, t2: VectorType) => VectorType(
         mux_type_and_widths(t1.tpe, t2.tpe), t1.size)
-      case (t1:BundleType, t2:BundleType) => BundleType((t1.fields zip t2.fields) map {
-        case (f1, f2) => Field(f1.name,f1.flip,mux_type_and_widths(f1.tpe, f2.tpe))
+      case (t1: BundleType, t2: BundleType) => BundleType(t1.fields zip t2.fields map {
+        case (f1, f2) => Field(f1.name, f1.flip, mux_type_and_widths(f1.tpe, f2.tpe))
       })
       case _ => UnknownType
     }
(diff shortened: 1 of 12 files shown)


src/main/scala/firrtl/Utils.scala
```diff
@@ -89,21 +89,14 @@ object Utils extends LazyLogging {
   val BoolType = UIntType(IntWidth(1))
   val one  = UIntLiteral(BigInt(1),IntWidth(1))
   val zero = UIntLiteral(BigInt(0),IntWidth(1))
-  def uint (i:Int) : UIntLiteral = {
-     val num_bits = req_num_bits(i)
-     val w = IntWidth(scala.math.max(1,num_bits - 1))
-     UIntLiteral(BigInt(i),w)
+  def uint(i: Int) : UIntLiteral = {
```
jackkoenig: To fully fix this spacing it should be:
```scala
def uint(i: Int): UIntLiteral = {
```
Also, I think we should move this to an actual apply method on UIntLiteral companion object itself. That doesn't have to be done in this commit though.

src/main/scala/firrtl/Utils.scala
```diff
@@ -89,21 +89,14 @@ object Utils extends LazyLogging {
...
-     val ix = if (i < 0) ((-1 * i) - 1) else i
-     ceil_log2(ix + 1) + 1
-  }
-
-  def create_mask(dt: Type): Type = dt match {
-    case t: VectorType => VectorType(create_mask(t.tpe),t.size)
-    case t: BundleType => BundleType(t.fields.map (f => f.copy(tpe=create_mask(f.tpe))))
-    case t: UIntType => BoolType
-    case t: SIntType => BoolType
+  def req_num_bits(i: Int) : Int = {
```
jackkoenig: Same comment as above on spacing, there should be no space between the closing parentheses of the parameter list and the colon.

src/main/scala/firrtl/passes/MemUtils.scala
```diff
@@ -64,71 +63,67 @@ object toBits {
...
   }
-  def hiermask(e: Expression, maskType: Type, dataType: Type): Expression = (maskType, dataType) match {
-    case (mt: VectorType, dt: VectorType) => seqCat((0 until mt.size).reverse.map(i => hiermask(WSubIndex(e, i, mt.tpe, UNKNOWNGENDER), mt.tpe, dt.tpe)))
-    case (mt: BundleType, dt: BundleType) => seqCat((mt.fields zip dt.fields).map { case (mf, df) =>
-      hiermask(WSubField(e, mf.name, mf.tpe, UNKNOWNGENDER), mf.tpe, df.tpe) } )
+  private def hiermask(e: Expression, maskType: Type, dataType: Type): Expression = (maskType, dataType) match {
+    case (mt: VectorType, dt: VectorType) => seqCat(0 until mt.size map { i =>
+      hiermask(WSubIndex(e, i, mt.tpe, UNKNOWNGENDER), mt.tpe, dt.tpe) })
+    case (mt: BundleType, dt: BundleType) => seqCat(mt.fields zip dt.fields map { case (mf, df) =>
+      hiermask(WSubField(e, mf.name, mf.tpe, UNKNOWNGENDER), mf.tpe, df.tpe) })
```
jackkoenig: I would much rather these take more lines  for clarity: eg.
```scala
    case (mt: VectorType, dt: VectorType) => 
      seqCat((0 until mt.size) map { i =>
        hiermask(WSubIndex(e, i, mt.tpe, UNKNOWNGENDER), mt.tpe, dt.tpe)
      })
    case (mt: BundleType, dt: BundleType) =>
      seqCat((mt.fields zip dt.fields) map { case (mf, df) =>
        hiermask(WSubField(e, mf.name, mf.tpe, UNKNOWNGENDER), mf.tpe, df.tpe)
      })
```
(I also tend to prefer parentheses around range creation and zipping)

Also, in the original code there was a reverse on the range, why was it removed? is this correct without it?

src/main/scala/firrtl/passes/MemUtils.scala
```diff
@@ -64,71 +63,67 @@ object toBits {
...
-      case ex: WSubIndex => getPart(ex, ex.tpe, rhs, 0)
-      case t => error("Invalid LHS expression for fromBits!")
+      case ex @ (_: WRef | _: WSubField | _: WSubIndex) => getPart(ex, ex.tpe, rhs, 0)
+      case _ => error("Invalid LHS expression for fromBits!")
     }
     Block(fbits._2)
   }
-  def getPartGround(lhs: Expression, lhst: Type, rhs: Expression, offset: BigInt): (BigInt, Seq[Statement]) = {
+  private def getPartGround(lhs: Expression, lhst: Type, rhs: Expression,
+      offset: BigInt): (BigInt, Seq[Statement]) = {
```
jackkoenig: Once we need to definition line because it is to long, prefer just going to 1 parameter per line

src/main/scala/firrtl/passes/MemUtils.scala
```diff
@@ -64,71 +63,67 @@ object toBits {
...
-          val (tmpOffset, substmts) = getPart(WSubField(lhs, f.name, f.tpe, UNKNOWNGENDER), f.tpe, rhs, currentOffset)
-          stmts = stmts ++ substmts
-          currentOffset = tmpOffset
-        }
-        (currentOffset, stmts)
-      }
-      case t: GroundType => getPartGround(lhs, t, rhs, offset)
-      case t => error("Unknown type encountered in fromBits!")
+  private def getPart(lhs: Expression, lhst: Type, rhs: Expression,
+      offset: BigInt): (BigInt, Seq[Statement]) = lhst match {
```
jackkoenig: one parameter per line

src/main/scala/firrtl/passes/MemUtils.scala
```diff
@@ -64,71 +63,67 @@ object toBits {
...
+    case t: VectorType => (0 until t.size foldRight (offset, Seq[Statement]())){ case (i, (curOffset, stmts)) =>
+      val subidx = WSubIndex(lhs, i, t.tpe, UNKNOWNGENDER)
+      val (tmpOffset, substmts) = getPart(subidx, t.tpe, rhs, curOffset)
+      (tmpOffset, stmts ++ substmts)
+    }
```
jackkoenig: There is way too much going on the line of the case, would prefer something like:
```scala
    case t: VectorType =>
      (0 until t.size).foldRight(offset, Seq[Statement]()) { 
        case (i, (curOffset, stmts)) =>
          val subidx = WSubIndex(lhs, i, t.tpe, UNKNOWNGENDER)
          val (tmpOffset, substmts) = getPart(subidx, t.tpe, rhs, curOffset)
          (tmpOffset, stmts ++ substmts)
      }
```
I also actually prefer calling folds with `.` anyway because it helps keep the number of parentheses down, that's not a huge deal but I think splitting up the lines a bit is

src/main/scala/firrtl/passes/MemUtils.scala
```diff
@@ -139,78 +134,88 @@ object MemPortUtils {
   def flattenType(t: Type) = UIntType(IntWidth(bitWidth(t)))
 
   def defaultPortSeq(mem: DefMemory) = Seq(
-    Field("addr", Default, UIntType(IntWidth(ceil_log2(mem.depth)))),
-    Field("en", Default, UIntType(IntWidth(1))),
+    Field("addr", Default, UIntType(IntWidth(ceil_log2(mem.depth) max 1))),
```
jackkoenig: This is a bug fix, it doesn't really belong in a "clean up" commit

src/main/scala/firrtl/passes/MemUtils.scala
```diff
@@ -139,78 +134,88 @@ object MemPortUtils {
...
-  }
+  )
 
   def memToBundle(s: DefMemory) = BundleType(
-    s.readers.map(p => Field(p, Default, rPortToBundle(s))) ++
-      s.writers.map(p => Field(p, Default, wPortToBundle(s))) ++
-      s.readwriters.map(p => Field(p, Default, rwPortToBundle(s))))
+    s.readers.map(Field(_, Flip, rPortToBundle(s))) ++
+    s.writers.map(Field(_, Flip, wPortToBundle(s))) ++
+    s.readwriters.map(Field(_, Flip, rwPortToBundle(s))))
```
jackkoenig: This is also a bug fix / functional change and doesn't belong in a cleaning up commit

src/main/scala/firrtl/passes/MemUtils.scala
```diff
@@ -139,78 +134,88 @@ object MemPortUtils {
...
   
   def memToFlattenBundle(s: DefMemory) = BundleType(
-    s.readers.map(p => Field(p, Default, rPortToFlattenBundle(s))) ++
-      s.writers.map(p => Field(p, Default, wPortToFlattenBundle(s))) ++
-      s.readwriters.map(p => Field(p, Default, rwPortToFlattenBundle(s))))
+    s.readers.map(Field(_, Flip, rPortToFlattenBundle(s))) ++
+    s.writers.map(Field(_, Flip, wPortToFlattenBundle(s))) ++
+    s.readwriters.map(Field(_, Flip, rwPortToFlattenBundle(s))))
```
jackkoenig: same here

src/main/scala/firrtl/passes/Uniquify.scala
```diff
@@ -56,6 +56,7 @@ import firrtl.Mappers._
   *      to rename a
   */
 object Uniquify extends Pass {
+  import MemPortUtils.memType
```
jackkoenig: I'd prefer this import at the top of the file rather than in the object

jackkoenig: Could also just not import it and use MemPortUtils.memType as the function call below

src/main/scala/firrtl/passes/InferTypes.scala
```diff
@@ -66,16 +66,16 @@ object InferTypes extends Pass {
...
         s copy (tpe = t)
       case s: DefWire =>
-        val t = remove_unknowns(get_type(s))
+        val t = remove_unknowns(s.tpe)
         types(s.name) = t
         s copy (tpe = t)
       case s: DefNode =>
-        val sx = s map infer_types_e(types)
-        val t = remove_unknowns(get_type(sx))
+        val sx = s map infer_types_e(types) match { case sx: DefNode => sx }
```
jackkoenig: all this match does is result in a match error if the result is not a DefNode, is that what's intended?
If so, I think .asInstanceOf[DefNode] is actually more clear

donggyukim: I prefer 'match' over 'asInstanceOf', but I'll follow you :)

src/main/scala/firrtl/passes/MemUtils.scala
donggyukim: Now this is in a separate commit.

jackkoenig:
By the way, I verified no functional changes (at least on rocket.fir) for any of the style commits

PR #288, opened 2016-09-09
Revert to keyset

jackkoenig:
Minor bug in Expand Whens that causes it to process the same connection lhs twice if it appears in both the consequence and alternate of a when statement. 

Also delete ReplSeqMemTest that breaks every time code generation breaks. @shunshou, we should talk about how we can make it more robust and readd it.


shunshou:
Fine by me?

azidar:
lgtm

donggyukim:
lgtm

PR #289, opened 2016-09-09
Fix read enables for smems

donggyukim:
* CHANGE BASE TO MASTER BEFORE MERGE
* Fixes #286

@terpstra Can you confirm this works well with your examples?

Issue #272, opened 2016-09-07
Cache IntWidths to avoid redundant object creation


donggyukim:
We need to refactor the IntWidth factory to cache small `IntWidth`s. (#271)

jackkoenig:
Example in Scala for BigInt:  https://github.com/scala/scala/blob/v2.11.8/src/library/scala/math/BigInt.scala#L39

Will mean IntWidth is no longer a case class but we can DIY that

PR #290, opened 2016-09-09
Cache IntWidths to avoid redudant object creations

donggyukim:
* CHANGE BASE TO MASTER
* SHOULD BE MERGED AFTER #289 
* Addresses #272

@aswaterman looks good?

### DIFF
--- a/src/main/scala/firrtl/ir/IR.scala
+++ b/src/main/scala/firrtl/ir/IR.scala
@@ -300,8 +300,39 @@ abstract class Width extends FirrtlNode {
   }
 }
 /** Positive Integer Bit Width of a [[GroundType]] */
-case class IntWidth(width: BigInt) extends Width {
+object IntWidth {
+  private val maxCached = 1024
+  private val cache = new Array[IntWidth](maxCached + 1)
+  def apply(width: BigInt): IntWidth = {
+    if (0 <= width && width <= maxCached) {
+      val i = width.toInt
+      var w = cache(i)
+      if (w eq null) {
+        w = new IntWidth(width)
+        cache(i) = w
+      }
+      w
+    } else new IntWidth(width)
+  }
+  // For pattern matching
+  def unapply(w: IntWidth): Option[BigInt] = Some(w.width)
+}
+class IntWidth(val width: BigInt) extends Width with Product {
   def serialize: String = s"<$width>"
+  override def equals(that: Any) = that match {
+    case w: IntWidth => width == w.width
+    case _ => false
+  }
+  override def hashCode = width.toInt
+  override def productPrefix = "IntWidth"
+  override def toString = s"$productPrefix($width)"
+  def copy(width: BigInt = width) = IntWidth(width)
+  def canEqual(that: Any) = that.isInstanceOf[Width]
+  def productArity = 1
+  def productElement(int: Int) = int match {
+    case 0 => width
+    case _ => throw new IndexOutOfBoundsException
+  }
 }
 case object UnknownWidth extends Width {
   def serialize: String = ""


donggyukim:
By the way, this speeds up `ExpandWhens` a lot:
* With BOOM on a6: ~18 secs -> ~10 secs

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -211,8 +211,29 @@ abstract class Width extends FirrtlNode {
...
 }
 /** Positive Integer Bit Width of a [[GroundType]] */
-case class IntWidth(width: BigInt) extends Width {
+object IntWidth {
+  private val maxCached = 1024
+  private val cache = new Array[IntWidth](maxCached)
+  def apply(width: BigInt): IntWidth = {
+    if (1 <= width && width <= maxCached) {
+      val i = width.toInt - 1
+      Option(cache(i)) match {
```
aswaterman: You should use null, not Option, as with scala's BigInt implementation.  Since the point of this optimization is to avoid object creation, making an Option is sort of self-defeating.

jackkoenig:
I found the size of the performance improvement surprising so I looked into it a little.
it seems that in `mux_width_and_types` we needlessly create a new IntWidth and UInt/SIntType and in AND and NOT we do the same. I'm going to fix that issue in a separate pull request but the improvement from this one is encouraging!

I still think the measured performance improvement is a bit high but I'm not going to complain

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -211,8 +211,28 @@ abstract class Width extends FirrtlNode {
...
-case class IntWidth(width: BigInt) extends Width {
+object IntWidth {
+  private val maxCached = 1024
+  private val cache = new Array[IntWidth](maxCached)
+  def apply(width: BigInt): IntWidth = {
+    if (1 <= width && width <= maxCached) {
+      val i = width.toInt - 1
+      var w = cache(i)
+      if (w eq null) {
+        w = new IntWidth(width) ; cache(i) = w
```
jackkoenig: Can this be two lines rather than use a semicolon?

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -211,8 +211,28 @@ abstract class Width extends FirrtlNode {
...
+      }
+      w
+    } else new IntWidth(width)
+  }
+  // For pattern matching
+  def unapply(w: IntWidth): Option[BigInt] = Some(w.width)
+}
+class IntWidth(val width: BigInt) extends Width {
```
jackkoenig: For the purposes of the node counting function IntWidth needs to implement Product.
Also do we need to override hashcode? What is the default hashcode of a Scala class? We could easily just return the width.toInt

sdtwigg: You generally need to override hashCode when you override equals. The default hashCode is the Java Object default which is basically hashing the memory location. This is improper for a class trying to establish value equality. The following (Java) article explains why (and, although not seemingly a problem here, other pitfalls): http://www.artima.com/lejava/articles/equality.html

Since other pull requests are indicating 0-width UInts are likely to be a better supported thing, it would probably make sense to just cache 0-width IntWidth as well. Also avoids needing to adjust the index. (Make sure to then change the check to < maxCached or +1 the array size.)

To complete the case class behavior you should also override toString and provide a copy method. I think the toString implementation usually uses parts of the Product extension Jack mentioned. I believe case classes also extend Serializable but am having trouble finding details on that.

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -211,8 +211,35 @@ abstract class Width extends FirrtlNode {
...
+  override def equals(that: Any) = that match {
+    case w: IntWidth => width == w.width
+    case _ => false
+  }
+  override def hashCode = width.toInt
+  override def toString = s"IntWidth($width)"
+  def copy(width: BigInt = width) = new IntWidth(width)
+  def canEqual(that: Any) = that.isInstanceOf[Width]
+  def productArity = 1
+  def productElement(int: Int) = int match { case 0 => width }
```
donggyukim: @sdtwigg @jackbackrack @aswaterman Is it now correct?

sdtwigg: The copy method should delegate to the object factory method so you benefit from caching.

You should also ' override def productPrefix = "IntWidth" ' and consider using that in the toString implementation (although IntWidth likely isn't meant to be subclasses so not as critical).
productElement should throw an IndexOutOfBoundsException instead of a MatchError.

I'm consulting with some scala colleagues on the semantics on marking a case class as 'abstract'. It seems like it only suppresses the factory method and copy method, which you can just re-add yourself. (You will need to append {} to the new operator since you can't instantiate an abstract class so you need to 'refine' the class to a concrete class.) It is unclear whether this behavior is reliable enough to be used.

azidar:
One worry I have is this adds a lot of implementation code to IR.scala, which should be as lean as possible. For now, the benefits are worth merging this, but I'd like to keep this on the horizon to see if we can factor it out of IR.scala, especially when we clarify the distinction between WIR and IR.

jackkoenig:
I'm running Jenkins on this which checks it and the branches it depends on

jackkoenig:
Commit 144041a23329f258541e9d461b5bcc7f2284e686 should be dropped from this, no? It was squashed and merged into master as 7c38199ce7a5d9dd7e27ffbb9b2b2770b972ed94

jackkoenig:
K I rebased on master and will merge after Travis

Issue #292, opened 2016-09-09
scala.MatchError on EmptyExpression; only manifests in rocket-chip vsim/ directory.


ccelio:
I suspect some sort of memory/SRAM pass is going wrong. 

I'm using version (6a05468ed0ece1ace3019666b16f2ae83ef76ef9), but it also manifests on master (765b880).

This is on rocket-chip (using BOOM on boom branch), and everything works in the emulator/ build directory, but fails in the vsim/ directory. 

**Error message:**
````
Exception in thread "main" scala.MatchError: EmptyExpression (of class firrtl.EmptyExpression$)
	at firrtl.passes.InferTypes$.firrtl$passes$InferTypes$$infer_types_e$1(Passes.scala:172)
	at firrtl.passes.InferTypes$$anonfun$7.apply(Passes.scala:172)
	at firrtl.passes.InferTypes$$anonfun$7.apply(Passes.scala:172)
	at firrtl.Mappers$ExpMagnet$$anon$5.map(Mappers.scala:112)
	at firrtl.Mappers$ExpMap.map(Mappers.scala:144)
	at firrtl.passes.InferTypes$.firrtl$passes$InferTypes$$infer_types_e$1(Passes.scala:172)
	at firrtl.passes.InferTypes$$anonfun$firrtl$passes$InferTypes$$infer_types_s$1$5.apply(Passes.scala:213)
	at firrtl.passes.InferTypes$$anonfun$firrtl$passes$InferTypes$$infer_types_s$1$5.apply(Passes.scala:213)
	at firrtl.Mappers$StmtMagnet$$anon$3.map(Mappers.scala:54)
	at firrtl.Mappers$StmtMap.map(Mappers.scala:95)
	at firrtl.passes.InferTypes$.firrtl$passes$InferTypes$$infer_types_s$1(Passes.scala:213)
	at firrtl.passes.InferTypes$$anonfun$firrtl$passes$InferTypes$$infer_types_s$1$3.apply(Passes.scala:213)
	at firrtl.passes.InferTypes$$anonfun$firrtl$passes$InferTypes$$infer_types_s$1$3.apply(Passes.scala:213)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
	at scala.collection.immutable.List.foreach(List.scala:381)
[...]

Issue #286, opened 2016-09-08
Unconditional SeqMem reads turn into never executed reads


terpstra:
Currently, if there is no 'when' infront of the address heading into a SeqMem read address, firrtl creates a constant '0' for the read enable. It should be a constant '1'!

eg:
    val mem = SeqMem(1 << addrBits.size, Vec(beatBytes, Bits(width = 8)))
    val bar = mem.read(counter) // read address never follows counter!

terpstra:
@shunshou I've been told this was probably broken by the new firrtl SRAM extraction code.

shunshou:
Let me look into it. Can you send me an e-mail with Chisel generated FIRRTL/chirrtl for whatever config you're using? (Before its run through actual FIRRTL).

Is this for fpga configuration?

shunshou:
I'm also wondering if this is a problem with the Chirrtl to high FIRRTL pass, because I recall that it initializes memory ports to 0 and then adds in logic from whens. But if it hasn't been a problem before, then maybe it's my pass doing something wonky.

shunshou:
This is an issue that's outside of my transform, I think.

```
class GCD extends Module {
  val io = new Bundle {
    val wen = Bool(INPUT)
    val in = Bits(INPUT,1)
    val counter = UInt(INPUT, 2)
    val out = Bits(OUTPUT,1)
  }
  val mem = SeqMem(4, Bits(width = 1))
  val temp = io.counter + UInt(1)
  val bar = mem.read(temp) 
  when (io.wen) {
    mem.write(io.counter,io.in)
  }
  io.out := bar
}
```

Produces the CHIRRTL:

```
circuit GCD : 
  module GCD : 
    input clk : Clock
    input reset : UInt<1>
    output io : {flip wen : UInt<1>, flip in : UInt<1>, flip counter : UInt<2>, out : UInt<1>}
    
    io is invalid
    smem mem : UInt<1>[4] @[GCD.scala 14:19] 
    node T_7 = add(io.counter, UInt<1>("h01")) @[GCD.scala 15:25] 
    node temp = tail(T_7, 1) @[GCD.scala 15:25] 
    read mport bar = mem[temp], clk
    when io.wen : @[GCD.scala 18:17] 
      write mport T_8 = mem[io.counter], clk
      T_8 <= io.in
      skip @[GCD.scala 18:17] 
    io.out <= bar @[GCD.scala 21:10] 
```

Which produces High FIRRTL:

```
[...]

PR #294, opened 2016-09-10
Bug fix -- remove all empty expressions after ReplSeqMem passes

shunshou:
Addresses #292 
@ccelio can you try this and see if it works  (compiles + simulates)?
@jackkoenig @azidar -- do you think it's better to do something like this or just case out WSubField, WSubIndex, and EmptyExpression (assuming those are the only 3 possible?)

### DIFF
--- a/src/main/scala/firrtl/passes/UpdateDuplicateMemMacros.scala
+++ b/src/main/scala/firrtl/passes/UpdateDuplicateMemMacros.scala
@@ -43,13 +43,23 @@ object MemTransformUtils {
   }
 
   def updateStmtRefs(s: Statement, repl: Map[String, Expression]): Statement = {
-    def updateRef(e: Expression): Expression = e map updateRef match {
-      case e: WSubField => repl getOrElse (e.serialize, e)
-      case e => e
+    def updateRef(e: Expression): Expression = e map updateRef match { 
+      case e => repl getOrElse (e.serialize, e) 
+    }
+    def hasEmptyExpr(stmt: Statement): Boolean = {
+      var foundEmpty = false
+      def testEmptyExpr(e: Expression): Expression = {
+        e map testEmptyExpr match {
+          case EmptyExpression => foundEmpty = true
+          case _ =>
+        }
+        e // map must return; no foreach
+      }
+      stmt map testEmptyExpr
+      foundEmpty
     }
     def updateStmtRefs(s: Statement): Statement = s map updateStmtRefs map updateRef match {
-      case Connect(info, EmptyExpression, exp) => EmptyStmt 
-      case Connect(info, WSubIndex(EmptyExpression, _, _, _), exp)  => EmptyStmt
+      case c: Connect if hasEmptyExpr(c) => EmptyStmt
       case s => s
     }
     updateStmtRefs(s)
--- a/src/test/scala/firrtlTests/ReplSeqMemTests.scala
+++ b/src/test/scala/firrtlTests/ReplSeqMemTests.scala
@@ -16,9 +16,54 @@ class ReplSeqMemSpec extends SimpleTransformSpec {
     new passes.ReplSeqMem(TransID(-2)),
     new MiddleFirrtlToLowFirrtl(),
     (new Transform with SimpleRun {
-     def execute(c: ir.Circuit, a: AnnotationMap) = run(c, passSeq) }),
+     def execute(c: ir.Circuit, a: AnnotationMap) = run(c, passSeq) } ),
     new EmitFirrtl(writer)
   )
+
+  "ReplSeqMem" should "generate blackbox wrappers" in {
+    val input = """
+circuit Top : 
+  module Top : 
+    input clk : Clock
+    input reset : UInt<1>
+    input head_ptr : UInt<5>
+    input tail_ptr : UInt<5>
+    input wmask : {takens : UInt<2>, history : UInt<14>, info : UInt<14>}
+    output io : {backend : {flip allocate : {valid : UInt<1>, bits : {info : {takens : UInt<2>, history : UInt<14>, info : UInt<14>}}}}, commit_entry : {valid : UInt<1>, bits : {info : {takens : UInt<2>, history : UInt<14>, info : UInt<14>}}}}
+    output io2 : {backend : {flip allocate : {valid : UInt<1>, bits : {info : {takens : UInt<2>, history : UInt<14>, info : UInt<14>}}}}, commit_entry : {valid : UInt<1>, bits : {info : {takens : UInt<2>, history : UInt<14>, info : UInt<14>}}}}
+
+    io is invalid
+    io2 is invalid
+
+    smem entries_info : {takens : UInt<2>, history : UInt<14>, info : UInt<14>}[24]
+    when io.backend.allocate.valid :
+      write mport W = entries_info[tail_ptr], clk
+      W <- io.backend.allocate.bits.info
+
+    read mport R = entries_info[head_ptr], clk
+    io.commit_entry.bits.info <- R
+
+    smem entries_info2 : {takens : UInt<2>, history : UInt<14>, info : UInt<14>}[24]
+    when io2.backend.allocate.valid :
+      write mport W1 = entries_info2[tail_ptr], clk
+      when wmask.takens :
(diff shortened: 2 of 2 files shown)


ccelio:
Looks like it's working.

azidar:
lgtm

shunshou:
Again a function test as opposed to some global FIRRTL thing?

azidar:
No, a global firrtl thing with a slimmed down circuit of Chris's example, which would have previously caused an Empty expression to leak from your pass.

shunshou:
Haha ok then I actually need to figure out the exact part of his code that caused the problem :( would be nice if there was an easier way to figure it out...

donggyukim:
@shunshou This probably happens with branch predictors having smems?

shunshou:
And that's somehow different between normal rocket and boom? Anyways, no worries. I'll dig it up.

ccelio:
Rockets branch predictor is flip flop based. Boom has a backing branch predictor which is much, much bigger and is stored in synchronous memories.

shunshou:
OK I'm about to push a version with tests. Something like:

```
circuit Top : 
  module Top : 
    input clk : Clock
    input reset : UInt<1>
    input head_ptr : UInt<5>
    input tail_ptr : UInt<5>
    input wmask : {takens : UInt<2>, history : UInt<14>, info : UInt<14>}
    output io : {backend : {flip allocate : {valid : UInt<1>, bits : {info : {takens : UInt<2>, history : UInt<14>, info : UInt<14>}}}}, commit_entry : {valid : UInt<1>, bits : {info : {takens : UInt<2>, history : UInt<14>, info : UInt<14>}}}}
    output io2 : {backend : {flip allocate : {valid : UInt<1>, bits : {info : {takens : UInt<2>, history : UInt<14>, info : UInt<14>}}}}, commit_entry : {valid : UInt<1>, bits : {info : {takens : UInt<2>, history : UInt<14>, info : UInt<14>}}}}

    io is invalid
    io2 is invalid

    smem entries_info : {takens : UInt<2>, history : UInt<14>, info : UInt<14>}[24]
    when io.backend.allocate.valid :
      write mport W = entries_info[tail_ptr], clk
      W <- io.backend.allocate.bits.info

    read mport R = entries_info[head_ptr], clk
    io.commit_entry.bits.info <- R

----

    smem entries_info2 : {takens : UInt<2>, history : UInt<14>, info : UInt<14>}[24]
    when io2.backend.allocate.valid :
      write mport W1 = entries_info2[tail_ptr], clk
      when wmask.takens :
        W1.takens <- io.backend.allocate.bits.info.takens
      when wmask.history :
        W1.history <- io.backend.allocate.bits.info.history
      when wmask.info :
        W1.info <- io.backend.allocate.bits.info.history
      
    read mport R1 = entries_info2[head_ptr], clk
    io2.commit_entry.bits.info <- R1
```

Slight digression: 

I looked at the **BranchReorderBuffer** module [which I think is the source of this FIRRTL bug], and the way the CHIRRTL is written would imply memory reads wouldn't work, as in #286, until Donggyu's changes get merged in.

Also, is there any desire to change the Chisel memory spec? Right now, you can only have write masks when data types are Vecs, but is there any meaning to have write masks when data types are other aggregates like Bundles? Chisel would just need to check somehow that the # of mask bits = # of elements in the aggregate, and the output conf file would need to be more general. @aswaterman ? 

I'm just saying that b/c the SeqMem in BOOM has data type Bundle. Even though FIRRTL always has a write mask, my check does verify that the mask is redundant and gets rid of it prior to black boxing. Theoretically, you could write valid CHIRRTL like in the second part of the code snippet and it should result in non-redundant wmasks, even if Chisel currently doesn't support it. 

Anyways, courtesy of Albert's awesome Utils, the generated Verilog is correct even for the 2nd half of the snippet, but the mask_gran for the conf files is incorrect, because you can't have a single # in that case [I just always assume it's # bits in data / # of elements in aggregate type = 30/3 in this case, whereas mask_gran should actually be {2,14,14}].

Should I then be erroring out b/c the Chirrtl doesn't match Chisel3 spec? Or... ?

donggyukim:
With tackling #165, I was frustrated that the current FIRRTL memory model cannot prevent aggregate data type memories from spliting. With @shunshou's example, the desired data type for `entries_info` after lowering types is `UInt<30>` without any write masks. However, we want to have write masks by lowering vector type memories, but there's no way to carry write mask information until verilog generation.

The problem is the data types for sequential memories are unbounded. These data types are very unlikely to be bundles of vectors with non-trivial write mask grans. We may also want FIRRTL memories to contain mask grans after lowering types. With these fixes, we can just replace memories with blackboxes as well as prevent them from splitting.

#165 is the biggest issue for Strober because it needs identical SRAM models both for FPGA and ASIC. It seems like @shunshou has an issue for the same reasons.

shunshou:
Any suggestions on how to handle unbounded datatypes or imposing some looser restrictions? 

That's why Chisel3 limits wmasks to Vecs as the top level aggregates, but it's just slightly weird that FIRRTL doesn't follow the same rules.

Well, courtesy of Albert's Utils, my current black boxing pass does something like:

[Concatenates bundle elements and properly handles at least Bundles of ground types, except the conf file is wrong...]

```
module entries_info2(
  input  [4:0] R0_addr,
  input   R0_en,
  input   R0_clk,
  output [1:0] R0_data_takens,
  output [13:0] R0_data_history,
  output [13:0] R0_data_info,
  input  [4:0] W0_addr,
  input   W0_en,
  input   W0_clk,
  input  [1:0] W0_data_takens,
  input  [13:0] W0_data_history,
  input  [13:0] W0_data_info,
  input   W0_mask_takens,
  input   W0_mask_history,
  input   W0_mask_info
);
  wire [4:0] entries_info2_ext_R0_addr;
  wire  entries_info2_ext_R0_en;
  wire  entries_info2_ext_R0_clk;
  wire [29:0] entries_info2_ext_R0_data;
  wire [4:0] entries_info2_ext_W0_addr;
  wire  entries_info2_ext_W0_en;
  wire  entries_info2_ext_W0_clk;
  wire [29:0] entries_info2_ext_W0_data;
  wire [2:0] entries_info2_ext_W0_mask;
  wire [27:0] GEN_0;
  wire [1:0] GEN_1;
  entries_info2_ext entries_info2_ext (
    .R0_addr(entries_info2_ext_R0_addr),
    .R0_en(entries_info2_ext_R0_en),
    .R0_clk(entries_info2_ext_R0_clk),
    .R0_data(entries_info2_ext_R0_data),
    .W0_addr(entries_info2_ext_W0_addr),
    .W0_en(entries_info2_ext_W0_en),
    .W0_clk(entries_info2_ext_W0_clk),
    .W0_data(entries_info2_ext_W0_data),
    .W0_mask(entries_info2_ext_W0_mask)
  );
  assign R0_data_takens = entries_info2_ext_R0_data[29:28];
  assign R0_data_history = entries_info2_ext_R0_data[27:14];
  assign R0_data_info = entries_info2_ext_R0_data[13:0];
  assign entries_info2_ext_R0_addr = R0_addr;
  assign entries_info2_ext_R0_en = R0_en;
  assign entries_info2_ext_R0_clk = R0_clk;
  assign entries_info2_ext_W0_addr = W0_addr;
  assign entries_info2_ext_W0_en = W0_en;
  assign entries_info2_ext_W0_clk = W0_clk;
  assign entries_info2_ext_W0_data = {W0_data_takens,GEN_0};
  assign entries_info2_ext_W0_mask = {W0_mask_takens,GEN_1};
  assign GEN_0 = {W0_data_history,W0_data_info};
  assign GEN_1 = {W0_mask_history,W0_mask_info};
endmodule
```

ccelio:
FYI: I'm amenable to changing my style regarding synchronous memories... there's not a lot of documentation to go off of, so I'm stumbling in the dark here.  The blessing and the curse of branch predictors is I may actually be getting the wrong functional behavior, but the processor is able to forge on ahead anyways. >.<

shunshou:
I don't think you'd need to change the style of your SeqMem's. I just think that it'd be nice if Chisel3 could support BundleTypes with legit write masks. The fact that you don't need a write mask makes this a non-issue, except for my failing to catch it as an edge case. 

Hmm... you should probably check to see if there's an example where not reading kills you? I'm fairly certain that in master, you'd just never be able to properly read from memories...

Issue #87, opened 2016-03-09
Supporting Verilog's "inout" construct


azidar:
There is a problem where existing Verilog boxes have ports that can be driven by different places, i.e. I2C bus. This is represented by the Verilog "inout" construct. While we don't want to directly support modeling tri-state buses, we do need a way to connect black boxed Verilog designs within FIRRTL to other black boxed Verilog designs.

As a result, we need a way to indicate that two ports are connected, but that we can make no other assumptions about it (i.e. you cannot buffer the signal, or introduce a mux, etc.)

My proposal is to add two new constructs:
1) An Analog type. An analog type has the property that no two Analog types are equivalent. Thus, the following FIRRTL causes a type error:
```
wire x : Analog
wire y : Analog
x <= y   ;ERROR: x and y do not have equivalent types
y <= x   ;ERROR: x and y do not have equivalent types
```
Another benefit is a bundle type that contains an Analog type will also fail when assigning to/from:
```
wire x : {a : Analog, b : UInt<32>}
wire y : {a : Analog, b : UInt<32>}
x <= y   ;ERROR: x and y do not have equivalent types
y <= x   ;ERROR: x and y do not have equivalent types
```
Finally, no primitive operation, mux, or node will accept an Analog-typed argument.

2) An `attach` operator. The `attach` operator takes a variable number of Analog-typed expressions, and attaches them together.
```
wire x : Analog
wire y : Analog
attach(x,y)
```
Assuming `a`, `b`, and `c` are all Analog types, the following two examples are equivalent:
```
attach(a,b)
attach(b,c)
```
and
```
attach(a,b,c)
```
Middle FIRRTL and Low FIRRTL will have the restriction that no two `attach` operators will have the same referenced expressions.

These two constructs give us the following property that Analog types will never interact with our existing connection semantics. This is for good reason - and Analog signal's direction is dynamic (or non-existent), while the connection semantics assumes a statically known direction. Rather than change our existing connection semantics and worry about how it interacts with other constructs, its better to never cause an interaction.

Thoughts?
@aswaterman @jackkoenig @jackbackrack @sdtwigg and any others?

aswaterman:
I haven't thought through the implications fully, but I think you're going to want to add the ability to attach analog types to other FIRRTL types.  Otherwise, people will just write black boxes that consist only of a single assign statement to connect them together, as a means to get around the FIRRTL limitation.  This is hokey and provides even less safety.

azidar:
[...]

PR #295, opened 2016-09-10
Spec features added: AnalogType and Attach

azidar:
AnalogType(width: Width):
  - Concrete syntax:
      wire x: Analog<10>
  - New groundtype, very restricted in use cases.
  - Can only declare ports and wires with Analog type
  - Analog types are never equivalent, thus if x and y have Analog
    types: x <= y is never legal.

Attach(info: Info, source: Expression, exprs: Seq[Expression]):
  - Concrete syntax:
      attach x to (y, z)
  - New statement
  - Source must be Analog and either a port or wire (not instance port)
  - Exprs must have an Analog type reference to an instance port
  - Source and exprs must have identical widths

Included WDefInstanceConnector to enable emission of Verilog inout
Should be mostly feature complete.
Need to update spec if PR gets accepted.


azidar:
See #87

src/main/scala/firrtl/Emitter.scala
```diff
@@ -115,6 +118,9 @@ class VerilogEmitter extends Emitter {
         val wx = long_BANG(t) - 1
         if (wx > 0) w write s"[$wx:0]"
       case ClockType =>
+      case t: AnalogType =>
+        val wx = long_BANG(t) - 1
```
donggyukim: `long_BANG` -> `bitWidth`. `long_BANG` will disappear.

src/main/scala/firrtl/Emitter.scala
```diff
@@ -425,21 +416,28 @@ class VerilogEmitter extends Emitter {
...
         }
       }
 
       def build_streams(s: Statement): Statement = s map build_streams match {
         case (s: DefWire) => 
           declare("wire",s.name,s.tpe)
           val e = wref(s.name,s.tpe)
-          assign(e,netlist(e))
+          netlist.get(e) match {
+            case Some(n) => assign(e,n)
```
donggyukim: Space after comma.

src/main/scala/firrtl/Visitor.scala
```diff
@@ -139,6 +139,8 @@ class Visitor(infoMode: InfoMode) extends FIRRTLBaseVisitor[FirrtlNode] {
           case "SInt" => if (ctx.getChildCount > 1) SIntType(IntWidth(string2BigInt(ctx.IntLit.getText)))
           else SIntType(UnknownWidth)
           case "Clock" => ClockType
+          case "Analog" => if (ctx.getChildCount > 1) AnalogType(IntWidth(string2BigInt(ctx.IntLit.getText)))
+          else AnalogType(UnknownWidth)
```
donggyukim: This can be:
```
case "Analog" if (ctx.getChildCount > 1) => AnalogType(IntWidth(string2BigInt(ctx.IntLit.getText)))
case "Analog" => AnalogType(UnknownWidth)
```
But you don't need to fix it :)

src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -585,7 +619,15 @@ object CheckWidths extends Pass {
 
     def check_width_s(minfo: Info, mname: String)(s: Statement): Statement = {
       val info = get_info(s) match { case NoInfo => minfo case x => x }
-      s map check_width_e(info, mname) map check_width_s(info, mname)
+      s map check_width_e(info, mname) map check_width_s(info, mname) match {
+        case Attach(info, source, exprs) => 
+          exprs foreach ( e =>
+            if(width_BANG(e) != width_BANG(source))
```
donggyukim: * Space after `if`
* `width_BANG` -> `bitWidth`

src/main/scala/firrtl/WIR.scala
```diff
@@ -147,6 +151,7 @@ class WrappedType(val t: Type) {
       case (_: UIntType, _: UIntType) => true
       case (_: SIntType, _: SIntType) => true
       case (ClockType, ClockType) => true
+      case (_: AnalogType, _: AnalogType) => false
```
donggyukim: Two AnalogTypes can never be the same? This can be confusing later, so can you comment this?

src/main/scala/firrtl/WIR.scala
```diff
@@ -147,6 +151,7 @@ class WrappedType(val t: Type) {
...
       case (t1: VectorType, t2: VectorType) =>
         t1.size == t2.size && wt(t1.tpe) == wt(t2.tpe)
       case (t1:BundleType,t2:BundleType) =>
```
donggyukim: Add space after comma when you fix this commit :)

src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -295,13 +303,12 @@ object CheckTypes extends Pass with LazyLogging {
...
       case (t) => true
     }
-
     def check_types_primop(info: Info, mname: String, e: DoPrim) {
       def all_same_type (ls:Seq[Expression]) {
         if (ls exists (x => wt(ls.head.tpe) != wt(e.tpe)))
           errors append new OpNotAllSameType(info, mname, e.op.serialize)
       }
-      def all_ground (ls: Seq[Expression]) {
+      def all_int (ls: Seq[Expression]) {
```
donggyukim: `all_int` -> `allInt`

src/main/scala/firrtl/LoweringCompilers.scala
```diff
@@ -150,7 +150,8 @@ class EmitVerilogFromLowFirrtl (val writer: Writer) extends Transform with Simpl
       passes.SplitExpressions,
       passes.CommonSubexpressionElimination,
       passes.DeadCodeElimination,
-      passes.VerilogRename)
+      passes.VerilogRename,
+      passes.VerilogPrep)
```
donggyukim: It's not clear to me why `VerilogPrep` is present here. Is it different from lowering types?

jackkoenig:
A few high-level questions:

1. Why AnalogType instead of just Analog? (analogous to how concrete syntax for UIntType and SIntType is UInt and SInt)
1. Why add a new attach statement instead of using connect? Is the idea that the widths have to be identical unlike in connect where you can connect from smaller width to larger width?
1. Why does attach go from Expression to Seq[Expression]? If you need to attach lots of things could you not just make multiple statements?
1. Can you have Vecs/Bundles of AnalogType? If so it might be useful to allow attaching aggregates. In particular I'm thinking of the day when we might emit Vecs of GroundType as arrays. This isn't a huge deal but something to keep in mind.

azidar:
> Why AnalogType instead of just Analog? (analogous to how concrete syntax for UIntType and SIntType is UInt and SInt)

I wrote it wrong, it is Analog<3>. I'll edit the comment.

> Why add a new attach statement instead of using connect? Is the idea that the widths have to be identical unlike in connect where you can connect from smaller width to larger width?

So the big problem is Analog is undirectioned, so just the concept of a Connection is at odds with the concept of Analog. This way, Analog type doesn't propagate anywhere, it is limited to other Analog types and Attach.

> Why does attach go from Expression to Seq[Expression]? If you need to attach lots of things could you not just make multiple statements?

Its so that you can lower multiple attach statements into one during high->mid transformation. This isn't written yet, whoops. I'll add it before I merge (I knew there was something I forgot). That way, you have the guarantee that the source is being used once and only once (much easier to emit).

> Can you have Vecs/Bundles of AnalogType? If so it might be useful to allow attaching aggregates. In particular I'm thinking of the day when we might emit Vecs of GroundType as arrays. This isn't a huge deal but something to keep in mind.

So you say I should allow aggregates where the leaf types must all be Analog, and otherwise all types are identical? Perhaps... So I guess it would get split up (like memories) during lowering?


Another thing to think about is how Attach interacts with Conditionally, and right now, they 100% don't interact. An attach is final, irregardless of where it is declared. You can't mux an Analog type (or conditionally attach an Analog type)

src/main/scala/firrtl/WIR.scala
azidar: That's correct. This is how Analog totally skips out of the Firrtl type system. The only way Analog can play with another Analog component is through Attach. Otherwise, we'd need to special case it during ExpandWhens, Lowering, ExpandConnects, etc.

src/main/scala/firrtl/LoweringCompilers.scala
azidar: There has been talk of rewriting the Verilog Emitter as a sequence of Firrtl passes, and then a very simple Verilog Emitter (as the Firrtl graph is already in a Verilog-esque state). This is me starting that by simplifying the Instance emission (which was also required to emit inout properly)

azidar:
Another important property that needs to be implemented is two Attach statements cannot share expressions (but they can share sources). Thus, every Analog signal can only be associated with one source.

jackkoenig:
> So the big problem is Analog is undirectioned, so just the concept of a Connection is at odds with the concept of Analog. This way, Analog type doesn't propagate anywhere, it is limited to other Analog types and Attach.

That makes sense. We *could* just overload the connection operator for Analog types. Connection is directional, but it does connect things in both directions. With regards to conditional connections turing into muxes, it could just error. This does mean that Analog types couldn't be connected in conditional blocks which seems reasonable but might be hard to target for generators?

Regardless a different operator for the undirectioned Analog type is reasonable. I'm not sure what attach would mean with regards to the other GroundTypes though since those types are directioned.

>> Why does attach go from Expression to Seq[Expression]? If you need to attach lots of things could you not just make multiple statements?

> Its so that you can lower multiple attach statements into one during high->mid transformation. This isn't written yet, whoops. I'll add it before I merge (I knew there was something I forgot). That way, you have the guarantee that the source is being used once and only once (much easier to emit).

I'm not sure I follow. Since Analog is undirectioned, what does "source" mean in this context? Would attach make more sense as Seq[Expression] since they're all attached to each other?

>>Can you have Vecs/Bundles of AnalogType? If so it might be useful to allow attaching aggregates. In particular I'm thinking of the day when we might emit Vecs of GroundType as arrays. This isn't a huge deal but something to keep in mind.

> So you say I should allow aggregates where the leaf types must all be Analog, and otherwise all types are identical? Perhaps... So I guess it would get split up (like memories) during lowering?

For Vecs the leaf types always have to be the same but for Bundles that's not the case. Regardless, if we don't use connect on Analog then in practice we would only ever have Bundles of all Analog types being attached or Bundles of no Analog types being connected.

> Another thing to think about is how Attach interacts with Conditionally, and right now, they 100% don't interact. An attach is final, irregardless of where it is declared. You can't mux an Analog type (or conditionally attach an Analog type)

Yeah I'm not sure how I see mux/conditional attaching of Analog types working. If we overloaded connect to work on Analog types then we could error if these connections are located in conditional blocks. Otherwise, attach can just be either illegal in conditional blocks or always unconditional.

shunshou:
Have you guys looked into common use cases for bidirectional ports? i.e. tristate buffer with Verilog similar to: https://www.altera.com/support/support-resources/design-examples/design-software/verilog/ver_bidirec.highResolutionDisplay.html

I'm not sure what other use cases @mwachs5 and others have, but the place I've had to use inout = for I2C interfaces. In that case, the inout is only at the interface, and you separate out the connection into input / output down the road. You have to have some notion of high impedance 'Z' in this case... so maybe, you should just blackbox a tristate buffer that wraps said verilog? 

1)
Signal for indicating which direction it's currently being used in [can only be one direction at a time]
2)
Signal that takes analog -> digital input
3)
Signal that takes digital output [output mode] or high 'Z' [input mode] -> analog

At least in the Chisel world, I don't think there's any other "analog"-y thing you'd want to implement?

azidar:
> I'm not sure what attach would mean with regards to the other GroundTypes though since those types are directioned.

Only the attach source can be an arbitrary ground type. The Seq[Exp] must all be analog.

> I'm not sure I follow. Since Analog is undirectioned, what does "source" mean in this context? Would attach make more sense as Seq[Expression] since they're all attached to each other?

That's actually what I started with, but it turns out in Verilog you can't just declare two inout ports and connect them in a way that they "pass through". There needs to be only one "source", like a wire declaration or an inout port, which can connect to multiple instance inout ports.

> Regardless, if we don't use connect on Analog then in practice we would only ever have Bundles of all Analog types being attached or Bundles of no Analog types being connected.

Exactly. :)

> Otherwise, attach can just be either illegal in conditional blocks or always unconditional.

Yeah, my vote is unconditional.

Re: shunshou

> Have you guys looked into common use cases for bidirectional ports?

Yes, this is the exact use case we are supporting. This proposal is not to be able to directly model a tristate buffer in Firrtl (the Firrtl interpreter doesn't need to add a high impedance state). Instead, we just want to allow ExtModules which implement tristate buffers to be able to talk to each other. These concepts are really just a formalization of allowing some signal with unknown properties to connect between instance ports.

mwachs5:
Great suggestion @shunshou and you're spot on with the use case, but I think it's even more critical to be able to hook up black boxes that are using inouts. I'd rather have that minimal functionality now.

azidar:
> then it makes sense to be able to black box all of the analog macros and connect things up in Chisel...

Yeah, that was the thought at least :)

shunshou:
Yeah, now that I think about it, being able to hook up black boxes that are using inouts = all analog IP -- will actually be really important for CRAFT-related future tapeouts. Digital top just seems so much easier when we have these complicated systems, and it'd get rid of a lot of the scripting overhead I had to contend with this summer -- even simple things like adding pads to the top level digital design. 

Cool cool :)

ben-k:
It took me a little while to understand that the intent of this is for tristate/inout signals specifically - I think the name may not be quite right.  There certainly exist unidirectional analog signals that can be wired to one another and even muxed (with transmission gates) - the distinction between regular digital signals is that it doesn't make sense to register them or to perform arbitrary logic on them.  

Perhaps call it InOut or TriState instead?

azidar:
> It took me a little while to understand that the intent of this is for tristate/inout signals specifically

That is the original need, but hopefully this feature can be used for other analog signals as well. However, I do get the point that it is not clear that "Analog" should be used for tristate stuff.

Perhaps "NonDigital"? Or "Other"?

shunshou:
Only one port to a tristate is inout, and that port is inherently analog. IMO inout == analog. If Chisel can interface with it directly, it needs to support 'Z'. If Chisel can't interface with it directly (i.e. No digital signals are connected in Chisel to it -- and it has to go through a black box, as the use case is here), then no high Z is needed and really, there's no functional need to propagate anything, as best as I can tell.

aswaterman:
I'd stick with Analog as it is not inherently limited to inout

ben-k:
@azidar Thanks for clarifying.  If the intent is that this can (in the future) be a general-purpose class, one use case of which is for tristates, then the name seems fine.

azidar:
@donggyukim @jackkoenig this is ready for review

src/main/scala/firrtl/WIR.scala
donggyukim: Sorry for confusion, but I meant comments in code :)

donggyukim:
LGTM

PR #296, opened 2016-09-12
Fix bits type redo

jackkoenig:
Resurrected #174. Split out Debug Compilers for a separate PR.

This commit fixes a host of minor bugs and adds the appropriate tests to find them. These bug fixes are required in order for the compiler to be rerun after lowering (eg. for transformations that want to operate on low FIRRTL then rerun the compiler).

I wasn't sure before, and I'm definitely not sure now whether or not the constant propogation of pad widths I added to Legalize should be there. This is a linting warning so it's not clear that it is a required "legalization" but the warning it was exercising used to be an error in Verilator so it's probably a good idea to not exercise it. I'm also not sure if it got fixed elsewhere somehow over the last few months but I wouldn't be surprised.

### DIFF
--- a/src/main/scala/firrtl/passes/Passes.scala
+++ b/src/main/scala/firrtl/passes/Passes.scala
@@ -211,23 +211,22 @@ object ExpandConnects extends Pass {
 // TODO replace UInt with zero-width wire instead
 object Legalize extends Pass {
   def name = "Legalize"
-  def legalizeShiftRight (e: DoPrim): Expression = e.op match {
-    case Shr => {
-      val amount = e.consts.head.toInt
-      val width = long_BANG(e.args.head.tpe)
-      lazy val msb = width - 1
-      if (amount >= width) {
-        e.tpe match {
-          case t: UIntType => UIntLiteral(0, IntWidth(1))
-          case t: SIntType =>
-            DoPrim(Bits, e.args, Seq(msb, msb), SIntType(IntWidth(1)))
-          case t => error(s"Unsupported type ${t} for Primop Shift Right")
-        }
-      } else {
-        e
+  private def legalizeShiftRight(e: DoPrim): Expression = {
+    require(e.op == Shr)
+    val amount = e.consts.head.toInt
+    val width = bitWidth(e.args.head.tpe)
+    lazy val msb = width - 1
+    if (amount >= width) {
+      e.tpe match {
+        case UIntType(_) => UIntLiteral(0, IntWidth(1))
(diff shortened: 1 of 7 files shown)


src/main/scala/firrtl/passes/Passes.scala
```diff
@@ -242,11 +241,12 @@ object Legalize extends Pass {
     }
   }
   def run (c: Circuit): Circuit = {
-    def legalizeE (e: Expression): Expression = {
-      e map (legalizeE) match {
-        case e: DoPrim => legalizeShiftRight(e)
-        case e => e
+    def legalizeE(expr: Expression): Expression = expr map legalizeE match {
+      case prim: DoPrim => prim.op match {
```
azidar: unapply the doprim: DoPrim(Shr, ....) etc

src/main/scala/firrtl/passes/Passes.scala
```diff
@@ -228,16 +228,18 @@ object Legalize extends Pass {
...
-  def legalizeConnect(c: Connect): Statement = {
+  private def legalizeConnect(c: Connect): Statement = {
     val t = c.loc.tpe
     val w = long_BANG(t)
-    if (w >= long_BANG(c.expr.tpe)) c
-    else {
-      val newType = t match {
-        case _: UIntType => UIntType(IntWidth(w))
-        case _: SIntType => SIntType(IntWidth(w))
+    if (w >= long_BANG(c.expr.tpe)) {
```
azidar: inferWidth instead of long_BANG

azidar:
lgtm after style fixes

src/main/scala/firrtl/passes/Passes.scala
jackkoenig: I didn't unapply the DoPrim because it gets passed to legalizeShiftRight. I could make the legalize functions (more come in a later commit) accept the args, consts, and tpe from the DoPrim but that seems a bit excessive.

jackkoenig:
I modified bitWidth to support ClockType via a new unapply method on GroundType.

Weirdly, this pull request shows two of the commits out of order (possible Github bug?), regardless, they are correct if you look at the actual branch or check out the branch manually. 
7f12a07d941bd08749ef350654fe9a324b882a6b is actually before 6c93f5e688554ff3432715d206351d2d84341eed. Kind of funny that there is dependency between them but they are shown here out of order.

jackkoenig:
PLEASE DO NOT DELETE THIS BRANCH

PR #305, opened 2016-09-16
fill empty module body with "begin end"

yunsup:
apparently vivado treats an empty module as a black box and triggers an error

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -612,7 +612,7 @@ class VerilogEmitter extends Emitter {
         }
         emit(Seq(");"))
 
-        if (declares.isEmpty && assigns.isEmpty) emit(Seq(tab, "begin end"))
+        if (declares.isEmpty && assigns.isEmpty) emit(Seq(tab, "always @(*) begin end"))
         for (x <- declares) emit(Seq(tab, x))
         for (x <- instdeclares) emit(Seq(tab, x))
         for (x <- assigns) emit(Seq(tab, x))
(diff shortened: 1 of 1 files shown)


yunsup:
I'm assuming that the module is empty if there's no declarations & assignments.

aswaterman:
This is pretty sad, but the fix seems reasonable, since even if the assumption happens to be incorrect, the effect is benign.

yunsup:
Can someone help me figure out why this change broke the build? @jackkoenig @azidar @donggyukim Thanks!

azidar:
Looks like verilator thinks thinks an empty module with only a "begin end" is a syntax error:
```
module BundleWire(
  input   clk,
  input   reset,
  input  [31:0] io_in_x,
  input  [31:0] io_in_y
);
  begin end
endmodule
```
produces

```
%Error: BundleWireTester.v:20: syntax error, unexpected begin
```

aswaterman:
This should make everyone happy:

    always @(*) begin end

yunsup:
do you want to force push to this branch overwriting my commit, which doesn't work with verilator?

azidar:
I was just going to squash and merge the commit

scottj97:
Cadence Incisive gives a compile warning on this:

```
  always @(*) begin end
         |
ncelab: *W,STARMT (../TestHarness.MyConfig.v,196147|9): This @* expands to empty list, will never wake up.

```
I'm investigating possible alternatives, including `initial begin end`.

scottj97:
I've found that `initial begin end` satisfies all five of: VCS, Incisive, Questa, Vivado, Verilator. I was able to reproduce the problems (described above) with Vivado and Verilator, and both problems disappear with `initial begin end`.

mwachs5:
Any idea if synthesis tools blow up on this?

scottj97:
@mwachs5 haven't tried but our backend expert wasn't concerned when I discussed it with him, and he preferred this change.

PR #306, opened 2016-09-17
Style Fixes for Memory Passes

donggyukim:
@shunshou This includes style fixes without functional changes. This works with rocket-chip master. I hope this also looks good to you.

I'll remove annotations to `Info` after merging this.

### DIFF
--- a/src/main/scala/firrtl/Namespace.scala
+++ b/src/main/scala/firrtl/Namespace.scala
@@ -57,8 +57,6 @@ class Namespace private {
 }
 
 object Namespace {
-  def apply(): Namespace = new Namespace
-
   // Initializes a namespace from a Module
   def apply(m: DefModule): Namespace = {
     val namespace = new Namespace
@@ -69,7 +67,7 @@ object Namespace {
       case s: Block => s.stmts flatMap buildNamespaceStmt
       case _ => Nil
     }
-    namespace.namespace ++= (m.ports collect { case dec: IsDeclaration => dec.name })
+    namespace.namespace ++= m.ports map (_.name)
     m match {
       case in: Module =>
         namespace.namespace ++= buildNamespaceStmt(in.body)
@@ -82,9 +80,14 @@ object Namespace {
   /** Initializes a [[Namespace]] for [[ir.Module]] names in a [[ir.Circuit]] */
   def apply(c: Circuit): Namespace = {
     val namespace = new Namespace
-    c.modules foreach { m =>
-      namespace.namespace += m.name
-    }
+    namespace.namespace ++= c.modules map (_.name)
+    namespace
+  }
+
+  /** Initializes a [[Namespace]] from arbitrary strings **/
+  def apply(names: Seq[String] = Nil): Namespace = {
+    val namespace = new Namespace
+    namespace.namespace ++= names
     namespace
   }
 }
--- a/src/main/scala/firrtl/passes/AnnotateMemMacros.scala
+++ b/src/main/scala/firrtl/passes/AnnotateMemMacros.scala
@@ -2,13 +2,13 @@
 
 package firrtl.passes
 
-import scala.collection.mutable
-import AnalysisUtils._
-import firrtl.WrappedExpression._
-import firrtl.ir._
 import firrtl._
+import firrtl.ir._
 import firrtl.Utils._
 import firrtl.Mappers._
+import WrappedExpression.weq
+import MemPortUtils.memPortField
+import AnalysisUtils._
 
 case class AppendableInfo(fields: Map[String, Any]) extends Info {
   def append(a: Map[String, Any]) = this.copy(fields = fields ++ a)
@@ -44,35 +44,40 @@ object AnalysisUtils {
   // use case: compare if two nodes have the same origin
   // limitation: only works in a module (stops @ module inputs)
   // TODO: more thorough (i.e. a + 0 = a)
-  def getConnectOrigin(connects: Connects, node: String): Expression = {
-    if (connects contains node) getOrigin(connects, connects(node))
-    else EmptyExpression
-  }
+  def getConnectOrigin(connects: Connects)(node: String): Expression =
+    connects get node match {
+      case None => EmptyExpression
+      case Some(e) => getOrigin(connects, e)
+    }
+  def getConnectOrigin(connects: Connects, e: Expression): Expression =
+    getConnectOrigin(connects)(e.serialize)
 
   private def getOrigin(connects: Connects, e: Expression): Expression = e match { 
     case Mux(cond, tv, fv, _) =>
       val fvOrigin = getOrigin(connects, fv)
       val tvOrigin = getOrigin(connects, tv)
       val condOrigin = getOrigin(connects, cond)
-      if (we(tvOrigin) == we(one) && we(fvOrigin) == we(zero)) condOrigin
-      else if (we(condOrigin) == we(one)) tvOrigin
-      else if (we(condOrigin) == we(zero)) fvOrigin
-      else if (we(tvOrigin) == we(fvOrigin)) tvOrigin
-      else if (we(fvOrigin) == we(zero) && we(condOrigin) == we(tvOrigin)) condOrigin
+      if (weq(tvOrigin, one) && weq(fvOrigin, zero)) condOrigin
+      else if (weq(condOrigin, one)) tvOrigin
+      else if (weq(condOrigin, zero)) fvOrigin
+      else if (weq(tvOrigin, fvOrigin)) tvOrigin
+      else if (weq(fvOrigin, zero) && weq(condOrigin, tvOrigin)) condOrigin
       else e
-    case DoPrim(PrimOps.Or, args, consts, tpe) if args.contains(one) => one
-    case DoPrim(PrimOps.And, args, consts, tpe) if args.contains(zero) => zero
+    case DoPrim(PrimOps.Or, args, consts, tpe) if args exists (weq(_, one)) => one
+    case DoPrim(PrimOps.And, args, consts, tpe) if args exists (weq(_, zero)) => zero
     case DoPrim(PrimOps.Bits, args, Seq(msb, lsb), tpe) =>  
-      val extractionWidth = (msb-lsb)+1
+      val extractionWidth = (msb - lsb) + 1
       val nodeWidth = bitWidth(args.head.tpe)
       // if you're extracting the full bitwidth, then keep searching for origin
-      if (nodeWidth == extractionWidth) getOrigin(connects, args.head)
-      else e
+      if (nodeWidth == extractionWidth) getOrigin(connects, args.head) else e
     case DoPrim((PrimOps.AsUInt | PrimOps.AsSInt | PrimOps.AsClock), args, _, _) => 
       getOrigin(connects, args.head)
     // note: this should stop on a reg, but will stack overflow for combinational loops (not allowed)
-    case _: WRef | _: SubField | _: SubIndex | _: SubAccess if connects.contains(e.serialize) && kind(e) != RegKind =>
-      getConnectOrigin(connects, e.serialize) 
+    case _: WRef | _: WSubField | _: WSubIndex | _: WSubAccess if kind(e) != RegKind =>
+       connects get e.serialize match {
+         case Some(ex) => getOrigin(connects, ex)
+         case None => e
+       }
     case _ => e
   }
 
@@ -92,55 +97,45 @@ object AnalysisUtils {
 
   // memories equivalent as long as all fields (except name) are the same
   def eqMems(a: DefMemory, b: DefMemory) = a == b.copy(name = a.name) 
-
 }
 
 object AnnotateMemMacros extends Pass {
+  def name = "Analyze sequential memories and tag with info for future passes(useMacro, maskGran)"
 
-  def name = "Analyze sequential memories and tag with info for future passes (useMacro,maskGran)"
+  // returns # of mask bits if used
+  def getMaskBits(connects: Connects, wen: Expression, wmask: Expression): Option[Int] = {
+    val wenOrigin = getConnectOrigin(connects, wen)
+    val wmaskOrigin = connects.keys filter
+      (_ startsWith wmask.serialize) map getConnectOrigin(connects)
+    // all wmask bits are equal to wmode/wen or all wmask bits = 1(for redundancy checking)
+    val redundantMask = wmaskOrigin forall (x => weq(x, wenOrigin) || weq(x, one))
+    if (redundantMask) None else Some(wmaskOrigin.size)
+  }
 
-  def run(c: Circuit) = {
-
-    def annotateModMems(m: Module) = {
-      val connects = getConnects(m)
-
-      // returns # of mask bits if used
-      def getMaskBits(wen: String, wmask: String): Option[Int] = {
-        val wenOrigin = we(getConnectOrigin(connects, wen))
-        val one1 = we(one)
-        val wmaskOrigin = connects.keys.toSeq.filter(_.startsWith(wmask)).map(x => we(getConnectOrigin(connects, x)))
-        // all wmask bits are equal to wmode/wen or all wmask bits = 1(for redundancy checking)
-        val redundantMask = wmaskOrigin.map( x => (x == wenOrigin) || (x == one1) ).foldLeft(true)(_ && _)
-        if (redundantMask) None else Some(wmaskOrigin.length)
+  def updateStmts(connects: Connects)(s: Statement): Statement = s match {
+    // only annotate memories that are candidates for memory macro replacements
+    // i.e. rw, w + r (read, write 1 cycle delay)
+    case m: DefMemory if m.readLatency == 1 && m.writeLatency == 1 &&
+        (m.writers.length + m.readwriters.length) == 1 && m.readers.length <= 1 =>
+      val dataBits = bitWidth(m.dataType)
+      val rwMasks = m.readwriters map (rw =>
+        getMaskBits(connects, memPortField(m, rw, "wmode"), memPortField(m, rw, "wmask")))
+      val wMasks = m.writers map (w =>
+        getMaskBits(connects, memPortField(m, w, "en"), memPortField(m, w, "mask")))
+      val memAnnotations = Map("useMacro" -> true)
+      val tempInfo = appendInfo(m.info, memAnnotations)
+      (rwMasks ++ wMasks).head match {
+        case None =>
+          m copy (info = tempInfo)
+        case Some(maskBits) =>
+          m.copy(info = tempInfo.append("maskGran" -> dataBits / maskBits))
       }
+    case s => s map updateStmts(connects)
+  }
 
-      def updateStmts(s: Statement): Statement = s match {
-        // only annotate memories that are candidates for memory macro replacements
-        // i.e. rw, w + r (read, write 1 cycle delay)
-        case m: DefMemory if m.readLatency == 1 && m.writeLatency == 1 && 
-            (m.writers.length + m.readwriters.length) == 1 && m.readers.length <= 1 =>
-          val dataBits = bitWidth(m.dataType)
-          val rwMasks = m.readwriters map (w => getMaskBits(s"${m.name}.$w.wmode", s"${m.name}.$w.wmask"))
-          val wMasks = m.writers map (w => getMaskBits(s"${m.name}.$w.en", s"${m.name}.$w.mask"))
-          val maskBits = (rwMasks ++ wMasks).head
-          val memAnnotations = Map("useMacro" -> true)
-          val tempInfo = appendInfo(m.info, memAnnotations)
-          if (maskBits == None) m.copy(info = tempInfo)
-          else m.copy(info = tempInfo.append("maskGran" -> dataBits/maskBits.get))       
-        case b: Block => b map updateStmts
-        case s => s
-      }
-      m.copy(body=updateStmts(m.body))
-    } 
+  def annotateModMems(m: DefModule) = m map updateStmts(getConnects(m))
 
-    val updatedMods = c.modules map {
-      case m: Module => annotateModMems(m)
-      case m: ExtModule => m
-    }
-    c.copy(modules = updatedMods)
-
-  }  
-
+  def run(c: Circuit) = c copy (modules = (c.modules map annotateModMems))
 }
 
 // TODO: Add floorplan info?
--- a/src/main/scala/firrtl/passes/AnnotateMemMacros.scala
+++ b/src/main/scala/firrtl/passes/AnnotateMemMacros.scala
@@ -21,22 +21,22 @@ case class AppendableInfo(fields: Map[String, Any]) extends Info {
 }
 
 object AnalysisUtils {
-
-  def getConnects(m: Module) = {
-    val connects = mutable.HashMap[String, Expression]()
-    def getConnects(s: Statement): Statement = {
-      s map getConnects match {
+  type Connects = collection.mutable.HashMap[String, Expression]
+  def getConnects(m: DefModule): Connects = {
+    def getConnects(connects: Connects)(s: Statement): Statement = {
+      s match {
         case Connect(_, loc, expr) =>
           connects(loc.serialize) = expr
         case DefNode(_, name, value) =>
           connects(name) = value
         case _ => // do nothing
       }
-      s // return because we only have map and not foreach
+      s map getConnects(connects)
     }
-    getConnects(m.body)
-    connects.toMap
-  }  
+    val connects = new Connects
+    m map getConnects(connects)
+    connects
+  }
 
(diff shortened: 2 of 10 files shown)


shunshou:
Should be fine; I'll give the code a stare in a bit. Are you planning on moving Annotations to the actual AnnotationMap?

shunshou:
Also, is there some reason to switch from if/else if, etc. to match with stuff like match true => match false => ? I guess is that something common to do? B/c it seems just like a weird personal choice to match to various combinations of true, false to me. Just curious...

src/main/scala/firrtl/passes/AnnotateValidMemConfigs.scala
```diff
@@ -249,43 +248,42 @@ class AnnotateValidMemConfigs(reader: Option[YamlFileReader]) extends Pass {
       sp: Option[SRAMCompiler] = None, 
       dp: Option[SRAMCompiler] = None) {
     def serialize = {
-      if (sp != None) sp.get.serialize
-      if (dp != None) dp.get.serialize
+      sp match {
```
shunshou: Isn't this just more lines of code for doing the same thing? I guess this is more of a stylistic preference, but I personally don't get why this is better.

donggyukim: As far as I know, Option types are designed to be matched to get its value.

donggyukim:
I switched `if` to `match` when both conditions and values are too long. I agree that this can be strange.

donggyukim:
Also, to remove info annotation, I'm going to use `DynamicVariable`. `AnnotationMap` is only for between transforms, so we cannot use it between passes.

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
```diff
@@ -21,35 +21,35 @@ case class AppendableInfo(fields: Map[String, Any]) extends Info {
 }
 
 object AnalysisUtils {
-
-  def getConnects(m: Module) = {
-    val connects = mutable.HashMap[String, Expression]()
-    def getConnects(s: Statement): Statement = {
-      s map getConnects match {
+  type Connects = collection.mutable.HashMap[String, Expression]
```
shunshou: I'm not really sure how this works, but the reason I wanted to do a "connects.toMap" at the end was to have getConnects return an immutable map even if, inside the function, you're appending to a mutable HashMap. Is, then, returning a "Connects" type = Mutable Hashmap of [String, Expression] bad? 

Not really sure how that works...

donggyukim: From software engineering point of view, it's evil to return mutable data structures. However, I noticed that converting HashMap to immutable.Map can be slow.(It may not be true though)

We can just change Connects to immutable.Map if we think it should be.

jackkoenig: I believe conversion from mutable map to immutable map is really slow, at least it was for LinkedHashMaps in ExpandWhens

shunshou: and therefore it's better to leave everything as mutable until some better way to do this is figured out?

src/main/scala/firrtl/Namespace.scala
```diff
@@ -69,7 +67,7 @@ object Namespace {
       case s: Block => s.stmts flatMap buildNamespaceStmt
       case _ => Nil
     }
-    namespace.namespace ++= (m.ports collect { case dec: IsDeclaration => dec.name })
+    namespace.namespace ++= m.ports map (_.name)
```
shunshou: Also, just curious why the former isn't needed anymore (IsDeclaration)

src/main/scala/firrtl/Namespace.scala
```diff
@@ -82,9 +80,14 @@ object Namespace {
...
     val namespace = new Namespace
-    c.modules foreach { m =>
-      namespace.namespace += m.name
-    }
+    namespace.namespace ++= c.modules map (_.name)
+    namespace
+  }
+
+  /** Initializes a [[Namespace]] from arbitrary strings **/
+  def apply(names: Seq[String] = Nil): Namespace = {
```
shunshou: And why do you need a default Nil?

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
```diff
@@ -44,35 +44,45 @@ object AnalysisUtils {
...
-      else if (we(fvOrigin) == we(zero) && we(condOrigin) == we(tvOrigin)) condOrigin
+      if (weq(tvOrigin, one) && weq(fvOrigin, zero)) condOrigin
+      else if (weq(condOrigin, one)) tvOrigin
+      else if (weq(condOrigin, zero)) fvOrigin
+      else if (weq(tvOrigin, fvOrigin)) tvOrigin
+      else if (weq(fvOrigin, zero) && weq(condOrigin, tvOrigin)) condOrigin
       else e
-    case DoPrim(PrimOps.Or, args, consts, tpe) if args.contains(one) => one
-    case DoPrim(PrimOps.And, args, consts, tpe) if args.contains(zero) => zero
+    case DoPrim(PrimOps.Or, args, consts, tpe) if args exists (weq(_, one)) => one
```
shunshou: Is there ever a case where "one" isn't general enough? vs. doing weq(_,one)

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
```diff
@@ -44,35 +44,45 @@ object AnalysisUtils {
...
+    case DoPrim(PrimOps.And, args, consts, tpe) if args exists (weq(_, zero)) => zero
     case DoPrim(PrimOps.Bits, args, Seq(msb, lsb), tpe) =>  
-      val extractionWidth = (msb-lsb)+1
+      val extractionWidth = (msb - lsb) + 1
       val nodeWidth = bitWidth(args.head.tpe)
       // if you're extracting the full bitwidth, then keep searching for origin
-      if (nodeWidth == extractionWidth) getOrigin(connects, args.head)
-      else e
+      nodeWidth == extractionWidth match {
```
shunshou: Especially for something like this when there are only 2 conditions, doing a match for true/false seems overkill o_o

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
```diff
@@ -44,35 +44,45 @@ object AnalysisUtils {
...
+        case true => getOrigin(connects, args.head)
+        case false => e
+      }
     case DoPrim((PrimOps.AsUInt | PrimOps.AsSInt | PrimOps.AsClock), args, _, _) => 
       getOrigin(connects, args.head)
     // note: this should stop on a reg, but will stack overflow for combinational loops (not allowed)
-    case _: WRef | _: SubField | _: SubIndex | _: SubAccess if connects.contains(e.serialize) && kind(e) != RegKind =>
-      getConnectOrigin(connects, e.serialize) 
+    case _: WRef | _: WSubField | _: WSubIndex => kind(e) match {
```
shunshou: Would SubAccess never be part of this?

shunshou: Lind of a style preference thing: do people prefer 2-layer nested match vs just being like if kind(e) != RegKind, because line 86 already gives the default if it's not RegKind...

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
```diff
@@ -92,55 +102,45 @@ object AnalysisUtils {
...
-    val updatedMods = c.modules map {
-      case m: Module => annotateModMems(m)
-      case m: ExtModule => m
-    }
-    c.copy(modules = updatedMods)
+    case s => s map updateStmts(connects)
+  }
 
-  }  
+  def annotateModMems(m: DefModule) = m map updateStmts(getConnects(m))
```
shunshou: Silly question, but what's the mechanism for annotateModMems to return a DefModule? m map --> works on the statements in m.body, but is there some implicit m copy (body = updatedStmts List) somewhere?

src/main/scala/firrtl/passes/ReplSeqMem.scala
```diff
@@ -91,50 +88,35 @@ Optional Arguments:
...
+          else if (new File(inputFileName).exists) Some(new YamlFileReader(inputFileName))
+          else error("Input configuration file does not exist!")
+        }
+        val outConfigFile = new ConfWriter(PassConfigUtil.getPassOptions(t)(OutputConfigFileName))
+        run(c, passSeq(inConfigFile, outConfigFile))
+      case _ => error("Unexpected transform annotation")
     }
-}
\ No newline at end of file
+    case _ => TransformResult(c)
```
shunshou: What happened to map? How does that get passed along? [even if this transform doesn't change it]

src/main/scala/firrtl/passes/ReplaceMemMacros.scala
```diff
@@ -95,75 +40,86 @@ class ReplaceMemMacros(writer: ConfWriter) extends Pass {
...
+      connectFields(bbPort, "wmode", wrapperPort, "wmode"), 
+      Connect(NoInfo, createSubField(bbPort, "wdata"), wrapperWData))
+    hasMask match {
+      case false => defaultSeq
+      case true => defaultSeq :+ Connect(
         NoInfo,
-        WSubField(bbPort, "wmask", bbMask, UNKNOWNGENDER),
-        rhs
+        createSubField(bbPort, "wmask"),
+        maskBits(createSubField(wrapperPort, "wmask"), wrapperWData.tpe, fillMask)
```
shunshou: Isn't wrapperWData.type [ground type] != wrapperMem.dataType? [aggregate]

src/main/scala/firrtl/passes/AnnotateValidMemConfigs.scala
```diff
@@ -249,43 +248,42 @@ class AnnotateValidMemConfigs(reader: Option[YamlFileReader]) extends Pass {
...
+  def updateStmts(s: Statement): Statement = s match {
+    case m: DefMemory if containsInfo(m.info, "useMacro") => sramCompilers match {
+      case None => m
+      case Some(compiler) if (m.readwriters.length == 1) =>
+        compiler.sp match {
+          case None => error("Design needs RW port memory compiler!")
+          case Some(p) => p append m
+        }
+      case Some(compiler) =>
+        compiler.sp match {
```
shunshou: Should be dp

src/main/scala/firrtl/Namespace.scala
donggyukim: All ports are `IsDeclaration` according to @jackkoenig

src/main/scala/firrtl/Namespace.scala
donggyukim: I just delete `def apply(): Namespace`. This emulates it.

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
donggyukim: Yeah, I think so

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
donggyukim: It's dangerous to check equality without `WrappedExpression`.

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
donggyukim: `SubAccess` should be gone before this transform. I can revert changes on `RegKind`.

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
donggyukim: No, this mapper applies the function to m.body. It does nothing for `ExtModule`.

src/main/scala/firrtl/passes/ReplSeqMem.scala
donggyukim: Nothing happens. `Transform` internally pass annotations.

src/main/scala/firrtl/passes/ReplaceMemMacros.scala
donggyukim: Yes, you're right. Thanks for pointing this :)

src/main/scala/firrtl/passes/AnnotateValidMemConfigs.scala
donggyukim: Thanks, I'll fix it.

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
shunshou: ok but the return type is a DefModule?
which is either the original ExtModule or "Module" with whatever changes?

Haha... I'm still such a functional programming noob

shunshou:
So I went through all of the files that I've worked on. I can look through InferReadWrite.scala too, but I figure you'd know it better since you wrote it. This was useful to me too from a "I learned more about functional programming" perspective haha.

shunshou:
Also, when you said "pass with rocket" did you run all the way until vsim with inferRW, replSeqMem passes enabled? I feel like "wrapperWData.type" would've had a match error if replSeqMem actually ran... unless there's something I was misunderstanding about toBitMask or w/e the function was...

src/main/scala/firrtl/passes/InferReadWrite.scala
```diff
@@ -50,153 +51,137 @@ case class InferReadWriteAnnotation(t: String, tID: TransID)
...
-      case s: Connect  =>
-        connects(s.loc.serialize) = s.expr
-      case s: PartialConnect =>
-        connects(s.loc.serialize) = s.expr
-      case s: DefNode =>
-        connects(s.name) = s.value
-      case s: Block =>
-        s.stmts foreach analyze
-      case _ =>
+  def getProductTerms(connects: Connects)(e: Expression): Seq[Expression] = e match {
```
shunshou: Technically, the Mux(cond,1,0) is only one of many possible things to test for without ConstProp... Say a + 0 could be optimized to just a... I feel like you should use some kind of getOrigin function [that you can add more things to check for] and then after you have all the origins do a "getProductTerms" kind of thing...

src/main/scala/firrtl/passes/InferReadWrite.scala
```diff
@@ -50,153 +51,137 @@ case class InferReadWriteAnnotation(t: String, tID: TransID)
...
+    // No ConstProp yet...
+    case Mux(cond, tval, fval, _) if weq(tval, one) && weq(fval, zero) =>
+      getProductTerms(connects)(cond)
```
shunshou: I noticed in your previous code you had cond +: getProductTerms(cond.serialize). In that case, you're making a seq of cond + condOrigin. But in this case, how does your HashMap lookup work if you never use serialize? [And do you only care about adding condOrigin instead of cond itself + condOrigin to the Seq?]

Actually, I see on line 68, you have connects get e, but I thought get matched the key, and e is of type Expession = value?

src/main/scala/firrtl/passes/InferReadWrite.scala
```diff
@@ -50,153 +51,137 @@ case class InferReadWriteAnnotation(t: String, tID: TransID)
...
-            stmts += Connect(NoInfo, WSubField(rw_exp, "addr", ut, FEMALE),
-              Mux(connects(s"${mem.name}.$w.en"), connects(s"${mem.name}.$w.addr"), 
-                  connects(s"${mem.name}.$r.addr"), ut))
-          }
-        }
-        if (readwriters.isEmpty) mem else DefMemory(mem.info,
-          mem.name, mem.dataType, mem.depth, mem.writeLatency, mem.readLatency,
-          mem.readers filterNot readers, mem.writers filterNot writers,
-          mem.readwriters ++ readwriters)
+  def replaceStmt(repl: Netlist)(s: Statement): Statement =
```
shunshou: I know I basically took this from you for my pass. Maybe this can be made into a pass utility? Same for replaceExp -- although I don't understand why you need to explicitly match to WSubField...

src/main/scala/firrtl/passes/UpdateDuplicateMemMacros.scala
```diff
@@ -33,35 +83,41 @@ object MemTransformUtils {
...
       def testEmptyExpr(e: Expression): Expression = {
-        e map testEmptyExpr match {
+        e match {
           case EmptyExpression => foundEmpty = true
           case _ =>
         }
-        e // map must return; no foreach
+        e map testEmptyExpr // map must return; no foreach
       }
       stmt map testEmptyExpr
```
shunshou: Just want to be sure, and I'm sure @jackkoenig already mentioned this to me. stmt map testEmptyExpr goes through all of the expressions in the stmt? So a map on a statement could be "sub statements" i.e. statements in Block or it can be Expressions?

src/main/scala/firrtl/passes/InferReadWrite.scala
```diff
@@ -50,153 +51,137 @@ case class InferReadWriteAnnotation(t: String, tID: TransID)
...
+      val readers = new PortSet
+      val writers = new PortSet
+      val readwriters = collection.mutable.ArrayBuffer[String]()
+      val namespace = Namespace(mem.readers ++ mem.writers ++ mem.readwriters)
+      for (w <- mem.writers ; r <- mem.readers) {
+        val wp = getProductTerms(connects)(memPortField(mem, w, "en"))
+        val rp = getProductTerms(connects)(memPortField(mem, r, "en"))
+        if (wp exists (a => rp exists (b => checkComplement(a, b)))) {
+          val rw = namespace newName "rw"
+          val rwExp = WSubField(WRef(mem.name, ut, MemKind, ug), rw, ut, ug)
```
shunshou: use createSubField here and get rid of ut, ug?

src/main/scala/firrtl/passes/InferReadWrite.scala
```diff
@@ -50,153 +51,137 @@ case class InferReadWriteAnnotation(t: String, tID: TransID)
...
+          repl(memPortField(mem, r, "clk"))  = EmptyExpression
+          repl(memPortField(mem, r, "addr")) = EmptyExpression
+          repl(memPortField(mem, r, "data")) = WSubField(rwExp, "rdata", mem.dataType, MALE)
+          repl(memPortField(mem, w, "en"))   = WSubField(rwExp, "wmode", BoolType, FEMALE)
+          repl(memPortField(mem, w, "clk"))  = EmptyExpression
+          repl(memPortField(mem, w, "addr")) = EmptyExpression
+          repl(memPortField(mem, w, "data")) = WSubField(rwExp, "wdata", mem.dataType, FEMALE)
+          repl(memPortField(mem, w, "mask")) = WSubField(rwExp, "wmask", ut, FEMALE)
+          stmts += Connect(NoInfo, WSubField(rwExp, "clk", ClockType, FEMALE),
+            WRef("clk", ClockType, NodeKind, MALE))
```
shunshou: Does this work when the memory clock isn't "clk"? i.e. the clk going into the top level module is named something different....

src/main/scala/firrtl/passes/InferReadWrite.scala
```diff
@@ -50,153 +51,137 @@ case class InferReadWriteAnnotation(t: String, tID: TransID)
...
+    InferReadWritePass,
+    CheckInitialization,
+    InferTypes,
+    ResolveKinds,
+    ResolveGenders
+  )
+  def execute(c: Circuit, map: AnnotationMap) = map get transID match {
+    case Some(p) => p get CircuitName(c.main) match {
+      case Some(InferReadWriteAnnotation(_, _)) => run(c, passSeq)
+      case _ => TransformResult(c)
```
shunshou: I forgot who said it (Jack or Adam) but this should maybe error out if the wrong annotation goes with the transID?

shunshou:
Ok, I think I stared at all of your changes... Sorry for bombarding you with questions

src/main/scala/firrtl/passes/InferReadWrite.scala
donggyukim: Eventually, the solution is ContProp on high firrtl. I'll fix both `getOrigin` and this with it.

src/main/scala/firrtl/passes/InferReadWrite.scala
donggyukim: There's an implicit conversion from `Expression` to `String` in this pass. Is it your question?

src/main/scala/firrtl/passes/InferReadWrite.scala
```diff
@@ -50,153 +51,137 @@ case class InferReadWriteAnnotation(t: String, tID: TransID)
...
+  def checkComplement(a: Expression, b: Expression) = (a, b) match {
+    // b ?= Not(a)
+    case (_, DoPrim(Not, args, _, _)) => weq(args.head, a)
+    // a ?= Not(b)
+    case (DoPrim(Not, args, _, _), _) => weq(args.head, b)
+    // b ?= Eq(a, 0) or b ?= Eq(0, a)
+    case (_, DoPrim(Eq, args, _, _)) =>
+      weq(args(0), a) && weq(args(1), zero) ||
+      weq(args(1), a) && weq(args(0), zero)
+    // a ?= Eq(b, 0) or b ?= Eq(0, a)
```
donggyukim: I just ported Chisel2's inference for backward compatibility and assumed enable signals are written according to the rules.

src/main/scala/firrtl/passes/UpdateDuplicateMemMacros.scala
donggyukim: It applies the function to its own sub-statements only.

src/main/scala/firrtl/passes/InferReadWrite.scala
donggyukim: Yes, we should fix it in a separate pull request.

src/main/scala/firrtl/passes/InferReadWrite.scala
donggyukim: But, it would be great if we could consider all the cases :)

src/main/scala/firrtl/passes/ReplaceMemMacros.scala
donggyukim: My guess here is rocket-chip actually don't need this, so it worked with this broken code...

src/main/scala/firrtl/passes/InferReadWrite.scala
donggyukim: I don't know how I handle it... @azidar?

src/main/scala/firrtl/Namespace.scala
jackkoenig: See definition of Port, it mixes in IsDeclaration
https://github.com/ucb-bar/firrtl/blob/master/src/main/scala/firrtl/ir/IR.scala#L407

src/main/scala/firrtl/passes/InferReadWrite.scala
donggyukim: I just don't want it to access the map except for `WSubField`(All memory ports are `WSubField`)

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
donggyukim: The mapper eventually returns `DefModule`, but internally different things according to its type.
(`DefModule` is either `Module` or `ExtModule`.)

src/main/scala/firrtl/passes/AnnotateMemMacros.scala
donggyukim: I see this should be general, so I'll also revive `WSubAccess`.

shunshou: Well, that's arguable and I guess up to you... We should make people aware where passes can/should be run though... And what things disappear at what step...

src/main/scala/firrtl/passes/InferReadWrite.scala
shunshou: Yep that was my question. Thanks!

src/main/scala/firrtl/passes/InferReadWrite.scala
shunshou: I think just throw an exception for case _ rather than using TranformResult(c) ?

shunshou:
Is the AnnotationMap the Annotation system? I just find it weird that it's only between transforms then... I'm curious what your DynamicVariable implementation will look like

donggyukim:
@shunshou is it lgtm, now? I'll start working on #309, #307 on top of this PR.

shunshou:
I'll stare again in an hour. Did you rebase stuff? Guess I should just look at where I made comments then?

donggyukim:
Yes, I rebased it. I think you only need to see around comments.

shunshou:
I think it's lgtm now.

Issue #308, opened 2016-09-20
Spec wrong for resultant bits of add


chick:
Firrtl spec says in section 7.1
```
add(UInt(w1), SInt(w2)) should result in SInt(w1.max(w2)+1)
```
But consider
```3.U<2> + 1.S<2>``` should have the numeric result +4 requiring 3 bits but because the result needs a zero sign bit so 4 can be positive, 
the actual SInt requires 4 bits or ```SInt(w1.max(w2)+1)```.
This is causing the interpreter to blow up.  Should the spec be changed, or if not what should the interpreter do here.

aswaterman:
Largest positive value is 2^w1 + 2^(w2-1) - 2, so max(w1, w2-1)+1+1 should suffice.

Largest negative value is -2^(w2-1), which is already dominated by the previous expression.

chick:
Where w2 is always the width of the SInt for both cases ```SInt + UInt``` and ```UInt + SInt```

aswaterman:
Yeah, that follows from commutativity

chick:
@jackkoenig @azidar, any plan on this?  This issue and #193 are causing users problems in the field when running the interpreter.  I don't think I can fix this in the interpreter without it agreeing with firrlt's behavior in this situation.  I think the fixes are fairly easy, plus an update to the manual

Issue #193, opened 2016-06-16
Width inference incorrect for subtraction of SInt by UInt


chick:
Consider
```
-4.SInt<3> -  5.UInt<3> = -9.SInt<5>
   100     -     101    =  11001
   -4      -       5    =     -9
```  
The Firrtl Spec incorrectly states that ```SInt<x> - UInt<y>  => SInt<max(x,y)+1>``` in 7.2 page 40.

aswaterman:
Should be max(x+1,y+2), right?

chick:
The same width error is true of UInt - SInt  should be ```max(w1, w2) + 2```

chick:
I think your are right, but making a test to convince myself

chick:
@aswaterman it looks to me like your rule is correct except when the SInt width is 1, e.g.
```i.S<1> + j.U<y>```  in which case the width is ```y + 1```. 
Example: ```-1.S<1> - 8.U<4> => -9.S<5>```  though predicted ```max(x+1, y+2) == 6```
Any problems with this?  I assume we want Firrtl behavior to be precise.

aswaterman:
Yeah, though it is asymmetric.  That is, S<1> - U<n> needs only n+1 bits, but U<n> - S<1> needs the full n+2.

I think taking advantage of the special case makes sense.

chick:
Yes, The ```j.U<y> - i.S<x>`` follows the rule even when x is 1
Verified both for Widths from 1 to 100

PR #312, opened 2016-09-22
Fixed width inference for add, sub

azidar:
#308 #193 

I did not special case subtraction for SInt<1> because it can't be expressed easily with regards to the width inferencer. Once ranges are implemented, this special case will be caught regardless, so it doesn't seem worthwhile to special case yet.

### DIFF
--- a/src/test/scala/firrtlTests/WidthSpec.scala
+++ b/src/test/scala/firrtlTests/WidthSpec.scala
@@ -1,0 +1,87 @@
+/*
+Copyright (c) 2014 - 2016 The Regents of the University of
+California (Regents). All Rights Reserved.  Redistribution and use in
+source and binary forms, with or without modification, are permitted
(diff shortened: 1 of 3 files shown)


src/test/scala/firrtlTests/WidthSpec.scala
```diff
@@ -0,0 +1,87 @@
...
+      InferTypes,
+      CheckTypes,
+      InferWidths)
+    val input =
+      """circuit Unit :
+        |  module Unit :
+        |    input x: UInt<3>
+        |    input y: SInt<2>
+        |    output z: SInt
+        |    z <= add(y, x)""".stripMargin
```
jackkoenig: Shouldn't this be a sub?

chick:
@azidar looks to me like you did not update spec.tex and spec.pdf

PR #314, opened 2016-09-23
Add support for Parameterized BlackBoxes

jackkoenig:
I also made minor changes to the integration testing style to make it easier to write external modules tests. They'll also be parallelized more and thus slightly faster.

### DIFF
--- a/src/main/antlr4/FIRRTL.g4
+++ b/src/main/antlr4/FIRRTL.g4
@@ -63,7 +63,7 @@ circuit
 
 module
   : 'module' id ':' info? INDENT port* moduleBlock DEDENT
-  | 'extmodule' id ':' info? INDENT port* DEDENT
+  | 'extmodule' id ':' info? INDENT port* defname? parameter* DEDENT
   ;
 
 port
@@ -87,6 +87,16 @@ type
 
 field
   : 'flip'? id ':' type
+  ;
+
+defname
+  : 'defname' '=' id NEWLINE
+  ;
+
+parameter
+  : 'parameter' id '=' IntLit NEWLINE
+  | 'parameter' id '=' StringLit NEWLINE
+  | 'parameter' id '=' DoubleLit NEWLINE
   ;
 
 moduleBlock
@@ -289,6 +299,10 @@ IntLit
   : '0'
   | ( '+' | '-' )? [1-9] ( Digit )*
   | '"' 'h' ( HexDigit )+ '"'
+  ;
+
+DoubleLit
+  : ( '+' | '-' )? Digit+ '.' Digit+ ( 'E' Digit+ )?
   ;
 
 fragment
--- a/src/main/antlr4/FIRRTL.g4
+++ b/src/main/antlr4/FIRRTL.g4
@@ -97,6 +97,7 @@ parameter
   : 'parameter' id '=' IntLit NEWLINE
   | 'parameter' id '=' StringLit NEWLINE
   | 'parameter' id '=' DoubleLit NEWLINE
+  | 'parameter' id '=' RawString NEWLINE
   ;
 
 moduleBlock
@@ -321,7 +322,16 @@ HexDigit
   ;
 
 StringLit
-  : '"' ('\\"'|.)*? '"'
+  : '"' UnquotedString? '"'
+  ;
+
+RawString
+  : '\'' UnquotedString? '\''
+  ;
+
+fragment
+UnquotedString
+  : ('\\"'|~[\r\n])+?
   ;
 
 FileInfo
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -82,6 +82,14 @@ class VerilogEmitter extends Emitter {
     }
     case _ => error("Shouldn't be here")
   }
+  /** Turn Params into Verilog Strings */
+  def stringify(param: Param): String = param match {
+    case IntParam(name, value) => s".$name($value)"
+    case DoubleParam(name, value) => s".$name($value)"
+    case StringParam(name, value) =>
+      val strx = "\"" + VerilogStringLitHandler.escape(value) + "\""
+      s".${name}($strx)"
+  }
   def emit(x: Any)(implicit w: Writer) { emit(x, 0) }
   def emit(x: Any, top: Int)(implicit w: Writer) {
     def cast(e: Expression): Any = e.tpe match {
@@ -237,7 +245,7 @@ class VerilogEmitter extends Emitter {
      }
    }
    
-    def emit_verilog(m: Module)(implicit w: Writer): DefModule = {
+    def emit_verilog(m: Module, moduleMap: Map[String, DefModule])(implicit w: Writer): DefModule = {
       val netlist = mutable.LinkedHashMap[WrappedExpression, Expression]()
       val addrRegs = mutable.HashSet[WrappedExpression]()
       val namespace = Namespace(m)
@@ -448,8 +456,13 @@ class VerilogEmitter extends Emitter {
           simulate(sx.clk, sx.en, printf(sx.string, sx.args), Some("PRINTF_COND"))
           sx
         case sx: WDefInstanceConnector =>
+          val (module, params) = moduleMap(sx.module) match {
+            case ExtModule(_, _, _, extname, params) => (extname, params)
+            case Module(_, name, _, _) => (name, Seq.empty)
+          }
           val es = create_exps(WRef(sx.name, sx.tpe, InstanceKind, MALE))
-          instdeclares += Seq(sx.module, " ", sx.name, " (")
+          val ps = if (params.nonEmpty) params map stringify mkString ("#(", ", ", ") ") else ""
+          instdeclares += Seq(module, " ", ps, sx.name ," (")
           (es zip sx.exprs).zipWithIndex foreach {case ((l, r), i) =>
             val s = Seq(tab, ".", remove_root(l), "(", r, ")")
             if (i != es.size - 1) instdeclares += Seq(s, ",")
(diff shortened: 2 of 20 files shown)


sdtwigg:
It would be beneficial to be able to supply unquoted string parameters. In particular, this should allow for specification of type parameters.

src/main/scala/firrtl/Emitter.scala
```diff
@@ -92,6 +92,14 @@ class VerilogEmitter extends Emitter {
     }
     case _ => error("Shouldn't be here")
```
sdtwigg: Just allowing a MatchError would be better here since at least then it will say what you got but shouldn't have....

src/main/scala/firrtl/Emitter.scala
```diff
@@ -92,6 +92,14 @@ class VerilogEmitter extends Emitter {
...
   }
+  /** Turn Params into Verilog Strings */
+  def stringify(param: Param): String = param match {
```
sdtwigg: If you are matching on cases then the Param type should be sealed.

src/main/scala/firrtl/Emitter.scala
```diff
@@ -372,8 +380,9 @@ class VerilogEmitter extends Emitter {
         initials += Seq("`endif")
       }
 
-      def instantiate(n: String,m: String, es: Seq[Expression]) {
-         instdeclares += Seq(m, " ", n, " (")
+      def instantiate(n: String, m: String, es: Seq[Expression], params: Seq[Param]): Unit = {
+         val ps = if (params.nonEmpty) params map stringify mkString ("#(", ", ", ") ") else ""
+         instdeclares += Seq(m, " ", ps, n ," (")
          es.zipWithIndex foreach {case (e, i) =>
```
sdtwigg: for((e, i) <- es.zipWithIndex) {
  // ...
}
would be easier to read (ideally with e and i better spelled out)

src/main/scala/firrtl/Emitter.scala
```diff
@@ -466,8 +475,12 @@ class VerilogEmitter extends Emitter {
           simulate(s.clk, s.en, printf(s.string, s.args), Some("PRINTF_COND"))
           s
         case (s: WDefInstance) =>
+          val (module, params) = moduleMap(s.module) match {
```
sdtwigg: The DefModule class is not sealed so you shouldn't be matching over it (unless you use a wildcard....)

src/main/antlr4/FIRRTL.g4
```diff
@@ -87,6 +87,16 @@ field
   : 'flip'? id ':' type
   ;
 
+extModuleName
```
sdtwigg: The FIRRTL IR definition call this extname but serializes as 'name ='. In the internal IR name is actually the overall defname. The inconsistency is undesirable.

src/main/scala/firrtl/Emitter.scala
jackkoenig: I didn't touch that line, I was kind of leaving it for whenever the Emitter gets ultimately reworked

src/main/scala/firrtl/Emitter.scala
jackkoenig: Nothing in Firrtl is currently sealed, I think this can/will be solved in a later pull request

src/main/scala/firrtl/Emitter.scala
jackkoenig: Similar to DefModule below, nothing is sealed, although since I'm adding this new IR node I think I will seal it

src/main/scala/firrtl/Emitter.scala
jackkoenig: I wasn't really rewriting the whole function although I can fix this adjacent line

src/main/antlr4/FIRRTL.g4
jackkoenig: Yeah I'd rather call it name everywhere but name is already taken by the actual Firrtl name of the ExtModule. I can make it extname everywhere or defname, what do you think?

jackkoenig:
The main reason why I didn't add unquoted strings for type parameter support is I'm not sure that it would work. I can't find many examples of instantiating type parameters by name. For example at:
https://www.doulos.com/knowhow/sysverilog/tutorial/classes/
They pass a type parameter with #(bit [7:0]), but would this work with #(.T(bit [7:0]))?

Section 5.8.1 of the SV LRM instantiates some type parameters by name but it seems to involve type defs before the module instantiation. It seems to be a bit more complicated then just supporting unquoted strings.

src/main/scala/firrtl/Emitter.scala
jackkoenig: Wasn't really touching the whole function but I guess I can improve adjacent lines

sdtwigg:
StackOverflow had some nice examples of people using 'parameter type' such as at http://stackoverflow.com/questions/18627235/override-size-of-a-parameter-that-is-an-array-of-a-struct-in-systemverilog 

If I recall correctly, the .NAME syntax versus just listing is similar to calling a function usingnamed arguments or just calling 'normally'.

Also, verilator supports it via http://www.veripool.org/issues/376-Verilator-Support-parameter-type- so you should be able to try out some examples with that.

aswaterman:
@sdtwigg our experience is that, in practice, third-party IP never uses the type-parameterization feature.  Perhaps your company would be willing to publish some non-trivial examples?

jackkoenig:
I have added a commit (to be rebased) that makes the overriding of Extmodule names more consistent. Renamed it everywhere to defname. I also fixed a bug where ExtModule serialization did not include defname and parameters

muojp:
@jackkoenig  Hi, it would be reasonable to update `README.md` for up-to-date prerequisites regarding #315 (last doc revision pointed at Verilator 3.880 https://github.com/ucb-bar/firrtl/commit/2bb33667826b7b072ce4cc7a54f4b6fc1a15d2bd#diff-04c6e90faac2675aa89e2176d2eec7d8R15 ).
I made a pull-req doing this (https://github.com/ucb-bar/firrtl/pull/317) for your convenience but, well, it's just 1-byte change, please feel free to close the pull-req and modify README.md by yourself.

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -448,9 +473,16 @@ case class Module(info: Info, name: String, ports: Seq[Port], body: Statement) e
...
   * Generally used for Verilog black boxes
+  * @param defname Defined name of the external module (ie. the name Firrtl will emit)
   */
-case class ExtModule(info: Info, name: String, ports: Seq[Port]) extends DefModule {
-  def serialize: String = serializeHeader("extmodule")
+case class ExtModule(
+    info: Info,
+    name: String,
+    ports: Seq[Port],
+    defname: String,
```
shunshou: Is the reason for having defname so you can have some kind of external annotation for a module name that might vary case-to-case?

src/main/scala/firrtl/Emitter.scala
```diff
@@ -82,6 +82,15 @@ class VerilogEmitter extends Emitter {
     }
     case _ => error("Shouldn't be here")
   }
+  /** Turn Params into Verilog Strings */
+  def stringify(param: Param): String = param match {
+    case IntParam(name, value) => s".$name($value)"
+    case DoubleParam(name, value) => s".$name($value)"
+    case StringParam(name, value) =>
+      val strx = "\"" + VerilogStringLitHandler.escape(value) + "\""
+      s".${name}($strx)"
```
shunshou: You don't need {} around name, right?

src/main/scala/firrtl/Visitor.scala
```diff
@@ -105,14 +105,29 @@ class Visitor(infoMode: InfoMode) extends FIRRTLBaseVisitor[FirrtlNode] {
...
+        val params = ctx.parameter map visitParameter
+        ExtModule(info, ctx.id.getText, ports, defname, params)
     }
   }
 
   private def visitPort[FirrtlNode](ctx: FIRRTLParser.PortContext): Port = {
     Port(visitInfo(Option(ctx.info), ctx), ctx.id.getText, visitDir(ctx.dir), visitType(ctx.`type`))
   }
 
+  private def visitParameter[FirrtlNode](ctx: FIRRTLParser.ParameterContext): Param = {
```
shunshou: In general, is the point of visiting to go through the input *.fir file, match to the FIRRTL.g4 "template" and then parse parameters via the below match? And therefore you can't do anything else except have all of these nulls?

shunshou:
Hm, I'll mess around more with actual test examples, but I think all of your changes look good to me? 

As someone who almost exclusively writes VHDL, it'd be interesting to see if I could stitch up a Chisel design with parameterized VHDL black boxes too... Not super sure how mixed Verilog/VHDL designs work, but I'd assume you'd just do instantiation in whatever is the top level language?

src/main/scala/firrtl/ir/IR.scala
jackkoenig: The motivation behind defname is that, from the perspective of Firrtl, blackboxes with different parameters that affect the IO are different external modules. We discussed a couple of ways to get around this. One way would have been to essentially implement Verilog parameterization in Firrtl. The alternative we chose was just to let the user override the blackbox module name via defname. Thus, "name" is used in the Firrtl netlist, and defname is what we actually emit.

src/main/scala/firrtl/Visitor.scala
jackkoenig: That's more or less correct. FIRRTL.g4 is used by ANTLR to generate a Java lexer and parser. The parser returns a concrete syntax tree which we then "visit" to turn into an abstract syntax tree (as described in IR.scala). I could box everything here with Option and then match on (Some(blah), None, ...) but given that we are interacting with Java here I decided to just go ahead and deal with the nulls.

jackkoenig:
This is ready to merge

jackkoenig:
Confirmed that it works with rocket-chip

PR #316, opened 2016-09-23
Style cleanup take 3

chick:
This is the next round of style changes. 
Nothing to dramatic here, that will come with the next two rounds


src/main/scala/firrtl/Namespace.scala
```diff
@@ -27,14 +27,15 @@ MODIFICATIONS.
 
 package firrtl
 
+import scala.collection.mutable
 import scala.collection.mutable.HashSet
```
jackkoenig: Could you also delete the import scala.collection.mutable.HashSet? I guess this can be a later style-fix commit deleting unnecessary imports

chick: I think I'll just do that in another pass

Issue #297, opened 2016-09-13
Dynamic left shift crashes firrtl in ConstProp


ucbjrl:
The following Chisel code:

    class OneHotTester(_a: Int) extends BasicTester {
      val addr = Wire(UInt(width = 32))
      addr := UInt(_a, 32)
      val t = UIntToOH(addr))
      val o = 1 << _a
      assert(UInt(o) === t)
      stop()
    }

crashes firrtl with following:

    Exception in thread "main" java.lang.IllegalArgumentException: requirement failed
    	at scala.Predef$.require(Predef.scala:207)
    	at firrtl.passes.ConstProp$.constPropPrim(ConstProp.scala:236)
    	at firrtl.passes.ConstProp$.firrtl$passes$ConstProp$$constPropExpression$1(ConstProp.scala:274)
    	at firrtl.passes.ConstProp$$anonfun$1.apply(ConstProp.scala:272)
    	at firrtl.passes.ConstProp$$anonfun$1.apply(ConstProp.scala:272)
    	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
    	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
    	at scala.collection.immutable.List.foreach(List.scala:381)
    	at scala.collection.TraversableLike$class.map(TraversableLike.scala:245)
    	at scala.collection.immutable.List.map(List.scala:285)
    	at firrtl.Mappers$ExpMagnet$$anon$5.map(Mappers.scala:111)
    	at firrtl.Mappers$ExpMap.map(Mappers.scala:144)

    ...

The relevant code is:

    case Bits => e.args(0) match {
      case lit: Literal => {
        val hi = e.consts(0).toInt
        val lo = e.consts(1).toInt
        require(hi >= lo)
        UIntLiteral((lit.value >> lo) & ((BigInt(1) << (hi - lo + 1)) - 1), width_BANG(e.tpe))
      }

`hi` == -1; `lo` == 0

aswaterman:
UIntToOH is making a 2^32-bit result.  FIRRTL probably shouldn't be able to handle that, but should error in a more friendly way.

jackkoenig:
Fair point, I "fixed" the problem but then it does allow for some comically large widths. Is there an upper bound on the width of wires in Verilog?

aswaterman:
[...]

PR #318, opened 2016-09-25
Added max width check to dshl shift amount

azidar:
Address #297


kevinkim06:
Hello,
There is a small issue. Dshl width check error message is "Width of dshl shift amount cannot be larger than 20 bits" now. But 20-bit also fails. I think either "larger" -> "equal or larger than" or condition in CheckWidths.scala should be "bitWidth(b.tpe) >= DshlMaxWidth" -> "bitWidth(b.tpe) > DshlMaxWidth".

PR #319, opened 2016-09-26
Style cleanup take 4

chick:
more style cleanup.
Big one here is the removing of unnecessary parentheses.  
It's hard to say sometimes whether unnecessary ones add some readability

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -90,14 +90,13 @@ class VerilogEmitter extends Emitter {
       case ClockType => e
       case AnalogType(w) => e
     }
-    (x) match {
+    x match {
       case (e: DoPrim) => emit(op_stream(e), top + 1)
       case (e: Mux) => emit(Seq(e.cond," ? ",cast(e.tval)," : ",cast(e.fval)),top + 1)
       case (e: ValidIf) => emit(Seq(cast(e.value)),top + 1)
       case (e: WRef) => w write e.serialize
       case (e: WSubField) => w write LowerTypes.loweredName(e)
-      case (e: WSubAccess) => w write (
-        s"${LowerTypes.loweredName(e.exp)}[${LowerTypes.loweredName(e.index)}]")
+      case (e: WSubAccess) => w write s"${LowerTypes.loweredName(e.exp)}[${LowerTypes.loweredName(e.index)}]"
       case (e: WSubIndex) => w write e.serialize
       case (e: Literal) => v_print(e)
       case (e: VRandom) => w write s"{${e.nWords}{$$random}}"
@@ -183,7 +182,7 @@ class VerilogEmitter extends Emitter {
        case Neq => Seq(cast_if(a0), " != ", cast_if(a1))
        case Pad =>
          val w = bitWidth(a0.tpe)
-         val diff = (c0 - w)
+         val diff = c0 - w
          if (w == BigInt(0)) Seq(a0)
          else doprim.tpe match {
            // Either sign extend or zero extend.
@@ -197,7 +196,7 @@ class VerilogEmitter extends Emitter {
        case AsClock => Seq("$unsigned(", a0, ")")
        case Dshlw => Seq(cast(a0), " << ", a1)
        case Dshl => Seq(cast(a0), " << ", a1)
-       case Dshr => (doprim.tpe) match {
+       case Dshr => doprim.tpe match {
          case (_: SIntType) => Seq(cast(a0)," >>> ", a1)
          case (_) => Seq(cast(a0), " >> ", a1)
        }
--- a/src/main/scala/firrtl/PrimOps.scala
+++ b/src/main/scala/firrtl/PrimOps.scala
@@ -102,7 +102,7 @@ object PrimOps extends LazyLogging {
   private lazy val builtinPrimOps: Seq[PrimOp] =
     Seq(Add, Sub, Mul, Div, Rem, Lt, Leq, Gt, Geq, Eq, Neq, Pad, AsUInt, AsSInt, AsClock, Shl, Shr,
         Dshl, Dshr, Neg, Cvt, Not, And, Or, Xor, Andr, Orr, Xorr, Cat, Bits, Head, Tail)
-  private lazy val strToPrimOp: Map[String, PrimOp] = builtinPrimOps map (op => op.toString -> op) toMap
+  private lazy val strToPrimOp: Map[String, PrimOp] = builtinPrimOps.map { case op : PrimOp=> op.toString -> op }.toMap
 
   /** Seq of String representations of [[ir.PrimOp]]s */
   lazy val listing: Seq[String] = builtinPrimOps map (_.toString)
--- a/src/main/scala/firrtl/PrimOps.scala
+++ b/src/main/scala/firrtl/PrimOps.scala
@@ -139,7 +139,7 @@ object PrimOps extends LazyLogging {
     def w2 = passes.getWidth(e.args(1).tpe)
     def c1 = IntWidth(e.consts.head)
     def c2 = IntWidth(e.consts(1))
-    e copy (tpe = (e.op match {
+    e copy (tpe = e.op match {
       case Add => (t1, t2) match {
         case (_: UIntType, _: UIntType) => UIntType(PLUS(MAX(w1, w2), IntWidth(1)))
         case (_: UIntType, _: SIntType) => SIntType(PLUS(MAX(w1, MINUS(w2, IntWidth(1))), IntWidth(2)))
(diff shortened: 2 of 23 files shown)


src/main/scala/firrtl/passes/ReplSeqMem.scala
```diff
@@ -88,7 +88,7 @@ Optional Arguments:
     error("No circuit name specified for ReplSeqMem!" + usage)
   )
   val target = CircuitName(passCircuit)
-  def duplicate(n: Named) = this copy (t = (t replace (s"-c:$passCircuit", s"-c:${n.name}")))
+  def duplicate(n: Named) = this copy (t = t replace(s"-c:$passCircuit", s"-c:${n.name}"))
```
jackkoenig: The deletion of the space after replace confuses me a bit. I think this might be illustrating another style-bug throughout Firrtl. I believe that you're only supposed to invoke methods using spaces if:

1. They are Arity-1:  Infix notation, eg. `names mkString ","`, and this should only be used for purely-functional methods or methods which take functions as parameters
2. They are higher-order functions, eg. `map` and `foreach`

replace is arity-2 and not higher-order so I think the right invocation may be `t.replace(...)`

This might also affect the style of copy since I do not believe it accepts high-order functions so it should generally be invoked with `.` syntax. Anyone have thoughts?

Source: http://docs.scala-lang.org/style/method-invocation.html

src/main/scala/firrtl/PrimOps.scala
```diff
@@ -102,7 +102,7 @@ object PrimOps extends LazyLogging {
   private lazy val builtinPrimOps: Seq[PrimOp] =
     Seq(Add, Sub, Mul, Div, Rem, Lt, Leq, Gt, Geq, Eq, Neq, Pad, AsUInt, AsSInt, AsClock, Shl, Shr,
         Dshl, Dshr, Neg, Cvt, Not, And, Or, Xor, Andr, Orr, Xorr, Cat, Bits, Head, Tail)
-  private lazy val strToPrimOp: Map[String, PrimOp] = builtinPrimOps map (op => op.toString -> op) toMap
+  private lazy val strToPrimOp: Map[String, PrimOp] = builtinPrimOps.map { case op : PrimOp=> op.toString -> op }.toMap
```
jackkoenig: I think this should either be
```scala
(builtinPrimOps map (op => op.toString -> op)).toMap
```
or
```scala
builtinPrimOps.map(op => op.toString -> op).toMap
```

The `.map` followed by a space looks really wrong to me

sdtwigg: Why was the case added?

If it was needed syntactically, I strongly advise for-comprehensions instead of 'map-case':
```scala
for(op: PrimOp <- builtInPrimOps) yield (op.toString, op)
```

src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -336,9 +336,9 @@ object CheckTypes extends Pass {
         case (e: WSubIndex) => e.exp.tpe match {
           case (t: VectorType) if e.value < t.size =>
           case (t: VectorType) =>
-            errors append (new IndexTooLarge(info, mname, e.value))
+            errors append new IndexTooLarge(info, mname, e.value)
```
jackkoenig: Rather than removing the parentheses we should actually shift these to `.` notation since they are modifying a mutable data structure

src/main/scala/firrtl/passes/ReplSeqMem.scala
sdtwigg: Similarly, copy should 'look more like a function' than replace since it uses named args.

I generally advocate for preferring using . and no space rather than spaces except in the rare cases where it is a short chaining of 'common binary operator-like functions' (like map, reduce, fold, and corollaries). Really, it is more of a: 'when in doubt, use function calls'.

Note that, in the case where you need to split things across multiple lines, going back to the function call notation is fairly helpful....
```scala
val myResult: Seq[Int] =
  seqInts
  .map(x => x - (x % 2))  // round down to nearest even
  .scanLeft(0)(_+_)  // take running sum of numbers
  .drop(1)  // drop the first sum
  .filter(_ % 2 == 0) // filter to only even sums
```

src/main/scala/firrtl/Emitter.scala
```diff
@@ -90,14 +90,13 @@ class VerilogEmitter extends Emitter {
       case ClockType => e
       case AnalogType(w) => e
     }
-    (x) match {
+    x match {
```
sdtwigg: Do you have plans to convert these to extractor-style? (or, at least, the  ones that are accessing every field of the case class anyway like Mux, SubAccess)

src/main/scala/firrtl/PrimOps.scala
```diff
@@ -139,7 +139,7 @@ object PrimOps extends LazyLogging {
     def w2 = passes.getWidth(e.args(1).tpe)
     def c1 = IntWidth(e.consts.head)
     def c2 = IntWidth(e.consts(1))
-    e copy (tpe = (e.op match {
+    e copy (tpe = e.op match {
```
sdtwigg: e.copy(

src/main/scala/firrtl/Utils.scala
```diff
@@ -114,22 +114,22 @@ object Utils extends LazyLogging {
...
        case (_: GroundType) => f
        case (t: BundleType) => 
-         val (_, flip) = ((t.fields foldLeft (i, None: Option[Orientation])){
-            case ((n, ret), x) if n < get_size(x.tpe) => ret match {
-              case None => (n, Some(get_flip(x.tpe,n,times(x.flip,f))))
-              case Some(_) => (n, ret)
-            }
-            case ((n, ret), x) => (n - get_size(x.tpe), ret)
-         })
+         val (_, flip) = (t.fields foldLeft(i, None: Option[Orientation])) {
```
sdtwigg: folds aren't arity-1 and should be more explicitly done as a function call.

(After seeing this, I realize that I misspoke earlier when I referred to fold as a common enough function to have this syntax.)

src/main/scala/firrtl/Utils.scala
```diff
@@ -220,25 +220,26 @@ object Utils extends LazyLogging {
...
               val ls = get_valid_points(f1_tpe, f2.tpe, f1_times, f2_times)
-              (points ++ (ls map {case (x, y) => (x + ilen, y + jlen)}), jlen + get_size(f2.tpe))
+              (points ++ (ls map { case (x, y) => (x + ilen, y + jlen) }), jlen + get_size(f2.tpe))
           }
-        })._1
+        }._1
       case (t1: VectorType, t2: VectorType) =>
         val size = math.min(t1.size, t2.size)
-        (((0 until size) foldLeft (Seq[(Int, Int)](), 0, 0)){case ((points, ilen, jlen), _) =>
+        ((0 until size) foldLeft(Seq[(Int, Int)](), 0, 0)) { case ((points, ilen, jlen), _) =>
```
sdtwigg: For future review (since your changes do make it slightly better), this fold is still a bit of a mess.
The map-case should be changed to a for-comprehension.
Instead of binding to ls, can do
```scala
val newPoints = for((x,y) <- get_valid_points(t1.tpe, t2.tpe, flip1, flip2)) yield (x + ilen, y + jlen)
```
Also, a lot of style guides discourage the use of ._1 to access Bundle fields.
The many small appends isn't very efficient for a Seq. Doing a scanLeft then flatten may be better.

src/main/scala/firrtl/passes/AnnotateValidMemConfigs.scala
```diff
@@ -58,7 +57,7 @@ case class SRAMConfig(
     val fieldMap = getClass.getDeclaredFields.map { f =>
```
sdtwigg: .map then space here is inappropriate. Either no . or no space.

src/main/scala/firrtl/passes/InferTypes.scala
```diff
@@ -110,7 +110,7 @@ object CInferTypes extends Pass {
 
     def infer_types_e(types: TypeMap)(e: Expression) : Expression =
       e map infer_types_e(types) match {
-         case (e: Reference) => e copy (tpe = (types getOrElse (e.name, UnknownType)))
+         case (e: Reference) => e copy (tpe = types getOrElse(e.name, UnknownType))
```
sdtwigg: getOrElse isn't arity-1..

If you wanted to chain arity-1 can do get(key): Option[B] and then getOrElse on that option.

src/main/scala/firrtl/passes/RemoveCHIRRTL.scala
```diff
@@ -56,7 +56,7 @@ object RemoveCHIRRTL extends Pass {
       (e1s zip e2s) map { case (e1, e2) => Mux(e.cond, e1, e2, mux_type(e1, e2)) }
     case (e: ValidIf) =>
       create_exps(e.value) map (e1 => ValidIf(e.cond, e1, e1.tpe))
-    case (e) => (e.tpe) match {
+    case (e) => e.tpe match {
```
sdtwigg: just 'case e'

chick: The case statements are going to be handle by a separate cleanup pass. I haven't touched any of that stuff yet.  I am not sure why these got wrapped in parens int the first place.

src/main/scala/firrtl/passes/AnnotateValidMemConfigs.scala
chick: I am more familiar and more comfortable with braces being preceded by a space, but not parens.  Mostly I'd here I am trying to get the source to pass the standard style guide without warning.  I don't have strong feelings about this but I think it makes code blocks stand out, and has a symmetry with match {

src/main/scala/firrtl/PrimOps.scala
chick: I had a hard time converting this statement.  my ide preferred using case.  I am not a fan of yield, it's clear enough if you really understand for comprehensions, but I find the asymmetry with ordinary usages of for such as the inability to do intermediate calculations before the yielded quantity and the difficulties in inserting breakpoints in the for comprehensions sufficient to keep me from it.

src/main/scala/firrtl/passes/ReplSeqMem.scala
chick: I have to say I prefer dot notation.  But I was trying to minimize the differences from the previous "un"-stylish code.  I have to admit to relying on IntelliJ to fix lot's of these things in batch.  It would be nice to include our own sensibilities in a style guide but I think that should be  separate or at an extended version of the style-cleanup-take-N passes

src/main/scala/firrtl/Emitter.scala
chick: I prefer extractor style

jackkoenig:
In the spirit of recognizing that this PR isn't intended to fix every style issue, I went ahead and fixed the ones that I thought were incorrectly removing a space but not changing to `.`-style invocation  (eg. `(list foldLeft (foo, bar)) { ... }` -> `(list foldLeft(foo, bar)) { ... }`, instead now replaced with `list.foldLeft(foo, bar) { ... }`) . 

As for the many style issues with copy, etc. Those are left as a future fix. I even left the space invocation used on mutable datastructures since those issues should be fixed together

PR #321, opened 2016-09-26
Fix CInferMDir

donggyukim:
This fixes a bug in CInferMDir revealed by #260 with some style fixes.

@ucbjrl Can you try this branch?


ucbjrl:
It passes Jenkins tests (i.e., fixes earlier chisel-testers failure reported with #260)

PR #322, opened 2016-09-27
Add fixed point type

azidar:
Preliminary implementation, not fully featured.


src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -304,12 +304,39 @@ object CheckTypes extends Pass {
         if (ls exists (x => wt(ls.head.tpe) != wt(e.tpe)))
           errors append new OpNotAllSameType(info, mname, e.op.serialize)
       }
-      def allInt(ls: Seq[Expression]) {
+      def all_USC(ls: Seq[Expression]) {
```
chick: Why mix in snake case here

src/main/scala/firrtl/passes/fixedType/FPInferTypes.scala
```diff
@@ -0,0 +1,204 @@
+//package dspTools.firrtlPasses
```
chick: What are all these commented out files for

PR #323, opened 2016-09-27
DO NOT MERGE - Prototype of fold and children on FirrtlNodes

jackkoenig:
Just starting the discussion of what children and fold mean and providing a simple implementation so that I can show a possible use. The real implementation will define fold on all nodes rather than using this slow general definition.

The idea is that using "map" when you don't really want to change the circuit is a bit clumsy and/or does not show intent very well. Here I'm trying to provide other techniques for traversing the graph and returning a different type (eg. B in foldLeft)


jackkoenig:
Maybe still worth revisiting but this has been dead for a while

PR #324, opened 2016-09-28
Scala style cleanup take 5

chick:
With a couple of minor exceptions this take is just about fixing the shadowing of variables in match statements.  The general strategy is simply to append an x to the variable name for each level of match
```scala
  e match {
    case ex: A => ex match {
      case exx: B =>
```
The dangerous cases seemed to center on usages like
```scala
  e map f(x) match {
    case e: A =>
      ...
    case e => e
```
It is tempting to map that last line to ```case _ => e``` but this tended to spectacularly blow up the test harness.  I have, to a fault, settled on converting it to `case ex => ex

### DIFF
--- a/src/main/scala/firrtl/Utils.scala
+++ b/src/main/scala/firrtl/Utils.scala
@@ -72,7 +72,7 @@ object Utils extends LazyLogging {
         case 1 => newStmts.head
         case _ => Block(newStmts)
       }
-    case _ => s
+    case sx => sx
   }
 
   /** Indent the results of [[ir.FirrtlNode.serialize]] */
--- a/src/main/scala/firrtl/Utils.scala
+++ b/src/main/scala/firrtl/Utils.scala
@@ -94,27 +94,27 @@ object Utils extends LazyLogging {
   def create_exps(n: String, t: Type): Seq[Expression] =
     create_exps(WRef(n, t, ExpKind, UNKNOWNGENDER))
   def create_exps(e: Expression): Seq[Expression] = e match {
-    case (e: Mux) =>
-      val e1s = create_exps(e.tval)
-      val e2s = create_exps(e.fval)
+    case ex: Mux =>
+      val e1s = create_exps(ex.tval)
+      val e2s = create_exps(ex.fval)
       e1s zip e2s map {case (e1, e2) =>
-        Mux(e.cond, e1, e2, mux_type_and_widths(e1, e2))
+        Mux(ex.cond, e1, e2, mux_type_and_widths(e1, e2))
       }
-    case (e: ValidIf) => create_exps(e.value) map (e1 => ValidIf(e.cond, e1, e1.tpe))
-    case (e) => e.tpe match {
-      case (_: GroundType) => Seq(e)
+    case ex: ValidIf => create_exps(ex.value) map (e1 => ValidIf(ex.cond, e1, e1.tpe))
+    case ex => ex.tpe match {
+      case (_: GroundType) => Seq(ex)
       case (t: BundleType) => (t.fields foldLeft Seq[Expression]())((exps, f) =>
-        exps ++ create_exps(WSubField(e, f.name, f.tpe,times(gender(e), f.flip))))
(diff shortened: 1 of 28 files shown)


src/main/scala/firrtl/passes/PadWidths.scala
```diff
@@ -34,30 +34,30 @@ object PadWidths extends Pass {
...
+        ex copy (op = Dshlw, args = Seq(fixup(width(ex.tpe))(ex.args.head), ex.args(1)))
       case Shl =>
         // special case as arg should be same width as result
-        e copy (op = Shlw, args = Seq(fixup(width(e.tpe))(e.args.head)))
-      case _ => e
+        ex copy (op = Shlw, args = Seq(fixup(width(ex.tpe))(ex.args.head)))
+      case _ => ex
     }
-    case e => e
+    case _ => e
```
azidar: This is wrong - we need to return ```e map onExp```, not ```e```.

src/main/scala/firrtl/Utils.scala
```diff
@@ -72,7 +72,7 @@ object Utils extends LazyLogging {
         case 1 => newStmts.head
         case _ => Block(newStmts)
       }
-    case s => s
+    case _ => s
```
azidar: Need to return ```s map squashEmpty```, not ```s```

src/main/scala/firrtl/passes/PadWidths.scala
```diff
@@ -34,30 +34,30 @@ object PadWidths extends Pass {
...
-    case s: Connect =>
-      s copy (expr = fixup(width(s.loc))(s.expr))
-    case s: DefRegister =>
-      s copy (init = fixup(width(s.tpe))(s.init))
-    case s => s map onStmt
+    case sx: Connect =>
+      sx copy (expr = fixup(width(sx.loc))(sx.expr))
+    case sx: DefRegister =>
+      sx copy (init = fixup(width(sx.tpe))(sx.init))
+    case _ => s map onStmt
```
azidar: Return ```s map onExp map onStmt``` not ```s map onStmt```

src/main/scala/firrtl/passes/Passes.scala
```diff
@@ -80,34 +80,34 @@ object PullMuxes extends Pass {
...
-           case (ex: ValidIf) => ValidIf(ex.cond,
-              WSubAccess(ex.value, e.index, e.tpe, e.gender), e.tpe)
-           case (ex) => e
+         case ex: WSubAccess => ex.exp match {
+           case exx: Mux => Mux(exx.cond,
+              WSubAccess(exx.tval, ex.index, ex.tpe, ex.gender),
+              WSubAccess(exx.fval, ex.index, ex.tpe, ex.gender), ex.tpe)
+           case exx: ValidIf => ValidIf(exx.cond,
+              WSubAccess(exx.value, ex.index, ex.tpe, ex.gender), ex.tpe)
+           case _ => ex  // case exx => exx causes failed tests
```
azidar: This is because we are returning the WSubAccess, not the sub expression. It is correct to return ```ex```

src/main/scala/firrtl/passes/SplitExpressions.scala
```diff
@@ -31,14 +31,14 @@ object SplitExpressions extends Pass {
...
-          case e => e
+          case _ => e
         }
 
         // Recursive. Splits compound nodes
         def onExp(e: Expression): Expression =
           e map onExp match {
             case ex: DoPrim => ex map split
-            case v => v
+            case _ => e
```
azidar: Need to return ```e map onExp``` not ```e```

src/main/scala/firrtl/Utils.scala
chick: So I assume that means the new line 75 is ok. Right?

azidar: No, it needs to change from
```
case _ => s
```

to 

```
case sx => sx
```

chick: But that is what it says now.

Issue #325, opened 2016-09-28
Bug in CHIRRTL mem handling


aswaterman:
The following bails out in RemoveCHIRRTL's collect_smems_and_mports:

```
circuit Queue :
  module Queue :
    input clock : Clock

    cmem ram : UInt<1>[2]
    infer mport T_107 = ram[UInt(0)], clock
```

The issue seems to be that an "infer" port is declared that isn't actually used.  The right behavior should be to disregard the port.  Here is a stack trace:

```
Exception in thread "main" scala.MatchError: MInfer (of class firrtl.MInfer$)
  at firrtl.passes.RemoveCHIRRTL$.collect_smems_and_mports(RemoveCHIRRTL.scala:76)
  at firrtl.passes.RemoveCHIRRTL$$anonfun$collect_smems_and_mports$1.apply(RemoveCHIRRTL.scala:84)
  at firrtl.passes.RemoveCHIRRTL$$anonfun$collect_smems_and_mports$1.apply(RemoveCHIRRTL.scala:84)
  at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
  at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
  at scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:59)
  at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:48)
  at scala.collection.TraversableLike$class.map(TraversableLike.scala:245)
  at scala.collection.AbstractTraversable.map(Traversable.scala:104)
  at firrtl.ir.Block.mapStmt(IR.scala:227)
  at firrtl.Mappers$StmtMagnet$$anon$1.map(Mappers.scala:41)
  at firrtl.Mappers$StmtMap$.map$extension(Mappers.scala:55)
  at firrtl.passes.RemoveCHIRRTL$.collect_smems_and_mports(RemoveCHIRRTL.scala:84)
  at firrtl.passes.RemoveCHIRRTL$$anonfun$remove_chirrtl_m$1.apply(RemoveCHIRRTL.scala:247)
  at firrtl.passes.RemoveCHIRRTL$$anonfun$remove_chirrtl_m$1.apply(RemoveCHIRRTL.scala:247)
  at firrtl.ir.Module.mapStmt(IR.scala:443)
  at firrtl.Mappers$ModuleMagnet$$anon$11.map(Mappers.scala:115)
[...]

PR #326, opened 2016-09-28
Fix RemoveCHIRRTL for unused memory ports

donggyukim:
This addresses #325.

@aswaterman 
Memory port directions were not inferred during CInferMDir if not being used, which was not accepted in RemoveCHIRRTL.

### DIFF
--- a/src/main/scala/firrtl/passes/RemoveCHIRRTL.scala
+++ b/src/main/scala/firrtl/passes/RemoveCHIRRTL.scala
@@ -77,6 +77,7 @@ object RemoveCHIRRTL extends Pass {
           case MRead => p.readers += MPort(s.name, s.exps(1))
           case MWrite => p.writers += MPort(s.name, s.exps(1))
           case MReadWrite => p.readwriters += MPort(s.name, s.exps(1))
+          case MInfer => // direction may not be inferred if it's not being used
         }
         mports(s.mem) = p
       case s =>
@@ -144,6 +145,7 @@ object RemoveCHIRRTL extends Pass {
               raddrs(e.name) = SubField(SubField(Reference(s.mem, ut), s.name, ut), "en", ut)
             case _ => ens += "en"
           }
+        case MInfer => // do nothing if it's not being used
       }
       Block(
         (addrs map (x => Connect(s.info, SubField(SubField(Reference(s.mem, ut), s.name, ut), x, ut), s.exps.head))) ++
--- a/src/test/scala/firrtlTests/ChirrtlSpec.scala
+++ b/src/test/scala/firrtlTests/ChirrtlSpec.scala
@@ -94,6 +94,9 @@ class ChirrtlSpec extends FirrtlFlatSpec {
     }
   }
 
+  it should "compile" in {
+    compileFirrtlTest("EmptyChirrtlMem", "/features")
+  }
   it should "compile and run" in {
     runFirrtlTest("ChirrtlMems", "/features")
   }
--- a/src/test/resources/features/EmptyChirrtlMem.fir
+++ b/src/test/resources/features/EmptyChirrtlMem.fir
@@ -1,0 +1,5 @@
+circuit Queue :
+  module Queue :
+    input clock : Clock
+    cmem ram : UInt<1>[2]
+    infer mport T_107 = ram[UInt(0)], clock


src/main/scala/firrtl/passes/RemoveCHIRRTL.scala
```diff
@@ -118,7 +119,7 @@ object RemoveCHIRRTL extends Pass {
         set_write(rws, "wdata", "wmask")
       val mem = DefMemory(s.info, s.name, s.tpe, s.size, 1, if (s.seq) 1 else 0,
                   rds map (_.name), wrs map (_.name), rws map (_.name))
-      Block(mem +: stmts)
+      if (stmts.nonEmpty) Block(mem +: stmts) else EmptyStmt
```
aswaterman: This seems like it's addressing a separate issue (delete the memory if it has no ports) -- I guess the problem is that other passes are unhappy if the memory has no ports?

shunshou: If a memory has no ports, why should it exist? If deleting the memory will make it so that other passes have no reason to be sad... isn't that the right decision?

I guess it'd still be problematic if someone wrote FIRRTL but not CHIRRTL and then passed it through some intermediate lowering compiler... You'd need some kind of clean up pass before every pass that can break.

donggyukim: @aswaterman @shunshou Isn't it weird to keep memories with no ports? Even the previous implementation doesn't allow them. This handles CHIRRTL memories only, so it shouldn't affect the other memories.

shunshou: But it's still possible to write memories without ports using FIRRTL, and if that code gets compiled, other passes would break... By only changing RemoveCHIRRTL, you wouldn't be sufficiently generic? Unless I'm mistaken...

donggyukim: @shunshou CHIRRTL memories and FIRRTL memories have different types(`CDefMemory` and `DefMemory`, respectively). And this works only for `CDefMemory`.

aswaterman: I wasn't saying this is the wrong thing to do; I was just trying to understand it.

Note that other passes still might have to understand Mems with no ports, though!  Future optimizations could delete the ports.

donggyukim: @aswaterman Yes, I agree. The reason I did this is not to expose those side effects for now.

shunshou: Ok, I see. Is it possible to create (in a pass or via a *.fir file) a DefMemory with an empty list of ports? And if so, will passes break? I feel like maybe mine would, but I'd need to check...

donggyukim: @shunshou It's not clear to see whether or not it would break, but you can create no-port FIRRTL memories and they won't disappear. We could delete no-port memories in DCE later.

src/test/scala/firrtlTests/ChirrtlSpec.scala
```diff
@@ -94,7 +94,37 @@ class ChirrtlSpec extends FirrtlFlatSpec {
     }
   }
 
-  it should "compile and run" in {
-    runFirrtlTest("ChirrtlMems", "/features")
```
jackkoenig: Why does this delete an existing test?

src/test/scala/firrtlTests/ChirrtlSpec.scala
```diff
@@ -94,7 +94,37 @@ class ChirrtlSpec extends FirrtlFlatSpec {
...
+        if (!check) {
+          errors append new PassException("Queue should be empty")
+          errors.trigger
+        }
+        c
+      }
+    }
+    (passes :+ CheckQueue).foldLeft(Parser.parse(input.split("\n").toIterator)) {
+      (c: Circuit, p: Pass) => p.run(c)
+    }
```
jackkoenig: This could also perhaps be more concisely described by writing a self-contained Firrtl Test (like features/ChirrtlMems.fir) and just check if it compiles or something via compileFirrtlTest.

src/test/scala/firrtlTests/ChirrtlSpec.scala
donggyukim: This is a big mistake...

src/test/scala/firrtlTests/ChirrtlSpec.scala
```diff
@@ -31,7 +31,7 @@ import java.io._
 import org.scalatest._
 import org.scalatest.prop._
 import firrtl.Parser
-import firrtl.ir.Circuit
+import firrtl.ir._
```
jackkoenig: Given the shift to a compileTest (thanks for that!), this change in import is unnecessary

src/main/scala/firrtl/passes/RemoveCHIRRTL.scala
jackkoenig: I'm actually of the opinion that the empty memory should be propagated then deleted later, my test in #327 shows that zero-port memories are, in general, supported

donggyukim:
Overridden by #355

PR #336, opened 2016-10-13
Reorganized memory blackboxing

azidar:
Moved to new package memlib
Added comments
Moved utility functions around
Removed unused AnnotateValidMemConfigs.scala (will add it back in when it is actually used)


src/main/scala/firrtl/passes/memlib/MemIR.scala
```diff
@@ -0,0 +1,43 @@
...
+    writeLatency: Int,
+    readLatency: Int,
+    readers: Seq[String],
+    writers: Seq[String],
+    readwriters: Seq[String],
+    readUnderWrite: Option[String],
+    maskGran: Option[BigInt],
+    memRef: Option[String]
+    //pins: Seq[Pin],
+    ) extends Statement with IsDeclaration {
```
shunshou: Still don't understand why you can't extend DefMemory instead? then you can just rely on DefMemory's serialize, map*, etc.

src/main/scala/firrtl/passes/memlib/MemIR.scala
```diff
@@ -0,0 +1,43 @@
...
+    s"mem $name :" + info.serialize +
+    indent(
+      (Seq("\ndata-type => " + dataType.serialize,
+          "depth => " + depth,
+          "read-latency => " + readLatency,
+          "write-latency => " + writeLatency) ++
+          (readers map ("reader => " + _)) ++
+          (writers map ("writer => " + _)) ++
+          (readwriters map ("readwriter => " + _)) ++
+       Seq("read-under-write => undefined")) mkString "\n")
```
shunshou: When read-under-write annotations actually get passed through from Chisel, it'd be nice if you only had to change in 1 place...

src/main/scala/firrtl/passes/memlib/MemIR.scala
jackkoenig: You can't extend case classes and DefMemory is a case class. We could consider making built-in Firrtl IR nodes "DIY" case classes (which would then be extendable), but that is quite a bit of code to write since we'd have to reimplement all of the methods that case classes generate

src/main/scala/firrtl/passes/memlib/MemIR.scala
```diff
@@ -0,0 +1,43 @@
...
+    readLatency: Int,
+    readers: Seq[String],
+    writers: Seq[String],
+    readwriters: Seq[String],
+    readUnderWrite: Option[String],
+    maskGran: Option[BigInt],
+    memRef: Option[String]
+    //pins: Seq[Pin],
+    ) extends Statement with IsDeclaration {
+  def serialize: String =
```
jackkoenig: So serialization on intermediate passes is nice for debugging. Once we have the annotation system a bit more complete we can serialize them as well, but for now I think it's fine for the thing to just serialize as a defmem. In fact rather than reimplementing the code from defmem's serialize, I'd just create one and call serialize on that. ie.

```scala
def serialize: String = DefMemory(info, name, dataType, depth,...).serialize
```

I think that would show the intent here better, and since serialize is constructing strings anyway, I'm sure creating an extra short-lived object is irrelevant for performance.

jackkoenig: I just noticed toMem below, you could even just write 

```scala
def serialize: String = this.toMem.serialize
```

src/main/scala/firrtl/passes/memlib/RenameAnnotatedMemoryPorts.scala
```diff
@@ -0,0 +1,82 @@
...
+
+/** Changes memory port names to standard port names (i.e. RW0 instead T_408)
+ *
+ *  TODO(shunshou): Module namespace can collide with new names!
+ */
+object RenameAnnotatedMemoryPorts extends Pass {
+
+  def name = "Rename Annotated Memory Ports"
+
+  type AnnotatedMemories = collection.mutable.ArrayBuffer[DefAnnotatedMemory]
```
shunshou: Where is this used?

src/main/scala/firrtl/passes/memlib/MemIR.scala
shunshou: I think doing that is better than copy+pasting functions from DefMemory into DefAnnotatedMemory

src/main/scala/firrtl/passes/memlib/RenameAnnotatedMemoryPorts.scala
```diff
@@ -0,0 +1,82 @@
...
+import firrtl.Utils._
+import firrtl.Mappers._
+import AnalysisUtils._
+import MemPortUtils._
+import MemTransformUtils._
+
+
+/** Changes memory port names to standard port names (i.e. RW0 instead T_408)
+ *
+ *  TODO(shunshou): Module namespace can collide with new names!
```
shunshou: Run uniquify @ end of transform

src/main/scala/firrtl/passes/memlib/ReplaceMemMacros.scala
```diff
@@ -1,28 +1,94 @@
...
+  def createMemModule(m: DefAnnotatedMemory, wrapperName: String): Seq[DefModule] = {
     assert(m.dataType != UnknownType)
     val wrapperIoType = memToBundle(m)
     val wrapperIoPorts = wrapperIoType.fields map (f => Port(NoInfo, f.name, Input, f.tpe))
+    // Creates a type with the write/readwrite masks omitted if necessary
     val bbIoType = memToFlattenBundle(m)
     val bbIoPorts = bbIoType.fields map (f => Port(NoInfo, f.name, Input, f.tpe))
     val bbRef = createRef(m.name, bbIoType)
-    val hasMask = containsInfo(m.info, "maskGran")
+    val hasMask = !m.maskGran.isEmpty
```
shunshou: Does isEmpty return true when maskGran = None?

shunshou: Ok I guess it does... but maybe you'd rather use "isDefined" instead of ! isEmpty

src/main/scala/firrtl/passes/memlib/ResolveMaskGranularity.scala
```diff
@@ -0,0 +1,124 @@
...
+    case DoPrim(PrimOps.Bits, args, Seq(msb, lsb), tpe) =>  
+      val extractionWidth = (msb - lsb) + 1
+      val nodeWidth = bitWidth(args.head.tpe)
+      // if you're extracting the full bitwidth, then keep searching for origin
+      if (nodeWidth == extractionWidth) getOrigin(connects)(args.head) else e
+    case DoPrim((PrimOps.AsUInt | PrimOps.AsSInt | PrimOps.AsClock), args, _, _) => 
+      getOrigin(connects)(args.head)
+    case ValidIf(cond, value, ClockType) => getOrigin(connects)(value)
+    // note: this should stop on a reg, but will stack overflow for combinational loops (not allowed)
+    case _: WRef | _: WSubField | _: WSubIndex | _: WSubAccess if kind(e) != RegKind =>
```
shunshou: Just want to make sure I understand your reasoning:

Originally, I had:

```
  def getConnectOrigin(connects: Connects)(node: String): Expression =
    connects get node match {
      case None => EmptyExpression
      case Some(e) => getOrigin(connects, e)
    }
  def getConnectOrigin(connects: Connects, e: Expression): Expression =
    getConnectOrigin(connects)(e.serialize)
```

Which kind of did things in a round-about way.

You got rid of those two and instead added 

```
def getOrigin(connects: Connects, s: String): Expression =
 +    getOrigin(connects)(WRef(s, UnknownType, ExpKind, UNKNOWNGENDER))
```

The substitutions aren't 100% functionally equivalent, but I think that's ok? 

Instead of using getConnectOrigin(connects: Connects, e: Expression), you directly use your new def getOrigin(connects: Connects)(e: Expression) [previous version, just no longer private].

2 things that are different:

for case _: WRef | _: WSubField | _: WSubIndex | _: WSubAccess if kind(e) != RegKind =>

You'll never get to "connects get node" -> "connects get e.serialize" if e is a RegKind. 

Also, if e.serialize isn't a key in connects, rather than returning EmptyExpression, you just return e. 

This is ok, b/c it makes sense that RegKinds and nodes that aren't found in the connects HashMap are just connected to themselves = unconnected to anything else?

src/main/scala/firrtl/passes/memlib/ResolveMaskGranularity.scala
```diff
@@ -0,0 +1,124 @@
...
+  */
+object ResolveMaskGranularity extends Pass {
+  def name = "Resolve Mask Granularity"
+
+  /** Returns the number of mask bits, if used
+    */
+  def getMaskBits(connects: Connects, wen: Expression, wmask: Expression): Option[Int] = {
+    val wenOrigin = getOrigin(connects)(wen)
+    val wmaskOrigin = connects.keys filter
+      (_ startsWith wmask.serialize) map {s: String => getOrigin(connects)(WRef(s, UnknownType, ExpKind, UNKNOWNGENDER))}
```
shunshou: I don't understand why you use WRef here when you have the function getOrigin(connects, string) that does the same thing?

src/main/scala/firrtl/passes/memlib/ResolveMemoryReference.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  def updateMemMods(m: DefModule) = {
+    val uniqueMems = new AnnotatedMemories
+    (m map updateMemStmts(uniqueMems))
+  }
+
+  def run(c: Circuit) = c copy (modules = c.modules map updateMemMods)
+}
+
+// vim: set ts=4 sw=4 et:
```
shunshou: ???

shunshou:
@azidar Ok, I'm done

src/main/scala/firrtl/passes/memlib/MemIR.scala
azidar: Fixed.

src/main/scala/firrtl/passes/memlib/MemIR.scala
azidar: Fixed.

src/main/scala/firrtl/passes/memlib/MemIR.scala
azidar: Agreed, leaving it as is.

src/main/scala/firrtl/passes/memlib/ReplaceMemMacros.scala
azidar: Fixed.

src/main/scala/firrtl/passes/memlib/ResolveMaskGranularity.scala
```diff
@@ -0,0 +1,124 @@
...
+    * Named components are serialized LHS of connections, nodes, invalids
+    */
+  def getConnects(m: DefModule): Connects = {
+    def getConnects(connects: Connects)(s: Statement): Statement = {
+      s match {
+        case Connect(_, loc, expr) =>
+          connects(loc.serialize) = expr
+        case DefNode(_, name, value) =>
+          connects(name) = value
+        case IsInvalid(_, value) =>
```
azidar: Discussed IRL, decided origin for invalid is WInvalid. Leaving as is.

src/main/scala/firrtl/passes/memlib/ResolveMaskGranularity.scala
azidar: That is correct (discussed offline). The behavior is the same if used on a correct circuit, the only difference is when called with an unknown reference. Leaving as is.

src/main/scala/firrtl/passes/memlib/ResolveMemoryReference.scala
azidar: Fixed.

PR #337, opened 2016-10-14
Create a simple system for executions and command line parameters

chick:
New model for tracking parameters and having those parameters
register scopt command to allow the parameters to be set by
command line args.
Create composable forms of the these parameters to allow separate
elements of the chisel3 toolchain to combine these parameters
Create execution return structures that simplify return values
to earlier toolchain elements


src/main/scala/firrtl/Driver.scala
```diff
@@ -1,189 +1,182 @@
...
+            success
+          case failure: FirrtlExecutionFailure =>
+            optionsManager.showUsageAsError()
+            failure
+          case result =>
+            throw new Exception(s"Error: Unknown Firrtl Execution result $result")
         }
-      case None => throw new Exception("No specified compiler option." + usage)
+      case _ =>
+        FirrtlExecutionFailure("")
```
azidar: showUsageAsError?

src/main/scala/firrtl/ExecutionOptionsManager.scala
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+  parser.opt[String]("info-mode")
+    .valueName ("<ignore|use|gen|append>")
+    .foreach { x =>
+      firrtlOptions = firrtlOptions.copy(infoModeName = x.toLowerCase)
+    }
+    .validate { x =>
+      if (Array("ignore", "use", "gen", "append").contains(x.toLowerCase)) parser.success
+      else parser.failure(s"$x not a legal compiler")
```
azidar: Incorrect failure message

src/main/scala/firrtl/ExecutionOptionsManager.scala
```diff
@@ -0,0 +1,327 @@
...
+  parser.opt[Seq[String]]("infer-rw")
+    .abbr("firw")
+    .valueName ("<circuit[,...]>")
+    .foreach { x =>
+      firrtlOptions = firrtlOptions.copy(
+        annotations = firrtlOptions.annotations ++
+          x.map { value => passes.InferReadWriteAnnotation(value, TransID(-1)) }
+      )
+    }.text {
+      "Enable readwrite port inference for the target circuit, for multiples separate with commas and no spaces"
```
azidar: Should only have one, multiples should not be allowed.

src/main/scala/firrtl/ExecutionOptionsManager.scala
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+  parser.opt[Seq[String]]("repl-seq-mem")
+    .abbr("frsq")
+    .valueName ("-c:<circuit>:-i:<filename>:-o:<filename>[,...]")
```
azidar: I don't think comma separated is legal

src/test/scala/firrtlTests/DriverSpec.scala
```diff
@@ -0,0 +1,157 @@
...
+        firrtlOptions.annotations.length should be (3)
+        firrtlOptions.annotations.foreach { annotation =>
+          annotation shouldBe a [passes.InlineAnnotation]
+        }
+      }
+      "infer-rw annotation" in {
+        val optionsManager = new ExecutionOptionsManager("test") with HasFirrtlOptions
+
+        optionsManager.parse(
+          Array("--infer-rw", "ignore,use,gen,append")
```
azidar: should be --file-info, not --infer-rw, and then a separate test for --infer-rw

src/main/scala/firrtl/ExecutionOptionsManager.scala
chick: I don't understand comment.  If this option can have multiple values then scopt wants them to be comma separated.  Or are you saying there cannot be multiple values

azidar:
Using option -fif without -fof will run the compiler, not write the output, but not error.

chick:
Attn: @azidar @aswaterman @palmer-dabbelt @colinschmidt @sdtwigg 
A coordinated set of changes to the *core four* (Firrtl, Chisel3, Firrtl-interpreter and Chisel-testers).  Each has a PR for execute-harness-plan-3
On each of the core four
 - A Driver.execute method with two signatures
  - On that takes a ExecutionOptionsManger which is a collection of startup parameters in case classes and runs the toolchain with them
  - One that takes an Array[String] command line arguments and builds a options manager and passes it to the other execute method.
 - Command line arguments are standardized around the [/scopt/scopt](/scopt/scopt) project
The standardization of options **will** introduce some api changes, particularly with firrtl.  I am hoping to get feedback on how painful this will be for the various constituencies.

For example:  Firrtl use to look like
```firrtl -i input.fir -o output.v -X verilog```
This would become in its simplest form
```firrtl --top-name dut```   
or being explicit about the target-dir
```firrtl --target-dir work_dir --top-name dut```
or
```firrtl -td work_directory -tn dut```
Both would take a file named work_dir/dut.fir and produces a work_dir/dut.v

Overrides to the filename are still available so you can still say, when using -i and -o the target-dir is ignored
```firrtl -i dut.fir -o dut.v -X verilog```

There are reasonable defaults for most of the settings.  In particular
target-dir is test_work_dir in line with the way the chisel-testers works
top-name defaults when possible to the dut default (mostly chisel and above where the Module is passed into the execute)

Please opine on this and if it looks like it will harm your work flow and make files.  But please do so quickly.

**Comment by colinschmidt** [comment]
It seems the only rocket-chip changes are: 
```
--inferRW => --infer-rw
--replSeqMem => --repl-seq-mem
```
which seems reasonable.

I had a question about how default values are handled though. After a quick look through this PR I couldn't find what the default for info-mode is now. I had submitted a PR recently to make it append and this code was deleted with no similar addition. So in addition to this specific case I just want to know whether other defaults for non-specified options are changing, and if that is intentional.

**Comment by chick** [comment]
The defaults for Firrtl are

case class FirrtlExecutionOptions(
    inputFileNameOverride:  String = "",
    outputFileNameOverride: String = "",
    compilerName:           String = "verilog",
    infoModeName:           String = "use",
    inferRW:                Seq[String] = Seq.empty,
    firrtlSource:           Option[String] = None,
    annotations:            List[Annotation] = List.empty)
  extends ComposableOptions {

Other than changing the file stuff a bit to try to make the different
core four repo's be more consistent.

I tried to keep all defaults the same.  I will try and confirm also
that the usage will show this too.


Chick Markley
Staff Programmer -- Aspire Lab
University of California, Berkeley
583-2 Soda Hall, Berkeley, CA 94720-1776
chick@berkeley.edu

**Comment by colinschmidt** [comment]
I'd prefer if we could keep the default info mode append as was accepted and merged via my PR #328

File: src/main/scala/firrtl/ExecutionOptionsManager.scala [hunk 6]
```diff
@@ -0,0 +1,324 @@
...
+                                   emitType: String,
+                                   emitted:  String
+                                 ) extends FirrtlExecutionResult
+
+/**
+  * The firrtl compilation failed.
+  *
+  * @param message  Some kind of hint as to what went wrong.
+  */
+case class FirrtlExecutionFailure(message: String) extends FirrtlExecutionResult
```
jackkoenig: This is essentially the same as changing FirrtlExecutionSuccess to FirrtlExecutionResult and then making execute return Try[FirrtlExecutionResult]. Thoughts on doing that instead? 

http://www.scala-lang.org/api/current/#scala.util.Try

src/main/scala/firrtl/ExecutionOptionsManager.scala
```diff
@@ -0,0 +1,324 @@
...
+  * Indicates a successful execution of the firrtl compiler, returning the compiled result and
+  * the type of compile
+  *
+  * @param emitType  The name of the compiler used, currently "high", "low", or "verilog"
+  * @param emitted   The text result of the compilation, could be verilog or firrtl text.
+  */
+case class FirrtlExecutionSuccess(
+                                   emitType: String,
+                                   emitted:  String
+                                 ) extends FirrtlExecutionResult
```
jackkoenig: the case class parameters should be tabbed with 4 spaces

src/main/scala/firrtl/ExecutionOptionsManager.scala
```diff
@@ -0,0 +1,324 @@
...
+    }
+    .text {
+      "Replace sequential memories with blackboxes + configuration file"
+    }
+
+  parser.note("")
+
+}
+
+trait FirrtlExecutionResult
```
jackkoenig: This trait should be sealed

src/main/scala/firrtl/ExecutionOptionsManager.scala
```diff
@@ -0,0 +1,324 @@
...
+  * Most of the chisel toolchain components require a topName which defines a circuit or a device under test.
+  * Much of the work that is done takes place in a directory.
+  * It would be simplest to require topName to be defined but in practice it is preferred to defer this.
+  * For example, in chisel, by deferring this it is possible for the execute there to first elaborate the
+  * circuit and then set the topName from that if it has not already been set.
+  */
+case class CommonOptions(
+                          topName:       String = "",
+                          targetDirName: String = "test_run_dir"
+                        ) extends ComposableOptions {
```
jackkoenig: These parameters should be tabbed by 4 spaces

src/main/scala/firrtl/ExecutionOptionsManager.scala
chick: sealed it

src/main/scala/firrtl/ExecutionOptionsManager.scala
chick: made one line

jackkoenig: missed one

src/main/scala/firrtl/ExecutionOptionsManager.scala
```diff
@@ -0,0 +1,324 @@
...
+    }
+    .validate { x =>
+      if (Array("ignore", "use", "gen", "append").contains(x.toLowerCase)) parser.success
+      else parser.failure(s"$x bad value must be one of ignore|use|gen|append")
+    }
+    .text {
+      s"specifies the source info handling, default is ${firrtlOptions.infoMode}"
+    }
+
+  parser.opt[Seq[String]]("in-line")
```
jackkoenig: I don't think "inline" should be hyphenated. It isn't hyphenated on Wikipedia:  https://en.wikipedia.org/wiki/Inline_function

colinschmidt:
Trying to use this PR as the firrtl commit in rocket-chip breaks the memory blackboxing.
I changed the firrtl invocation from
```
$(FIRRTL) -i $< -o $(generated_dir)/$(long_name).v -X verilog --inferRW $(MODEL) --replSeqMem -c:$(MODEL):-o:$(generated_dir)/$(long_name).conf
```
to
```
$(FIRRTL) -i $< -o $(generated_dir)/$(long_name).v -X verilog --infer-rw $(MODEL) --repl-seq-mem -c:$(MODEL):-o:$(generated_dir)/$(long_name).conf
```
But it no longer produces the .conf file. Perhaps some defaults are different now?

PR #338, opened 2016-10-18
Logger 1

chick:
Note: This is built on top of execute-harness-plan-3, do not merge until that is resolved.

#### Implement a log4scala like logging system
This system has the rather remarkable property
that it is possible to turn it on conveniently when
you want it.  It also provides for class level granularity
as well as the traditional Error, Warn, Info, Debug


jackkoenig:
Could you stage this on execute-harness-plan-3 instead of master? It would make the diff much easier to read.

ucbjrl:
lgtm - passes jenkins

azidar:
lgtm

Issue #333, opened 2016-10-05
replseqmem/inferRW generates needless extra modules


colinschmidt:
In Chisel and Firrtl I have a circuit that creates identical SeqMem/smem's in multiple distinct modules and in the resulting verilog I end up with a bunch of modules that contain the memories each named differently.

This is just a nice to have not some high priority issue. In chisel2 this did not happen.

shunshou:
Can you give me a Chirrtl snippet of a circuit that does that and your expected instance/module names? I can optimize that.

PR #341, opened 2016-10-19
fix duplicate memory detection

donggyukim:
Addresses #333: `uniqueueMems` should be constructed in terms of the entire circuit.

@shunshou I might have changed the algorithm with my refactoring. Also `eqMems` changes in accordance with the recent refactoring.

### DIFF
--- a/src/main/scala/firrtl/passes/memlib/ResolveMaskGranularity.scala
+++ b/src/main/scala/firrtl/passes/memlib/ResolveMaskGranularity.scala
@@ -77,7 +77,8 @@ object AnalysisUtils {
 
   /** Checks whether the two memories are equivalent in all respects except name
     */
-  def eqMems(a: DefAnnotatedMemory, b: DefAnnotatedMemory) = a == b.copy(name = a.name) 
+  def eqMems(a: DefAnnotatedMemory, b: DefAnnotatedMemory) =
+    a == b.copy(info = a.info, name = a.name, memRef = a.memRef)
 }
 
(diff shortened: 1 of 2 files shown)


shunshou:
Hm ok... I don't remember any more... info was always a field of DefMemory, right? o_O Strange that I didn't exclude it...... Now info is just source locator, so I definitely don't want to look at it for uniquness... And other added annotations definitely still need to match. 

Not sure why the tests get an UnknownType error... code seems reasonable...

donggyukim:
I fix `eqMems` to pass tests

colinschmidt:
Is this branch ready for a trial run on my codebase?

donggyukim:
@colinschmidt Yeah, let me know if this also works for your case.

jackkoenig:
Note that due to a change to chisel3, all Travis tests are currently failing. I'm trying to fix it.

colinschmidt:
Something weird is happening on this branch, I don't actually get any conf file out. I think that this could be due to several recent PRs(rocket-chip is far behind firrtl/master) of  so I'll try to do a bisect and see what I come up with.

colinschmidt:
Ok so #337 was this issue according to bisect. Once thats resolved I'll test this again.

colinschmidt:
with help from @chick I got a fix for #337 and was able to verify that this branch fixes my issue. Thanks!

shunshou:
Looks good to me pending Travis pass? Sorry, should've caught the ref myself.

PR #344, opened 2016-10-20
add verilator vcd command line option

vegaluisjose:
I have created a pull-request for rocket that support passing VCD filename from command line

### DIFF
--- a/src/test/resources/top.cpp
+++ b/src/test/resources/top.cpp
@@ -2,7 +2,11 @@
 #include <iostream>
 
 #if VM_TRACE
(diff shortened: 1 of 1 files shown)


jackkoenig:
Also, I'm pretty sure rocket-chip doesn't use this file so the reference to Makefrag-verilator doesn't really make sense in this repo

vegaluisjose:
Hi Jack,

Fair enough

jackkoenig:
Thanks for this suggestion though! When I finally move this out of Firrtl I'll be sure to include it!

Issue #345, opened 2016-10-21
FIRRTL generating broken bit-masked sequential memories


ccelio:
Here's my Chisel code:

```scala
val h_table = SeqMem(num_entries, Vec(fetch_width, Bool()))                    
...                                          
                                                                           
   val waddr = hwq.io.deq.bits.index                                           
   val wmask = hwq.io.deq.bits.executed                                        
   val wdata = hwq.io.deq.bits.takens                                          
   h_table.write(waddr, wdata, wmask)                                          
...                                                                      
```

Which turns into perfectly acceptable FIRRTL:
````verilog
smem h_table : UInt<1>[2][32768] @[2bc-table.scala 164:24]                                   
...                                             
  write mport T_99 = h_table[hwq.io.deq.bits.index], hwq.clock                               
  when hwq.io.deq.bits.executed[0] :                                                         
    T_99[0] <= hwq.io.deq.bits.takens[0]                                                     
    skip                                                                                     
  when hwq.io.deq.bits.executed[1] :                                                         
    T_99[1] <= hwq.io.deq.bits.takens[1]                                                     
    skip                                                                                     
  skip @[2bc-table.scala 172:4]                                                              

````

But then gets turned into the broken Verilog code in which it swaps the bit-mask bits and the write-data bits:

````verilog
module h_table(                                            
  input  [14:0] RW0_addr,                                  
  input   RW0_en,                                          
  input   RW0_clk,                                         
  input   RW0_wmode,                                       
  input   RW0_wdata_0,                                     
  input   RW0_wdata_1,                                     
  output  RW0_rdata_0,                                     
  output  RW0_rdata_1,                                     
  input   RW0_wmask_0,                                     
  input   RW0_wmask_1                                      
);                                                         
  wire [14:0] h_table_ext_RW0_addr;                        
  wire  h_table_ext_RW0_en;                                
  wire  h_table_ext_RW0_clk;                               
  wire  h_table_ext_RW0_wmode;                             
  wire [1:0] h_table_ext_RW0_wdata;                        
  wire [1:0] h_table_ext_RW0_rdata;                        
  wire [1:0] h_table_ext_RW0_wmask;                        
  h_table_ext h_table_ext (                                
    .RW0_addr(h_table_ext_RW0_addr),                       
    .RW0_en(h_table_ext_RW0_en),                           
    .RW0_clk(h_table_ext_RW0_clk),                         
    .RW0_wmode(h_table_ext_RW0_wmode),                     
[...]

PR #346, opened 2016-10-22
Fix bitmask

shunshou:
Should solve #345 . I flipped the order that the bitmasks were concatenated when "toBitMask" was used [special case when maskGran = 1]. @ccelio lmk if Boom is ok now. 

@azidar @donggyukim I moved mem files into memlib. Would you agree that VerilogMemDelays.scala also goes there?


colinschmidt:
It seems that some time ago you changed both the mask and data to be reversed, and then @donggyukim  removed one of these reverses in commit b8ee3179ed8070211c95ecbcceda0f7dbf635a13. Are you sure you don't want to solve this issue by reversing the other one so both data and mask are reversed?

shunshou:
You mean going back to what I originally had? 

In the end, it's functionally equivalent. You shouldn't need to care about the order the data is stored in in memory, as long as, @ the memory interface [wrapper IO, not blackbox IO], it corresponds with what you expect. 

I guess, when you have a Vec of UInts, do you prefer to concatenate like:

uint_0 uint_1 uint_2

OR 

uint_2 uint_1 uint_0

---

I think I decided I'm ok with Vec elements being stored index 0 on the left, whereas the LSB of a UInt is on the right. 

Regardless, the endian-ness shouldn't matter to the hardware designer b/c that's all sorted for you internally. They just need to be consistent...

ccelio:
Doesn't the endian-ness matters when looking at the waveforms? I expect element 0 to be the LSB when concatenated. 

This is the current (fixed) output:

```verilog
assign h_table_ext_RW0_wdata = {RW0_wdata_0,RW0_wdata_1};
assign h_table_ext_RW0_wmask = {RW0_wmask_0,RW0_wmask_1};
```

I think I'm going to be confused by this, as I believe it is inconsistent with other treatments of Vecs and UInts.

shunshou:
So you prefer that if your input data is a 2 element Vec of UInts with width 4, the waveform would look like (vec_index, uint_index):

1_3 1_2 1_1 1_0 0_3 0_2 0_1 0_0

Rather than

0_3 0_2 0_1 0_0 1_3 1_2 1_1 1_0

?

---

By treatment of Vecs, are you referring to how something like vec.asUInt would operate? 

Anyways, if you find that the first one is less confusing, I can just change everything to use reverse.

shunshou:
Also, if you take for granted that the waveform @ the SRAM black box will be right (probably bad assumption if the firrtl pass screwed up), it's probably more intuitive for you to just probe the wrapper IO, which just would have separate signals for each Vec element?

ccelio:
> 1_3 1_2 1_1 1_0 0_3 0_2 0_1 0_0
>
> Rather than
>
> 0_3 0_2 0_1 0_0 1_3 1_2 1_1 1_0

Yes, I believe the order should be {a1[3:0], a0[3:0]}, your first suggestion.

I like that bit X can be found by `element_i*w + bit_j`. And if `a1 = 0x3`, and `a0 = 0xc`, then the waveform gives me `0x3c`, and adding more elements just adds more bits on to the front, leaving the LSBs untouched (`0x...3c`).


My apologies if my opinion is clouded by the 1-bit element scenario, because I'm playing with Vec(Bools()) and the internals are switching around between storing them as UInts() and as arrays of 1-bit elements.  It'd make my brain hurt having to switch between `0xc` and `0x3` being the same thing.

Do @colinschmidt and @donggyukim have opinions?

shunshou:
Ok, I guess the Vec of Bools case is particularly annoying -- kind of a side question, then is... Are you planning to implement this w/ SRAM? I think the minimum widths are > 2, so you'd be wasting bits?

ccelio:
> Are you planning to implement this w/ SRAM? I think the minimum widths are > 2, so you'd be wasting bits?

If I understand your question correctly, I'm using SRAMs in this case for my branch predictor. Each bit in the SRAM is a branch prediction.  Conceptually, it is a super tall and very skinny sequential memory.  The idea is something... a FIRRTL pass, a python script?... will translate it transparently into a rectangular SRAM that matches what is available from the technology library.  Worst case, I'll just have to come in later and refactor the code.  Just a "small matter" of bitmasking and indexing...

shunshou:
I see. I didn't think about masking to make skinny memories fatter. Anyways, seems like Colin and you both probably prefer reversed, so I'll make that change...

ccelio:
Yah, I think there's a really cool story here for the Berkeley Chisel/FIRRTL methodology in how we handle memories.  

And thanks for hammering this out!

donggyukim:
lgtm

PR #347, opened 2016-10-24
revert the last commit of #346

donggyukim:
@colinschmidt @ccelio @shunshou 
Turns out that #346 breaks rocket-chip because Chisel assumes Vec(0) is MSB when a mask is given as UInt.(Or rocket-chip's behavioral mem model is broken if it's not). We should fix Chisel if we don't like concat order, or merge this PR.

### DIFF
--- a/src/main/scala/firrtl/passes/memlib/MemUtils.scala
+++ b/src/main/scala/firrtl/passes/memlib/MemUtils.scala
@@ -27,7 +27,7 @@ object toBits {
     case t => error("Invalid operand expression for toBits!")
   }
   private def hiercat(e: Expression): Expression = e.tpe match {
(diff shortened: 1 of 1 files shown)


aswaterman:
@donggyukim I'm confused; how can you give a mask as UInt?  Chisel only takes Seq[Bool].

shunshou:
Regardless, I thought everything above the actual black box should be agnostic to the direction I fed signals into the black box... unless I screwed up the interface???

Or you mean that vlsi_mem_gen flipped the order again when concatenating b/c of some assumption?

donggyukim:
Seems like there's still a bug in `memlib`, so I'm closing this.

PR #351, opened 2016-10-25
Refactor Transforms and Compilers

jackkoenig:
- Transform Ids now handled by Class[_ <: Transform] instead of magic numbers
- Transforms define inputForm and outputForm
- Custom transforms can be inserted at runtime into compiler or the Driver
- Current "built-in" custom transforms handled via above mechanism
- Verilog-specific passes moved to the Verilog emitter


azidar:
Random thought - what if we got rid of TransformId's entirely, and just relied on the actual type of the annotation instead? So each transform matches on annotation type instead of transformid.

jackkoenig:
I've tried similar things before when playing with how to insert additional checks into the compilers. This causes problems because you can't programmatically work with the types except via reflection, and if we can avoid reflection here, we should.

For example, we need to be able to collect all Annotations associated with our custom transform. With TransformId this is straightforward since we can just use a Map from TransformId to annotations. Since Transforms are generally classes (and Passes honestly should be as well), you can't just use the name of the class like you can with a singleton object. This TransformId trait can be mixed in by a Transform's companion object though, and then you can just use that.

azidar:
> For example, we need to be able to collect all Annotations associated with our custom transform. 

Just to push back a little bit - we can still collect all Annotations by iterating through the annotations and matching on their type. We don't necessarily need to write a function that does this for an arbitrary custom transform, correct? We could push this obligation to the transform writer, because the transform writer needs to handle their custom annotations anyways.

I think the transform/inserting checks is a little different, because here it is necessary to write a function that schedules an arbitrary Transform.

jackkoenig:
@azidar @donggyukim 

Alright I rebased this on master, removed TransformId in favor of using the class of a given Transform (essentially a "built-in" id), and added the dedup pass (fixing some issues along the way). 

Please check that my modifications to existing transforms are faithful, and I will also probably code review the dedup pass here (although I already made some style changes haha)

src/main/scala/firrtl/Compiler.scala
```diff
@@ -28,47 +28,147 @@ MODIFICATIONS.
...
 
-// - Transforms a circuit
-// - Can consume multiple CircuitAnnotation's
-trait Transform {
-  def execute(circuit: Circuit, annotationMap: AnnotationMap): TransformResult
+/** Current form of the Firrtl Circuit */
+sealed abstract class CircuitForm(private val value: Int) extends Ordered[CircuitForm] {
+  def compare(that: CircuitForm): Int = this.value - that.value
 }
+final case object ChirrtlForm extends CircuitForm(3)
```
azidar: Explain what the numbers mean in a comment

src/main/scala/firrtl/Compiler.scala
```diff
@@ -28,47 +28,147 @@ MODIFICATIONS.
...
+        case HighForm =>
+          Seq(new IRToWorkingIR, new ResolveAndCheck, new transforms.DedupModules,
+              new HighFirrtlToMiddleFirrtl) ++ getLoweringTransforms(MidForm, outputForm)
+        case MidForm => Seq(new MiddleFirrtlToLowFirrtl) ++ getLoweringTransforms(LowForm, outputForm)
+        case LowForm => error("Internal Error! This shouldn't be possible") // should be caught by if above
+      }
+    }
+  }
+
+  def mergeTransforms(lowering: Seq[Transform], custom: Seq[Transform]): Seq[Transform] = {
```
azidar: Add comment with example to show how the scheduling works.

src/test/scala/firrtlTests/InferReadWriteSpec.scala
```diff
@@ -61,19 +61,19 @@ class InferReadWriteSpec extends SimpleTransformSpec {
     }
   }
 
-  object InferReadWriteCheck extends Transform with SimpleRun {
-    def execute (c: Circuit, map: AnnotationMap) =
-      run(c, Seq(InferReadWriteCheckPass))
+  class InferReadWriteCheck extends PassBasedTransform {
```
azidar: Why is this a class, not an object?

src/test/scala/firrtlTests/fixed/RemoveFixedTypeSpec.scala
```diff
@@ -204,14 +203,15 @@ class RemoveFixedTypeSpec extends FirrtlFlatSpec {
         |    io_out <= io_in
       """.stripMargin
 
-    class CheckChirrtlTransform extends Transform with SimpleRun {
+    case object CheckChirrtlTransformId
```
azidar: Remove

src/main/scala/firrtl/Compiler.scala
```diff
@@ -49,13 +49,38 @@ case class CircuitState(
...
+  *
+  * Thus, [[ChirrtlForm]] has the fewest requirements while [[LowForm]] has the most.
+  */
 sealed abstract class CircuitForm(private val value: Int) extends Ordered[CircuitForm] {
   def compare(that: CircuitForm): Int = this.value - that.value
 }
+/** Chirrtl Form
+  *
+  * The form of the circuit emitted by Chisel. Not a true Firrtl form.
+  * @todo Detailed description
```
azidar: Includes cmem, smem, and mport IR nodes. Enables declaring memories separately from their ports.

src/main/scala/firrtl/Compiler.scala
```diff
@@ -49,13 +49,38 @@ case class CircuitState(
...
+  */
 final case object ChirrtlForm extends CircuitForm(3)
+/** High Form
+  *
+  * @todo Detailed description
```
azidar: As detailed in the firrtl specification.

src/main/scala/firrtl/Compiler.scala
```diff
@@ -49,13 +49,38 @@ case class CircuitState(
...
+  */
 final case object HighForm extends CircuitForm(2)
+/** Middle Form
+  *
+  * @todo Detailed description
```
azidar: All widths have been inferred, when's have been expanded, and all connections have been expanded.

src/main/scala/firrtl/Compiler.scala
```diff
@@ -49,13 +49,38 @@ case class CircuitState(
...
+  */
 final case object MidForm extends CircuitForm(1)
+/** Low Form
+  *
+  * @todo Detailed description
```
azidar: All vector/bundle types have been lowered. All implicit truncation has been made explicit.

jackkoenig:
I added a test and a substantial ScalaDoc to merge transforms. I also added some other documentation. @azidar

jackkoenig:
There seem to be problems with this branch and rocket-chip, don't merge yet.

jackkoenig:
Okay, problem was that VerilogMemDelays inserts some compound expressions, I moved it back to when it was originally run (ie. before SplitExpressions) with a TODO to move it to the verilog emitter. 

Current version confirmed to pass assembly tests with rocket-chip.

Issue #350, opened 2016-10-24
FIRRTL Match Error when building in vsim


zhemao:
I tried compiling the unittests in vsim on the latest rocket-chip master and got an internal FIRRTL exception.

```
java -Xmx2G -Xss8M -XX:MaxPermSize=256M -cp /scratch/zhemao/rocket-chip/firrtl/utils/bin/firrtl.jar firrtl.Driver -i /scratch/zhemao/rocket-chip/vsim/generated-src/unittest.UncoreUnitTestConfig.fir -o /scratch/zhemao/rocket-chip/vsim/generated-src/unittest.UncoreUnitTestConfig.v -X verilog --infer-rw TestHarness --repl-seq-mem -c:TestHarness:-o:/scratch/zhemao/rocket-chip/vsim/generated-src/unittest.UncoreUnitTestConfig.conf
Exception in thread "main" scala.MatchError: UnknownType (of class firrtl.ir.UnknownType$)
    at firrtl.Utils$.create_exps(Utils.scala:104)
    at firrtl.Utils$.create_exps(Utils.scala:95)
    at firrtl.passes.LowerTypes$.lowerTypesStmt(LowerTypes.scala:187)
    at firrtl.passes.LowerTypes$$anonfun$1.apply(LowerTypes.scala:145)
    at firrtl.passes.LowerTypes$$anonfun$1.apply(LowerTypes.scala:145)
    at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
    at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
    at scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:59)
    at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:48)
    at scala.collection.TraversableLike$class.map(TraversableLike.scala:245)
    at scala.collection.AbstractTraversable.map(Traversable.scala:104)
    at firrtl.ir.Block.mapStmt(IR.scala:237)
    at firrtl.Mappers$StmtMagnet$$anon$1.map(Mappers.scala:41)
    at firrtl.Mappers$StmtMap$.map$extension(Mappers.scala:55)
    at firrtl.passes.LowerTypes$.lowerTypesStmt(LowerTypes.scala:145)
    at firrtl.passes.LowerTypes$$anonfun$1.apply(LowerTypes.scala:145)
[...]

PR #352, opened 2016-10-25
Improve reference & name resolution in ReplSeqMem

donggyukim:
This addresses #350, which is caused by bad reference resolutions for the same name memories. Do not merge until I add tests for this PR.

@zhemao This compiles your example, but not sure it's functionally working. Can you run the test with this branch?

### DIFF
--- a/src/main/scala/firrtl/passes/memlib/ReplaceMemMacros.scala
+++ b/src/main/scala/firrtl/passes/memlib/ReplaceMemMacros.scala
@@ -148,7 +148,9 @@ class ReplaceMemMacros(writer: ConfWriter) extends Pass {
     }
   }
 
+  /** Mapping from (module, memory name) pairs to blackbox names */
   private type NameMap = collection.mutable.HashMap[(String, String), String]
+  /** Construct NameMap by assigning unique names for each memory blackbox */
   def constructNameMap(namespace: Namespace, nameMap: NameMap, mname: String)(s: Statement): Statement = {
     s match {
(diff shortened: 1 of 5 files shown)


src/main/scala/firrtl/passes/memlib/ReplaceMemMacros.scala
```diff
@@ -148,7 +148,21 @@ class ReplaceMemMacros(writer: ConfWriter) extends Pass {
     }
   }
 
+  private type NameMap = collection.mutable.HashMap[(String, String), String]
```
azidar: Comment - what are each of those strings?

src/main/scala/firrtl/passes/memlib/ReplaceMemMacros.scala
```diff
@@ -148,7 +148,21 @@ class ReplaceMemMacros(writer: ConfWriter) extends Pass {
...
+  def constructNameMap(namespace: Namespace, nameMap: NameMap, mname: String)(s: Statement): Statement = {
```
azidar: Comment - what does this function do?

zhemao:
I've confirmed that this fixes the issue with running RocketChip unit tests in vsim.

aswaterman:
@donggyukim can you address @azidar's changes so we can pull this into rocket-chip?

shunshou:
How about the tests?

donggyukim:
@shunshou I'll write them soon :)

PR #354, opened 2016-10-28
Keep package name + directory structure consistent

colinschmidt:
This annoyed me so heres a PR


colinschmidt:
I guess this wasn't the only place in firrtl to change it perhaps I'll update this sometime

donggyukim:
@colinschmidt Why did it annoy you? Does it affect the build system?

colinschmidt:
Other passes in that folder I have been referencing by passes.memlib.blah so when I saw this one moved after pulling firrtl I added memlib to its import but then that wasn't correct. Its not a big deal it just seems like an inconsistency.

jackkoenig:
Colin is right in that by convention the package structure should match the
directory structure.

On Fri, Oct 28, 2016 at 5:43 PM Colin Schmidt <notifications@github.com>
wrote:

Other passes in that folder I have been referencing by passes.memlib.blah
so when I saw this one moved after pulling firrtl I added memlib to its
import but then that wasn't correct. Its not a big deal it just seems like
an inconsistency.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/ucb-bar/firrtl/pull/354#issuecomment-257059715>, or mute
the thread

Issue #332, opened 2016-10-04
"shouldn't get here" in RemoveCHIRRTL


aswaterman:
```
circuit Test :
  module Test :
    input clock : Clock

    cmem rf : UInt<64>[31]
    node rf_wdata = mux(UInt(0), UInt(0), UInt(0))
    infer mport port = rf[UInt(0)], clock
    port <- rf_wdata
```

results in

```
Exception in thread "main" firrtl.FIRRTLException: shouldn't be here
at firrtl.Utils$.error(Utils.scala:204)
at firrtl.Utils$.get_valid_points(Utils.scala:241)
at firrtl.passes.RemoveCHIRRTL$.remove_chirrtl_s(RemoveCHIRRTL.scala:229)
at firrtl.passes.RemoveCHIRRTL$$anonfun$remove_chirrtl_s$3.apply(RemoveCHIRRTL.scala:238)
at firrtl.passes.RemoveCHIRRTL$$anonfun$remove_chirrtl_s$3.apply(RemoveCHIRRTL.scala:238)
at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:245)
at scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:59)
at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:48)
at scala.collection.TraversableLike$class.map(TraversableLike.scala:245)
at scala.collection.AbstractTraversable.map(Traversable.scala:104)
at firrtl.ir.Block.mapStmt(IR.scala:227)
at firrtl.Mappers$StmtMagnet$$anon$4.map(Mappers.scala:41)
at firrtl.Mappers$StmtMap$.map$extension(Mappers.scala:55)
at firrtl.passes.RemoveCHIRRTL$.remove_chirrtl_s(RemoveCHIRRTL.scala:238)
at firrtl.passes.RemoveCHIRRTL$$anonfun$remove_chirrtl_m$5.apply(RemoveCHIRRTL.scala:250)
at firrtl.passes.RemoveCHIRRTL$$anonfun$remove_chirrtl_m$5.apply(RemoveCHIRRTL.scala:250)
at firrtl.ir.Module.mapStmt(IR.scala:433)
at firrtl.Mappers$ModuleMagnet$$anon$11.map(Mappers.scala:115)
[...]

PR #355, opened 2016-10-30
Fix CHIRRTL bugs

donggyukim:
Addresses #325, #332, and overrides #326.

It's a bit surprising that it has been working with wrong type inferences for CHIRRTL.


azidar:
The types get over-written later on in the InferTypes pass, so that probably cleared out the incorrectly inferred types.

src/main/scala/firrtl/passes/InferTypes.scala
```diff
@@ -103,8 +103,9 @@ object CInferTypes extends Pass {
         types(sx.name) = sx.tpe
         sx
       case sx: DefNode =>
-        types(sx.name) = sx.value.tpe
-        sx
+        val sxx = (sx map infer_types_e(types)).asInstanceOf[DefNode]
+        types(sxx.name) = sxx.value.tpe
+        sxx
```
jackkoenig: Why does `DefRegister` call `infer_types_e` on it's child `Expressions` but `DefWire` does not?

donggyukim: I have no idea, but `InferTypes` is written in this manner...

jackkoenig: @azidar?

donggyukim: I see, `DefWire` has no expressions.

jackkoenig: Haha whoops, duh

Issue #356, opened 2016-10-31
Reductions are broken


jwright6323:
From ucb-bar/chisel3#345

I noticed specifically through xorR() (it seems andR and orR don't actually use FIRRTL reductions). Right now, the operation just ties the node to the low bit.

John

PR #357, opened 2016-10-31
Fixed Verilog emission of andr, orr, and xorr

azidar:
#356


azidar:
Worth discussing whether we should remove andr and orr entirely from spec.

aswaterman:
I think removing andr and orr makes sense.

Note, there's a much more efficient way to emit these in Verilog: just prefix the name of the signal with the operator, and you've got a reduction.

http://www.asic-world.com/verilog/operators2.html

PR #358, opened 2016-11-01
bug fix for mems

sbeamer:
This removes a spurious connect statement generated by the VerilogMemDelays pass. After the following ConstProp pass, this can turn into an identity assignment (T_880 <= T_880), and this cyclic dependency can cause problems. The current Verilog emitter inadvertently hides this bug since its use of mutable state overwrites this bogus assignment.


donggyukim:
Yes, it shouldn't be there. Thanks, Scott!

PR #361, opened 2016-11-02
make default dir be current directory

chick:
Firrtl will default to looking for and writing to files in the current directory.  
This was done by changing the common options targetDirName to be "."
Other projects, in particular chisel-testers' may override this.


azidar:
Looks like you need to fix your tests as well:

> [error] Failed tests:
> [error] 	firrtlTests.DriverSpec
> [error] 	firrtlTests.ExecutionOptionManagerSpec

PR #364, opened 2016-11-04
Cleanup license at top of every file

jackkoenig:
Replace with more sensible comment to see LICENSE rather than including the
whole license in every file


aswaterman:
yes please

PR #365, opened 2016-11-07
Clock List Transform

azidar:
Given a "top" module in the design, list the driving clock signal for all `input clock: Clock` ports in every child module. Valid sources are also listed, including the top "clock" signal, as well as output clock-typed ports from external modules.

### DIFF
--- a/src/main/scala/firrtl/LoweringCompilers.scala
+++ b/src/main/scala/firrtl/LoweringCompilers.scala
@@ -126,7 +126,22 @@ class LowFirrtlCompiler extends Compiler {
 
 /** Emits Verilog */
 class VerilogCompiler extends Compiler {
+//<<<<<<< HEAD
   def emitter = new VerilogEmitter
   def transforms: Seq[Transform] =
     getLoweringTransforms(ChirrtlForm, LowForm) :+ (new LowFirrtlOptimization)
+//=======
+//  def transforms(writer: Writer): Seq[Transform] = Seq(
+//    new Chisel3ToHighFirrtl,
+//    new IRToWorkingIR,
+//    new ResolveAndCheck,
+//    new HighFirrtlToMiddleFirrtl,
+//    new passes.InferReadWrite(TransID(-1)),
+//    new passes.memlib.ReplSeqMem(TransID(-2)),
+//    new MiddleFirrtlToLowFirrtl,
+//    new passes.clocklist.ClockListTransform(TransID(-3)),
+//    new passes.InlineInstances(TransID(0)),
+//    new EmitVerilogFromLowFirrtl(writer)
+//  )
+//>>>>>>> Added clocklist transform
 }
--- a/src/main/scala/firrtl/LoweringCompilers.scala
+++ b/src/main/scala/firrtl/LoweringCompilers.scala
@@ -126,22 +126,7 @@ class LowFirrtlCompiler extends Compiler {
 
 /** Emits Verilog */
 class VerilogCompiler extends Compiler {
-//<<<<<<< HEAD
   def emitter = new VerilogEmitter
   def transforms: Seq[Transform] =
     getLoweringTransforms(ChirrtlForm, LowForm) :+ (new LowFirrtlOptimization)
-//=======
-//  def transforms(writer: Writer): Seq[Transform] = Seq(
-//    new Chisel3ToHighFirrtl,
-//    new IRToWorkingIR,
-//    new ResolveAndCheck,
-//    new HighFirrtlToMiddleFirrtl,
-//    new passes.InferReadWrite(TransID(-1)),
-//    new passes.memlib.ReplSeqMem(TransID(-2)),
-//    new MiddleFirrtlToLowFirrtl,
-//    new passes.clocklist.ClockListTransform(TransID(-3)),
-//    new passes.InlineInstances(TransID(0)),
-//    new EmitVerilogFromLowFirrtl(writer)
-//  )
-//>>>>>>> Added clocklist transform
 }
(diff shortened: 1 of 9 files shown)


src/main/scala/firrtl/LoweringCompilers.scala
```diff
@@ -126,7 +126,22 @@ class LowFirrtlCompiler extends Compiler {
...
+//    new ResolveAndCheck,
+//    new HighFirrtlToMiddleFirrtl,
+//    new passes.InferReadWrite(TransID(-1)),
+//    new passes.memlib.ReplSeqMem(TransID(-2)),
+//    new MiddleFirrtlToLowFirrtl,
+//    new passes.clocklist.ClockListTransform(TransID(-3)),
+//    new passes.InlineInstances(TransID(0)),
+//    new EmitVerilogFromLowFirrtl(writer)
+//  )
+//>>>>>>> Added clocklist transform
```
jackkoenig: Remove merge stuff

src/main/scala/firrtl/passes/clocklist/ClockList.scala
```diff
@@ -0,0 +1,72 @@
...
+
+    // Remove everything from the circuit, unless it has a clock type
+    // This simplifies the circuit drastically so InlineInstances doesn't take forever.
+    val onlyClockCircuit = RemoveAllButClocks.run(c)
+    
+    // Inline the clock-only circuit up to the specified top module
+    val modulesToInline = (c.modules.collect { case Module(_, n, _, _) if n != top => ModuleName(n, CircuitName(c.main)) }).toSet
+    val inlineTransform = new InlineInstances
+    val inlinedCircuit = inlineTransform.run(onlyClockCircuit, modulesToInline, Set()).circuit
+    val topModule = inlinedCircuit.modules.collectFirst { case m if m.name == top => m }.get
```
jackkoenig: I think this would be better as a find:
```scala
inlinedCircuit.modules.find(_.name == top).get
```
And possibly it should be `.getOrElse(throwInternalError)`

src/main/scala/firrtl/passes/clocklist/ClockListTransform.scala
```diff
@@ -0,0 +1,78 @@
...
+  )
+  passOptions.get(InputConfigFileName) match {
+    case Some(x) => error("Unneeded input config file name!" + usage)
+    case None =>
+  }
+  val target = ModuleName(passModule, CircuitName(passCircuit))
+  def duplicate(n: Named) = n match {
+    case ModuleName(m, CircuitName(c)) => this copy (t = t.replace(s"-c:$passCircuit", s"-c:$c").replace(s"-m:$passModule", s"-m:$m"))
+    case _ => error("Cannot move ClockListAnnotation to a non-module")
+  }
```
jackkoenig: Generally, it makes more sense for this kind of stuff to be located in an apply method in the companion object. Then the annotation itself can be:
```scala
case class ClockListAnnotation(target: ModuleName) extends Annotation with Loose with Unstable {
  def transform = classOf[ClockListTransform]
}
```
Thus, we can construct `ClockListAnnotation`s from the actual `ModuleName` instead of having to do it from the command-line argument String. Perhaps more importantly, we can match and unapply on the information we care about.

src/main/scala/firrtl/passes/clocklist/ClockListTransform.scala
```diff
@@ -0,0 +1,78 @@
...
+        case _ => error(s"Found an unexpected annotation: $cla")
+      }
+    }) match {
+      case Seq((top, out)) => 
+        val outputFile = new PrintWriter(out)
+        val newC = (new ClockList(top, outputFile)).run(state.circuit)
+        outputFile.close()
+        CircuitState(newC, state.form)
+      case _ => error(s"Found too many (or too few) ClockListAnnotations!")
+    }
```
jackkoenig: I find this really hard to follow, especially because there is a match inside of a collect chained with a match inside of a match.

src/main/scala/firrtl/passes/clocklist/ClockListUtils.scala
```diff
@@ -0,0 +1,62 @@
...
+  def getSourceList(moduleMap: Map[String, DefModule])(lin: Lineage): Seq[String] = {
+    val s = lin.foldLeft(Seq[String]()){case (sL, (i, l)) =>
+      val sLx = getSourceList(moduleMap)(l)
+      val sLxx = sLx map (i + "$" + _)
+      sL ++ sLxx
+    }
+    val sourceList = moduleMap(lin.name) match {
+      case ExtModule(i, n, ports, dn, p) =>
+        val portExps = ports.flatMap{p => create_exps(WRef(p.name, p.tpe, PortKind, to_gender(p.direction)))}
+        portExps.filter(e => (e.tpe == ClockType) && (gender(e) == FEMALE)).map(e => e.serialize)
```
jackkoenig: Doesn't really matter but could just be `map (_.serialize)`

azidar: done

src/main/scala/firrtl/passes/clocklist/ClockListTransform.scala
azidar: done

src/main/scala/firrtl/passes/clocklist/ClockListTransform.scala
azidar: done

src/main/scala/firrtl/LoweringCompilers.scala
azidar: done

PR #366, opened 2016-11-07
Fix annotations

azidar:
Fix bug where calling state.copy(circuit = newCircuit) would duplicate annotations.
Also changed getMyAnnotations to return Seq[Annotation]

### DIFF
--- a/src/main/scala/firrtl/Annotations.scala
+++ b/src/main/scala/firrtl/Annotations.scala
@@ -232,21 +232,6 @@ object Annotations {
    * Container of all annotations for a Firrtl compiler.
    */
   case class AnnotationMap(annotations: Seq[Annotation]) {
-    type NamedMap = Map[Named, Map[Class[_], Annotation]]
-    type IDMap = Map[Class[_], Map[Named, Annotation]]
-
-    val (namedMap: NamedMap, idMap:IDMap) =
-      //annotations.foldLeft(Tuple2[NamedMap, IDMap](Map.empty, Map.empty)){
-      annotations.foldLeft((Map.empty: NamedMap, Map.empty: IDMap)){
-        (partialMaps: (NamedMap, IDMap), annotation: Annotation) => {
-          val transformToAnn = partialMaps._1.getOrElse(annotation.target, Map.empty)
-          val pNMap = partialMaps._1 + (annotation.target -> (transformToAnn + (annotation.transform -> annotation)))
-
-          val nToAnn = partialMaps._2.getOrElse(annotation.transform, Map.empty)
-          val ptransformMap = partialMaps._2 + (annotation.transform -> (nToAnn + (annotation.target -> annotation)))
-          Tuple2(pNMap, ptransformMap)
-        }
-      }
     def get(id: Class[_]): Seq[Annotation] = annotations.filter(a => a.transform == id)
     def get(named: Named): Seq[Annotation] = annotations.filter(n => n == named)
   }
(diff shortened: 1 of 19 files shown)


src/main/scala/firrtl/passes/memlib/ReplaceMemMacros.scala
```diff
@@ -219,12 +219,12 @@ class ReplaceMemMacros(writer: ConfWriter) extends Transform {
...
-          case PinAnnotation(c, pin) => pin
-          case _ => error(s"Bad Annotations: ${p.values}")
-        }
-      case None => "pin"
+      case Nil => "pin"
+      case Seq(p) => p match {
+        case PinAnnotation(c, pin) => pin
+        case _ => error(s"Bad Annotation: ${p}")
+      }
+      case _ => error("Shouldn't be here")
```
jackkoenig: throwInternalError?

src/main/scala/firrtl/passes/memlib/ReplaceMemTransform.scala
```diff
@@ -96,8 +96,9 @@ Optional Arguments:
 class SimpleTransform(p: Pass, form: CircuitForm) extends Transform {
   def inputForm = form
   def outputForm = form
-  def execute(state: CircuitState): CircuitState = state.copy(circuit = p.run(state.circuit))
+  def execute(state: CircuitState): CircuitState = state.copy(circuit = p.run(state.circuit), annotations = None)
```
jackkoenig: copy here will still copy the rename map though right? Perhaps it should just be

```scala
def execute(state: CircuitState): CircuitState = CircuitState(p.run(state.circuit), outputForm)
```

src/main/scala/firrtl/passes/wiring/WiringTransform.scala
```diff
@@ -84,6 +85,5 @@ class WiringTransform extends Transform with SimpleRun {
           state.copy(circuit = runPasses(state.circuit, passSeq(winfo)))
         case _ => error("Wrong number of sources, tops, or sinks!")
       }
```
jackkoenig: I believe there is an implicit match so you can just write:
```scala
p.foreach {
  case SinkAnnotation(...
```

src/main/scala/firrtl/Annotations.scala
```diff
@@ -247,8 +247,8 @@ object Annotations {
           Tuple2(pNMap, ptransformMap)
         }
       }
-    def get(id: Class[_]): Option[Map[Named, Annotation]] = idMap.get(id)
-    def get(named: Named): Option[Map[Class[_], Annotation]] = namedMap.get(named)
+    def get(id: Class[_]): Seq[Annotation] = annotations.filter(a => a.transform == id)
```
azidar: Yeah, I think we should cross that bridge when we get there. Another idea is for the transforms overwrite getMyAnnotations to pass a different class.

src/main/scala/firrtl/passes/memlib/ReplaceMemTransform.scala
azidar: The rename map is generally not passed to a new transform, so state.renamemap should always be none. But you are right, we should probably play it safe and just return CircuitState(circuit, form)

src/main/scala/firrtl/passes/wiring/WiringTransform.scala
azidar: done

src/main/scala/firrtl/passes/memlib/ReplaceMemMacros.scala
azidar: done

PR #372, opened 2016-11-10
Bugfix inferwidth

azidar:
This will *probably* lead to more uninferred width errors, but now widths
that were previously incorrectly inferred are now correctly uninferred.

An example is:
  reg r : UInt, clock with: (reset => (reset, UInt<2>(3)))
  node x = add(r, r)
  r <= x

Here, r's width follows the following formula, which cannot be solved:

rWidth >= max(max(rWidth, rWidth) + 1, 2)

### DIFF
--- a/src/main/scala/firrtl/passes/InferWidths.scala
+++ b/src/main/scala/firrtl/passes/InferWidths.scala
@@ -81,7 +81,7 @@ object InferWidths extends Pass {
 
     def remove_cycle(n: String)(w: Width): Width = {
       //;println-all-debug(["Removing cycle for " n " inside " w])
-      w map remove_cycle(n) match {
+      w match {
         case wx: MaxWidth => MaxWidth(wx.args filter {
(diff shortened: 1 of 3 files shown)


src/main/scala/firrtl/passes/InferWidths.scala
```diff
@@ -81,7 +93,7 @@ object InferWidths extends Pass {
 
     def remove_cycle(n: String)(w: Width): Width = {
       //;println-all-debug(["Removing cycle for " n " inside " w])
-      w map remove_cycle(n) match {
```
chick: Is this related to the Issue.  It's hard for me to see how it is.  Maybe a comment on how this works

azidar: I think the entire width inference needs to be better documented, and I think that's a separate PR. I'll submit an issue for it.

azidar:
@chick can you approve?

Issue #371, opened 2016-11-09
Firrtl deduplication pass does not work


colinschmidt:
See the firrtl file at /nscratch/colins/hwacha-broken-dedupe.fir

Arbiter_23 and Arbiter_15

as well as many other things have no difference in their body but have different names. This chains up the hierarchy to break de-duplication throughout the design.

jackkoenig:
Closed via #375

PR #375, opened 2016-11-15
Bugfix: Dedup aggressively (ignore fileinfo)

azidar:
FileInfo is merged
Addresses #371


colinschmidt:
This fixes my issue. Thanks!

colinschmidt:
@jackkoenig Now that the chisel removal is merged this really should be merged ASAP to provide feature parity

jackkoenig:
I agree, looking at it

jackkoenig:
This PR seems to have an exponential effect on runtime. With the following code connecting `n` identical queues together:

```scala
class QueueTop(n: Int) extends Module {
  val io = IO(new Bundle {
    val in = Input(Decoupled(UInt(64.W)))
    val out = Output(Decoupled(UInt(64.W)))
  })

  var ref = io.in
  for (i <- 0 until n) {
    ref = Queue(ref, 8)
  }
  //io.out <> ref // the fact that this bulk connect isn't working is also concerning
  io.out.bits := ref.bits
  io.out.valid := ref.valid
  ref.ready := io.out.ready
}
```

I got the following performance results (for chisel3 + Firrtl):

On firrtl master (9a967a27aa8bb51f4b62969d2889f9a9caa48e31):
```
For param 1: 43.6 ms
For param 2: 52.8 ms
For param 4: 83.3 ms
For param 8: 122.2 ms
For param 16: 174.2 ms
For param 32: 244.8 ms
For param 64: 417.3 ms
For param 128: 814.4 ms
```
Note that the resulting Verilog is *not* deduped (presumably because of the source locators which this PR is fixing).

On this branch (ab8440c045ef619db45f64e49263cbb08eea14a0):
```
For param 1: 51.0 ms
For param 2: 48.6 ms
For param 4: 80.1 ms
For param 8: 121.9 ms
For param 16: 379.8 ms
For param 32: 2443.0 ms
For param 64: 18830.0 ms
For param 128: (it was taking too long so I killed it)
```
Note at the end it's growth of 6.6x despite 2x the increase in the amount of work to do.
I will also point out that it does at least actually dedup.

aswaterman:
Growing 6.6x per 2x isn't exponential, but is curious in its own right.  The runtime is polynomial, so say T(n)=x^p.  Then we have that T(2n)/T(n)=6.6, so 2^p=6.6, so p=2.72... which is _e_!

colinschmidt:
Is this PR blocked on the performance regression? If so I think its worth noting that as a user with a large design I did not notice any slow down when using this branch, but I leave the decision up to you.

jackkoenig:
Yeah considering it isn't a problem for rocket chip or hwacha in practice I
think it's fine to merge. I made the PR on chisel3 to help us track the
problem so it can ultimately get fixed. We should add it to the Firrtl task
list too.

jackkoenig:
Per previous comment, I'm going to approve this but we should make sure to track the performance issue

PR #377, opened 2016-11-15
Rewrote inline xform to fix quadratic perf. bug

azidar:
Turns out caching previously inlined modules is not useful

The previous algorithm in a module, would flatten an instance's children, then
flatten that instance. This caused all instances to be effectively
inlined the number of times of its depth in the instance hierarchy,
making it O(n*d*s), where n is the number of instances, and d is the depth
of the instance, and s is the number of statements in the instance.

The new algorithm directly inlines a module by keeping track of the
parents of that instance, making it constant time with the number of
instantiated instances.

### DIFF
--- a/src/main/scala/firrtl/passes/Inline.scala
+++ b/src/main/scala/firrtl/passes/Inline.scala
@@ -28,7 +28,10 @@ class InlineInstances extends Transform {
    private def collectAnns(circuit: Circuit, anns: Iterable[Annotation]): (Set[ModuleName], Set[ComponentName]) =
      anns.foldLeft(Set.empty[ModuleName], Set.empty[ComponentName]) {
        case ((modNames, instNames), ann) => ann match {
-         case InlineAnnotation(CircuitName(c)) => (circuit.modules.collect { case Module(_, name, _, _) if name != circuit.main => ModuleName(name, CircuitName(c)) }.toSet , instNames)
+         case InlineAnnotation(CircuitName(c)) =>
+           (circuit.modules.collect {
+             case Module(_, name, _, _) if name != circuit.main => ModuleName(name, CircuitName(c))
+           }.toSet, instNames)
          case InlineAnnotation(ModuleName(mod, cir)) => (modNames + ModuleName(mod, cir), instNames)
          case InlineAnnotation(ComponentName(com, mod)) => (modNames, instNames + ComponentName(com, mod))
          case _ => throw new PassException("Annotation must be InlineAnnotation")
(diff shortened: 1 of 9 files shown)


azidar:
Inlining all of hwacha now takes 3 seconds, instead of not terminating after an hour. However, emitting the fully inlined hwacha requires a LOT of heap space.

jackkoenig:
We can use the fully inlined versions of things for things like combinational loop detection even if we don't end up emitting it.

src/main/scala/firrtl/passes/Inline.scala
```diff
@@ -25,9 +25,10 @@ class InlineInstances extends Transform {
    val inlineDelim = "$"
    override def name = "Inline Instances"
 
-   private def collectAnns(anns: Iterable[Annotation]): (Set[ModuleName], Set[ComponentName]) =
+   private def collectAnns(circuit: Circuit, anns: Iterable[Annotation]): (Set[ModuleName], Set[ComponentName]) =
      anns.foldLeft(Set.empty[ModuleName], Set.empty[ComponentName]) {
        case ((modNames, instNames), ann) => ann match {
+         case InlineAnnotation(CircuitName(c)) => (circuit.modules.collect { case Module(_, name, _, _) if name != circuit.main => ModuleName(name, CircuitName(c)) }.toSet , instNames)
```
chick: Style guide complains about long lines like this

src/main/scala/firrtl/passes/Inline.scala
```diff
@@ -84,99 +85,60 @@ class InlineInstances extends Transform {
...
-              inlinedModules(name) = mx
-              mx
-            case mx: ExtModule =>
-              inlinedModules(mx.name) = mx
-              mx
-         }
-      }
+    def appendNamePrefix(prefix: String)(name:String): String = prefix + name
+    def appendRefPrefix(prefix: String, currentModule: String)(e: Expression): Expression = e match {
+      case WSubField(WRef(ref, _, InstanceKind, _), field, tpe, gen) if (flatInstances.contains(currentModule + "." + ref)) =>
```
chick: Style guide says you have unnecessary parens here

src/main/scala/firrtl/passes/Inline.scala
azidar: done

src/main/scala/firrtl/passes/Inline.scala
azidar: done

PR #379, opened 2016-11-18
Bugfix: exponential runtime of pull muxes

azidar:


src/main/scala/firrtl/passes/Passes.scala
```diff
@@ -54,35 +54,32 @@ object ToWorkingIR extends Pass {
...
+     def pull_muxes_e(e: Expression): Expression = e map pull_muxes_e match {
+       case ex: WSubField => ex.exp match {
+         case exx: Mux => Mux(exx.cond,
+           WSubField(exx.tval, ex.name, ex.tpe, ex.gender),
+           WSubField(exx.fval, ex.name, ex.tpe, ex.gender), ex.tpe)
+         case exx: ValidIf => ValidIf(exx.cond,
+           WSubField(exx.value, ex.name, ex.tpe, ex.gender), ex.tpe)
+         case _ => ex  // case exx => exx causes failed tests
        }
-       exxx map pull_muxes_e
```
donggyukim: It's not clear why we needed this before ;)

PR #385, opened 2016-11-22
Bugfix: pad is max(width,const) during constprop

azidar:


aswaterman:
Ah, right.

Prefer a max b, rather than math.max(a, b).

src/main/scala/firrtl/passes/ConstProp.scala
```diff
@@ -198,8 +198,8 @@ object ConstProp extends Pass {
       case _ => e
     }
     case Pad => e.args.head match {
-      case UIntLiteral(v, _) => UIntLiteral(v, IntWidth(e.consts.head))
-      case SIntLiteral(v, _) => SIntLiteral(v, IntWidth(e.consts.head))
+      case UIntLiteral(v, IntWidth(w)) => UIntLiteral(v, IntWidth(max(e.consts.head, w)))
+      case SIntLiteral(v, IntWidth(w)) => SIntLiteral(v, IntWidth(max(e.consts.head, w)))
```
jackkoenig: change to `e.consts.head max w`

azidar:
Lol not sure what happened with the commit history, but the last commit is the update. @jackkoenig, please review.

jackkoenig:
This looks like the commit history got all messed up. @azidar can you fix it?

azidar:
subsumed by #465

PR #387, opened 2016-11-23
Clk2clock - rename the implicit "clk" module input "clock"

ucbjrl:
This applies the chisel naming scheme to firrtl tests.
**NOTE** - This is implemented as two commits: rename "clk" to "clock" in:
* `top.cpp` and tests that use `.fir` resource files,
* "self-contained" tests that incorporate firrtl source as Strings.

In both cases, memory clock enables are left as "clk", and `.fir` files in the `regress` directory are not changed.

There should be no functional firrtl changes.


jackkoenig:
Can this PR be squashed and merged or should the two commits be preserved?

ucbjrl:
They could be squashed and merged. I wan't to have the option to do the first (externally visible uses of "clk"), and I wasn't sure if the second should be augment to deal with memory "clk"s.

PR #390, opened 2016-11-30
Fixes for Annotation serialized/deserialize

chick:
Made serializer agree with deserializer on text representation
This is part of providing annotation support for chisel3
Re-ordered serializations of Named subclasses to be C or C.m or C.m.c where C=circuit, m=module, c=component
Note: component may contain dots
Added serialize deserialize tests to AnnotationSpec
Did some style cleanup on AnnotationSpec
Added explicit return type on SimpleTransformSpec#execute


src/main/scala/firrtl/annotations/AnnotationYamlProtocol.scala
```diff
@@ -8,30 +8,28 @@ import net.jcazevedo.moultingyaml._
...
+          case _ => deserializationError("Color expected")
       }
     }
-    def toTarget(string: String) = string.split('.').toSeq match {
+    def toTarget(string: String): Named = string.split('.').toSeq match {
       case Seq(c) => CircuitName(c)
       case Seq(c, m) => ModuleName(m, CircuitName(c))
-      case Nil => error("BAD")
-      case s => ComponentName(s.drop(2).mkString("."), ModuleName(s(1), CircuitName(s(0))))
+      case Nil => sys.error("BAD")
```
jackkoenig: I think this is intending to refer to Utils.error rather than sys.error.

src/test/scala/firrtlTests/AnnotationTests.scala
```diff
@@ -74,4 +57,43 @@ class AnnotationTests extends AnnotationSpec with Matchers {
...
+
+    annos.zip(readAnnos).foreach { case (beforeAnno, afterAnno) =>
+      beforeAnno.targetString should be (afterAnno.targetString)
+      beforeAnno.target should be (afterAnno.target)
+      beforeAnno.transformClass should be (afterAnno.transformClass)
+      beforeAnno.transform should be (afterAnno.transform)
+
+      beforeAnno should be (afterAnno)
+    }
+  }
```
jackkoenig: I would like to see a negative test of some sort. In particular, I'm curious what the error message is if someone annotates something like `circuit.module..` or `.circuit.module.component`

src/main/scala/firrtl/annotations/Named.scala
```diff
@@ -21,11 +21,12 @@ final case class CircuitName(name: String) extends Named {
 
 final case class ModuleName(name: String, circuit: CircuitName) extends Named {
   if(!validModuleName(name)) throw AnnotationException(s"Illegal module name: $name")
-  def serialize: String = name + "." + circuit.serialize
+//  def serialize: String = name + "." + circuit.serialize
```
jackkoenig: Probably no reason to keep around the commented out part

src/main/scala/firrtl/annotations/AnnotationYamlProtocol.scala
chick: Will change to do that, but as written it was using Predef.error which is deprecated

src/test/scala/firrtlTests/AnnotationTests.scala
chick: @jackkoenig not sure how that would happen unless someone hand crafts something with dot's in it, because otherwise these names are generated in chisel, but if this is a concern the constructors for the Named family should probably test for this.

PR #394, opened 2016-12-05
Fix validation print for log-level

colinschmidt:


colinschmidt:
Feel free to cherry pick this commit onto some other clean-up branch if you want. Just stumbled across this and thought I should log it somehow.

PR #396, opened 2016-12-06
Bugfix: apply/unapply of PinAnnotation broken

azidar:


colinschmidt:
This needs validation that it still works? I remember some issue with one of these annotations a while ago, but I can't find it right now, so I'll just test this.

PR #398, opened 2016-12-08
"Scope" test resource (top.cpp).

ucbjrl:
Jar resources (unlike classes) are typically not scoped. This can create collisions if we have similarly named resources in multiple jars, especially when merging multiple projects in an IDE. Give this resource a distinct name to avoid colliding with chisel3 top.cpp (and shouldn't that resource move to chisel-testers?).


ucbjrl:
ping

PR #402, opened 2016-12-12
Zero width

azidar:
Semantics:
 - No change to width inference rules, e.g.
     a<0> + b<2> = c<3>
 - Replace zero width wires with UInt<1>(0) or SInt<1>(0)
 - Performs constant prop.
 - Redo width/type inference


aswaterman:
Cool!  I don't think I'll have time to try it out soon, but I'm glad to see it's moving forward.

src/test/scala/firrtlTests/ZeroWidthTests.scala
```diff
@@ -0,0 +1,98 @@
...
+         |    x <= y""".stripMargin
+    val check = 
+      """module Top(
+        |  output  [2:0] x
+        |);
+        |  assign x = 3'h0;
+        |endmodule
+        |""".stripMargin.split("\n") map normalized
+    executeTest(input, check, compiler)
+  }
```
jackkoenig: Do we really want to remove ports?

aswaterman: Removing them is the right call, IMO.  Including the ports and artificially making them 1-bit wide is even weirder.

jackkoenig: Yeah, upon further thought I agree

src/main/scala/firrtl/passes/ZeroWidth.scala
```diff
@@ -0,0 +1,65 @@
...
+      case (Port(info, name, dir, _), Some(t)) => Port(info, name, dir, t)
+    }
+    m match {
+      case ext: ExtModule => ext.copy(ports = ports)
+      case in: Module => in.copy(ports = ports, body = onStmt(in.body))
+    }
+  }
+  def run(c: Circuit): Circuit = {
+    val cx = c.copy(modules = c.modules map onModule)
+    ConstProp.run(InferTypes.run(cx))
```
jackkoenig: Does `ConstProp` work on `MiddleFirrtl`? Also do you think running these extra passes should be organized more in the "SimpleRun" style? This could instead be organized as a transform, or they could be added after this pass in the list of passes in `HighFirrtlToMiddleFirrtl`. Perhaps this is best, but it make looking at the runtime kind of confusing since what's actually happening is 3 passes being run but this will be printed as just 1 pass.

src/main/scala/firrtl/passes/ZeroWidth.scala
```diff
@@ -0,0 +1,65 @@
...
+      fields.zip(fields.map(f => removeZero(f.tpe))) collect {
+        case (Field(name, flip, _), Some(t)) => Field(name, flip, t)
+      } match {
+        case Nil => None
+        case seq => Some(BundleType(seq))
+      }
+    case VectorType(t, size) => removeZero(t) match {
+      case None => None
+      case Some(tx) => Some(VectorType(tx, size))
+    }
```
jackkoenig: This is what map is for on Options:
```scala
case VectorType(t, size) => removeZero(t) map (VectorType(_, size))
```

src/main/scala/firrtl/passes/ZeroWidth.scala
```diff
@@ -0,0 +1,65 @@
...
+import firrtl.Utils.throwInternalError
+
+
+object ZeroWidth extends Pass {
+  def name = this.getClass.getName
+  private val ZERO = BigInt(0)
+  private def removeZero(t: Type): Option[Type] = t match {
+    case GroundType(IntWidth(ZERO)) => None
+    case BundleType(fields) =>
+      fields.zip(fields.map(f => removeZero(f.tpe))) collect {
```
jackkoenig: I think this is better expressed as
```scala
fields map (f => (f, removeZero(f.tpe)) collect { ...
```

src/test/scala/firrtlTests/ZeroWidthTests.scala
```diff
@@ -0,0 +1,98 @@
...
+        |    output x : UInt
+        |    x <= mux(y, UInt<2>(2), UInt<2>(1))""".stripMargin
+    val check =
+      """circuit Top :
+        |  module Top :
+        |    output x : UInt<2>
+        |    x <= UInt<2>(1)""".stripMargin
+      (parse(exec(input)).serialize) should be (parse(check).serialize)
+  }
+}
```
jackkoenig: I'd like to see a test where a zero width wire is in a Bundle and one with it in a Vec

src/main/scala/firrtl/passes/ZeroWidth.scala
```diff
@@ -0,0 +1,65 @@
...
+      def applyRemoveZero(t: Type): Type = removeZero(t) match {
+        case None => removed = true; t
+        case Some(tx) => tx
+      }
+      val sxx = sx map applyRemoveZero
+      if(removed) EmptyStmt else sxx
+    case sx => sx map onStmt
+  }
+  private def onModule(m: DefModule): DefModule = {
+    val ports = m.ports.zip(m.ports.map(p => removeZero(p.tpe))) collect {
```
jackkoenig: Similar to above, would be better as map to tuple instead of zip of map

src/main/scala/firrtl/passes/ZeroWidth.scala
azidar: Yeah, I looked at the code and `ConstProp` should work on `MiddleFirrtl`. However, I got rid of it because you shouldn't _have_ to run `ConstProp` for it to work. However, you do need to run InferTypes, so I am running it at the end of the pass.

PR #405, opened 2016-12-14
Delete sample annotations file at the end of test

leonardt:
Minor change cleans up a sample file at the end of the test case.


lenny-t:
;) some interesting stuff in the works on this side of the bay. FIRRTL is a candidate backend for a new project, more to come as things become concrete.

PR #407, opened 2016-12-20
Fix stack overflow from massive MaxWidth chains during width inference

jackkoenig:
Fixes ucb-bar/chisel3#420

Also added a test, would be better as a Chisel test


jackkoenig:
Removed huge file, will add test to Chisel instead

PR #408, opened 2017-01-05
Fix ScalaDoc complaints; add sbt-site, sbt-ghpages boilerplate.

ucbjrl:


ucbjrl:
Why is travis killing these builds? It would be nice if it said something more than:

```
/home/travis/build.sh: line 57:  3022 Killed                  sbt clean test

travis_time:end:094683e4:start=1484931337479114285,finish=1484931789166992676,duration=451687878391
The command "sbt clean test" exited with 137.
```

Running it a second time seems to succeed.
Is there a timeout we need to bump?

colinschmidt:
Its worth noting that as an external consumer of firrtl ucb-bar/pfpmp required changes to compile with this change. see https://github.com/ucb-bar/pfpmp/commit/2fb5fe08bec0b1eb9a492d17722bcc80d84fc8b7

PR #410, opened 2017-01-13
Emit blackboxes inline

chick:
Do not merge!
This is an example implementation of emitting black boxes inline
Please check it out and comment

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -36,7 +36,7 @@ case class VRandom(width: BigInt) extends Expression {
   def mapWidth(f: Width => Width): Expression = this
 }
 
-class VerilogEmitter extends Emitter with PassBased {
+class VerilogEmitter(verilogBlackBoxes: Map[String, String] = Map.empty) extends Emitter with PassBased {
   val tab = "  "
   private val blackBoxesEmitted = new mutable.HashSet[String]
 
@@ -566,26 +566,15 @@ class VerilogEmitter extends Emitter with PassBased {
   def emitBlackBoxInline(blackBoxName: String)(implicit w: Writer): Unit = {
     if(! blackBoxesEmitted.contains(blackBoxName)) {
       blackBoxesEmitted += blackBoxName
-      w.write(
-        """
-        |
-        |module BBFAdd(
-        |    input  [63:0] in1,
-        |    input  [63:0] in2,
-        |    output reg [63:0] out
-        |);
-        |  always @* begin
-        |  out <= $realtobits($bitstoreal(in1) + $bitstoreal(in2));
-        |  end
-        |endmodule
-        |
-      """.stripMargin
-      )
+      if(verilogBlackBoxes.contains(blackBoxName)) {
+        w.write("\n")
+        w.write(verilogBlackBoxes(blackBoxName)
+        )
+      }
     }
   }
 
-  def
-  passSeq = Seq(
+  def passSeq = Seq(
     passes.VerilogWrap,
     passes.VerilogRename,
     passes.VerilogPrep)
@@ -596,6 +585,10 @@ class VerilogEmitter extends Emitter with PassBased {
     val moduleMap = (circuit.modules map (m => m.name -> m)).toMap
     circuit.modules foreach {
       case (m: Module) => emit_verilog(m, moduleMap)(writer)
+      case (m: ExtModule) =>
+    }
+    circuit.modules foreach {
+      case (m: Module) =>
       case (m: ExtModule) => emitBlackBoxInline(m.defname)(writer)
     }
   }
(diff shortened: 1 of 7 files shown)


shunshou:
I think you should pass in a config file that contains a list of files rather than a big command line or other argument of files...

sdtwigg:
Can you explain what the higher level intent is?

grebe:
I think this looks pretty nifty and a lot like what I had in mind. Some thoughts:

1) Ideally you'd be able to pass in something like a StringBuffer so you can easily include verilog snippets from the resources folder
2) I think the firrtl folks would prefer s/blackbox/extmodule
3) I'm a little leery about parsing verilog files, even for something as simple as extracting a module. Maybe we should enforce 1 module per file

shunshou:
The way I thought of this: Pass in a a file (or files) of Verilog modules (1 module per file) that you want sucked into Chisel/FIRRTL generated stuff. If Verilog module names matched any of the module names inside the Chisel-generated stuff, swap out Chisel module for ExtModule of the same name + copy paste Verilog content into big Chisel-generated Verilog file. 

So say you wanted Black boxed float ops vs. normal fixed point stuff: You either include a file with a list of BBFloats or don't include it.

grebe:
@sdtwigg we've been using blackboxes to wrap operations around verilog real (this is for dsp, being able to switch between floating point and fixed point is useful). We have a mechanism to pass blackbox implementations to the firrtl interpreter automagically, but the current way we do this when using verilator is to manually copy/paste a verilog file (or do so in a Makefile). It works, but is somewhat fragile. Ideally the chisel+firrtl compiler would just emit the floating point wrappers with the rest of the circuit.

jackkoenig:
One of the most requested and highest priority feature requests we have in Firrtl is to emit multiple Verilog files (in particular, one per module). While these are not mutually exclusive ideas, if we're simply trying to reduce the number of Verilog files here I'm not sure if there's much of a point. As @grebe has suggested, it seems that this is simply to help manage additional Verilog files generated by a Chisel library. This pattern is going to become more common as people develop more libraries. 
It seems strange to me that the solution to this is to modify the compiler itself rather than the build process.

Perhaps, rather than trying to inline all of our blackboxes we instead just provide a more organized way of keeping track of additional files for compilation. The compile options could essentially create a filelist that libraries have some way of appending to that the build process can then query at the end in order to generate a compilation command (or, copy/paste as @grebe mentioned).

colinschmidt:
I agree with @jackkoenig. My use of black boxes has always been dealt with by changing the build system. I'm not sure how much time savings will be accomplished by simply moving the blackbox verilog integration to before the chisel/firrtl build process rather than after. You still need to point chisel/firrtl to your source files just as you need to point verilator to your source files. Perhaps I'm missing some killer feature of this integration though,  I'm struggling to think outside the (black)box.

grebe:
I think I agree with what both @jackkoenig and @colinschmidt said. I just want to emphasize that the reason it makes sense to modify firrtl is that the code using blackboxes is in a library. We have all sorts of build systems using the library- chisel-testers using firrtl-interpreter or verilator as a backend, chisel's BasicTesters, and the rocket build system for example. As an author of the library I understand what's going on and I can deal with the annoyance of making it work with whatever build system I run into.

The problem is that a library user won't understand what is going on- they'll just be upset that it isn't working. This is especially problematic for our particular use case- floating point is meant to be a feature that allows you to quickly get started on your dsp circuit without sweating the details. If the "quick start" involves reading four paragraphs of chisel+firrtl explanation and having to figure out how the build system works, that's a problem.

Unless you are using the firrtl-interpreter (which has the ability for a library to pass scala models of blackboxes), it seems to me that every build system has to go through firrtl, so it seemed natural to modify the compiler to allow libraries to add black box implementations. One way to do this is by inlining verilog during verilog emission. Rather than inlining, you could just emit separate files at some point. I think this is what Jack's second paragraph was talking about. Regardless, I think it needs to happen in firrtl.

aswaterman:
Definitely should use the build system for this (& separately firrtl should emit one module per file).

chick:
I agree with @grebe that one of the key motivations of this approach is to make a way that the implementors of a library (like dsptools) can ensure that black boxes that underly the implementation can be transparently placed where they are accessible to the overall toolchain

chick:
This implementation is just a simple API, chisel3, testers, dsptools etc can overly this api with their own.  In this implementation directories or files containing verilog can be passed down to firrtl but equally importantly verilog blocks can just be passed in through the map contained in the FirrtlOptions

colinschmidt:
I feel that if you are integrating someone else's hardware library into your design you will already be in a position to pass multiple files into verilator. I can imagine it being useful for your library to explicitly tell you which files you need to include. This could be done in the library itself and for example generate a file that contains a list of these full paths. It seems to me that adding this kind of interface will make it harder to use firrtl as a general tool. How do passes handle the verilog?

shunshou:
Ok, maybe rather than using FIRRTL, can you just include as a TesterOption a string that points to a file with the list of Verilog BB's that somehow get passed to the Verilator backend for idk what kind of command is used to work with Verilator (probably also copied into test_run_dir)? [which is wholly independent of FIRRTL and more of a tester option]. BB Reals aside, this would be useful for testing other BB's with Verilator.

chick:
Replaced by different implementation

PR #412, opened 2017-01-18
Speed up deduplication. Fixes #391

### LINKED ISSUE #391: Deduplication Performance Bug
From #375 

Deduplication has polynomial time complexity

This PR seems to have an exponential effect on runtime. With the following code connecting `n` identical queues together:

```scala
class QueueTop(n: Int) extends Module {
  val io = IO(new Bundle {
    val in = Flipped(Decoupled(UInt(64.W)))
    val out = Decoupled(UInt(64.W))
  })

  var ref = io.in
  for (i <- 0 until n) {
    ref = Queue(ref, 8)
  }
  io.out <> ref
}
```

I got the following performance results (for chisel3 + Firrtl):

On firrtl master (9a967a27aa8bb51f4b62969d2889f9a9caa48e31):
```
For param 1: 43.6 ms
For param 2: 52.8 ms
For param 4: 83.3 ms
For param 8: 122.2 ms
For param 16: 174.2 ms
For param 32: 244.8 ms
For param 64: 417.3 ms
For param 128: 814.4 ms
```
Note that the resulting Verilog is *not* deduped (presumably because of the source locators which this PR is fixing).

On this branch (ab8440c045ef619db45f64e49263cbb08eea14a0):
```
For param 1: 51.0 ms
For param 2: 48.6 ms
For param 4: 80.1 ms
For param 8: 121.9 ms
For param 16: 379.8 ms
For param 32: 2443.0 ms
For param 64: 18830.0 ms
For param 128: (it was taking too long so I killed it)
```

Note at the end it's growth of 6.6x despite 2x the increase in the amount of work to do.
I will also point out that it does at least actually dedup.

Fixing this will allow https://github.com/ucb-bar/chisel3/pull/383 to be merged

jackkoenig:
Removes combining of source locators since the benefit is dubious.
Adds MultiInfo for faster appending of source locators in future.

I ended up removing the combining of source locators even though I had fixed the performance pathology because it's not clear to me if it's even beneficial. I looked at the results of deduplicating 64 identical Queues and all of the concatinations made the results... not pretty. Anyway we can always add it back in a sensible way in the future if we want to.

### DIFF
--- a/src/main/scala/firrtl/ir/IR.scala
+++ b/src/main/scala/firrtl/ir/IR.scala
@@ -21,10 +21,17 @@ case object NoInfo extends Info {
 }
 case class FileInfo(info: StringLit) extends Info {
   override def toString: String = " @[" + info.serialize + "]"
-  def ++(that: Info): Info = that match {
-    case NoInfo => this
-    case FileInfo(otherInfo) => FileInfo(FIRRTLStringLitHandler.unescape(info.serialize + " " + otherInfo.serialize))
+  def ++(that: Info): Info = MultiInfo(Seq(this, that))
+}
+case class MultiInfo(infos: Seq[Info]) extends Info {
+  private def collectStringLits(info: Info): Seq[StringLit] = info match {
+    case FileInfo(lit) => Seq(lit)
+    case MultiInfo(seq) => seq flatMap collectStringLits
+    case NoInfo => Seq.empty
   }
+  override def toString: String =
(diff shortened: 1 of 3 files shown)


colinschmidt:
Checking if this solves my use case.

colinschmidt:
Yup this works. There is an unrelated issue I had to fix to get my use case working that I will submit an issue/PR for.

Thanks for fixing this.

jackkoenig:
Done. Also rebased on master

PR #415, opened 2017-01-20
Rework Attach to work on arbitrary Analog hierarchies

jackkoenig:
Now you can just attach Analogs without any concept of "source". It's just representing an electrical connection.

### DIFF
--- a/src/main/scala/firrtl/passes/ExpandWhens.scala
+++ b/src/main/scala/firrtl/passes/ExpandWhens.scala
@@ -10,21 +10,27 @@ import firrtl.PrimOps._
 import firrtl.WrappedExpression._
 
 import annotation.tailrec
+import collection.mutable
+import collection.immutable.ListSet
 
 /** Expand Whens
 *
-* @note This pass does three things: remove last connect semantics,
-* remove conditional blocks, and eliminate concept of scoping.
+* This pass does the following things:
+* $ - Remove last connect semantics
+* $ - Remove conditional blocks
+* $ - Eliminate concept of scoping
+* $ - Consolidate attaches
+*
 * @note Assumes bulk connects and isInvalids have been expanded
 * @note Assumes all references are declared
 */
 object ExpandWhens extends Pass {
   def name = "Expand Whens"
-  type NodeMap = collection.mutable.HashMap[MemoizedHash[Expression], String]
-  type Netlist = collection.mutable.LinkedHashMap[WrappedExpression, Expression]
-  type Simlist = collection.mutable.ArrayBuffer[Statement]
-  type Attachlist = collection.mutable.ArrayBuffer[Statement]
-  type Defaults = Seq[collection.mutable.Map[WrappedExpression, Expression]]
+  type NodeMap = mutable.HashMap[MemoizedHash[Expression], String]
+  type Netlist = mutable.LinkedHashMap[WrappedExpression, Expression]
+  type Simlist = mutable.ArrayBuffer[Statement]
+  // Defaults ideally would be immutable.Map but conversion from mutable.LinkedHashMap to mutable.Map is VERY slow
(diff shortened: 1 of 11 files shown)


jackkoenig:
Unfortunately it appears that alias doesn't work in design compiler. Looking into alternatives....

jackkoenig:
Now with improvements!

* Verified that generated code is synthesizable (test case lives in Chisel, will push to that branch soon)
* Found that Verilator really doesn't support the general case
    * emit ``` `ifdef verilator `error ``` in those cases since the generated Verilog is still synthesizable and works with alias 
    * Will bump [Verilator Issue](https://www.veripool.org/boards/3/topics/1041-Verilator-Implementing-SystemVerilog-alias-) so that this restriction can be lifted
* Added code to infer zero or single source attaches so that simpler Verilog can be emitted
    * These cases are supported by Verilator (so no ``` `error``` emitted)
    * This should work for the majority of use cases (notable excluding a module that tries to attach two inout ports)
    * Useful improvement would be to dead code eliminate / constant propagate extra wires on an attach so that they don't prevent single source inference

src/main/scala/firrtl/passes/ExpandWhens.scala
```diff
@@ -45,6 +52,8 @@ object ExpandWhens extends Pass {
       case (k, WInvalid) => IsInvalid(NoInfo, k.e1)
       case (k, v) => Connect(NoInfo, k.e1, v)
     }
+  private def expandAttachList(attachList: Attachlist): Seq[Statement] =
+    attachList.reverse map (set => Attach(NoInfo, set.toList))
```
azidar: Why reverse?

src/main/scala/firrtl/passes/ExpandWhens.scala
```diff
@@ -90,7 +99,13 @@ object ExpandWhens extends Pass {
         case c: IsInvalid =>
           netlist(c.expr) = WInvalid
           EmptyStmt
-        case c: Attach => c
+        case Attach(_, exprs) =>
```
azidar: What are the semantics for:

Example 1:
```
attach(a, b)
when p:
  attach(b, c)
```
Example 2:
```
when p:
  attach(d, e)
```

I'd recommend example 2 to simply ignore the when statement, so the expanded form would simply be ```attach(d,e)```.

Example 1 is confusing, but I believe its best to just define the semantics that attach statements always ignore when statements, resulting in ```attach(a,b,c)```. This still abides by the instance-replacement property (arbitrary code can get wrapped in a module and instantiated, and things will still work).

src/test/scala/firrtlTests/AttachSpec.scala
```diff
@@ -147,6 +111,36 @@ class InoutVerilog extends FirrtlFlatSpec {
...
+        |  `ifdef SYNTHESIS
+        |    assign x = a1;
+        |    assign a1 = x;
+        |    assign x = a2;
+        |    assign a2 = x;
+        |    assign a1 = a2;
+        |    assign a2 = a1;
+        |  `else
+        |    alias x = a1 = a2;
+        |  `endif
```
azidar: Add the verilator check

src/main/scala/firrtl/passes/ExpandWhens.scala
jackkoenig: So the current semantics in Firrtl are that all whens are ignored when attaching, Example 1 will indeed turn in to `attach(a, b, c)`. (Chisel errors if you try to attach conditionally, but that's technically not relevant here)

jackkoenig:
@azidar I believe I've addressed all of your concerns. I reworked how ExpandWhens combines the attaches because it wasn't preserving order like I had intended. There are now tests to illustrate that the order is preserved in an intuitive way.

Issue #413, opened 2017-01-19
Wiring pass broken after annotation changes


colinschmidt:
I put the fix here https://github.com/ucb-bar/firrtl/tree/fix-wiring-pins
I wanted to use Jack's PR so I made the branch based off that rather than master which makes the PR look strange.

Also, I'm not sure if thats the right way to do it, but it seems like the fewest changes.

colinschmidt:
closed by #416

PR #416, opened 2017-01-20
use new annotations correctly for wiring

colinschmidt:
This fixes the wiring transform


colinschmidt:
This fixes #413

colinschmidt:
The travis failure is reproducible but I'm not sure why its related to my change...investigating.

colinschmidt:
Actually I spoke too soon. My local reproduction failed because my verilator was too old. I've re-run the test on travis and it passed, so I'm merging.

PR #418, opened 2017-01-24
Blackboxhelper

chick:
FOR REVIEW:
Adds a new mechanism for placing verilator implementations of blackboxes into the directory where verilator or other backend compiler can get to them.
This uses annotations passed from the blackbox modules to determine the source.
The target is based on the execution harness


src/main/scala/firrtl/Driver.scala
```diff
@@ -90,7 +92,20 @@ object Driver {
       val annotationFile = new File(annotationFileName)
       if (annotationFile.exists) {
         val annotationsYaml = io.Source.fromFile(annotationFile).getLines().mkString("\n").parseYaml
-        val annotationArray = annotationsYaml.convertTo[Array[Annotation]]
+        val annotationArray = {
+          val annos = annotationsYaml.convertTo[Array[Annotation]]
+
+          if (annos.nonEmpty) {
+            annos ++ List(Annotation(
+              ModuleName("All", CircuitName("All")),
```
azidar: I dislike this... I assume it isn't possible to get the Circuit name here. At the very least, make it just a CircuitName, not ModuleName(..).

chick: @azidar Will fix.  Problem is that this annotation does not even target a module at all.  Changed for now to CircuitName("All")
probably better but would take more time to follow through would be to create
```scala
final case class NoCircuit extends Named {
    def serialize: String = "NoCircuitSpecified"
}
```

PR #424, opened 2017-01-27
Replacematcherror - catch exceptions and convert to internal error.

ucbjrl:


src/main/scala/firrtl/Utils.scala
```diff
@@ -12,11 +12,55 @@ import scala.collection.mutable.{StringBuilder, ArrayBuffer, LinkedHashMap, Hash
...
+      message
+    } else if (cause != null) {
+      cause.toString
+    } else {
+      null
+    }
+  }
+}
+class FIRRTLException(str: String, cause: Throwable = null)
+  extends RuntimeException(FIRRTLException.defaultMessage(str, cause), cause)
```
jackkoenig: What's the motivation for making FIRRTLException extend RuntimeException? Also, shouldn't we avoid null? I know this essentially wraps Java so perhaps it's unavoidable.

ucbjrl: Here's the motivation:
[extending-exception-runtimeexception-in-java](http://stackoverflow.com/questions/19857008/extending-exception-runtimeexception-in-java), [runtimeexception controversy](https://docs.oracle.com/javase/tutorial/essential/exceptions/runtime.html)

Ideally we should have two categories of FIRRTL exceptions/errors:
- parsing, input errors
- programming - "can't happen here" errors

and we should throw a different exception for each of those conditions.

The "parsing, input errors" can be addressed (at least potentially), by the client, either by changing the input or environment. Programming errors can only be addressed by a developer and require updating firrtl source code.

I'm not sure if it's clear which of those categories a MatchError falls into.

The nulls are inherited from Java.

ucbjrl:
Update existing matches to add case _ => throw InternalError(arg), and keep this catch/throw making sure we output the complete stack trace (from the initial throw).
Add optional arg to InternalError.

PR #427, opened 2017-01-30
Replace createTempDirectory with createTestDirectory

jackkoenig:
Will place tests in ./test_run_dir/ instead of /tmp/

### DIFF
--- a/src/main/scala/firrtl/util/BackendCompilationUtilities.scala
+++ b/src/main/scala/firrtl/util/BackendCompilationUtilities.scala
@@ -4,14 +4,25 @@ package firrtl.util
 
 import scala.sys.process._
 import java.io._
+import java.nio.file.Files
+import java.text.SimpleDateFormat
+import java.util.Calendar
 
 import firrtl._
 import firrtl.{Driver, ExecutionOptionsManager}
 
 import scala.sys.process.{ProcessBuilder, ProcessLogger}
-
-
+ 
 trait BackendCompilationUtilities {
+  /** Parent directory for tests */
+  lazy val TestDirectory = new File("test_run_dir")
+
+  def timeStamp: String = {
+    val format = new SimpleDateFormat("yyyyMMddHHmmss")
+    val now = Calendar.getInstance.getTime
+    format.format(now)
+  }
+
   /** Copy the contents of a resource to a destination file.
     */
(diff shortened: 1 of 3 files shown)


src/main/scala/firrtl/util/BackendCompilationUtilities.scala
```diff
@@ -24,17 +37,25 @@ trait BackendCompilationUtilities {
...
+      val dir: Future[File] = Future {
+        var _dir = new File(outer, timeStamp)
+        while (!_dir.mkdirs()) {
+          Thread.sleep(1)
+          _dir = new File(outer, timeStamp)
+        }
+        _dir
+      }
+      Await.result(dir, 10 seconds)
     }
```
jackkoenig: is there a more elegant way to write this? Should I just have done a random hash instead?

ucbjrl: Can't you use this (in JDK >= 1.7):

    public static Path createTempDirectory(Path dir,
                           String prefix,
                           FileAttribute<?>... attrs)
                                    throws IOException

    Creates a new directory in the specified directory, using the given prefix to generate its
    name. The resulting Path is associated with the same FileSystem as the given directory.
    The details as to how the name of the directory is constructed is implementation
    dependent and therefore not specified. Where possible the prefix is used to construct
    candidate names.

    As with the createTempFile methods, this method is only part of a temporary-file facility. A
    shutdown-hook, or the File.deleteOnExit() mechanism may be used to delete the directory
    automatically.
    
    The attrs parameter is optional file-attributes to set atomically when creating the directory.
    Each attribute is identified by its name. If more than one attribute of the same name is
    included in the array then all but the last occurrence is ignored.

jackkoenig: I thought about using that but I thought it might be nice to have the filenames be based on the current time instead of just being a bunch of random numbers. Do you think using `createTempDirectory` would be better?

ucbjrl: I think we should use standard utility methods where they are available. You can use time as part of the prefix if you like, but be careful about making the absolute path too long.

I can see an argument that this isn't really a temporary directory - it's not clear when it can be deleted - but we don't really care what the name is, as long is it's unique.

PR #428, opened 2017-02-01
Fix anno in backend

chick:
moving backend compilation utilities inadvertently lost the addition of a the black box verilog file list.
This fixes, that as well, as a couple of implementation bugs that slipped in.  Motivated by chisel PR [Move Backend Compilation Utilities](https://github.com/ucb-bar/chisel3/pull/400)

### DIFF
--- a/src/main/scala/firrtl/Driver.scala
+++ b/src/main/scala/firrtl/Driver.scala
@@ -229,19 +229,32 @@ object FileUtils {
     *
     * @param directoryPathName a directory hierarchy to delete
     */
-  def deleteDirectoryHierarchy(directoryPathName: String): Unit = {
-    if(directoryPathName.isEmpty || directoryPathName.startsWith("/")) {
-      // don't delete absolute path
+  def deleteDirectoryHierarchy(directoryPathName: String): Boolean = {
+    deleteDirectoryHierarchy(new File(directoryPathName))
+  }
+  /**
+    * recursively delete all directories in a relative path
+    * DO NOT DELETE absolute paths
+    *
+    * @param file: a directory hierarchy to delete
+    */
+  def deleteDirectoryHierarchy(file: File, atTop: Boolean = true): Boolean = {
+    if(file.getPath.split("/").last.isEmpty ||
+      file.getAbsolutePath == "/" ||
+      file.getPath.startsWith("/")) {
+      Driver.dramaticError(s"delete directory ${file.getPath} will not delete absolute paths")
+      false
     }
     else {
-      val directory = new java.io.File(directoryPathName)
-      if(directory.isDirectory) {
-        directory.delete()
-        val directories = directoryPathName.split("/+").reverse.tail
-        if (directories.nonEmpty) {
-          deleteDirectoryHierarchy(directories.reverse.mkString("/"))
+      val result = {
+        if(file.isDirectory) {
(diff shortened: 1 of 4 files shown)


src/main/scala/firrtl/Driver.scala
```diff
@@ -229,19 +229,32 @@ object FileUtils {
...
+    *
+    * @param file: a directory hierarchy to delete
+    */
+  def deleteDirectoryHierarchy(file: File, atTop: Boolean = true): Boolean = {
+    if(file.getPath.split("/").last.isEmpty ||
+      file.getAbsolutePath == "/" ||
+      file.getPath.startsWith("/")) {
+      Driver.dramaticError(s"delete directory ${file.getPath} will not delete absolute paths")
+      false
     }
```
jackkoenig: Could we not use deleteOnExit() instead? https://docs.oracle.com/javase/7/docs/api/java/io/File.html#deleteOnExit()

src/main/scala/firrtl/util/BackendCompilationUtilities.scala
```diff
@@ -93,8 +89,22 @@ trait BackendCompilationUtilities {
...
+
+    val blackBoxVerilogList = {
+      val list_file = new File(dir, firrtl.transforms.BlackBoxSourceHelper.FileListName)
+      if(list_file.exists()) {
+        Seq("-f", list_file.getAbsolutePath)
+      }
+      else {
+        Seq.empty[String]
+      }
+    }
```
jackkoenig: I'm not convinced that this is the right way to do this. Looking at the function, this adds some magical arguments from a file with a certain name, would it not be better to have the caller pass the filename to the function itself?

src/main/scala/firrtl/Driver.scala
chick: deleteOnExit only deletes a directory if it is empty.  I don't see any particular value to using it over delete.

src/main/scala/firrtl/util/BackendCompilationUtilities.scala
chick: @jackkoenig I agree this could be more elegant, but this works and there are things backed up behind this.  The problem is one we have discussed about the bottom of the toolchain needing to pass stuff back up.  Annotations are a candidate mechanism, there are probably others.  In this case firrtl moves the files, because it knows about what's going on and there is stuff inline that it is well qualified to see.  It does not *know* that these files are being placed there for some subset of the available backends, so the backend knows that if such a file exists that it needs to add a command line options to use them.

src/main/scala/firrtl/Driver.scala
jackkoenig: I guess what I mean is rather than having this function to delete things, we just mark every single file we make with deleteOnExit(), that's generally what you do with temporary test files I believe. Regardless, I'm not going to block on it, just *might* be a better approach

src/main/scala/firrtl/util/BackendCompilationUtilities.scala
jackkoenig: We can discuss improving in the future, I'm fine with merging this to get stuff working. Hopefully once we find a solution for how to pass data back "up" we can refactor this as well.

chick:
@jackkoenig can you change your review to approve then, please

PR #433, opened 2017-02-08
Return a new circuit object after execution

colinschmidt:
This fixes an issue I was having with my ClockListAnnotations being duplicated.

h/t @azidar


chick:
@azidar can you explain how this fix works?

azidar:
Yes, so right now, the contract for Transformations is that they receive all annotations, but only pass on the ones they add. Then, between running transformations, the compiler will merge the new annotations with the other annotations (it saves them before passing the CircuitState to the transformation).

I think this is a bad API, and most certainly should be changed, because it makes this error super easy to do, and debugging it is very difficult unless you know where to look. This has been on my plate for some time, but seemed like a low priority item.

chick:
@azidar thanks for the explanation. Will probably escalate as people start writing their own transforms

PR #435, opened 2017-02-09
Add support for Analog types in partial connect

jackkoenig:
I have the checks still check that the flips are the same despite Analog being bidirectional, thoughts?

This will allow bulk connecting of Analog types in the compatibility wrapper. One consequence is that I believe it will bypass the current Chisel restriction that you can only bulk connect Analog types once per module to prevent strange consequences of last connect semantics.

### DIFF
--- a/src/main/scala/firrtl/passes/Checks.scala
+++ b/src/main/scala/firrtl/passes/Checks.scala
@@ -392,7 +392,7 @@ object CheckTypes extends Pass {
         case (_: UIntType, _: UIntType) => flip1 == flip2
         case (_: SIntType, _: SIntType) => flip1 == flip2
         case (_: FixedType, _: FixedType) => flip1 == flip2
-        case (_: AnalogType, _: AnalogType) => false
+        case (_: AnalogType, _: AnalogType) => true
         case (t1: BundleType, t2: BundleType) =>
           val t1_fields = (t1.fields foldLeft Map[String, (Type, Orientation)]())(
             (map, f1) => map + (f1.name -> (f1.tpe, f1.flip)))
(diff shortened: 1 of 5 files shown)


src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -392,7 +392,7 @@ object CheckTypes extends Pass {
         case (_: UIntType, _: UIntType) => flip1 == flip2
         case (_: SIntType, _: SIntType) => flip1 == flip2
         case (_: FixedType, _: FixedType) => flip1 == flip2
-        case (_: AnalogType, _: AnalogType) => false
+        case (_: AnalogType, _: AnalogType) => flip1 == flip2 // Should this ignore flip?
```
azidar: I think it should ignore flip, but I don't have a strong preference.

src/main/scala/firrtl/passes/InferWidths.scala
```diff
@@ -229,6 +229,7 @@ object InferWidths extends Pass {
       case (t1: SIntType, t2: SIntType) => Seq(WGeq(t1.width, t2.width))
       case (ClockType, ClockType) => Nil
       case (FixedType(w1, p1), FixedType(w2, p2)) => Seq(WGeq(w1,w2), WGeq(p1,p2))
+      case (AnalogType(w1), AnalogType(w2)) => Seq(WGeq(w1,w2))
```
azidar: You need both constraints, return: ```Seq(WGeq(w1,w2), WGeq(w2,w1))```

src/main/scala/firrtl/passes/Checks.scala
colinschmidt: Since you can't flip BlackBoxIO it would be good to ignore flips and connect regardless. Or perhaps this is irrelevant since thats a chisel constraint....

jackkoenig: I agree with ignoring flip, I'll change it

src/main/scala/firrtl/passes/InferWidths.scala
jackkoenig: Why does this need both constraints but the above UIntTypes and SIntTypes don't?

azidar: The semantics of attach are that all wires must have exactly the same width, because there aren't sign/zero-extend semantics for mismatched widths.

jackkoenig: Hahaha for some reason I was reading it as `Weq`, not `WGeq`... my b

jackkoenig:
Changes made, rebased on master

PR #437, opened 2017-02-12
Fixed the unsigned port issue for generating verilog for signed signals

hqjenny:
Added sign directives to the input/output ports to fix the issue that the Verilator returned unsigned value for signed integers and caused disparity in simulation results compared to firrtl simulator.

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -386,6 +386,12 @@ class VerilogEmitter extends Emitter with PassBased {
         case (p, i) => (p.tpe, p.direction) match {
           case (AnalogType(_), _) =>
             Seq("inout", "  ", p.tpe, " ", p.name)
+          case (SIntType(_), Input) =>
+            Seq(p.direction, " signed  ", p.tpe, " ", p.name)
+          case (SIntType(_), Output) =>
+            val ex = WRef(p.name, p.tpe, PortKind, FEMALE)
+            assign(ex, netlist(ex))
+            Seq(p.direction, " signed ", p.tpe, " ", p.name)
           case (_, Input) =>
             Seq(p.direction, "  ", p.tpe, " ", p.name)
           case (_, Output) =>


shunshou:
I talked to Jenny about this earlier. I should've, at least at https://github.com/ucb-bar/chisel-testers/blob/master/src/main/scala/chisel3/iotesters/VerilatorBackend.scala fixed the sign issue for peek, however, I've been testing it in the context of my own custom DspTester (not PeekPokeTester!) with overriden poke/peek/expects. I've definitely tried with negative numbers and Verilator (poke/peek/expect), and at least for my DspTester, I don't see the sign issue (due to just using the VerilatorBackend peek instead of expect, maybe? Although I can't see why... I'm really not sure when the path: String versions of any of the functions get used vs. the signal: InstanceId versions -- not even sure why there need to be two versions). Merging this will require that the changes I made to the VerilatorBackend peek be undone (along with some other things in dsptools). Please let me know when this is merged, so I can make sure I don't have double negatives in my code!

shunshou:
Actually, I think I see what's going on -- chisel-template (or whatever it's called), by default, uses the release version of chisel-testers. And that version doesn't contain any of my VerilatorBackend sign fixes @ucbjrl . It's probably still better to fix things at the Firrtl emissions level (?), but doing so requires that all of my sign fixes in chisel-testers get undone.

shunshou:
One more thing I'd like to add: For whatever reason, the Chisel2 C++ simulator also returned unsigned everythings and required an explicit sign_fix.

ucbjrl:
The release branch of chisel-template (and chisel-tutorial) uses the
release branch of the other repos (chisel3, firrtl, firrtl-interpreter,
chisel-testers). The master branch of chisel-template uses the master
branch of the other repos.

Currently, the default branch for chisel-template (and chisel-tutorial) is
the release branch. You need to explicitly check out the master branch to
work with it.

The thinking was release should be the default - it is "guaranteed" to
work. If you know what you're doing (and prepared to deal with bleeding
edge issues), you can check out the master branch.

The master brach of chisel-template should work with whatever fixes are in
the master branch of chisel-testers. If it doesn't, we need to fix it -
chisel-template that is.

shunshou:
Essentially, the problem boils down to the release branches of FIRRTL/chisel-testers not being correct for test benches with signed outputs. In the master branch of chisel-testers, I fixed the problem. Alternatively, as Jenny as done, you can fix the problem by changing the way FIRRTL emits signed outputs. That's probably the better approach, but require that I remove my sign fixes from chisel-testers. Regardless, the combinations of the various release branches currently produce incorrect testing

ucbjrl:
We'll be pushing out a new release (based on current masters) this coming
week.

shunshou:
Cool! I do wish there were some more obvious way for people to realize that potential bugs in release versions have actually been fixed on the master branch. It wasn't until Jenny pointed me to her code that I realized she was on release... and I was pretty confused why my solution hadn't worked for her.

azidar:
Sounds like we should remove Angie's fix from chisel-testers. @shunshou, can you submit a PR removing the changes to chisel-testers, and then merge both at the same time?

shunshou:
Will do. It'll prob happen late tomorrow tho. Kind of swamped with some stuff I promised people that I haven't had time to deal with until now. ~_~

ucbjrl:
I don't understand. All the master branches seem consistent, and as far as
I can see, all their tests pass.

Why do we need to back out Angie's fix to chisel-testers? Shouldn't we
update the releases to support it (i.e. merge masters into releases)?

shunshou:
There are two "correct" solutions:

My fix to chisel-testers + current FIRRTL master
Chisel-testers without my fix + This PR

However, I think it's better to solve the problem at the FIRRTL level, so it's best to undo my changes to chisel-testers and merge this PR. I'll double check that this works as intended and make a new PR to chisel-testers.

ucbjrl:
Got it. Thanks for the clarification Angie.

shunshou:
Can someone sanity check that this fix is compatible with chisel-testers for me? @hqjenny @azidar 

Probably best done by someone familiar with writing Verilator CPP testers? I removed the sign_fix stuff on Chisel testers and am using these Firrtl fixes, but the result is that everything is unsigned... Were these changes maybe accidentally tested with master branch of chisel-testers (that I patched up?) instead of with the release branch so it looked like these changes fixed the problem? 

-- Looks like the signed directive is ignored by the generated CPP TB, unless I'm mistaken. If so, then, maybe this is still good to have as reference (for other simulators?) but ineffective without sign_fix in Chisel Testers / VerilatorBackend.

aswaterman:
I believe this PR is incorrect, or is at least incomplete.  Anywhere the port is used in unsigned context in Verilog (which I believe happens for SInts in some cases), we'd have to change code generation to cast it to $unsigned.  And since ports and wires are used interchangeably as data sources, if you make SInt ports signed, you'd need to symmetrically make wires and regs signed.

I contend FIRRTL's code generation is already correct.  Something external should handle the mismatch with the interpreter (e.g. truncating the sign bits in the interpreter case).

shunshou:
Without this PR, my fix to chisel-testers properly handles the mismatch.

aswaterman:
Yeah, we should go with that approach.  OK to close?

shunshou:
I'm not really familiar with Verilog (VHDL is more my thing), but is it worthwhile having something external reinterpret the bits vs. just changing the way Firrtl emits Verilog (even if that might mean more places need to be changed)?

aswaterman:
I continue to believe FIRRTL is doing the right thing here.  Externally to Verilog, the ports are just bits, and the simulation context is responsible for interpreting them appropriately.  For example, it must know if a 64-bit bus is to be interpreted as an address, an IEEE-754 double, a signed integer, an opaque type, an enum, or whatever.  Just adding signed the signed designation addresses only one specific case, not the general problem, in exchange for additional complexity and likely lower simulation performance (lots of extra casts).

jackkoenig:
Given that an alternative solution exists and that this PR not sufficient as is, I'm going to close it.

PR #438, opened 2017-02-13
Emit memories larger than 512 MB with a sparse annotation

colinschmidt:
This comment causes vcs to treat this as a spare memory,
so it will dynamically allocate the required memory, and can
support very large reg constructs this way. This is useful for
test bench memories that might be simulating back DRAM or the like.

I tested this on rocket-chip's TestRAM.

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -454,7 +454,11 @@ class VerilogEmitter extends Emitter with PassBased {
           instdeclares += Seq(");")
           sx
         case sx: DefMemory =>
-          declare("reg", sx.name, VectorType(sx.dataType, sx.depth))
+          val fullSize = sx.depth*(sx.dataType match { case GroundType(i:IntWidth) => i.width })
+          if(fullSize > (1 << 29))
+            declare("reg /* sparse */", sx.name, VectorType(sx.dataType, sx.depth))
+          else
+            declare("reg", sx.name, VectorType(sx.dataType, sx.depth))
           initialize_mem(sx)
           if (sx.readLatency != 0 || sx.writeLatency != 1)
             throw EmitterException("All memories should be transformed into " +
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -454,11 +454,9 @@ class VerilogEmitter extends Emitter with PassBased {
           instdeclares += Seq(");")
           sx
         case sx: DefMemory =>
-          val fullSize = sx.depth*(sx.dataType match { case GroundType(i:IntWidth) => i.width })
-          if(fullSize > (1 << 29))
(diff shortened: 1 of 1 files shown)


src/main/scala/firrtl/Emitter.scala
```diff
@@ -454,7 +454,11 @@ class VerilogEmitter extends Emitter with PassBased {
           instdeclares += Seq(");")
           sx
         case sx: DefMemory =>
-          declare("reg", sx.name, VectorType(sx.dataType, sx.depth))
+          val fullSize = sx.depth*(sx.dataType match { case GroundType(i:IntWidth) => i.width })
+          if(fullSize > (1 << 29))
```
jackkoenig: Space after the if

src/main/scala/firrtl/Emitter.scala
```diff
@@ -454,7 +454,11 @@ class VerilogEmitter extends Emitter with PassBased {
           instdeclares += Seq(");")
           sx
         case sx: DefMemory =>
-          declare("reg", sx.name, VectorType(sx.dataType, sx.depth))
+          val fullSize = sx.depth*(sx.dataType match { case GroundType(i:IntWidth) => i.width })
```
jackkoenig: Space around `*` and please unapply `IntWidth` as well (ie. ` case GroundType(IntWidth(width)) => width`)

src/main/scala/firrtl/Emitter.scala
```diff
@@ -454,7 +454,11 @@ class VerilogEmitter extends Emitter with PassBased {
...
+          if(fullSize > (1 << 29))
+            declare("reg /* sparse */", sx.name, VectorType(sx.dataType, sx.depth))
+          else
+            declare("reg", sx.name, VectorType(sx.dataType, sx.depth))
```
jackkoenig: since the if is only changing the first arg of declare I think it would be better if you rewrote this as something like:
```scala
val decl = if (fullSize > (1 << 29)) "reg /* sparse */" else "reg"
declare(decl, sx.name, VectorType(sx.dataType, sx.depth))
```

colinschmidt:
Made all 3 changes

src/main/scala/firrtl/Emitter.scala
```diff
@@ -454,7 +454,9 @@ class VerilogEmitter extends Emitter with PassBased {
           instdeclares += Seq(");")
           sx
         case sx: DefMemory =>
-          declare("reg", sx.name, VectorType(sx.dataType, sx.depth))
+          val fullSize = sx.depth * (sx.dataType match { case GroundType(IntWidth(width)) => i.width })
```
jackkoenig: Should just be width instead of i.width no?

colinschmidt: Yup...I've now actually tested it again.

PR #439, opened 2017-02-13
Add chisel2 isVCSAvailable, isCommandAvailable to FileUtils.

ucbjrl:


azidar:
Can you explain why this is needed? Also, please assign a reviewer :)

ucbjrl:
We would like to conditionally enable VCS backend tests in various components, if `vcs` is available (i.e., on the millennium machines under Jenkins).

PR #443, opened 2017-02-14
Add ability to emit 1 file per module

jackkoenig:
This changes code that relied on this assumption. It also adds new types
for emission that understand emitting modules separately. Also adds a
commandline option.


jackkoenig:
This refactors basically what it means to emit, removing the idea of a Writer. I left the old functions, but they should probably be deprecated and removed.

colinschmidt:
Is this PR mis-titled? Should be "more than 1 file per circuit" no?

jackkoenig:
Hahaha yes it was. Fixed

src/main/scala/firrtl/Emitter.scala
```diff
@@ -22,8 +22,77 @@ import scala.collection.mutable.{ArrayBuffer, LinkedHashMap, HashSet}
...
+sealed abstract class EmittedModule {
+  def name: String
+  def emit: String
+}
+class EmittedVerilogModule(val name: String, val value: String) extends EmittedModule {
+  def emit: String = EmittedVerilogModule.preamble + "\n" + value
+}
+case object EmittedVerilogModule {
+  /** Preamble for every emitted Verilog file */
+  def preamble: String =
```
chick: Should users be able to customize this preamble?

jackkoenig:
Lots of changes but I think they make a lot of sense.

* I made Emitters Transforms
    * For practical reasons, the emitter is still a special transform in Compilers, but this PR is a step in the direction of making them entirely the same as transforms
* I added Annotations for telling the Emitters to do emission and Annotations for the results of emission. 
    * For performance reasons, the results of emission are kept in a private datastructure (otherwise they have to be serialized into the Annotation and deserialized out, this can be changed once we enhance Annotations to allow arbitrary data structures)
* I added a command-line option for emitting each module to a different file.
    * The plumbing to pull this off can be extended to support emitting just 1 module to a different file and other alternates like that
* I deprecated using writers and added some helper convenience method for extracting the result of emission from the `CircuitState`
* I updated the tests to reflect the new API

@azidar please take a look when you get a chance. You will note I actually changed the command-line option to be more like the current style where there are implicit dependencies between them. Refactoring this is a separate discussion so I decided not to try to change the API partially in this PR for not a lot of benefit.

jackkoenig:
Any thoughts on the updated form of this @azidar? Some big deprecations so let's be sure we like the API replacing it (I think the helpers `Compiler.compileAndEmit` along with `CircuitState.emittedCircuitOption` are sufficient).

@colinschmidt and @shunshou, any thoughts on the changes to compiling and emitting here?

shunshou:
I guess I never directly run Firrtl in my setup (always go through Chisel driver execute), so as long as the output is still the same, I have no strong preferences?

jackkoenig:
Yeah cool there are no real changes to execute except a new field. 1 thing that might be relevant is that FirrtlExecutionSuccess returns a String result. Currently in this PR if you specify to emit modules separately, it just returns an empty string in the FirrtlExecutionSuccess. I believe the right way to deal with this is to change that type, but yeah I avoided modifying that API here.

shunshou:
So does that mean all modules in a circuit are now split into new files? And you're saying that, if that's the case, I won't be able to get the full Verilog via FirrtlExecutionSuccess? Guess that might require some changes to some of my (and others') testers?

jackkoenig:
No, modules in a circuit are split into new files *if you provide the option to do so*, otherwise the behavior is identical to before and the FirrtlExecutionSuccess will still return the full Verilog.

shunshou:
Ok cool.

jackkoenig:
I do wonder what we should do in FirrtlExecutionSuccess for when splitting up the Modules is desired though. I think that's an API that should be changed in a separate PR though.

shunshou:
List of files? Or you could concatenate all the Verilog...

colinschmidt:
Ok so I'm trying to see how I can use this to replace GenerateTop and GenerateHarness.

If I split all modules to separate files then all I need to construct "Top" is a list of all modules under the hierarchy of a given module. To construct "Harness" I need the inverse of that list.
It seems like this would cause GenerateTop and GenerateHarness to be much simpler in terms of time/complexity because they don't need to do any of the circuit modification stuff, just use this new feature and enumerate some modules.

I also use Driver.compile now so I'll need to figure out how to migrate to Driver.execute. Is there an example of this somewhere?

jackkoenig:
I usually just look at DriverSpec. For example, see the optionsManager for the test I added:
https://github.com/ucb-bar/firrtl/blob/9a6a2a05abca97a4ec178112083a036712146d03/src/test/scala/firrtlTests/DriverSpec.scala#L188

This shows how to set various options, to see what is available see:
https://github.com/ucb-bar/firrtl/blob/9a6a2a05abca97a4ec178112083a036712146d03/src/main/scala/firrtl/ExecutionOptionsManager.scala#L31
and
https://github.com/ucb-bar/firrtl/blob/9a6a2a05abca97a4ec178112083a036712146d03/src/main/scala/firrtl/ExecutionOptionsManager.scala#L140

We need a ScalaDoc hahaha

colinschmidt:
Ok looks good to me. I agree scala doc would help me.

src/main/scala/firrtl/annotations/Named.scala
```diff
@@ -14,6 +14,13 @@ sealed trait Named {
...
 }
 
+// TODO Is there a better way to just indicate the top?
+// Perhaps this should be a termporary thing that gets turned into CircuitName below
+final case object CircuitTopName extends Named {
+  def name: String = "CircuitTop"
+  def serialize: String = name
+}
+
 final case class CircuitName(name: String) extends Named {
```
azidar: Can you just delete CircuitName?

src/main/scala/firrtl/ExecutionOptionsManager.scala
```diff
@@ -125,6 +125,10 @@ trait HasCommonOptions {
   parser.help("help").text("prints this usage text")
 }
 
+sealed abstract class OutputConfig
```
azidar: So this is the post-processed execution options? A comment would be nice.

src/main/scala/firrtl/Emitter.scala
```diff
@@ -22,10 +23,147 @@ import scala.collection.mutable.{ArrayBuffer, LinkedHashMap, HashSet}
...
+  // Old style, deprecated
   def emit(state: CircuitState, writer: Writer): Unit = writer.write(state.circuit.serialize)
 }
 
+// ***** Start actual Emitters *****
+class HighFirrtlEmitter extends FirrtlEmitter(HighForm)
+class MiddleFirrtlEmitter extends FirrtlEmitter(HighForm)
+class LowFirrtlEmitter extends FirrtlEmitter(HighForm)
+
+//class FirrtlEmitter extends Emitter {
```
azidar: Remove comments

src/main/scala/firrtl/Emitter.scala
```diff
@@ -22,10 +23,147 @@ import scala.collection.mutable.{ArrayBuffer, LinkedHashMap, HashSet}
 
 case class EmitterException(message: String) extends PassException(message)
 
-class FirrtlEmitter extends Emitter {
+// ***** Annotations for telling the Emitters what to emit *****
```
azidar: I think its time to split Emitter into separate files, and probably rework the directory structure. This can wait for another PR, though.

src/test/scala/firrtlTests/CInferMDirSpec.scala
```diff
@@ -70,9 +70,8 @@ circuit foo :
 """.stripMargin
 
     val annotationMap = AnnotationMap(Nil)
-    val writer = new java.io.StringWriter
-    compile(CircuitState(parse(input), ChirrtlForm, Some(annotationMap)), writer)
+    val res = compileAndEmit(CircuitState(parse(input), ChirrtlForm, Some(annotationMap)))
     // Check correctness of firrtl
-    parse(writer.toString)
+    parse(res.emittedCircuitOption.get.value)
```
azidar: Can you add a helper method to do this? Because its done all over the place in testing, like `res.getEmittedCircuit`?

src/test/scala/firrtlTests/CompilerTests.scala
```diff
@@ -170,6 +167,9 @@ circuit Top :
       "endmodule\n"
    ).reduce(_ + "\n" + _)
    "A circuit's verilog output" should "match the given string" in {
-      (getOutput) should be (check)
+      // For some reason just comparing the Strings doesn't work, but comparing characters does
+      for ((x, y) <- getOutput.zip(check)) {
```
azidar: That's weird that comparing Strings doesn't work....

src/test/scala/firrtlTests/DriverSpec.scala
```diff
@@ -66,7 +66,7 @@ class DriverSpec extends FreeSpec with Matchers with BackendCompilationUtilities
       val firrtlOptions = optionsManager.firrtlOptions
       val inputFileName = optionsManager.getBuildFileName("fir", firrtlOptions.inputFileNameOverride)
       inputFileName should be ("./cat.fir")
-      val outputFileName = optionsManager.getBuildFileName("v", firrtlOptions.outputFileNameOverride)
+      val outputFileName = firrtlOptions.getOutputConfig(optionsManager).asInstanceOf[SingleFile].targetFile
```
azidar: Can you add a helper method, like `firrtlOptions.getTargetFile`? And issue a nice error message if one doesn't exist?

src/test/scala/firrtlTests/DriverSpec.scala
```diff
@@ -79,7 +79,7 @@ class DriverSpec extends FreeSpec with Matchers with BackendCompilationUtilities
       val firrtlOptions = optionsManager.firrtlOptions
       val inputFileName = optionsManager.getBuildFileName("fir", firrtlOptions.inputFileNameOverride)
       inputFileName should be ("./bob.fir")
-      val outputFileName = optionsManager.getBuildFileName("v", firrtlOptions.outputFileNameOverride)
+      val outputFileName = firrtlOptions.getOutputConfig(optionsManager).asInstanceOf[SingleFile].targetFile
```
azidar: Use helper method.

src/test/scala/firrtlTests/InferReadWriteSpec.scala
```diff
@@ -76,10 +77,9 @@ circuit sram6t :
 """.stripMargin
 
     val annotationMap = AnnotationMap(Seq(memlib.InferReadWriteAnnotation("sram6t")))
-    val writer = new java.io.StringWriter
-    compile(CircuitState(parse(input), ChirrtlForm, Some(annotationMap)), writer)
+    val res = compileAndEmit(CircuitState(parse(input), ChirrtlForm, Some(annotationMap)))
     // Check correctness of firrtl
-    parse(writer.toString)
+    parse(res.emittedCircuitOption.get.value)
```
azidar: Fix as well.

src/test/scala/firrtlTests/MultiThreadingSpec.scala
```diff
@@ -13,10 +13,9 @@ class MultiThreadingSpec extends FirrtlPropSpec {
   property("The FIRRTL compiler should be thread safe") {
     // Run the compiler we're testing
     def runCompiler(input: Seq[String], compiler: firrtl.Compiler): String = {
-      val writer = new java.io.StringWriter
       val parsedInput = firrtl.Parser.parse(input)
-      compiler.compile(CircuitState(parsedInput, ChirrtlForm), writer)
-      writer.toString
+      val res = compiler.compileAndEmit(CircuitState(parsedInput, ChirrtlForm))
+      res.emittedCircuitOption.get.value
```
azidar: fix as well.

src/test/scala/firrtlTests/PassTests.scala
```diff
@@ -9,50 +9,28 @@ import org.scalatest.junit.JUnitRunner
...
    def parse(s: String): Circuit = Parser.parse(s.split("\n").toIterator, infoMode = UseInfo)
    def squash(c: Circuit): Circuit = RemoveEmpty.run(c)
 
    // Executes the test. Call in tests.
-   def execute(writer: Writer, annotations: AnnotationMap, input: String, check: String): Unit = {
-      compile(CircuitState(parse(input), ChirrtlForm, Some(annotations)), writer)
-      val actual = RemoveEmpty.run(parse(writer.toString)).serialize
+   def execute(annotations: AnnotationMap, input: String, check: String): Unit = {
+      val finalState = compileAndEmit(CircuitState(parse(input), ChirrtlForm, Some(annotations)))
+      val actual = RemoveEmpty.run(parse(finalState.emittedCircuitOption.get.value)).serialize
```
azidar: fix as well.

src/test/scala/firrtlTests/ReplSeqMemTests.scala
```diff
@@ -62,10 +63,9 @@ circuit Top :
 """.stripMargin
     val confLoc = "ReplSeqMemTests.confTEMP"
     val aMap = AnnotationMap(Seq(ReplSeqMemAnnotation("-c:Top:-o:"+confLoc)))
-    val writer = new java.io.StringWriter
-    compile(CircuitState(parse(input), ChirrtlForm, Some(aMap)), writer)
+    val res = compileAndEmit(CircuitState(parse(input), ChirrtlForm, Some(aMap)))
     // Check correctness of firrtl
-    parse(writer.toString)
+    parse(res.emittedCircuitOption.get.value)
```
azidar: fix as well.

src/test/scala/firrtlTests/ReplSeqMemTests.scala
```diff
@@ -85,10 +85,9 @@ circuit Top :
...
```
azidar: fix as well.

src/test/scala/firrtlTests/ReplSeqMemTests.scala
```diff
@@ -111,10 +110,9 @@ circuit CustomMemory :
 """.stripMargin
     val confLoc = "ReplSeqMemTests.confTEMP"
     val aMap = AnnotationMap(Seq(ReplSeqMemAnnotation("-c:CustomMemory:-o:"+confLoc)))
-    val writer = new java.io.StringWriter
-    compile(CircuitState(parse(input), ChirrtlForm, Some(aMap)), writer)
+    val res = compileAndEmit(CircuitState(parse(input), ChirrtlForm, Some(aMap)))
     // Check correctness of firrtl
-    parse(writer.toString)
+    parse(res.emittedCircuitOption.get.value)
```
azidar: fix

src/test/scala/firrtlTests/ReplSeqMemTests.scala
```diff
@@ -137,10 +135,9 @@ circuit CustomMemory :
...
```
azidar: fix

src/test/scala/firrtlTests/ReplSeqMemTests.scala
```diff
@@ -213,10 +210,9 @@ circuit CustomMemory :
     val aMap = AnnotationMap(Seq(
       ReplSeqMemAnnotation("-c:CustomMemory:-o:"+confLoc),
       NoDedupMemAnnotation(ComponentName("mem_0", ModuleName("CustomMemory",CircuitName("CustomMemory"))))))
-    val writer = new java.io.StringWriter
-    compile(CircuitState(parse(input), ChirrtlForm, Some(aMap)), writer)
+    val res = compileAndEmit(CircuitState(parse(input), ChirrtlForm, Some(aMap)))
     // Check correctness of firrtl
-    val circuit = parse(writer.toString)
+    val circuit = parse(res.emittedCircuitOption.get.value)
```
azidar: fix

src/test/scala/firrtlTests/ReplSeqMemTests.scala
```diff
@@ -254,10 +250,9 @@ circuit CustomMemory :
     val aMap = AnnotationMap(Seq(
       ReplSeqMemAnnotation("-c:CustomMemory:-o:"+confLoc),
       NoDedupMemAnnotation(ComponentName("mem_1", ModuleName("CustomMemory",CircuitName("CustomMemory"))))))
-    val writer = new java.io.StringWriter
-    compile(CircuitState(parse(input), ChirrtlForm, Some(aMap)), writer)
+    val res = compileAndEmit(CircuitState(parse(input), ChirrtlForm, Some(aMap)))
     // Check correctness of firrtl
-    val circuit = parse(writer.toString)
+    val circuit = parse(res.emittedCircuitOption.get.value)
```
azidar: fix etc.

src/main/scala/firrtl/annotations/Named.scala
jackkoenig: I'm thinking that should be done in a separate commit because it touches quite a bit of stuff and the conversion isn't completely mechanical. Things like inline instances work with circuitName[.moduleName[.componentName]] are kind of strange to change since circuitName would now be ignored. I think we should do this separately so we can discuss it independently of all of this

src/main/scala/firrtl/Emitter.scala
jackkoenig: Agreed, but yeah I agree it should be done in a follow on PR.

jackkoenig:
I've updated with implementing most of the review comments and merging master.

PR #444, opened 2017-02-14
[skip chisel tests] Add ability to skip Chisel tests in Travis

jackkoenig:
Since #443 changes an API that chisel tests use, the Chisel tests can't pass. This PR adds a tag that we can use to skip the chisel tests. I could just [skip ci] but I want #443 to run the Firrtl tests so I thought this would be helpful

### DIFF
--- a/.run_chisel_tests.sh
+++ b/.run_chisel_tests.sh
@@ -1,0 +1,11 @@
+set -e
+# Skip chisel tests if the commit message says to
+if [[ $TRAVIS_COMMIT_MESSAGE == *"[skip chisel tests]"* ]]; then
+  exit 0
+else
+  git clone https://github.com/ucb-bar/chisel3.git
+  mkdir -p chisel3/lib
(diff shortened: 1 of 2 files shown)


colinschmidt:
LGTM. It might be worth double checking that they will run if you don't include it by including another commit on the branch that doesn't have the flagging note.

jackkoenig:
Good question, let me experiment with this a little more before merging

jackkoenig:
[My check failed](https://travis-ci.org/ucb-bar/firrtl/builds/202364747) so I'm thinking this is Gucci. We will have to enforce that people use `[skip chisel tests]` correctly but that is what code reviews are for.

jackkoenig:
(The check was supposed to fail btw, it was making sure that a commit that followed a [skip chisel tests] commit still triggered running the chisel tests)

colinschmidt:
Yup lgtm. Merge whenever

PR #447, opened 2017-02-19
Implementation of nodedupe mem

colinschmidt:
This allows the replseqmem transform to not deduplicate
some memories, based on their name.

### DIFF
--- a/src/main/scala/firrtl/passes/memlib/ResolveMemoryReference.scala
+++ b/src/main/scala/firrtl/passes/memlib/ResolveMemoryReference.scala
@@ -2,35 +2,55 @@
 
 package firrtl.passes
 package memlib
+import firrtl._
 import firrtl.ir._
 import AnalysisUtils.eqMems
 import firrtl.Mappers._
+import firrtl.annotations._
 
+/** A component, e.g. register etc. Must be declared only once under the TopAnnotation
+  */
+object NoDedupMemAnnotation {
+  def apply(target: ComponentName): Annotation = Annotation(target, classOf[ResolveMemoryReference], s"nodedupmem!")
+
+  def unapply(a: Annotation): Option[ComponentName] = a match {
+    case Annotation(ComponentName(n, mn), _, "nodedupmem!") => Some(ComponentName(n, mn))
+    case _ => None
+  }
+}
 
 /** Resolves annotation ref to memories that exactly match (except name) another memory
  */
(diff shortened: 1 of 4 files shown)


src/main/scala/firrtl/passes/memlib/ResolveMemoryReference.scala
```diff
@@ -2,35 +2,55 @@
...
-    c copy (modules = c.modules map (m => m map updateMemStmts(m.name, uniqueMems)))
+    c copy (modules = c.modules map (m => m map updateMemStmts(m.name, uniqueMems, noDeDupeMems)))
+  }
+  def execute(state: CircuitState): CircuitState = {
+    getMyAnnotations(state) match {
+      case Nil => CircuitState(run(state.circuit, Seq.empty), state.form)
+      case annos =>
+        val noDedups = annos.collect { case NoDedupMemAnnotation(ComponentName(cn, _)) => cn }
+        CircuitState(run(state.circuit, noDedups), state.form)
+    }
```
jackkoenig: I think this would be better if only 1 `CircuitState` is constructed, the `getMyAnnotations... match` can return the Seq of annotations that is then passed to run.

src/test/scala/firrtlTests/ReplSeqMemTests.scala
```diff
@@ -187,10 +188,121 @@ circuit Top :
...
+    read mport _T_17 = mem_0[io.rAddr], clock
+    read mport _T_19 = mem_1[io.rAddr], clock
+    io.dO <= _T_17
+    when io.wEn :
+      write mport _T_18 = mem_0[io.wAddr], clock
+      write mport _T_20 = mem_1[io.wAddr], clock
+      _T_18 <= io.dI
+      _T_20 <= io.dI
+      skip
+""".stripMargin
```
jackkoenig: No need to stripMargin if you aren't using a margin

colinschmidt:
Review comments fixed.

Yes it solves my issue in hwacha/hurricane.

colinschmidt:
Investigating travis error.

colinschmidt:
I'm going to just re-run it since locally sbt test succeeds and the other travis run also succeeded.
It just seems to have spuriously hung.

jackkoenig:
Travis has spurious failures that I think arise from running out of memory or something. Need to look into but don't have time right now. People should just rerun their tests if they fail.

PR #451, opened 2017-02-21
Memoize hashCode for all FirrtlNodes

jackkoenig:
This PR adds some additional state to every FirrtlNode but speeds up passes that need to hash stuff (like ExpandWhens and CSE)

I benchmarked this branch using my trusty new [firrtlbench](https://github.com/jackkoenig/firrtlbench) repo.

Looking at regress/Rob.fir and rocketchip.DefaultL2Config.fir (built with [rocket-chip master](https://github.com/ucb-bar/rocket-chip/commit/45d016a76f155900d2e267072b804c20100dafb1))

Runtime with current Firrtl master b69e787c0a698b7fb703ccd8d24003f83207e296
[info] BenchmarkVerilogCompiler.measureRob     avgt    5  10.282 ± 0.823   s/op
[info] BenchmarkVerilogCompiler.measureRocket  avgt    5  14.875 ± 1.497   s/op

Runtime with this branch
[info] BenchmarkVerilogCompiler.measureRob     avgt    5   6.975 ± 1.667   s/op
[info] BenchmarkVerilogCompiler.measureRocket  avgt    5  12.722 ± 0.501   s/op

Speedup of 15-30%, not bad!


jackkoenig:
To illustrate why this has such a large impact, hashCode is called 1,428,192 times on WRef but only 226,259 of those calls are on unique WRefs.

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -328,7 +333,7 @@ class IntWidth(val width: BigInt) extends Width with Product {
     case w: IntWidth => width == w.width
     case _ => false
   }
-  override def hashCode = width.toInt
+  override lazy val hashCode = width.toInt
```
aswaterman: This one should be faster if not lazy, since computing this simple hash code is probably less costly than the lazy val bookkeeping.

sdtwigg: I originally wanted to complain that this will lead to very clustered hashCodes (since width tends to be clustered) and a bad distribution. However, it seems scala map implementations (like HashMap) will 'improve' the hashCode anyway:
e.g. https://github.com/scala/scala/blob/v2.12.2/src/library/scala/collection/immutable/HashMap.scala#L86

so this might not be an issue. Although, it would be good to get a collections expert to affirm this.

jackkoenig:
Performance specified as: mean (stdev), both in ms

| design | master | this branch |
| --- | --- | --- |
| rocket | 19.7 (0.43) |  19.3 (0.28) |
| boom | 40.4 (0.55) | 35.0 (0.31) |
| hwacha | 72.2 (1.6) | 67.9 (2.4) |

Not as big a win as it was originally unfortunately

azidar:
Is this still relevant?

jackkoenig:
I think its worth revisiting/rebenchmarking

aswaterman:
Agreed, still want to fix the lazy val -> def thing for IntWidth though

aswaterman:
Disregard my comment - changing lazy val to def is too onerous in this organization, and it's minor anyway.

jackkoenig:
I'm running some benchmarks

jackkoenig:
As far as I can tell, performance is a wash. I'm going to table this.

PR #460, opened 2017-02-23
Add shorthands

shunshou:


src/main/scala/firrtl/passes/memlib/MemUtils.scala
```diff
@@ -154,6 +155,10 @@ object createSubField {
   def apply(exp: Expression, n: String) = WSubField(exp, n, field_type(exp.tpe, n), UNKNOWNGENDER)
 }
 
+object createInstance {
```
azidar: It would be better to put the helpers near their corresponding class declarations in WIR.scala. See WRef for an example.

src/main/scala/firrtl/passes/memlib/MemUtils.scala
```diff
@@ -94,6 +94,7 @@ object castRhs {
     lhst match {
       case _: SIntType => DoPrim(AsSInt, Seq(rhs), Seq.empty, lhst)
       case FixedType(_, IntWidth(p)) => DoPrim(AsFixedPoint, Seq(rhs), Seq(p), lhst)
+      case ClockType => DoPrim(AsClock, Seq(rhs), Seq.empty, lhst)
```
azidar: When are you casting a clock type?

azidar:
@jackkoenig any thoughts on this? You are planning on refactoring WIR, so is this an acceptable solution in the mean time?

jackkoenig:
Looks fine to me

PR #462, opened 2017-02-23
Add support for bundle fields to start with digits

jackkoenig:
Also remove parsing support for ids with characters not supported in
Verilog nor in the Firrtl spec

As discussed in https://github.com/ucb-bar/chisel3/pull/517, it is useful for Firrtl to support digits as legal identifiers in fields since chisel Records can be used as heterogeneous Vecs

Adding this support was more complicated than I expected since it involved refactoring the lexer rules a bit but here it is!

I also removed "support" for parsing ids that aren't actually supported

### DIFF
--- a/src/main/antlr4/FIRRTL.g4
+++ b/src/main/antlr4/FIRRTL.g4
@@ -52,17 +52,17 @@ dir
   ;
 
 type 
-  : 'UInt' ('<' IntLit '>')?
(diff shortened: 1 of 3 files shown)


seldridge:
It would be great to update the BNF in the spec to add `fieldId` and/or to add some section on what are legal identifiers (as this makes the notion of an "identifier" different for fields vs. everything else).

It's also possible that I missed it in the spec.

PR #463, opened 2017-02-23
Align types and names of ports in emitted Verilog

jackkoenig:
I'm thinking padding the directions may be a little enthusiastic since there are exactly 3 and they aren't going to change. Perhaps I should get rid of `stringify` for `Direction` and instead just turn the directions into strings (with padding) in `build_ports`?

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -67,11 +67,7 @@ class VerilogEmitter extends Emitter with PassBased {
       val wx = bitWidth(tpe) - 1
       if (wx > 0) s"[$wx:0]" else ""
     case ClockType => ""
-    case _ => error("Shouldn't be here")
-  }
-  def stringify(dir: Direction): String = dir match {
-    case Input => "input"
-    case Output => "output"
+    case _ => error("Trying to write unsupported type in the Verilog Emitter")
   }
   def emit(x: Any)(implicit w: Writer) { emit(x, 0) }
   def emit(x: Any, top: Int)(implicit w: Writer) {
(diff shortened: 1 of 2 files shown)


src/main/scala/firrtl/Emitter.scala
```diff
@@ -62,6 +62,17 @@ class VerilogEmitter extends Emitter with PassBased {
       s".${name}($strx)"
     case RawStringParam(name, value) => s".$name($value)"
   }
+  def stringify(tpe: GroundType): String = tpe match {
+    case (_: UIntType | _: SIntType | _: AnalogType) =>
+      val wx = bitWidth(tpe) - 1
+      if (wx > 0) s"[$wx:0]" else ""
+    case ClockType => ""
+    case _ => error("Shouldn't be here")
```
colinschmidt: Make this more descriptive... "Trying to write an unsupported type in the Emitter"

src/main/scala/firrtl/Emitter.scala
```diff
@@ -382,16 +383,26 @@ class VerilogEmitter extends Emitter with PassBased {
...
+            case (_, AnalogType(_)) => "inout"
+            case (Input, _) => stringify(dir)
+            case (Output, _) =>
+              // Assign to the Port
+              val ex = WRef(name, tpe, PortKind, FEMALE)
+              assign(ex, netlist(ex))
+              stringify(dir)
+          }
+        }
+        val tpes = m.ports map (p => stringify(p.tpe.asInstanceOf[GroundType]))
```
colinschmidt: Is this a reasonable cast to make? Should we check something first? Before the refactor it didn't need to make a cast.

jackkoenig:
@colinschmidt I believe I have addressed your feedback

Issue #459, opened 2017-02-23
Update to CircuitState Annotations


azidar:
I will be reworking in the next couple of days how CircuitState Annotations will be passed.

Currently, the API is to only return the new annotations. This is extremely confusing, because a transformation that does nothing would naturally pass the input CircuitState without modifying it.

This causes all annotations to be duplicated, because we assume all the annotations contained in CircuitState are new, and add them to the existing annotations we saved before calling the transform.

The new proposal is to assume that the annotations that get passed by a transform are all of the annotations. If an annotation is not passed on, it is assumed to be deleted.

I will add a deleted field to all annotations, and this will be marked if a transform failed to pass it along. This should enable passes to error if they see their annotation get deleted (and they weren't expecting it to be).

This is just a PSA for people who are writing transforms, including @shunshou, @colinschmidt, @donggyukim, @jackkoenig, @chick.

I will try to update all transforms that I know about to the new API.

azidar:
Fixed by #464

PR #464, opened 2017-02-23
Addresses #459 Transform/Annotation API change

azidar:
Now, any annotation not propagated by a transform is considered deleted.
A new DeletedAnnotation is added in place of it.


src/main/scala/firrtl/Compiler.scala
```diff
@@ -113,6 +113,7 @@ trait SimpleRun extends LazyLogging {
   def runPasses(circuit: Circuit, passSeq: Seq[Pass]): Circuit =
     passSeq.foldLeft(circuit) { (c: Circuit, pass: Pass) =>
       val x = Utils.time(pass.name) { pass.run(c) }
+      logger.debug(s"** Pass ${pass.name} **")
```
jackkoenig: So this actually **should** be unnecessary. Utils.time prints the name and timing of the pass in info, and info **should** print when debug is enabled but I think there's a bug in the logger.

colinschmidt:
Whats the status on this PR? Just waiting on @jackkoenig ?

src/test/scala/firrtlTests/ReplSeqMemTests.scala
```diff
@@ -7,8 +7,11 @@ import firrtl.ir._
 import firrtl.passes._
 import firrtl.passes.memlib._
 import annotations._
+import logger.Logger
+import logger.LogLevel.Debug
 
 class ReplSeqMemSpec extends SimpleTransformSpec {
+  //Logger.setClassLogLevels(Map(this.getClass.getName -> Debug))
```
jackkoenig: Delete comment and unused imports above

src/main/scala/firrtl/annotations/Annotation.scala
```diff
@@ -26,3 +29,15 @@ final case class Annotation(target: Named, transform: Class[_ <: Transform], val
   def check(from: Named, tos: Seq[Named], which: Annotation): Unit = {}
   def duplicate(n: Named) = Annotation(n, transform, value)
 }
+
+object DeletedAnnotation {
+  def apply(xFormName: String, anno: Annotation): Annotation =
+    Annotation(anno.target, classOf[Transform], s"""DELETED by $xFormName\n${AnnotationUtils.toYaml(anno)}""")
```
jackkoenig: We really need to move to a better serialization library hahaha

src/main/scala/firrtl/Compiler.scala
azidar: Ok, well I'll leave it in for now, and a different fix and change it.

src/main/scala/firrtl/Compiler.scala
```diff
@@ -250,28 +251,31 @@ trait Compiler {
...
-        case None => Nil
-        case Some(p) => p.annotations
+
+      // For each annotation, rename all annotations.
+      val renames = result.renames.getOrElse(RenameMap()).map
+      val remappedAnnotations: Seq[Annotation] = for {
+        anno <- newAnnotations.toSeq
+        newAnno <- anno.update(renames.getOrElse(anno.target, Seq(anno.target)))
+      } yield newAnno
+      logger.debug(s"*** ${xform.name} ***")
```
azidar: Hmm, I guess I can look into it.

src/main/scala/firrtl/annotations/AnnotationUtils.scala
```diff
@@ -3,9 +3,15 @@
...
 package annotations
 
+import net.jcazevedo.moultingyaml._
+import firrtl.annotations.AnnotationYamlProtocol._
+
 import firrtl.ir._
 
 object AnnotationUtils {
+  def toYaml(a: Annotation): String = a.toYaml.prettyPrint
+  def fromYaml(s: String): Annotation = s.parseYaml.convertTo[Annotation]
```
azidar: Probably, i'll move them

azidar: Annotation is a case class, so I can split it into an object and class?

src/main/scala/firrtl/annotations/Annotation.scala
```diff
@@ -26,3 +29,15 @@ final case class Annotation(target: Named, transform: Class[_ <: Transform], val
...
+
+  private val deletedRegex = """(?s)DELETED by ([^\n]*)\n(.*)""".r
+  def unapply(a: Annotation): Option[Tuple2[String, Annotation]] = a match {
```
azidar: done

src/test/scala/firrtlTests/ReplSeqMemTests.scala
azidar: done

src/main/scala/firrtl/annotations/AnnotationUtils.scala
azidar: Splitting it doesn't work, because we rely on the Yaml protocol to serialize a case class. I'll move `toYaml` to the class, but keep the `fromYaml` here.

azidar: Actually, the toYaml method is already implemented, that's how its called on the class, so its better to leave this as is.

shunshou: FYI -- I'm using yaml to serialize/deserialize the annotation value string now too, so I can stick a lot of parameters into an Annotation. (I also think it's wonky that an Annotation *has* to be tied to a Transform).

jackkoenig: With a better serialization library (probably https://github.com/scala/pickling) specifying a transform won't be required. But even now, you don't **have** to use that field, you just can't use `getMyAnnotations` in a given Transform (which kind of makes sense).

shunshou: Hm I see. I made a fake transform that didn't do anything for some global params I used across multiple transforms.

PR #465, opened 2017-02-24
Fix pad, second try

azidar:
Bugfix: pad is max(width,const) during constprop


jackkoenig:
This seems to be causing an error in FixedPoint in chisel3

jackkoenig:
Where are we on a possible test to illustrate the bug?

azidar:
The bug was a mismatch with chisel, the version of master that had been pulled in was too dated. Just pulling from master fixed the error. So... I think a test is unnecessary.

jackkoenig:
By bug I don't mean the break with chisel3, I mean the bug this PR is fixing

Issue #409, opened 2017-01-05
Source locators don't work on multi-line reg


jackkoenig:
The following works
```
  reg r : UInt<32>, clock with : (reset => (reset, UInt(0))) @[MySourceFile.scala 0:0]
```
The following does **not** work:
```
  reg r : UInt<32>, clock with : 
    reset => (reset, UInt(0)) @[MySourceFile.scala 0:0]
```

The problem is in the ANTLR grammar:

```
stmt
  : ...
  | 'reg' id ':' type exp ('with' ':' reset_block)? info?
  | ...

simple_reset0:  'reset' '=>' '(' exp exp ')';

simple_reset
  : simple_reset0
  | '(' simple_reset0 ')'
  ;

reset_block
  : INDENT simple_reset NEWLINE DEDENT
  | '(' +  simple_reset + ')'
  ;
```

Basically, for register definitions where the reset is on a second line, the parser expects a newline so any source locator will cause a parsing error.

PR #468, opened 2017-02-27
Add info on reset block lines to ANTLR grammar

albert-magyar:
Some of the test examples (including Rocket) raise errors when run through more than one instance of FIRRTL compilation. This is due to the parser not recognizing info on reset blocks that gets produced by the FIRRTL emitter as legal FIRRTL. The errors can be reproduced with the command sequence below, which produces `extraneous input` error on the info attached to the reset block.

```
 utils/bin/firrtl -i src/test/resources/integration/RightShiftTester.fir -o RightShiftTester.lo.fir -X low
 utils/bin/firrtl -i RightShiftTester.lo.fir -o RightShiftTester.v -X verilog
```

If the following code produced by the FIRRTL emitter is indeed meant to be legal, this grammar change allows it. If it shouldn't be legal, perhaps an emitter change is in order.

```
  module RightShiftTester : @[:RightShiftTester.fir@29.2]
    ...
    reg T_6 : UInt<2>, clock with :
      reset => (reset, UInt<2>("h0")) @[:RightShiftTester.fir@39.4]
```

Maybe @jackkoenig could refine the change, since this may very well not be the best place to allow the extra info. It seems undesirable to put the `info` before the `NEWLINE DEDENT`, but this way seems to avoid allowing multiple infos per line, which may or may not be the desirable.


jackkoenig:
The issue this is addressing is #409

jackkoenig:
So layout what I think needs to be done:

Change `FIRRTL.g4` diff to:
```
 reset_block
 -	: INDENT simple_reset NEWLINE DEDENT
 +	: INDENT simple_reset info? NEWLINE DEDENT
 - 	| '(' +  simple_reset + ')'
 +      | '(' +  simple_reset + ')' info?
    ;
  
 stmt
   : 'wire' id ':' type info?
-   | 'reg' id ':' type exp ('with' ':' reset_block)? info?
+   | 'reg' id ':' type exp ('with' ':' reset_block)?
```

And make the Visitor find these new infos inside of rule `reset_block`.

I can probably make these changes at some point but I don't have time right this second.

PR #477, opened 2017-03-07
Pass now subclasses Transform

azidar:
- `Pass` now subclasses `Transform`
- Removed all names from Passes (now use class name)
- Added `UnknownForm`, which all passes by default take and return.
- Added `SeqTransform`
- Added `runTransform` method - `execute` now is protected, and is called by `runTransform`. `runTransform` takes a complete `CircuitState` (no renameMap, updated annotations), and returns a complete `CircuitState` (no renameMap, updated annotations), while execute returns an incomplete `CircuitState` (renameMap, non-updated annotations).


src/main/scala/firrtl/passes/Passes.scala
```diff
@@ -12,9 +12,21 @@ import firrtl.PrimOps._
...
 
-trait Pass extends LazyLogging {
-  def name: String
+trait Pass extends Transform {
+  def inputForm: CircuitForm = UnknownForm
+  def outputForm: CircuitForm = UnknownForm
   def run(c: Circuit): Circuit
+  def execute(state: CircuitState): CircuitState = {
+    val result = (state.form, inputForm) match {
+      case (UnknownForm, UnknownForm) => run(state.circuit)
```
jackkoenig: This case would be caught by either of the below two

src/main/scala/firrtl/passes/Passes.scala
```diff
@@ -12,9 +12,21 @@ import firrtl.PrimOps._
 
 import scala.collection.mutable
 
-trait Pass extends LazyLogging {
-  def name: String
+trait Pass extends Transform {
```
jackkoenig: Would appreciate a comment explaining that Pass is intended to be used as kind of a sub-transform which is why it's inputForm and outputForm are UnknownForm

src/main/scala/firrtl/passes/PadWidths.scala
```diff
@@ -9,7 +9,6 @@ import firrtl.Mappers._
 
 // Makes all implicit width extensions and truncations explicit
 object PadWidths extends Pass {
-  def name = "Pad Widths"
```
jackkoenig: Did you want to delete all of these names? You could change them all to override if you liked the names we gave.

src/main/scala/firrtl/Compiler.scala
```diff
@@ -110,61 +110,110 @@ final case object MidForm extends CircuitForm(1)
...
-/** For transformations that are simply a sequence of passes */
-abstract class PassBasedTransform extends Transform with PassBased {
+/** For transformations that are simply a sequence of transforms */
+abstract class SeqTransform extends Transform {
+  def transforms: Seq[Transform]
   def execute(state: CircuitState): CircuitState = {
     require(state.form <= inputForm,
       s"[$name]: Input form must be lower or equal to $inputForm. Got ${state.form}")
-    CircuitState(runPasses(state.circuit), outputForm, state.annotations)
+    val ret = transforms.foldLeft(state) { (in, xform) => xform.runTransform(in) }
```
jackkoenig: `transforms.foldLeft(state) { (in, xform) => xform.runTransform(in) }` shows up a lot, maybe there should be a `runTransforms` like the old `runPasses`?

src/main/scala/firrtl/Compiler.scala
```diff
@@ -110,61 +110,110 @@ final case object MidForm extends CircuitForm(1)
...
+
+  /** Propagate annotations and update their names.
+    *
+    * @param inAnno input AnnotationMap
+    * @param resAnno result AnnotationMap
+    * @param renameOpt result RenameMap
+    * @return the updated annotations
+    */
+  final private def propagateAnnotations(inAnno: Option[AnnotationMap], resAnno: Option[AnnotationMap],
+      renameOpt: Option[RenameMap]): Seq[Annotation] = {
```
jackkoenig: It's not a huge deal but I would prefer:

```scala
final private def propagateAnnotations(inAnno: Option[AnnotationMap],
                                       resAnno: Option[AnnotationMap],
                                       renameOpt: Option[RenameMap]): Seq[Annotation] = {
```
or
```scala
final private def propagateAnnotations(
    inAnno: Option[AnnotationMap], // Note the 4 spaces instead of 2
    resAnno: Option[AnnotationMap],
    renameOpt: Option[RenameMap]): Seq[Annotation] = {
```

src/test/scala/firrtlTests/AnnotationTests.scala
```diff
@@ -17,7 +17,8 @@ import org.scalatest.Matchers
  */
 trait AnnotationSpec extends LowTransformSpec {
   // Dummy transform
-  def transform = new CustomResolveAndCheck(LowForm)
+  def transform = new ResolveAndCheck
+  //def transform = new CustomResolveAndCheck(LowForm)
```
jackkoenig: delete comment

src/test/scala/firrtlTests/ReplSeqMemTests.scala
```diff
@@ -65,6 +65,7 @@ circuit Top :
     val aMap = AnnotationMap(Seq(ReplSeqMemAnnotation("-c:Top:-o:"+confLoc)))
     val res = compileAndEmit(CircuitState(parse(input), ChirrtlForm, Some(aMap)))
     // Check correctness of firrtl
+    println(res.annotations)
```
jackkoenig: Remove println

src/main/scala/firrtl/passes/PadWidths.scala
azidar: Yeah, I want people to just use the name of the class.

PR #481, opened 2017-03-09
infer-rw for exclusive when statements

### LINKED ISSUE #474: InferReadWrite pass fails to infer a single-port RAM for SyncReadMem
Case as attached.

[chisel1.zip](https://github.com/ucb-bar/firrtl/files/815095/chisel1.zip)

@donggyukim

donggyukim:
Should fix #474 .
@edwardcwang can you confirm this works for your memories?


edwardcwang:
Seems to work, thanks!

PR #482, opened 2017-03-09
Tutorial

azidar:
- Added Circuit mappers
- Added commandline option to specify a custom transform
- Added a tutorial pass in a new tutorial package

### DIFF
--- a/src/main/scala/tutorial/AnalyzeCircuit.scala
+++ b/src/main/scala/tutorial/AnalyzeCircuit.scala
@@ -9,9 +9,14 @@ import firrtl.Mappers._
 // Scala's mutable collections
 import scala.collection.mutable
 
+/** Ledger
+  *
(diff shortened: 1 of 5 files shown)


src/main/scala/tutorial/AnalyzeCircuit.scala
```diff
@@ -0,0 +1,49 @@
...
+class AnalyzeCircuit extends Transform {
+  def inputForm = LowForm
+  def outputForm = LowForm
+  def execute(state: CircuitState): CircuitState = {
+    val ledger = new Ledger()
+    state.circuit map walkModule(ledger)
+    println(ledger.serialize)
+    state
+  }
+  def walkModule(ledger: Ledger)(m: DefModule): DefModule = {
```
chick: As a tutorial a few more comments in here would be good.

PR #488, opened 2017-03-11
Upgrade Logging facility

chick:
Make thread-safe
Make logging by package name work
Use caching of class names to level for performance
Make some tests to show this working


src/main/scala/logger/Logger.scala
```diff
@@ -36,45 +38,103 @@ trait LazyLogging {
...
+        }
+      }
+      state.classToLevelCache(className) = matchPathToFindLevel(packageNameList)
+      state.classToLevelCache(className)
+    })
+    levelForThisClassName >= level
+  }
+
+  //scalastyle:off regex
   def showMessage(level: LogLevel.Value, className: String, message: => String): Unit = {
```
azidar: Comment

chick:
@azidar Took longer than anticipated, I hope this works for you now

src/main/scala/logger/Logger.scala
```diff
@@ -36,45 +38,103 @@ trait LazyLogging {
   val logger = new Logger(this.getClass.getName)
 }
 
+class LoggerState {
```
azidar: and make private please

azidar:
Looks like the default logger mode isn't set right - the travis build fails because it ran out of space to hold the log, as all tests are printing a lot.

chick:
@azidar I have pushed another fix, we should discuss in morning

chick:
@azidar I think the current commit of this PR addresses the issues we went over.  The chisel3 tutorial branch uses the latest change to show a straightforward way of capturing logging.

shunshou:
Does  commonOptions = CommonOptions(globalLogLevel = LogLevel.Debug)
work?

It seems like I'm not getting any useful logger info -- even tho I was able to in the past?

shunshou:
@chick -- seems like chisel3.driver passes logger info down correctly, but dsptools.driver (and I assume iotesters.driver) loses the logger option...

chick:
@shunshou I don't see where these would be lost when I take a quick look at 
dspTools.Driver -> iotestesters.Driver -> chisel3.Driver.  But I might be missing something
BTW, where does the code ```commonOptions = CommonOptions(globalLogLevel = LogLevel.Debug)``` come from.

chick:
@shunshou I think I understand the problem.  The Driver.execute methods of dsptools and iotesters and chisel3 will need a change to have them automatically set the logger options.  It's simple, it just needs to wrap the execute method's code in a Logger.makeScope.  If you think this might be the problem I can create PR's to introduce the compatibility change

PR #490, opened 2017-03-11
Eliminate useless ports

terpstra:
This adds a pass which prunes disconnected input and output ports. These were causing us problems with achieving coverage.

Please pay special attention and review the DeadCodeElimination changes. This PR makes the dead code elimination significantly more aggressive, which may have corner cases I didn't exercise in Rocket.

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -748,7 +748,8 @@ class VerilogEmitter extends Transform with PassBased with Emitter {
     passes.VerilogModulusCleanup,
     passes.VerilogWrap,
     passes.VerilogRename,
-    passes.VerilogPrep)
+    passes.VerilogPrep,
+    passes.DeadCodeElimination)
 
   def emit(state: CircuitState, writer: Writer): Unit = {
     writer.write(preamble)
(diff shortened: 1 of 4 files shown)


terpstra:
@jackkoenig Are you the correct person to point at this pass?

terpstra:
Hmm. I see there are unit tests which don't use their ports, but depend on the result having exactly the specified ports. I could resolve these by inserting a body that makes the ports actually used. Or this port removal pass could be made into a command-line option. What's best?

azidar:
I think aggressive dce and port elimination should be an optional feature. Often, new users of Chisel/Firrtl get extremely confused when they end up with an empty module.

To do this, follow the example of InlineInstances (e.g. extend Transform, not Pass). Ideally, I'd like to see the following annotations:

- Turn on the pass (like InlineInstancesAnnotation)
- Preserve a wire/port/node/etc. <- this allows future transforms, libraries, etc. to run your pass, even if they have an unused port they want to hook up later.

In order for the PR to be accepted, you'll need to write multiple tests in `src/test/scala/firrtlTests`, as well as write some more comments in your transform explaining the major functions.

shunshou:
@azidar idk how long it'll take to approve this PR. 

1) It'd be useful for me since who knows when RawModule will ever be merged into chisel. 
2) Even if I don't get something that eliminates unconnected ports, it'd be useful at least to see which ports are actually unconnected...

azidar:
@terpstra @shunshou I'll go ahead and take care of this PR.

azidar:
Superceded by #549

Issue #489, opened 2017-03-11
Internal Error: From dsptools ResetInverterSpec


chick:
In ucb-bar/dsptools on branch signBit-div2-mul2
The test file ResetNSpec crashes with 
```
[info] [0.004] Elaborating design...
[info] [0.085] Done elaborating.

Internal Error! Please file an issue at https://github.com/ucb-bar/firrtl/issues
firrtl.FIRRTLException: Internal Error! Please file an issue at https://github.com/ucb-bar/firrtl/issues
	at firrtl.Utils$.error(Utils.scala:363)
	at firrtl.Utils$.throwInternalError(Utils.scala:132)
	at firrtl.Driver$.execute(Driver.scala:183)
	at chisel3.Driver$.execute(Driver.scala:161)
	at dsptools.ResetNSpec$$anonfun$1.apply$mcV$sp(ResetInverterSpec.scala:26)
	at dsptools.ResetNSpec$$anonfun$1.apply(ResetInverterSpec.scala:22)
	at dsptools.ResetNSpec$$anonfun$1.apply(ResetInverterSpec.scala:22)
	at org.scalatest.Transformer$$anonfun$apply$1.apply$mcV$sp(Transformer.scala:22)
	at org.scalatest.OutcomeOf$class.outcomeOf(OutcomeOf.scala:85)
	at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
	at org.scalatest.Transformer.apply(Transformer.scala:22)
	at org.scalatest.Transformer.apply(Transformer.scala:20)
	at org.scalatest.FreeSpecLike$$anon$1.apply(FreeSpecLike.scala:373)
	at org.scalatest.Suite$class.withFixture(Suite.scala:1122)
	at org.scalatest.FreeSpec.withFixture(FreeSpec.scala:1740)
	at org.scalatest.FreeSpecLike$class.invokeWithFixture$1(FreeSpecLike.scala:370)
	at org.scalatest.FreeSpecLike$$anonfun$runTest$1.apply(FreeSpecLike.scala:382)
	at org.scalatest.FreeSpecLike$$anonfun$runTest$1.apply(FreeSpecLike.scala:382)
	at org.scalatest.SuperEngine.runTestImpl(Engine.scala:306)
	at org.scalatest.FreeSpecLike$class.runTest(FreeSpecLike.scala:382)
	at org.scalatest.FreeSpec.runTest(FreeSpec.scala:1740)
[...]

PR #497, opened 2017-03-16
Give better error message if missing emittedcircuit

azidar:
Addresses #489


src/main/scala/firrtl/Compiler.scala
```diff
@@ -55,6 +56,13 @@ case class CircuitState(
     })
     emittedOpt.getOrElse(Seq.empty)
   }
+  def deletedAnnotations: Seq[Annotation] = {
+    val deletedOpt = annotations map (_.annotations collect {
+      case DeletedAnnotation(xformName, anno) =>
+        DeletedAnnotation(xformName, anno)
```
chick: Nit, isn't the idiom for this ```case deletedAnno : DeletedAnnotation => deletedAnno```
Save object creataion

azidar: I can't match on DeletedAnnotation because it isn't a class - its an object.

chick:
Got it.  Thought I was looking at a case class, not an unapply

Issue #498, opened 2017-03-20
Firrtl Match Error Shlw


shunshou:
I can't get my design through Firrtl :(

Looks like the width padding is adding shlw PrimOps that aren't matched properly? I'm super confused as to why this is a problem for me but hasn't been seen by anyone else.

```
[info] - should read in ADC inputs *** FAILED ***
[info]   scala.MatchError: shlw (of class firrtl.Shlw$)
[info]   at firrtl.PrimOps$.set_primop_type(PrimOps.scala:129)
[info]   at firrtl.passes.InferTypes$.firrtl$passes$InferTypes$$infer_types_e$1(InferTypes.scala:32)
[info]   at firrtl.passes.InferTypes$$anonfun$firrtl$passes$InferTypes$$infer_types_s$1$7.apply(InferTypes.scala:60)
[info]   at firrtl.passes.InferTypes$$anonfun$firrtl$passes$InferTypes$$infer_types_s$1$7.apply(InferTypes.scala:60)
[info]   at firrtl.ir.Connect.mapExpr(IR.scala:240)
[info]   at firrtl.Mappers$StmtMagnet$$anon$2.map(Mappers.scala:19)
[info]   at firrtl.Mappers$StmtMap$.map$extension(Mappers.scala:30)
[info]   at firrtl.passes.InferTypes$.firrtl$passes$InferTypes$$infer_types_s$1(InferTypes.scala:60)
[info]   at firrtl.passes.InferTypes$$anonfun$firrtl$passes$InferTypes$$infer_types_s$1$5.apply(InferTypes.scala:60)
[info]   at firrtl.passes.InferTypes$$anonfun$firrtl$passes$InferTypes$$infer_types_s$1$5.apply(InferTypes.scala:60)
```

shunshou:
After further debugging:

My ClkSrcTransform (LowForm) runs an InferTypes pass before starting. Does that mean InferTypes doesn't work in LowFirrtl? I never caught it before, b/c I wasn't using it with designs that required padding widths...

```
Form: LowForm
======== Finished Transform MiddleFirrtlToLowFirrtl ========

======== Starting Transform LowFirrtlOptimization ========
-------- Starting Pass Remove ValidIfs --------
Time: 19.4 ms
-------- Finished Pass Remove ValidIfs --------
-------- Starting Pass Constant Propagation --------
Time: 48.1 ms
-------- Finished Pass Constant Propagation --------
-------- Starting Pass Pad Widths --------
[...]

PR #500, opened 2017-03-20
Fix 498

azidar:
Reworked Dshlw and Shlw to stay in Verilog emitter.

Moved SplitExpression to Veirlog Emitter. This may have a negative effect on CSE.

### DIFF
--- a/src/test/scala/firrtlTests/FirrtlSpec.scala
+++ b/src/test/scala/firrtlTests/FirrtlSpec.scala
@@ -95,7 +95,6 @@ trait FirrtlMatchers extends Matchers {
   def executeTest(input: String, expected: Seq[String], compiler: Compiler) = {
     val finalState = compiler.compileAndEmit(CircuitState(parse(input), ChirrtlForm))
     val lines = finalState.getEmittedCircuit.value split "\n" map normalized
-    println(finalState.getEmittedCircuit.value)
     for (e <- expected) {
       lines should contain (e)
     }
--- a/src/test/scala/firrtlTests/FirrtlSpec.scala
+++ b/src/test/scala/firrtlTests/FirrtlSpec.scala
@@ -95,6 +95,7 @@ trait FirrtlMatchers extends Matchers {
   def executeTest(input: String, expected: Seq[String], compiler: Compiler) = {
     val finalState = compiler.compileAndEmit(CircuitState(parse(input), ChirrtlForm))
     val lines = finalState.getEmittedCircuit.value split "\n" map normalized
+    println(finalState.getEmittedCircuit.value)
     for (e <- expected) {
       lines should contain (e)
     }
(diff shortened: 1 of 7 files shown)


src/test/scala/firrtlTests/FirrtlSpec.scala
```diff
@@ -95,6 +95,7 @@ trait FirrtlMatchers extends Matchers {
   def executeTest(input: String, expected: Seq[String], compiler: Compiler) = {
     val finalState = compiler.compileAndEmit(CircuitState(parse(input), ChirrtlForm))
     val lines = finalState.getEmittedCircuit.value split "\n" map normalized
+    println(finalState.getEmittedCircuit.value)
```
jackkoenig: No println

azidar:
* master with splitexp before cse - 47866 eliminated
* current PR with splitexp after cse - 37746 eliminated
* include split/cse/dce in Verilog emitter, no splitexp in optimization - 47573 eliminated (9827 in second cse/dce run)

I think this is acceptable, to include more optimization in the VerilogEmitter. Alternatively, we can change SplitExpression to not split tail(add(...)), tail(sub(...)), bits(shl(...)), and bits(dshl(...)).

jackkoenig:
So we definitely don't want a regression on how much gets removed.

To make sure I understand why we're seeing what we're seeing:
1. SplitExpressions helps CSE
1. Moving it to Verilog Emitter causes CSE to get less done in the optimization transform
1. We need SplitExpressions to run after shift wraps are inserted so that we don't emit nested expression.

Instead of running CSE again in the VerilogEmitter (which I think is fine if that's the only way to keep QOR, but should be avoided if possible) could we:
1. Run SplitExpressions both before CSE and in the VerilogEmitter, or
1. Be careful to not introduce nested expressions in the VerilogEmitter?

azidar:
No, the problem is that split expression is splitting up:
```
node x = tail(add(a, b), 1)
```
to
```
node y = add(a, b)
node x = tail(y, 1)
```
And so VerilogWrap actually does nothing right now.

azidar:
@jackkoenig I believe we can kill this PR, as the new phase stuff should make this no longer needed?

jackkoenig:
I think maybe this PR is outdated but I don't think the Phase stuff fixes #498, we still need to stop inserting Dshlw in SplitExpressions, because we still want SplitExpressions in LowFirrtlOptimization

seldridge:
Yeah, I was going to fix #498 in stage/phase, but didn't. It seemed like that was already enough of an omnibus PR... It can be updated if needed, though.

jackkoenig:
Thanks for the heads up @seldridge. I think we should close this PR

albert-magyar:
Closing because Jack said it should be closed, and since the future fix to #498 will now need to be Dependency API-compliant.

PR #503, opened 2017-03-21
Add TargetDirAnnotation to give transforms access

jackkoenig:
Also add GlobalCircuitAnnotation for creating similar annotations

* Thoughts about the "GlobalCircuitAnnotation" API? It's just a helper for annotations that are globally available and hold a single String. 
* Given the concept of "global annotations" or even annotations that may target a superclass transform, should `getMyAnnotations` return these? Or should `getMyAnnotations` continue to just return ones that target a given transform directly? We could add a new function like `getGlobalAnnotations` for ones that target `Transform` itself.
* Note that `Driver.execute` adds the annotation. Think the implementation is okay? Some annotations like this only make sense in the context of using execution options, which means we're going to be shifting people toward using them. As we had discussed @azidar, if we want to do any cleanup here it's only going to become higher priority.

### DIFF
--- a/src/main/scala/firrtl/Driver.scala
+++ b/src/main/scala/firrtl/Driver.scala
@@ -164,8 +164,13 @@ object Driver {
     // Output Annotations
     val outputAnnos = firrtlConfig.getEmitterAnnos(optionsManager)
 
+    // Should these and outputAnnos be moved to loadAnnotations?
+    val globalAnnos = Seq(TargetDirAnnotation(optionsManager.targetDirName))
+
     val finalState = firrtlConfig.compiler.compile(
-      CircuitState(parsedInput, ChirrtlForm, Some(AnnotationMap(firrtlConfig.annotations ++ outputAnnos))),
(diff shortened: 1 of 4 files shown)


shunshou:
I'm super ready for something like this!

But in general, there should be global annotations for inputs and outputs, I think. 

For example, I have a bunch of tapeout passes that take in configuration files. Originally, I had you specify configuration files per pass, but I think a more reasonable thing would be to specify a "tapeout config" directory to populate config files. Like clking config file should be tapeout/clocking.config; pads config file should be tapeout/pads.config, etc.

azidar:
> there should be global annotations for inputs and outputs
@shunshou, what do you mean inputs and outputs?

I think this is good, and I like adding a new function, not changing getMyAnnotations.

I have mixed feelings about whether outputAnnos and globalAnnos should be moved to loadAnnotations. I feel like we should think about what _are_ the global annotations we think we will likely see, and then go from there.

shunshou:
Input config directories, output dump directory

Issue #502, opened 2017-03-21
Java heap space OutOfMemoryError in ZeroWidth pass.


ccelio:
**Output:**
````

[info] [11.192] Done elaborating.
[success] Total time: 121 s, completed Mar 21, 2017 2:02:53 PM
mkdir -p /scratch/celio/firrtl-chip-deleteme/emulator/generated-src/
java -Xmx2G -Xss8M -XX:MaxPermSize=256M -cp /scratch/celio/firrtl-chip-deleteme/firrtl/utils/bin/firrtl.jar firrtl.Driver -i /scratch/celio/firrtl-chip-deleteme/emulator/generated-src/rocketchip.BOOMConfig.fir -o /scratch/celio/firrtl-chip-deleteme/emulator/generated-src/rocketchip.BOOMConfig.v -X verilog
Exception in thread "main" java.lang.OutOfMemoryError: Java heap space
	at scala.collection.mutable.ResizableArray$class.ensureSize(ResizableArray.scala:103)
	at scala.collection.mutable.ArrayBuffer.ensureSize(ArrayBuffer.scala:48)
	at scala.collection.mutable.ArrayBuffer.$plus$eq(ArrayBuffer.scala:84)
	at scala.collection.mutable.ArrayBuffer.$plus$eq(ArrayBuffer.scala:48)
	at scala.collection.TraversableLike$$anonfun$collect$1.apply(TraversableLike.scala:282)
	at scala.collection.TraversableLike$$anonfun$collect$1.apply(TraversableLike.scala:282)
	at scala.PartialFunction$$anonfun$runWith$1.apply(PartialFunction.scala:142)
	at scala.PartialFunction$$anonfun$runWith$1.apply(PartialFunction.scala:140)
	at scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:59)
	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:48)
	at scala.collection.TraversableLike$class.collect(TraversableLike.scala:282)
	at scala.collection.AbstractTraversable.collect(Traversable.scala:104)
	at firrtl.passes.ZeroWidth$.firrtl$passes$ZeroWidth$$removeZero(ZeroWidth.scala:19)
	at firrtl.passes.ZeroWidth$.firrtl$passes$ZeroWidth$$removeZero(ZeroWidth.scala:25)
[...]

PR #505, opened 2017-03-21
Fixed zero width perf bug #502

azidar:
Now remove DefNodes of zero width
Don't deeply walk nodes (was the source of the bug)


jackkoenig:
Was walking the types of DefNodes the problem?

azidar:
Yup, the types of DefNode values will be fixed with the InferTypes run.

PR #506, opened 2017-03-22
Improve dce

azidar:
Started from @terpstra's PR #490, but only the first commit (improvements to DCE).
Added tests a small bugfix.

### DIFF
--- a/src/main/scala/firrtl/passes/DeadCodeElimination.scala
+++ b/src/main/scala/firrtl/passes/DeadCodeElimination.scala
@@ -12,8 +12,8 @@ import annotation.tailrec
 object DeadCodeElimination extends Pass {
   def name = "Dead Code Elimination"
 
-  private def dceOnce(s: Statement): (Statement, Long) = {
-    val referenced = collection.mutable.HashSet[String]()
+  private def dceOnce(ports: Seq[Port], s: Statement): (Statement, Long) = {
+    val referenced = collection.mutable.HashSet[String](ports.map(_.name):_*)
     var nEliminated = 0L
 
     def checkExpressionUse(e: Expression): Expression = {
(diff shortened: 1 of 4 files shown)


src/main/scala/firrtl/passes/DeadCodeElimination.scala
```diff
@@ -34,22 +34,22 @@ object DeadCodeElimination extends Pass {
       out
     }
 
-    def maybeEliminate(x: Statement, name: String) =
-      if (referenced(name)) x
+    def maybeEliminate(x: Statement, name: String, exception: String): Statement =
+      if (referenced(name) && name != exception) x
```
terpstra: Why is this necessary? The effectively neuters the DefRegister maybeEliminate. It now never happens.

terpstra:
Is your intent with the 'exception' to detect cycles of logic that are otherwise unreferenced? If so, you should just use a mark and sweep algorithm after discovering edges.

azidar:
No, it is just to catch the case where a register is initialized to itself. Otherwise, these registers will never be eliminated.

terpstra:
Yes. That's exactly what I mean: a logic loop unreferenced anywhere else. If that is what you wanted to detect, then your algorithm is just a bad version of the mark-sweep algorithm, which is what we should use. The only reason I modified the pass as I did was that I wanted to keep my changes minimally invasive. However, the right way to do DCE if you want to detect cases like the one you are addressing is to do a complete mark-sweep pass. It will also be faster because it only needs to explore the netlist once.

azidar:
In Firrtl, the way we describe an uninitialized register is by initializing it to itself. Thus, this specific use case is extremely common, and is very easy to check for. Otherwise, no uninitialized register will ever be DCE'd by this pass. This is also in the spirit of keeping things minimally invasive.

Perhaps "bugfix" was the wrong word, for that I apologize.

I agree with you that the more general case is a mark and sweep algorithm, for any logic loop, and this would catch this (as well as any other dead code this pass collects).

jackkoenig:
Replaced by #549

Issue #511, opened 2017-03-28
Stack Overflow in Check Combinational Loops


jackkoenig:
You can see Firrtl that will exercise the problem [here](https://gist.github.com/jackkoenig/d8a0854701f75799968de68994d9c116) (along with Chisel to help generate other sized test cases).

Long chains of logic seem to be problematic

aswaterman:
I guess the recursion needs to be replaced with iteration on a heap-allocated stack.

albert-magyar:
Interesting, I had an iterative version that I ended up not using because Chisel2 used a recursive implementation. I'll fix this today.

jackkoenig:
Fixed by #513

PR #513, opened 2017-03-29
Change findSCCs to iterative implementation

albert-magyar:
This is to address #511


src/main/scala/firrtl/graph/DiGraph.scala
```diff
@@ -203,36 +203,59 @@ class DiGraph[T] (val edges: Map[T, Set[T]]) extends DiGraphLike[T] {
...
+        val frame = callStack.top
+        val v = frame.v
+        if (frame.childCall.isEmpty) {
+          indices(v) = counter
+          lowlinks(v) = counter
+          counter = counter + 1
+          stack.push(v)
+          onstack += v
+        }
+        for (w <- frame.childCall) {
```
jackkoenig: Unless I'm very much mistaken, these two branches are mutually exclusive. Perhaps this should be made more explicit with just a match?

albert-magyar: Good call

Issue #517, opened 2017-03-30
Error in Check Combinational Loops


jackkoenig:
The (not quite minimized) example: https://gist.github.com/jackkoenig/9cfee1e5a0f702df03b950b6401bc9c4

Gives the error:
```
[error] (run-main-20) firrtl.graph.PathNotFoundException: Unreachable node
```

jackkoenig:
Fixed via #518

PR #518, opened 2017-03-31
Find a single cycle from potentially many in a combinational SCC

albert-magyar:
This addresses #517 

The underlying issue is that an SCC may contain _multiple_ loops (that sometimes never form a single simple cycle), in which case the SCC algorithm will return them in an order that may not be traversable as a single path. Since the combinational loop detection pass does a lot of post-processing to pretty-print the loops, the loop-reporting steps failed under these conditions.

An extra step was added to extract a single simple cycle to report in the exception.

### DIFF
--- a/src/main/scala/firrtl/passes/CheckCombLoops.scala
+++ b/src/main/scala/firrtl/passes/CheckCombLoops.scala
@@ -104,29 +104,50 @@ object CheckCombLoops extends Pass {
   }
 
   /*
-   * Recover the full path of the cycle. Since cycles may pass through
-   * simplified instances, the hierarchy that the path passes through
-   * must be recursively recovered.
+   * Recover the full path from a path passing through simplified
+   * instances. Since edges may pass through simplified instances, the
+   * hierarchy that the path passes through must be recursively
+   * recovered.
    */
-  private def recoverCycle(
+  private def expandInstancePaths(
     m: String,
     moduleGraphs: mutable.Map[String,DiGraph[LogicNode]],
     moduleDeps: Map[String, Map[String,String]], 
     prefix: Seq[String],
-    cycle: Seq[LogicNode]): Seq[String] = {
+    path: Seq[LogicNode]): Seq[String] = {
     def absNodeName(prefix: Seq[String], n: LogicNode) =
       (prefix ++ n.inst ++ n.memport :+ n.name).mkString(".")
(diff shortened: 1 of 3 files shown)


src/main/scala/firrtl/passes/CheckCombLoops.scala
```diff
@@ -104,29 +104,42 @@ object CheckCombLoops extends Pass {
...
+        expandInstancePaths(child,moduleGraphs,moduleDeps,newprefix,subpath)
       } else {
         Seq(absNodeName(prefix,a))
       }
     }
-    cycNodes.flatten :+ absNodeName(prefix, cycle.last)
+    pathNodes.flatten :+ absNodeName(prefix, path.last)
+  }
+
+  private def findCycleInSCC[T](sccGraph: DiGraph[T]): Seq[T] = {
```
jackkoenig: A one-liner describing what this does would be great, also mentioning that you've already pruned the graph, I was confused by the fact that you only visit the head edge of a given vertex.

src/main/scala/firrtl/passes/CheckCombLoops.scala
```diff
@@ -169,8 +182,13 @@ object CheckCombLoops extends Pass {
       moduleGraphs(m.name) = DiGraph(internalDeps)
       simplifiedModuleGraphs(m.name) = moduleGraphs(m.name).simplify((m.ports map { p => LogicNode(p.name) }).toSet)
       for (scc <- moduleGraphs(m.name).findSCCs.filter(_.length > 1)) {
-        val cycle = recoverCycle(m.name,moduleGraphs,moduleDeps,Seq(m.name),scc :+ scc.head)
-        errors.append(new CombLoopException(m.info, m.name, cycle))
+        val sccSubgraph = moduleGraphs(m.name).subgraph(scc.toSet)
+        val cycle = findCycleInSCC(sccSubgraph)
+        (cycle zip cycle.tail) foreach {
```
jackkoenig: In Scala style, It used to be infix notation was reserved for purely functional things, foreach should always use a `.`. 

**But Scala changed it on me**. Now `.` notation is **always** recommended.

jackkoenig: You should change the foreach but by god they will take infix `zip` from my **cold dead hands**

PR #520, opened 2017-04-01
Fix typo in ExecutionOptionsManager comment

colinschmidt:
Now you should be able to copy paste the example code :)


colinschmidt:
I think this isn't the only typo will update after I get something working.

PR #522, opened 2017-04-03
Change implicit value classes to prefix with _

azidar:
Fixes bug where arbitrary expressions, if called with .expr, return the
same expression.


jackkoenig:
Implicit conversions ftw

PR #525, opened 2017-04-03
Build Java 1.7-compatible libraries.

ucbjrl:
Ensure release builds will work with Java 1.7. This could be restricted to the release branch, but since we don't require (currently) any 1.8 features, we shouldn't let them sneak in unannounced.
We could add this to all BIG4 projects but currently firrtl is the only once that generates Java code.


ucbjrl:
Superseded by #639

PR #528, opened 2017-04-05
Make sure ReplaceMemTransform sets its outputform

colinschmidt:
This crashes barstools otherwise, when trying to compare to unknown form.


colinschmidt:
@azidar I believe we talked about this a while ago and this was the right fix?

jackkoenig:
@colinschmidt is this still relevant?

colinschmidt:
I honestly don't know but we can just close it for now and if I run into this again I'll reopen.

Issue #547, opened 2017-04-25
Verilog Mem Delays Pass Emits Firrtl with Use Before Declaration


jackkoenig:
Here's a relatively simple test case:
```
circuit test :
  module test :
    input clock : Clock
    input reset : UInt<1>
    input in : UInt<32>
    output out : UInt<32>
  
    mem ram :
      data-type => UInt<32>
      depth => 2
      read-latency => 0
      write-latency => 1
      reader => r
      writer => w
      read-under-write => undefined

    wire value_1: UInt<1>    ; Watch value_1
    ram.r.addr <= value_1

    ram.r.en <= UInt<1>("h1")
    ram.r.clk <= clock
    ram.w.addr <= bits(in, 1, 0)
    ram.w.en <= bits(in, 0, 0)
    ram.w.clk <= clock
    ram.w.data <= in
    ram.w.mask <= UInt(1)

    out <= ram.r.data
    value_1 <= UInt(1)
```

After VerilogMemDelays the ordering gets messed up to be the following:
Specifically look at value_1.

```
circuit test : 
  module test : 
    input clock : Clock
    input reset : UInt<1> 
    input in : UInt<32> 
    output out : UInt<32> 
  
    mem ram : 
      data-type => UInt<32>
      depth => 2
      read-latency => 0
      write-latency => 1
      reader => r
      writer => w
      read-under-write => undefined
    node _GEN_0 = UInt<1>("h1")
    node _GEN_1 = value_1     ; value_1 is used here!!!
    ram.r.clk <= clock
    ram.r.en <= _GEN_0
    ram.r.addr <= _GEN_1
    node _GEN_2 = bits(in, 0, 0)
    node _GEN_3 = UInt<1>("h1")
    node _GEN_4 = bits(bits(in, 1, 0), 0, 0)
    node _GEN_5 = in
    ram.w.clk <= clock
    ram.w.en <= _GEN_2
    ram.w.mask <= _GEN_3
    ram.w.addr <= _GEN_4
    ram.w.data <= _GEN_5
    wire value_1 : UInt<1>    ; value_1 defined here now
    out <= ram.r.data

    value_1 <= UInt<1>("h1")
```

PR #548, opened 2017-04-25
Run CheckHighForm after all non-emitter transforms in firrtl tests

jackkoenig:
This should fail for now. It exposes a few bugs, one of which is #547.


jackkoenig:
This will probably fail because `Shlw` no longer exists, but `Dshlw` still does

PR #549, opened 2017-04-26
New dce

jackkoenig:
AT.LONG.LAST.

Aggressive DCE with a DontTouchAnnotation.

Current limitations:
* DontTouchAnnotation ignored by other optimization passes
* DontTouchAnnotation only works on leaf nodes, should LowerTypes also "lower" annotations of aggregates?
* Performance isn't amazing but as far as I can tell it's linear and it's faster than some other passes


src/main/scala/firrtl/transforms/DeadCodeElimination.scala
```diff
@@ -0,0 +1,244 @@
...
+    val depGraph = new MutableDiGraph[LogicNode]
+    c.modules.foreach {
+      case mod: Module => setupDepGraph(depGraph, instMaps(mod.name))(mod)
+      case ext: ExtModule =>
+        // Connect all inputs to all outputs
+        val node = LogicNode(ext.name, ext.name)
+        val ports = ext.ports.groupBy(_.direction)
+        depGraph.addVertex(node)
+        ports.get(Output).foreach(_.foreach(output => depGraph.addEdge(LogicNode(ext.name, output.name), node)))
+        ports.get(Input).foreach(_.foreach(input => depGraph.addEdge(node, LogicNode(ext.name, input.name))))
```
jackkoenig: Currently, this treats all external modules as a single node that depends on all of its inputs and is depended on by all of its outputs. This means that they can be dead code eliminated.

I'm thinking instead inputs to extmodules should all be treated as sinks of the circuit?
Alternatively, we could require the user to annotate which extmodules they don't want removed, I'm leaning toward the former though.

aswaterman: Some ExtModules should be preserved even if their outputs are unused (e.g., monitors) and some should be deleted (e.g., RAMs).  So, if don't-prune is the default, we'd need to add a DoTouch annotation.

Superficially, it seems better to prune by default and use the existing DontTouch annotation to treat them as sinks.

jackkoenig:
This passes the tests now and I'm running the rocket-chip Travis regressions on it: https://travis-ci.org/ucb-bar/rocket-chip/builds/228170902

jackkoenig:
Alright so the rocket-chip regressions passed. I think this is good to go.

src/main/scala/firrtl/transforms/DeadCodeElimination.scala
```diff
@@ -0,0 +1,251 @@
...
+        modulesxMap += m.name -> m
+      }
+    }
+
+    // Preserve original module order
+    c.copy(modules = c.modules.flatMap(m => modulesxMap.get(m.name)))
+  }
+
+  def execute(state: CircuitState): CircuitState = {
+		val dontTouches: Seq[LogicNode] = state.annotations match {
```
azidar: weird indentation

src/main/scala/firrtl/transforms/DeadCodeElimination.scala
```diff
@@ -0,0 +1,251 @@
...
+    depGraph
+  }
+
+  private def deleteDeadCode(instMap: collection.Map[String, String],
+                             deadNodes: Set[LogicNode],
+                             moduleMap: collection.Map[String, DefModule])
+                            (mod: DefModule): Option[DefModule] = {
+    // Gets all dependencies and constructs LogicNodes from them
+    // TODO this is a duplicate from setupDepGraph, remove by improving how we lookup
+    def getDeps(expr: Expression): Seq[LogicNode] =
```
azidar: Can you merge with previous getDeps, and give instMap as an argument?

src/main/scala/firrtl/transforms/DeadCodeElimination.scala
```diff
@@ -0,0 +1,251 @@
...
+          val (inst, tail) = Utils.splitRef(e)
+          LogicNode(instMap(inst.name), tail)
+        } else {
+          LogicNode(mod.name, e)
+        }
+      }
+
+    var emptyBody = true // TODO is there a better way to do this?
+
+    // TODO Delete unused writers from DefMemory???
```
azidar: Should probably be moved to DefMemory in setupDepGraph - also, I don't think you can dce write ports without optimizing the entire memory away, but you can dce read ports.

src/main/scala/firrtl/transforms/DeadCodeElimination.scala
```diff
@@ -0,0 +1,251 @@
...
+      extractRefs(expr).map { e =>
+        if (kind(e) == InstanceKind) {
+          val (inst, tail) = Utils.splitRef(e)
+          LogicNode(instMap(inst.name), tail)
+        } else {
+          LogicNode(mod.name, e)
+        }
+      }
+
+    var emptyBody = true // TODO is there a better way to do this?
```
azidar: No, I think this is the best way IMO.

src/main/scala/firrtl/transforms/DeadCodeElimination.scala
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    val portsx = mod.ports.filterNot(p => deadNodes.contains(LogicNode(mod.name, p.name)))
+
+    mod match {
+      case Module(info, name, _, body) =>
+        val bodyx = onStmt(body)
+        if (emptyBody && portsx.isEmpty) None else Some(Module(info, name, portsx, bodyx))
+      case ext: ExtModule =>
+        if (portsx.isEmpty) None else Some(ext.copy(ports = portsx))
```
azidar: if ports != portsx, isn't that an error? Maybe check that condition?

donggyukim:
BTW, I tried this branch, and the top-level pins were also pruned, which makes writing testbenches difficult. Is it intended?

jackkoenig:
Which top-level pins are pruned? If outputs are being pruned then it is definitely a bug. If inputs are being pruned, then they must not be doing anything. Of course, since sometimes you may want to design your top to fit in some interface where some ports are unused. I'll change it to not prune top-level inputs.

donggyukim:
Some fields in rocketchip AXI busses were pruned. I think they were all inputs.

jackkoenig:
Yeah that's pretty undesirable, I'll make that not happen

jackkoenig:
@donggyukim can you try this?

jackkoenig:
Alright, I have added OptimizableExtModuleAnnotation and stopped DCE-ing external modules unless they have the annotation. I have stopped DCEing top-level inputs. I have address Adam's reviewer feedback (and I think the ScalaDoc comment is pretty good, at least by codebase standards 😆). I have added support for annotation renaming although due to a bug in annotation renaming in LowerTypes, the test is currently on ignore, but I inspected the results in that test and DCE is doing the right thing.

jackkoenig:
Once these tests pass, I'm going to merge.

src/main/scala/firrtl/transforms/DeadCodeElimination.scala
```diff
@@ -0,0 +1,251 @@
...
+import firrtl.Mappers._
+import firrtl.WrappedExpression._
+import firrtl.Utils.{throwInternalError, toWrappedExpression, kind}
+import firrtl.MemoizedHash._
+import wiring.WiringUtils.getChildrenMap
+
+import collection.mutable
+import java.io.{File, FileWriter}
+
+class DeadCodeElimination extends Transform {
```
jackkoenig: done

src/main/scala/firrtl/transforms/DeadCodeElimination.scala
```diff
@@ -0,0 +1,251 @@
...
+        val node = LogicNode(mod.name, name)
+        depGraph.addVertex(node)
+        getDeps(value).foreach(ref => depGraph.addEdge(node, ref))
+      case DefWire(_, name, _) =>
+        depGraph.addVertex(LogicNode(mod.name, name))
+      case mem: DefMemory =>
+        // Treat DefMems as a node with outputs depending on the node and node depending on inputs
+				// From perpsective of the module or instance, MALE expressions are inputs, FEMALE are outputs
+        val memRef = WRef(mem.name, MemPortUtils.memType(mem), ExpKind, FEMALE)
+				val exprs = Utils.create_exps(memRef).groupBy(Utils.gender(_))
```
jackkoenig: Sometimes when I turn paste on to copy paste something in I forget to turn it off and weird stuff like this happens. Good catch.

src/main/scala/firrtl/transforms/DeadCodeElimination.scala
```diff
@@ -0,0 +1,251 @@
...
+				val sources = exprs.getOrElse(MALE, List.empty).flatMap(getDeps(_))
+				val sinks = exprs.getOrElse(FEMALE, List.empty).flatMap(getDeps(_))
```
jackkoenig: Yeah I agree, but saving this is an exercise for future me

src/main/scala/firrtl/transforms/DeadCodeElimination.scala
```diff
@@ -0,0 +1,251 @@
...
+        val memNode = getDeps(memRef) match { case Seq(node) => node }
+        depGraph.addVertex(memNode)
+        sinks.foreach(sink => depGraph.addEdge(sink, memNode))
+        sources.foreach(source => depGraph.addEdge(memNode, source))
+      case Attach(_, exprs) => // Add edge between each expression
+        exprs.flatMap(getDeps(_)).toSet.subsets(2).map(_.toList).foreach {
```
jackkoenig: 🕺

src/main/scala/firrtl/transforms/DeadCodeElimination.scala
jackkoenig: Yep, done.

src/main/scala/firrtl/transforms/DeadCodeElimination.scala
jackkoenig: It is now checked. Actually ran into this bug when I was making extmodules removed only when specified to be removed.

src/main/scala/firrtl/transforms/DeadCodeElimination.scala
jackkoenig: fixed

PR #553, opened 2017-04-27
Add checks on register clock and reset types (#33)

### LINKED ISSUE #33: Add clock check to high firrtl check
Actually, should be added to CheckTypes. Also, add check that initialization value is the same as the register type, and the type of reset is UInt<1> (or Clock for asynchronous reset? Needs more thought)

albert-magyar:
This fixes #33 

One issue is that CheckTypes can run before InferWidths, so _theoretically_ it's possible that the reset could be of unknown width. Therefore, this check ensures that the reset signal has `UIntType` type with either unknown or one-bit width.

I also added tests for both checks.

### DIFF
--- a/src/main/scala/firrtl/passes/Checks.scala
+++ b/src/main/scala/firrtl/passes/Checks.scala
@@ -61,11 +61,11 @@ object CheckHighForm extends Pass {
       def correctNum(ne: Option[Int], nc: Int) {
         ne match {
           case Some(i) if e.args.length != i =>
-            errors append new IncorrectNumArgsException(info, mname, e.op.toString, i)
+            errors.append(new IncorrectNumArgsException(info, mname, e.op.toString, i))
           case _ => // Do Nothing
         }
         if (e.consts.length != nc)
-          errors append new IncorrectNumConstsException(info, mname, e.op.toString, nc)
+          errors.append(new IncorrectNumConstsException(info, mname, e.op.toString, nc))
       }
 
       e.op match {
@@ -89,25 +89,25 @@ object CheckHighForm extends Pass {
         case ((percentx, n), b) if percentx && (validFormats contains b) =>
           (false, n + 1)
         case ((percentx, n), b) if percentx && b != '%' =>
-          errors append new BadPrintfException(info, mname, b.toChar)
+          errors.append(new BadPrintfException(info, mname, b.toChar))
           (false, n)
         case ((percentx, n), b) =>
           (if (b == '%') !percentx else false /* %% -> percentx = false */, n)
       }
-      if (percent) errors append new BadPrintfTrailingException(info, mname)
-      if (npercents != i) errors append new BadPrintfIncorrectNumException(info, mname)
+      if (percent) errors.append(new BadPrintfTrailingException(info, mname))
+      if (npercents != i) errors.append(new BadPrintfIncorrectNumException(info, mname))
     }
 
     def checkValidLoc(info: Info, mname: String, e: Expression) = e match {
       case _: UIntLiteral | _: SIntLiteral | _: DoPrim =>
-        errors append new InvalidLOCException(info, mname)
+        errors.append(new InvalidLOCException(info, mname))
       case _ => // Do Nothing
     }
 
     def checkHighFormW(info: Info, mname: String)(w: Width): Width = {
       w match {
         case wx: IntWidth if wx.width < 0 =>
-          errors append new NegWidthException(info, mname)
+          errors.append(new NegWidthException(info, mname))
         case wx => // Do nothing
       }
       w
@@ -116,7 +116,7 @@ object CheckHighForm extends Pass {
     def checkHighFormT(info: Info, mname: String)(t: Type): Type =
       t map checkHighFormT(info, mname) match {
         case tx: VectorType if tx.size < 0 => 
-          errors append new NegVecSizeException(info, mname)
+          errors.append(new NegVecSizeException(info, mname))
           t
         case _ => t map checkHighFormW(info, mname)
       }
@@ -124,7 +124,7 @@ object CheckHighForm extends Pass {
     def validSubexp(info: Info, mname: String)(e: Expression): Expression = {
       e match {
         case _: WRef | _: WSubField | _: WSubIndex | _: WSubAccess | _: Mux | _: ValidIf => // No error
-        case _ => errors append new InvalidAccessException(info, mname)
+        case _ => errors.append(new InvalidAccessException(info, mname))
       }
       e
     }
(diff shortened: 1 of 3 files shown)


src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -414,10 +418,20 @@ object CheckTypes extends Pass {
           errors append new InvalidConnect(info, mname, sx.loc.serialize, sx.expr.serialize)
         case sx: PartialConnect if !bulk_equals(sx.loc.tpe, sx.expr.tpe, Default, Default) =>
           errors append new InvalidConnect(info, mname, sx.loc.serialize, sx.expr.serialize)
-        case sx: DefRegister => sx.tpe match {
-          case AnalogType(w) => errors append new IllegalAnalogDeclaration(info, mname, sx.name)
-          case t if wt(sx.tpe) != wt(sx.init.tpe) => errors append new InvalidRegInit(info, mname)
-          case t =>
+        case sx: DefRegister => {
```
jackkoenig: Stylistically, we generally don't put brackets around the code following from a case, ie.
```scala
case sx: DefRegister => // we don't usually put { here
```

src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -414,10 +418,20 @@ object CheckTypes extends Pass {
...
+          sx.tpe match {
+            case AnalogType(w) => errors append new IllegalAnalogDeclaration(info, mname, sx.name)
+            case t if wt(sx.tpe) != wt(sx.init.tpe) => errors append new InvalidRegInit(info, mname)
```
jackkoenig: While I know the rest of the function does it this way, please use dot notation for `errors.append(...)`

src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -414,10 +418,20 @@ object CheckTypes extends Pass {
           errors append new InvalidConnect(info, mname, sx.loc.serialize, sx.expr.serialize)
         case sx: PartialConnect if !bulk_equals(sx.loc.tpe, sx.expr.tpe, Default, Default) =>
           errors append new InvalidConnect(info, mname, sx.loc.serialize, sx.expr.serialize)
-        case sx: DefRegister => sx.tpe match {
-          case AnalogType(w) => errors append new IllegalAnalogDeclaration(info, mname, sx.name)
-          case t if wt(sx.tpe) != wt(sx.init.tpe) => errors append new InvalidRegInit(info, mname)
-          case t =>
+        case sx: DefRegister => {
+          sx.tpe match {
+            case AnalogType(w) => errors append new IllegalAnalogDeclaration(info, mname, sx.name)
```
jackkoenig: If you don't care about the `w`, please indicate it like so:
```scala
case AnalogType(_) =>
```

src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -414,10 +418,20 @@ object CheckTypes extends Pass {
...
+            case t if wt(sx.tpe) != wt(sx.init.tpe) => errors append new InvalidRegInit(info, mname)
+            case t =>
+          }
+          sx.reset.tpe match {
+            case UIntType(IntWidth(w)) if (w == 1) =>
```
jackkoenig: I believe this can be better written as:
```scala
 case UIntType(IntWidth(1)) =>
```

src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -414,10 +418,20 @@ object CheckTypes extends Pass {
...
+            case UIntType(UnknownWidth) =>
```
jackkoenig: Better yet, you could combine these like so:
```scala
case UIntType(IntWidth(1) | UnknownWidth) =>
```

src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -414,10 +418,20 @@ object CheckTypes extends Pass {
...
+            case _ => errors append new IllegalResetType(info,mname,sx.name)
+          }
+          if (sx.clock.tpe != ClockType) {
+            errors append new RegReqClk(info,mname,sx.name)
```
jackkoenig: Also, spaces after commas

src/main/scala/firrtl/passes/Checks.scala
albert-magyar: The IntWidth takes a BigInt, so that won't work

src/main/scala/firrtl/passes/Checks.scala
albert-magyar: The IntWidth takes a BigInt, so that won't work

src/main/scala/firrtl/passes/Checks.scala
albert-magyar: This was pre-existing code; I'll change it

src/main/scala/firrtl/passes/Checks.scala
sdtwigg: Annoyingly, BigInt has no extractor/unapply methods.

If you think this type of extraction is sufficiently common, could also add an extractor in IntWidth for this itself. (Accessible with, like IntWidth.AsInt(1) and making sure to return None when can't return an Int.)

At the very least, should combine the cases via:
case UIntType(w) if w == IntWidth(1) || w == UnknownWidth

src/main/scala/firrtl/passes/Checks.scala
sdtwigg: See above.

src/main/scala/firrtl/passes/Checks.scala
albert-magyar: I think leaving the cases separate is good for clarity. The fact that the case for UnknownWidth results in no error is an artifact of the pass ordering; I think it would be better to leave it separate (with a comment explaining that the check is imperfect) than to just go for maximum brevity.

albert-magyar: @sdtwigg that does seem reasonable, but I think coming up with a de-facto solution for width matching at a meeting might be better than fixing it in this PR

src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -225,6 +225,8 @@ object CheckTypes extends Pass {
     s"$info: [module $mname]  Printf arguments must be either UIntType or SIntType.")
   class ReqClk(info: Info, mname: String) extends PassException(
     s"$info: [module $mname]  Requires a clock typed signal.")
+  class RegReqClk(info: Info, mname: String, name: String) extends PassException(
+    s"$info: [module $mname]  Register $name equires a clock typed signal.")
```
azidar: "equires"

albert-magyar:
Good idea @azidar, I added that

PR #557, opened 2017-05-02
Prep for Scala 2.12

ucbjrl:
This incorporates changes required to support Scala 2.12 - mostly bumping versions and replacing deprecated usage.


src/main/scala/firrtl/Compiler.scala
```diff
@@ -8,7 +8,8 @@ import annotations._
 
 import firrtl.ir.Circuit
 import passes.Pass
-import Utils.throwInternalError
+import firrtl.Utils.error
+import firrtl.Utils.throwInternalError
```
sdtwigg: Perhaps make these imports a bit more consistent: (this suggestion adds firrtl to all the associated lines but could also drop everywhere since in subpackage of firrtl)

import firrtl.annotations._  // Note that wildcard imports are not great....
import firrtl.ir.Circuit
import firrtl.passes.Pass
import firrtl.Utils.{error, throwInternalError}

azidar:
Are we waiting until pickling is pushed to 2.12?

ucbjrl:
I think we should merge this (assuming it passes reviews). The changes mostly clean up deprecation warnings and bump library dependencies to more modern versions.

jackkoenig:
Actually, according to @ericlove, scala picking is not only very buggy, but more or less dead. I think we should not worry about that code base and find an alternative.

Issue #566, opened 2017-05-10
Remove mixed primop input types, change sub uint return type, PR=#587


azidar:
And update spec.

PR #587, opened 2017-05-11
Change primop arg type

azidar:
Now all primops don't accept mixed-input arguments (except Dshl and Dshr, for the shift amount), addresses #566.

This works with rocket-chip running Chisel with the corresponding changes.


jackkoenig:
Are there any negative tests for trying to provide mixed primop arg types?

jackkoenig:
Also why does the new rocket.fir have so many more lines than the old one? Pretty big difference.

colinschmidt:
Perhaps there should be a helper script to generate rocket.fir so everyone knows how its done and its done consistently?

azidar:
My guess for why the new rocket.fir has more lines is that Chisel doesn't deduplicate modules any more. I just regenerated rocket.fir again with the DefaultConfig, and its still ~200,000 loc. I believe this is ready to merge.

@colinschmidt That's a good idea, it can be done in another PR.

azidar:
Conclusion: create firrtl-regress repo, put large regression files there. Move integration tests from firrtl to firrtl-regress. Add it as a sub repo. Delete current regressions.  Update the files with a new commit. In the future, can put instructions for how to clone/build rocketchip.

For this PR: delete rocket.fir, and manually edit RocketCore.fir to change mixed-type ops.

src/test/scala/firrtlTests/WidthSpec.scala
```diff
@@ -121,4 +85,46 @@ class WidthSpec extends FirrtlFlatSpec {
...
+        |    input y: SInt<2>
+        |    output z: SInt
+        |    z <= add(x, y)""".stripMargin
+    val check = Seq( "output z : SInt<4>")
+    intercept[PassExceptions] {
+      executeTest(input, check, passes)
+    }
+  }
+
+  "SInt<2> - UInt<3>" should "return SInt<5>" in {
```
jackkoenig: Shouldn't both of these mention that these should actually throw an exception?

src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -284,65 +288,42 @@ object CheckTypes extends Pass {
...
+              (isUInt, isSInt, isClock, isFix)
+            case _ => throwInternalError
+          }
+        } match {
+          //   (UInt,  SInt,  Clock, Fixed)
+          case (isAll, false, false, false) if isAll == okUInt  =>
+          case (false, isAll, false, false) if isAll == okSInt  =>
+          case (false, false, isAll, false) if isAll == okClock =>
+          case (false, false, false, isAll) if isAll == okFix   =>
+          case x => println(x); errors.append(new OpNotCorrectType(info, mname, e.op.serialize, exprs.map(_.tpe.serialize)))
```
jackkoenig: Delete println

jackkoenig:
Have the corresponding changes been made to chisel?

azidar:
No changes to chisel are needed.

azidar:
For multiply between an SInt and UInt, Chisel does the conversion to SInt:
https://github.com/freechipsproject/chisel3/blob/master/chiselFrontend/src/main/scala/chisel3/core/Bits.scala#L622

Other operators between SInt and UInt aren't allowed.

PR #588, opened 2017-05-12
Bugfix: renaming instance ports was broken.

azidar:


src/test/scala/firrtlTests/AnnotationTests.scala
```diff
@@ -445,6 +445,15 @@ class AnnotationTests extends AnnotationSpec with Matchers {
...
     val result = compiler.compile(CircuitState(parse(input), ChirrtlForm, getAMap(annos)), Nil)
+    /* Uncomment to help debug
+    println(result.circuit.serialize)
+    result.annotations.get.annotations.foreach{ a =>
+      a match {
+        case DeletedAnnotation(xform, anno) => println(s"$xform deleted: ${a.target}")
+        case Annotation(target, _, _) => println(s"not deleted: $target")
+      }
+    }
+    */
```
jackkoenig: Do you really think this is so likely to break we need to keep the printlns around? We could also make it logger.debug stuff.

azidar: Its tests and this has come up a bit, so I think its ok to leave. It will also give other people debugging a starting point.

PR #592, opened 2017-05-22
Reorganize FIRRTL directory/package structure

azidar:


src/main/scala/firrtl/transforms/analysis/Netlist.scala
```diff
@@ -1,12 +1,11 @@
-package firrtl.analyses
+package firrtl.transforms.analysis
```
jackkoenig: I'm not sure that analyses should be a sub package of transform. They are pretty different imo. @albert-magyar

src/main/scala/firrtl/util/Utils.scala
```diff
@@ -1,16 +1,15 @@
 // See LICENSE for license details.
```
jackkoenig: In moving this to a Utils package, I think we should also rename `object Utils` -> `package object util` so that we don't have to write things like `firrtl.util.Utils.throwInternalError`, instead writing `firrtl.util.throwInternalError`.

src/test/scala/firrtlTests/CustomTransformSpec.scala
```diff
@@ -2,10 +2,9 @@
 
 package firrtlTests
 
-import firrtl.ir.Circuit
 import firrtl._
-import firrtl.passes.Pass
-import firrtl.ir._
+import firrtl.ir.{Circuit, _}
```
jackkoenig: What is the difference between writing `import firrtl.ir._` and `import firrtl.ir.{Circuit, _}`?

PR #594, opened 2017-05-24
Fix performance bug in ZeroWidth

jackkoenig:
We were recursing on Types of Expressions which not only is super inefficient
(especially since we are already calling InferTypes afterward), but also
duplicates each of the Type objects that need changing.

This also means default BOOM configs can be run through firrtl without having to increase the default amount of memory given to sbt.


ccelio:
Thanks guys!

PR #595, opened 2017-05-26
Improvements to Verilog Randomization Emission

jackkoenig:
1. Add some comments to `endif
    Makes Verilog more readable to both humans and scripts
1. Change base of randomization values to _RAND instead of _GEN


jackkoenig:
I removed the refactoring of how the randomization is emitted since it didn't seem to work quite right, but I think that changing the base to `_RAND` is very helpful for people who want to filter these signals out in coverage, etc.

azidar:
lgtm

PR #596, opened 2017-05-26
Fix performance bug in DCE

jackkoenig:
We walked the whole set of vertices for every dont touch


jackkoenig:
Note for reviewers, vertices of a DiGraph is a Set

PR #599, opened 2017-06-05
Display the total time firrtl took to compile

colinschmidt:
I find it nice that chisel prints this info out.

Ideally I'd also like to print this out at a different log level so we can just get total compile time without getting individual pass timings. Open to opinions on how/whether to do that.


jackkoenig:
Yeah I think this is higher bar than log-level info, we could use log-level warning or just println

colinschmidt:
Now it is printed out by default.

PR #603, opened 2017-06-08
Fixes a typo in the verilog `elsif code generation

ss2783:


edwardcwang:
Is there a particular verilog simulator that doesn't like elseif?

ss2783:
I ran into this issue when using the vivado compiler but isn't it supposed to be `elsif as per the language standards and not `esleif?

edwardcwang:
Hmm, I've used firrtl generated countlessly many times with both Vivado and Synopsys DC and they haven't had any issues. What version of Vivado are you using?

jackkoenig:
Keep in mind this code is only generated with very particular use of Analog
(which is generally ill advised because it doesn't work in Verilator)

ss2783:
vivado 2016.1

sdtwigg:
System Verilog 1800-2012 Section 22.6 defines `ifdef, `else, `elsif, `endif, `ifndef . I don't think using `elseif is standards compliant.

wsong83:
VCS is very generous at tolerating non-standard Verilog features. So passing VCS is not really a proof of standard compliant.
Also for Vivado, the Vivado synthesiser is  more generous than the Vivado ISim simulator. Even if some code can be synthesised by Vivado, it might fail in simulation.

So if this is not compliant with standard and changing it is not a big deal, I would suggest to just fix it.

Issue #597, opened 2017-06-01
Registers which have dangling input generates incorrect verilog


jchang0:
Ticket migrated from:

https://github.com/ucb-bar/chisel3/issues/621

Correct behavior for register declaration in chisel is for register to keep the previous value even if the input is dangling.  Current verilog generation does not have that behavior, instead the register is taken out to be a wire, so in synthesis, the wire is always floating instead of able to keep the initial random value.

Relevent portion from the chisel thread:
Firrtl:

```
circuit RandTop :
  module RandTop :
    input clock : Clock
    input reset : UInt<1>
    output io : {out : UInt<32>}

    clock is invalid
    reset is invalid
    io is invalid
    reg rand : UInt<32>, clock @[Rand.scala 8:17]
    io.out <= rand @[Rand.scala 9:10]

```

Current Verilog:

```
module RandTop(
  input         clock,
  input         reset,
  output [31:0] io_out
);
  reg [31:0] rand$;
  reg [31:0] _GEN_0;
  assign io_out = rand$;
`ifdef RANDOMIZE
  integer initvar;
  initial begin
    `ifndef verilator
      #0.002 begin end
    `endif
  `ifdef RANDOMIZE_REG_INIT
  _GEN_0 = {1{$random}};
  rand$ = _GEN_0[31:0];
  `endif
  end
`endif
endmodule

```

Should have:

```
module RandTop(
  input         clock,
  input         reset,
  output [31:0] io_out
);
  reg [31:0] rand$;
  reg [31:0] _GEN_0;
  assign io_out = rand$;
`ifdef RANDOMIZE
  integer initvar;
  initial begin
    `ifndef verilator
      #0.002 begin end
    `endif
  `ifdef RANDOMIZE_REG_INIT
  _GEN_0 = {1{$random}};
  rand$ = _GEN_0[31:0];
  `endif
  end
`endif
  always @(posedge clock) begin
    rand$ <= rand$;
    // or perhaps
    rand$ <= _GEN_0;
  end
endmodule
```

aswaterman:
The always ```@(posedge clock)``` should actually contain no assignments at all, as the register is neither reset nor assigned.  The register should only be set in the initial block.

edwardcwang:
I thought that @jackkoenig said that the default next value of a `Reg` should be its current value?

aswaterman:
That's right, but ```always @(posedge clock) r <= r;``` is equivalent to no assignment at all.

[...]

PR #605, opened 2017-06-13
Zero floating

jackkoenig:
Partial resolution to #597 

This PR replaces IsInvalids with a connection to zero. Unfortunately, our constant propagation does not remove registers or wires, so a full resolution requires more substantial changes. This PR is still useful, however, because connecting these invalid wires and registers to zero simplifies the formal semantics and allows other tools to do that constant propagation.


jackkoenig:
I don't think we want to replace validif with a mux because it's extra hardware

aswaterman:
right

PR #610, opened 2017-06-21
Add --no-dce command-line option to skip DCE

jackkoenig:


jackkoenig:
As we discussed in person, I agree the Marker Pattern should be shared somehow as should some common patterns (NoDCE and DontConstProp are very similar in that they are annotations that target a specific transform but no particular named thing and it only matters if they are present or not).

PR #612, opened 2017-06-21
Add logger printing for declarations removed by DCE

jackkoenig:
Not sure if this is the best implementation or if it should be log-level debug (as PRed) or perhaps some other additional flag? Could also be log-level trace although the whole circuit is printed on that level which is a bit excessive.


colinschmidt:
I think debug is a good level for this.

PR #614, opened 2017-06-23
Better const prop

jackkoenig:


jackkoenig:
Don't merge until sure this passes for rocket-chip, there have been problems.

src/main/scala/firrtl/LoweringCompilers.scala
```diff
@@ -97,7 +97,6 @@ class LowFirrtlOptimization extends CoreTransform {
   def outputForm = LowForm
   def transforms = Seq(
     passes.RemoveValidIf,
-    passes.ConstProp,
     passes.PadWidths,
```
jackkoenig: @azidar Do you know why ConstProp was running before PadWidths? This ordering goes waaaay back. I think there is still a problem with this PR in rocket-chip but deleting this first run of ConstProp fixes the following bug:
```
wire w : { a : UInt<4>, b : UInt<4> }
w.a <= UInt<2>(3)
w.b <= UInt<2>(3)
node x = cat(w.a, w.b)
```
Without constant prop across wires this was fine, but with these changes ConstProp turns it into:
```
node x = UInt<4>("hf")
```
We need to run PadWidths before constant propagating such a concatenation.

An alternative I'm also trying out is padding connections to wires in ConstProp since there seem to be problems with not running constprop before padwidths

src/main/scala/firrtl/passes/ConstProp.scala
```diff
@@ -234,33 +234,52 @@ object ConstProp extends Pass {
...
-      if (old ne propagated)
-        nPropagated += 1
       propagated
     }
 
     def constPropStmt(s: Statement): Statement = {
-      s match {
+      val stmtx = s map constPropStmt map constPropExpression
+      stmtx match {
         case x: DefNode => nodeMap(x.name) = x.value
```
aswaterman: It looks like the old code did a preorder traversal and the new code does a postorder traversal.  Is that change benign because we are only matching on DefNode and Connect?

jackkoenig: Yeah this PR changes from postorder to preorder traversal. As far as I can tell it's functionally equivalent but results in a vast performance improvement because something like
```
node x = and(and(and(x, UInt<1>(1))))
```
can be done in one iteration than in multiple as it had been before. This is fixing the same type of performance bug as in https://github.com/freechipsproject/firrtl/pull/543 (why is why I separated that part out into a separate commit)

jackkoenig:
I've pushed a new commit that reverts to original behavior of running ConstProp before PadWidths because rocket-chip seems to break when that doesn't happen. I've also added some width padding  to ConstProp for Connections to wires so that things work properly. I'm kicking off a rocket-chip Travis run to check this PR overnight but I think it should work this way.

jackkoenig:
Okay given that https://github.com/freechipsproject/rocket-chip/pull/820 passes I think this is probably good to go, going to rebase the last commit then merge after Travis passes

jackkoenig:
This even improves compilation time despite adding a second pass to each iteration of ConstProp (probably due to fixing the performance bug which is technically orthogonal but reducing the amount of firrtl code should also make later passes run faster):

master (5135e9c6aeb1c9d472a5b29adb44a4acaf504fb8)
[info] BenchmarkVerilogCompiler.measureRob     avgt   10  11.647 ± 0.064   s/op
[info] BenchmarkVerilogCompiler.measureRocket  avgt   10  15.127 ± 0.237   s/op

better-const-prop (978d54599d92070158f941ea974b65f0ed383a53)
[info] BenchmarkVerilogCompiler.measureRob     avgt   10  11.052 ± 0.116   s/op
[info] BenchmarkVerilogCompiler.measureRocket  avgt   10  13.420 ± 0.276   s/op

PR #615, opened 2017-06-28
Remove reset

jackkoenig:
1. Fixed a kind of scary bug where a register that drives another register could be deleted from the latter register's update in emitted Verilog. See PipeTester for an example. Note that as far as I can tell, this bug has yet to manifest since Chisel cannot generate Firrtl that exhibits the bug. Improved constant propagation could have made it show up though, and my second commit here also made it show up.
1. Replace register reset with an explicit mux that is connected to the register. Removing reset enables more powerful constant propagation and other optimizations (like replacing all wires with nodes)

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -355,16 +355,11 @@ class VerilogEmitter extends SeqTransform with Emitter {
 
     def emit_verilog(m: Module, moduleMap: Map[String, DefModule])(implicit w: Writer): DefModule = {
       val netlist = mutable.LinkedHashMap[WrappedExpression, Expression]()
-      val addrRegs = mutable.HashSet[WrappedExpression]()
       val namespace = Namespace(m)
       namespace.newName("_RAND") // Start rand names at _RAND_0
       def build_netlist(s: Statement): Statement = s map build_netlist match {
         case sx: Connect =>
           netlist(sx.loc) = sx.expr
-          (kind(sx.loc), kind(sx.expr)) match {
-            case (MemKind, RegKind) => addrRegs += sx.expr
-            case _ =>
-          }
           sx
         case sx: IsInvalid =>
           netlist(sx.expr) = wref(namespace.newTemp, sx.expr.tpe)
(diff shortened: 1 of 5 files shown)


ben-k:
I can see why turning reset into an explicit mux would help with other optimizations, but does this mean Firrtl will no longer emit the verilog idiom for a resettable FF?  If so, this could definitely impact QoR.

jackkoenig:
This doesn't affect the emitted Verilog at all except for improvements where additional optimizations can occur. Firrtl will still walk the mux tree to emit register update using if else inside of the always @(posedge clock) block.

jackkoenig:
To be a little more concrete, I generated Verilog using rocket-chip DefaultConfig before and after this change

https://github.com/freechipsproject/rocket-chip/commit/c9cfe46604ff61fddf923f9d6754cd5b2230069e: 

You can see the Verilog (and diff) at the gist here: https://gist.github.com/jackkoenig/4f8b832565f7a1cd91e90bbf7d70226f

I've also pasted the short diff here. Left is without this PR, right is with this PR (ie. implicit register reset is replaced with an explicit mux):
```
133573,133577c133573
<     if (reset) begin
<       pending_0 <= 1'h0;
<     end else begin
<       pending_0 <= 1'h0;
<     end
---
>     pending_0 <= 1'h0;
169978,169982c169974
<     if (reset) begin
<       s1_slaveValid <= 1'h0;
<     end else begin
<       s1_slaveValid <= 1'h0;
<     end
---
>     s1_slaveValid <= 1'h0;
198917c198909,198925
<       ex_ra_0 <= _GEN_64;
---
>       if (fp_decoder_io_sigs_ren2) begin
>         if (fp_decoder_io_sigs_swap12) begin
>           ex_ra_0 <= _T_639;
>         end else begin
>           if (fp_decoder_io_sigs_ren1) begin
>             if (_T_636) begin
>               ex_ra_0 <= _T_637;
>             end
>           end
>         end
>       end else begin
>         if (fp_decoder_io_sigs_ren1) begin
>           if (_T_636) begin
>             ex_ra_0 <= _T_637;
>           end
>         end
>       end
198920c198928,198944
<       ex_ra_1 <= _GEN_66;
---
>       if (fp_decoder_io_sigs_ren2) begin
>         if (_T_645) begin
>           ex_ra_1 <= _T_639;
>         end else begin
>           if (fp_decoder_io_sigs_ren1) begin
>             if (fp_decoder_io_sigs_swap12) begin
>               ex_ra_1 <= _T_637;
>             end
>           end
>         end
>       end else begin
>         if (fp_decoder_io_sigs_ren1) begin
>           if (fp_decoder_io_sigs_swap12) begin
>             ex_ra_1 <= _T_637;
>           end
>         end
>       end
198923c198947,198955
<       ex_ra_2 <= _GEN_67;
---
>       if (fp_decoder_io_sigs_ren3) begin
>         ex_ra_2 <= _T_647;
>       end else begin
>         if (fp_decoder_io_sigs_ren2) begin
>           if (fp_decoder_io_sigs_swap23) begin
>             ex_ra_2 <= _T_639;
>           end
>         end
>       end
```

The amount of difference here is pretty small, primarily illustrating a couple of cases where additional optimizations could be applied and a few others where the mux tree was actually expanded further (although seemingly only in cases for registers without a reset).

jackkoenig:
(For the record, I ran this experiment before PRing the change but I should have pasted this here so that others could see that I wasn't changing things without checking)

ben-k:
Great, thanks for looking into it.

PR #616, opened 2017-06-29
Use Travis Build Stages

jackkoenig:
Splits up the chisel and firrtl parts of the test. This means they can run in parallel (and thus be faster). I'm also hopefully it'll prevent the jobs from getting spuriously killed.

### DIFF
--- a/.run_chisel_tests.sh
+++ b/.run_chisel_tests.sh
@@ -7,5 +7,5 @@ else
   mkdir -p chisel3/lib
   cp utils/bin/firrtl.jar chisel3/lib
   cd chisel3
-  sbt clean test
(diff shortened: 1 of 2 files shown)


colinschmidt:
LGTM

jackkoenig:
Well, I can confirm it has not fixed the spurious kills problem. I'm going to figure that out and fix it in this PR

jackkoenig:
I limited the parallelism allowed for chisel tests to 2 since Travis only gives 2 cores. This should prevent us from running out of memory and having spurious failures (although it does hurt performance a little, it still takes ~ 15 min total)

colinschmidt:
LGTM 👍

PR #620, opened 2017-06-30
Preserve "better" names in Constant Propagation

### LINKED ISSUE #378: Nice names getting removed
Per @ccelio 

Given simple Firrtl like the following:

```
circuit Mod :
  module Mod :
    input in : UInt<32>[4]
    input idx : UInt<2>
    output out : UInt<32>
    node named = in[idx]
    out <= named
```

We lose the nice name "named" in favor of a generated name:

```verilog
module Mod(
  input  [31:0] in_0,
  input  [31:0] in_1,
  input  [31:0] in_2,
  input  [31:0] in_3,
  input  [1:0] idx,
  output [31:0] out
);
  wire [31:0] _GEN_0;
  wire [31:0] _GEN_1;
  wire [31:0] _GEN_2;
  wire [31:0] _GEN_3;
  assign out = _GEN_0;
  assign _GEN_0 = _GEN_3;
  assign _GEN_1 = 2'h1 == idx ? in_1 : in_0;
  assign _GEN_2 = 2'h2 == idx ? in_2 : _GEN_1;
  assign _GEN_3 = 2'h3 == idx ? in_3 : _GEN_2;
endmodule
```

It looks like it's being removed in DCE. I think this could easily be solved via a heuristic that names that start in `_` lose to names that don't. We could also introduce a "temp name" annotation that's inserted whenever a temp name is inserted and those always lose out to non-temp names.

jackkoenig:
Names that do not start with '_' are "better" than those that do

Fixes #378 -- at least the original problem statement, as well as resolves how ConstProp was making good name removal much worse.

There is a reasonable debate to be had about how to pick between names that do not start with `_`. This does happen a bit in rocket-chip but most of them are replacing something like `in_0_d_bits_source` with `io_out_0_d_bits_source` although there are examples of things like `nackResponseMessage_size` being replaced by `probe_bits_size` or `pmp0_cfg_x`, `pmp0_cfg_w` and `pmp0_cfg_r` all being replaced by `default`.

In any case, this improves tons of names, we can fight later about how to differentiate between names that don't start with `_`.

### DIFF
--- a/src/main/scala/firrtl/transforms/ConstantPropagation.scala
+++ b/src/main/scala/firrtl/transforms/ConstantPropagation.scala
@@ -11,6 +11,7 @@ import firrtl.Mappers._
 import firrtl.PrimOps._
 
 import annotation.tailrec
+import collection.mutable
 
 class ConstantPropagation extends Transform {
   def inputForm = LowForm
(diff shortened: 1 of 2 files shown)


mwachs5:
Thanks. I tested this on my PLIC PR and it is So. Much. Better.

PR #621, opened 2017-06-30
Const prop regs

jackkoenig:
Now with a meaningful branch name.

This PR connects unconnected registers to zero and constant propagates across registers if they are only reset to a constant OR only connected to a constant (or reset to and connected to the same constant).

An extension of https://github.com/freechipsproject/firrtl/pull/605

### DIFF
--- a/src/main/scala/firrtl/passes/RemoveValidIf.scala
+++ b/src/main/scala/firrtl/passes/RemoveValidIf.scala
@@ -4,6 +4,8 @@ package firrtl
 package passes
 import firrtl.Mappers._
 import firrtl.ir._
+import Utils.throwInternalError
+import WrappedExpression.weq
 
 /** Remove ValidIf and replace IsInvalid with a connection to zero */
 object RemoveValidIf extends Pass {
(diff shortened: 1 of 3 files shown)


colinschmidt:
Is this a change from unconnected registers being random? Or is code coming from chisel always going to result in randomized unconnected registers?

colinschmidt:
Also I like that you added more tests than code 😄

aswaterman:
Yeah, this change makes it match synthesis tools' behavior.

jackkoenig:
This does mean you can't use registers with no reset value to inject random values anymore, but this should be done via a randomization library anyway.

colinschmidt:
You are saying there should be a transform that turns the registers into random values?

jackkoenig:
I'm not saying anything about implementation, it could be a firrtl pass, it could be a library that injects a blackbox. Just that this is something that you should be able to write and that the old behavior for an unconnected and not reset register is no longer sufficient.

API could be something like:
```scala
val rand = RandomReg(UInt(32.W))
```

colinschmidt:
Yeah I wasn't talking about implementation exactly just trying to clarify. May I ask why the old behavior was no longer sufficient?

edwardcwang:
Also, would something like RandomReg even be synthesizable?

jackkoenig:
@colinschmidt Like #605, X pessimism makes it really hard for formal tools to deal with Chisel generated Verilog. It turns out that dc wires things left floating to zero anyway.

@edwardcwang There are papers on circuits that generate random numbers so you could come up with something. It is my understanding that a register powers on with random-ish values for its bits but I suspect it's not truly random.

edwardcwang:
I guess my point was that I'm not sure it would be safe to rely on the synthesizability of RandomReg, so I'm not necessarily sure if it should be necessarily be a hardware type, but maybe an annotation or something instead.

colinschmidt:
I guess if DC is already wiring to zero and not leaving hanging then I see no issue.

PR #626, opened 2017-07-14
Fix ConstProp bug where multiple names would swap with one

jackkoenig:
Fixes issue in https://github.com/freechipsproject/rocket-chip/pull/848

PR #630, opened 2017-07-17
Do not swap wire names with node names

### LINKED ISSUE #627: Invalid firrtl from const prop
It seems like const prop generates invalid firrtl with this change: a43486b65506620f89f3e171101353b2dde65351 (e.g. It uses references before declared by node or wire.)

Thus, I couldn't use it before running custom transforms although it's just fine before feeding it into the emitter.

donggyukim:
This resolves #627. Can we merge it to #626?


jackkoenig:
I think this is the right solution.

The plan is to ultimately remove all wires, replacing them with topologically sorted nodes which would allow us to do the name swapping without the bug in #627

PR #633, opened 2017-08-01
Constant propagation across module boundaries

jackkoenig:
The long awaited. This could perhaps be improved and suggestions are welcome.

The process of constant propagating within a module is expanded to include possible constants on inputs and outputs of the module itself as well as instances. The biggest change involves the new "outer loop" of constant propagating modules in a topologically sorted order and repeating when necessary. Explanation pasted from a comment below. Important thing to note **This preserves deduplication**.

This outer loop works by applying constant propagation to the modules in a topologically
sorted order from leaf to root
Modules will register any outputs they drive with a constant in constOutputs which is then
checked by later modules in the same iteration (since we iterate from leaf to root)
Since Modules can be instantiated multiple times, for inputs we must check that all instances
are driven with the same constant value. Then, if we find a Module input where each instance
is driven with the same constant (and not seen in a previous iteration), we iterate again

### DIFF
--- a/src/main/scala/firrtl/transforms/ConstantPropagation.scala
+++ b/src/main/scala/firrtl/transforms/ConstantPropagation.scala
@@ -9,7 +9,9 @@ import firrtl.ir._
 import firrtl.Utils._
 import firrtl.Mappers._
 import firrtl.PrimOps._
+import firrtl.graph.DiGraph
 import firrtl.WrappedExpression.weq
+import firrtl.analyses.InstanceGraph
 
 import annotation.tailrec
 import collection.mutable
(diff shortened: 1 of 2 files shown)


jackkoenig:
I think this should be merged by squashing, the commits are interesting history in the development but not important for posterity

donggyukim:
Why doesn't this dedup when necessary?

jackkoenig:
Because deduplication is critical for hierarchical place and route. We can discuss changing the API for when deduplication does or does not occur but it seems prudent for this PR to preserve the status quo that is relied upon by users (in rocket-chip for example, [see hartid](https://github.com/freechipsproject/rocket-chip/blob/7eeb9dfd8892f5d0a8eb2f57ca38bc852e218de1/src/main/scala/coreplex/RocketCoreplex.scala#L118) which is an input to preserve deduplication)

azidar:
Can you make it an annotation as to whether a module is ok be duplicated?

donggyukim:
I'm happy with this PR. Are there any updates on this?

jackkoenig:
I think it's good as is. There is more work to be done for the dedup vs. okay to duplicate API, but I think that's more of a next step than a necessary thing for this PR.

donggyukim:
I totally agree :)

PR #636, opened 2017-08-14
Add work in progress of new emitter from google

dt27182:
This is the work we have so far on rewritting the emitter. It is currently just a set of functions that can turn a compound firrtl expression into a string. Since this is only currently used to emit boolean expressions, there is no logic to make sure the inferred widths are correct.

Beyond emitting compound expressions, we are also looking to better preserve the when structure in chisel inside the emitted verilog.

### DIFF
--- a/src/main/scala/firrtl/NewEmitter.scala
+++ b/src/main/scala/firrtl/NewEmitter.scala
@@ -1,0 +1,344 @@
+package firrtl
+
+import firrtl.ir._
+import firrtl.Utils._
+import firrtl.Mappers._
+import firrtl.PrimOps._
+import scala.collection.mutable
+import scala.collection.mutable.ListBuffer
+
+object CompoundExprEmitter {
+  def genVerilogReferenceName(expr: Expression): Option[String] = {
+    expr match {
+      case Reference(name, tpe) => Some(name)
+      case SubField(expr, name, tpe) =>
+        genVerilogReferenceName(expr) match {
+          case Some(exprName) => Some(exprName + "_" + name)
+          case None => None
+        }
+      case SubIndex(expr, value, tpe) =>
+        genVerilogReferenceName(expr) match {
+          case Some(exprName) => Some(exprName + "_" + value)
+          case None => None
+        }
+      case _ => //return None for expression types that don't have matching verilog reference names
+        None
+    }
+  }
+
+  /**
+   * Checks if expr is emittable by current new emitter framework. Remove when
+   * more complete new emitter is put into place.
+   */
+  def checkExprEmittable(
+    expr: Expression
+  ): Boolean = {
+    expr match {
+      case DoPrim(op, args, consts, tpe) => {
+        op match {
+          // TODO(b/38357475): remove when twigg's inline verilog refactor is checked in
+          case Rem => false
+          case _ => {
+            args.foldLeft(true)(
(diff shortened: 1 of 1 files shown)


azidar:
So this PR doesn't compile, its using `tpe(...)` and `long_BANG(...)` functions, which have since been removed. You may need to use `expr.tpe` and `bitWidth(expr.tpe)` instead.

src/main/scala/firrtl/NewEmitter.scala
```diff
@@ -0,0 +1,344 @@
...
+import firrtl.Utils._
+import firrtl.Mappers._
+import firrtl.PrimOps._
+import scala.collection.mutable
+import scala.collection.mutable.ListBuffer
+
+object CompoundExprEmitter {
+  def genVerilogReferenceName(expr: Expression): Option[String] = {
+    expr match {
+      case Reference(name, tpe) => Some(name)
```
azidar: Why are you matching on Reference instead of WRef? Same for SubField/WSubField, WSubIndex/SubIndex?

src/main/scala/firrtl/NewEmitter.scala
```diff
@@ -0,0 +1,344 @@
...
+      case SubField(expr, name, tpe) =>
+        genVerilogReferenceName(expr) match {
```
azidar: For this idiom, use `genVerilogReferenceName(expr).map( _ + "_" + name)`

src/main/scala/firrtl/NewEmitter.scala
```diff
@@ -0,0 +1,344 @@
...
+        throw new IllegalArgumentException("CompoundExprEmitter cannot yet handle expr: " + expr)
+      }
+    }
+  }
+
+  def genVerilog(
+    doprim: DoPrim,
+    resultText: StringBuilder
+  ): Unit = {
+    def getArg0 () : Expression = doprim.args(0)
```
azidar: should be `def getArg0(): Expression`

src/main/scala/firrtl/NewEmitter.scala
```diff
@@ -0,0 +1,344 @@
...
+
+    def genBinOpSigned (
+      arg0: Expression,
+      opString: String,
+      arg1: Expression
+    ): Unit = {
+      val signed = doprim.args.find(x => tpe(x).typeof[SIntType])
+      if (signed != None) {
+        tpe(arg0) match {
+          case (t:SIntType) => resultText.append("$signed(")
```
azidar: `case t: SIntType =>`

src/main/scala/firrtl/NewEmitter.scala
```diff
@@ -0,0 +1,344 @@
...
+          case (t:UIntType) => resultText.append("$signed({1'b0,")
+        }
+      }
+      resultText.append("(")
+      genVerilog(arg0, resultText)
```
azidar: You are recursively calling genVerilog, which could theoretically produce incorrect code:
```
wire [15: 0] answer;
wire [15: 0] a;
wire [15: 0] a;
assign answer = (a + b) >> 1; //will not work properly, because interim result is only 16 bits.
```

This is the central problem, which is why we haven't improved the verilog emitter. You need to ensure that all intermediate results are smaller than any wire/reg referenced in the expression.

PR #637, opened 2017-08-18
Add NodeCount analysis for helping with performance debugging

jackkoenig:


jackkoenig:
This PR doesn't *use* the analysis because it's pretty slow, an example of how one might use it is to add the following to runTransform in class Transform:

```scala
    logger.info {
      val oldNodeCount = analyses.NodeCount(state.circuit)
      val nodeCount = analyses.NodeCount(result.circuit)
      s"Nonequivalent Node Count: ${nodeCount.nonequivalent}\n" +
      s"Unique Node Count: ${nodeCount.unique}\n" +
      s"New Unique Nodes: ${nodeCount.uniqueFrom(oldNodeCount)}"
    }
```

(Yes, NodeCount gets run on every circuit twice, this way requires the minimum amount of changes to run the check)

PR #638, opened 2017-08-18
support -X sverilog to output xxxx.sv file

wsong83:
The assertion issued from Chisel is actually a SV feature.
EDA tools rely on file suffix to tell the language.
At least now I find the Vivado simulator is picky at the using SV features in Verilog files.
To save the trouble of manually setting language argument to simulators, I would like FIRRTL to issue .sv rather than .v
I can see in most situations, .v is still preferred.
So I change the FIRRTL arguements. If `-X sverilog` is used, FIRRTL still uses the same Verilog compiler but generate output with the .sv suffix. I hope this change is totally independent of other functions.


edwardcwang:
Seems okay, though it feels a bit strange that the only difference is a file extension (as opposed to stripping all the systemverilog features from -X verilog, etc). Perhaps document this really well with a warning? (otherwise as a user I might expect more differences between them)

PR #639, opened 2017-08-23
Update sbt to 0.13.16; add Scala 2.12 support.

ucbjrl:
To clean, test, and build Scala 2.11 and Scala 2.12 versions:
  % sbt +clean +test +publishLocal

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,13 +2,11 @@
 
 // sbt-site - sbt-ghpages
 
-site.settings
+enablePlugins(SiteScaladocPlugin)
 
-site.includeScaladoc()
+enablePlugins(GhpagesPlugin)
 
-ghpages.settings
-
-git.remoteRepo := "git@github.com:ucb-bar/firrtl.git"
+git.remoteRepo := "git@github.com:freechipsproject/firrtl.git"
 
 // Firrtl code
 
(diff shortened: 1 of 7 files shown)


ucbjrl:
NOTE: You need to be running Java 8 to build the 2.12 version.

jackkoenig:
How do you build just the Scala 2.11 version? Also, if we use 2.12 features, will the 2.11 cross building still work?

ucbjrl:
The default is to build the 2.11 version.
```shell
% sbt clean test publishLocal
```
will only build the 2.11 version.

If the 2.12 features require source code changes which the 2.11 compiler doesn't understand, then cross building won't work.

jackkoenig:
Fantastic news, Scala 2.11.11 has a flag to let you use 2.12 features like the only one I really care about, implementing apply for case classes!!!

jackkoenig:
If the default is 2.11, how do you build 2.12? Also should we maybe make 2.12 the default since we want people to switch to it? Maybe make it the default on the official 1.0.0 release?

ucbjrl:
Check out [cross building](http://www.scala-sbt.org/0.13/docs/Cross-Build.html). It looks like:
```shell
  % sbt "++ 2.12.2" clean test
```
should do it.

We will publish both so there isn't a default as far as clients/users are concerned. The default build (plain sbt commands) is really for developers. Eventually 2.12 will be the default. Submit a PR to make it the default `scalaVersion := "2.12.3"` whenever you think we (developers) are ready.

build.sbt
```diff
@@ -20,19 +18,49 @@ version := "1.1-SNAPSHOT"
...
+crossScalaVersions := Seq("2.11.11", "2.12.3")
+
+def scalacOptionsVersion(scalaVersion: String): Seq[String] = {
+  Seq() ++ {
+    // If we're building with Scala > 2.11, enable the compile option
+    //  switch to support our anonymous Bundle definitions:
+    //  https://github.com/scala/bug/issues/10047
+    CrossVersion.partialVersion(scalaVersion) match {
+      case Some((2, scalaMajor: Int)) if scalaMajor < 12 => Seq()
+      case _ => Seq("-Xsource:2.11")
```
jackkoenig: Is this necessary in Firrtl? It's necessary in chisel code using anonymous Bundles, but do the options have to be the same?

I ask because I actually want to add the "-Xsource:2.12" option when we compile with 2.11.11

ucbjrl:
It's not strictly necessary.

ucbjrl:
This is failing because Travis appears to be waiting to run the "prepare cache-verilator" stage until the test stages complete, but they require verilator. I suspect this is due to the cross versioning, even though the travis.yml isn't requesting a cross version build.

Could a Travis expert take a look at this please?

ucbjrl:
Could someone approve this now that it's passing Travis? I'd like to get it in next weeks release.

jackkoenig:
Is there an equivalent rocket-chip branch that uses this firrtl branch as well as the related chisel one?

ucbjrl:
Not yet, but do you want to hold off a Scala 2.12 release until we've
successfully build rocketchip with it?

jackkoenig:
Fair point. I don't think we should hold off but we should verify that rocket-chip works with these branches just to be sure we're cross-building correctly

The branch of rocket-chip that has bumped chisel3 and firrtl is: https://github.com/freechipsproject/rocket-chip/tree/bump-chisel-firrtl

jackkoenig:
Using the above rocket-chip branch with this branch of chisel3 merged into master (and the equivalent in firrtl), I get the following error:

```
/scratch/jack/rocket-chip-bump/chisel3/build.sbt:128: error: not found: value ScalaUnidocPlugin
  enablePlugins(ScalaUnidocPlugin).
```
Seems like the way plugins get enabled changed? Probably a fairly simple update of sbt configuration in rocket-chip should do it?

ucbjrl:
Updated plugins have a different interface. I've created
a Scala212-almostcompatible branch in rocket-chip with the Scala212
versions of chisel3 and firrtl, updated plugins, use of Scala 2.11.11 and
sbt 0.13.16 and it builds and runs the emulation tests (so far).

It's "almostcompatible" since it will require changes for an updated
sbt-pack plugin, and eventual replacement of the project/build.scala with
an appropriate build.sbt file (for sbt 1.x).

PR #640, opened 2017-08-23
Cleaned emitter to not emit preamble if unneeded

azidar:
Did the following experiment to check if there was any performance impact on rocket.fir:

New emitter
2137.1 ms
2133.6 ms
2051.0 ms
2166.3 ms

Old emitter
2009.8 ms
1556.9 ms
2030.8 ms
1596.6 ms

Looks like they are about the same, and small, so I think its good.


src/main/scala/firrtl/Emitter.scala
```diff
@@ -751,13 +754,23 @@ class VerilogEmitter extends SeqTransform with Emitter {
...
+    val (hasRANDOMIZE, emissions) = circuit.modules.foldLeft((false, Seq[String]())) { (trackers, m) =>
+      (trackers, m) match {
+        case ((allRANDOMIZE, seq), m: Module) =>
+          val (emission, rands) = emit_verilog(m, moduleMap)
+          (allRANDOMIZE || rands, seq ++ Seq(emission))
+        case ((allRANDOMIZE, seq), _: ExtModule) => (allRANDOMIZE, seq)
+        case _ => sys.error("Shouldn't be here")
+      }
+    }
+    if(hasRANDOMIZE) {
```
jackkoenig: Space after `if`

src/main/scala/firrtl/Emitter.scala
```diff
@@ -775,8 +788,9 @@ class VerilogEmitter extends SeqTransform with Emitter {
         circuit.modules flatMap {
           case module: Module =>
             val writer = new java.io.StringWriter
-            writer.write(preamble)
-            emit_verilog(module, moduleMap)(writer)
+            val (emission, hasRANDOMIZE) = emit_verilog(module, moduleMap)
+            if(hasRANDOMIZE) writer.write(preamble)
```
jackkoenig: space after `if`

src/main/scala/firrtl/Emitter.scala
```diff
@@ -751,13 +754,23 @@ class VerilogEmitter extends SeqTransform with Emitter {
...
-    circuit.modules.foreach {
-      case m: Module => emit_verilog(m, moduleMap)(writer)
-      case _: ExtModule => // do nothing
+    val (hasRANDOMIZE, emissions) = circuit.modules.foldLeft((false, Seq[String]())) { (trackers, m) =>
+      (trackers, m) match {
+        case ((allRANDOMIZE, seq), m: Module) =>
+          val (emission, rands) = emit_verilog(m, moduleMap)
+          (allRANDOMIZE || rands, seq ++ Seq(emission))
+        case ((allRANDOMIZE, seq), _: ExtModule) => (allRANDOMIZE, seq)
+        case _ => sys.error("Shouldn't be here")
```
jackkoenig: I believe we want to use the `error` method defined in Utils rather than `sys.error` right?

src/main/scala/firrtl/Emitter.scala
```diff
@@ -751,13 +754,23 @@ class VerilogEmitter extends SeqTransform with Emitter {
...
 
   def emit(state: CircuitState, writer: Writer): Unit = {
-    writer.write(preamble)
 
     val circuit = runTransforms(state).circuit
     val moduleMap = circuit.modules.map(m => m.name -> m).toMap
-    circuit.modules.foreach {
-      case m: Module => emit_verilog(m, moduleMap)(writer)
-      case _: ExtModule => // do nothing
+    val (hasRANDOMIZE, emissions) = circuit.modules.foldLeft((false, Seq[String]())) { (trackers, m) =>
```
jackkoenig: All of the parenthesis on the foldleft is one reason why I prefer using `.empty`, instead this could be:
```scala
val (hasRANDOMIZE, emissions) = circuit.modules.foldLeft((false, Seq.empty[String])) { // ...
```
that's just my opinion though

src/main/scala/firrtl/Emitter.scala
```diff
@@ -751,13 +754,23 @@ class VerilogEmitter extends SeqTransform with Emitter {
...
+      (trackers, m) match {
```
jackkoenig: As with most higher-order functions, I believe there is an implicit match, so I think you can basically delete
`(trackers, m) =>` and `(trackers, m) match {` and just move the cases below left a tab

src/main/scala/firrtl/Emitter.scala
```diff
@@ -751,13 +754,23 @@ class VerilogEmitter extends SeqTransform with Emitter {
...
+        case ((allRANDOMIZE, seq), m: Module) =>
+          val (emission, rands) = emit_verilog(m, moduleMap)
+          (allRANDOMIZE || rands, seq ++ Seq(emission))
```
jackkoenig: can just be `seq :+ emission`

src/main/scala/firrtl/Emitter.scala
```diff
@@ -751,13 +754,23 @@ class VerilogEmitter extends SeqTransform with Emitter {
...
+        case ((allRANDOMIZE, seq), _: ExtModule) => (allRANDOMIZE, seq)
```
jackkoenig: No reason to unapply the tuple if you're just passing it on:
```scala
  case (res, _: ExtModule) => res
```

azidar:
Addressed comments, thanks!

azidar:
Redid to emit preamble before initializations, not at top of file. Still need to check whether it works with rocketchip

azidar:
Works with rocketchip, can merge.

PR #641, opened 2017-08-23
Reorder port and wire assignments in Verilog

azidar:
Made a simple change to make wire and port assignments print out in the correct order.

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -533,11 +533,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
           (dir, tpe) match {
             case (_, AnalogType(_)) => "inout " // padded to length of output
             case (Input, _) => "input "
-            case (Output, _) =>
-              // Assign to the Port
-              val ex = WRef(name, tpe, PortKind, FEMALE)
-              assign(ex, netlist(ex))
-              "output"
+            case (Output, _) => "output"
           }
         }
         // Turn types into strings, all ports must be GroundTypes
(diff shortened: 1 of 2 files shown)


src/main/scala/firrtl/Emitter.scala
```diff
@@ -553,24 +549,26 @@ class VerilogEmitter extends SeqTransform with Emitter {
       }
 
       def build_streams(s: Statement): Statement = s map build_streams match {
+        case sx@Connect(info, loc@WRef(_, _, PortKind|WireKind|InstanceKind, _), expr) =>
```
jackkoenig: I generally prefer spaces around the `@`s and `|`s, ie.
```scala
case sx @ Connect(info, loc @ WRef(_, _, PortKind | WireKind | InstanceKind, _), expr) =>
```

src/main/scala/firrtl/Emitter.scala
```diff
@@ -553,24 +549,26 @@ class VerilogEmitter extends SeqTransform with Emitter {
...
           sx
         case sx: DefRegister =>
           declare("reg", sx.name, sx.tpe)
           val e = wref(sx.name, sx.tpe)
           update_and_reset(e, sx.clock, sx.reset, sx.init)
           initialize(e)
           sx
-        case sx: IsInvalid =>
-          val wref = netlist(sx.expr) match { case e: WRef => e }
+        case sx@IsInvalid(info, expr) =>
```
jackkoenig: same here with spaces

azidar:
Finished with syntax changes.

PR #648, opened 2017-08-30
WiringTransform Refactor

### LINKED ISSUE #647: Use Lowest Common Ancestor for Wiring Transform (Relax Explicit Top Requirement?)
I'm reading through [`WiringTransform`](https://github.com/freechipsproject/firrtl/blob/master/src/main/scala/firrtl/passes/wiring/WiringTransform.scala) and it's unclear to me why this requires that the user provides a [`TopAnnotation`](https://github.com/freechipsproject/firrtl/blob/master/src/main/scala/firrtl/passes/wiring/WiringTransform.scala#L41). This currently creates one problem and one annoyance:
  * __Problem__: Incorrectly specifying the `TopAnnotation` as being higher than the lowest common ancestor (LCA) of sinks and sources will result in the addition of source signals going all the way up to the specified top
  * __Annoyance__: The user has to specify the `TopAnnotation` correctly as the LCA. The user should know what this is, but I've always envisioned this type of pass as computing it for you and allowing you to place sinks and sources arbitrarily.

__Proposed Solution__: 
  1. Remove the requirement on a top annotation being specified
  2. Construct all data structures from the circuit-level top module
  3. Compute the [`sharedParent`](https://github.com/freechipsproject/firrtl/blob/master/src/main/scala/firrtl/passes/wiring/WiringUtils.scala#L115) as the LCA when running [`setFields`](https://github.com/freechipsproject/firrtl/blob/master/src/main/scala/firrtl/passes/wiring/WiringUtils.scala#L104) and remove the `setSharedParent` method

__Questions__:
  * Is the exposure of the source all the way up to the top and not using the LCA a bug or a feature? 
  * Is there a reason why the top is expected to be specified explicitly? The only thing that comes to mind here is that this is intended to solve the problem of multiple instances of a source existing in the hierarchy and the explicit top module designation is used for scoping (e.g., you have a dual core rocket and there are pins that need to be wired internal to both that are named the same thing). In this case, I think it's equivalent to say that all sinks should be connected to all named sources via their respective LCAs.

I can get a PR together. The modifications for LCA in `3` are easy, but doing doing the `1` and `2` changes are slightly more involved.

seldridge:
This refactors the `WiringTransform` to use `InstanceGraph` with the use of a new `EulerTour` class to speed things up via fast least common ancestor (LCA) queries:
  * All Sinks are connected to their closest Sources as determined by BFS. This removes the need for `TopAnnotation`.
  * Sources may be connected to sinks anywhere in the hierarchy (source as a sink submodule, sink as a source submodule, and source and sink in the same module)
  * An `EulerTour` class is added that enables `O(1)` LCA queries, or equivalently range minimum queries (RMQ), with `O(n)` preprocessing. This uses Berkman--Vishkin as described in the simplified version [here](https://rd.springer.com/content/pdf/10.1007/10719839.pdf#page=101).
  * Tests are added for the `EulerTour` class
  * Some of the `WiringTests` are renamed (to better reflect what is going on) and an intentionally failing `WiringTest` now passes.

This extends the existing `WiringTransform` to allow:
  * Connections to sink `Components` as well as `Modules` to enable "tunneled" connections from one module's internal signals to another's

This passes all old `WiringTests`, some new `WiringTests`, and the new `EulerTourTests`. I've done a spot-check of tunneled wiring from one RoCC to another RoCC in Rocket to trigger some printfs and things work there.

__Notes:__
The `EulerTour` makes an annoyingly frequent use of `ArrayBuffer` for data structures. It's my understanding that this is required to get constant-time lookups and a correspondingly correct implementation of Berkman--Vishkin. If there's a better way to do this, let me know and I can update the PR.

This nearly deprecates `Lineage` (and most everything in `WiringUtils`) except for the fact that `ClockList` still uses it.

Fixes #647 

__TODO__:
  - [x] Support source under sink
  - [x] Update documentation


seldridge:
@azidar: This should be ready to review.

seldridge:
Note: for my own needs (which can either be included in this or as a separate PR) I need to be able to wire up to existing signals while `WiringTransform` currently wires an existing source component to new sink ports (that the transform creates) on a specified module. The expected use case is effectively `Module` -> `ExtModule` connections (though it'll work just fine for connecting to a sink `Module` and creating a port for that connection). 

Effectively, what should the `WiringTransform` be able to wire? Two approaches:
  * Assume some simple primitive for the transform. A possibility within this space is to keep it the way it is. For additional connections, the user is expected to handle these with other passes. As an example, the `WiringTransform` creates sink ports whose names are recorded and used to generate _other_ annotations that will be wired up by other transforms.
  * Make `WiringTransform` capture all cases and wire from any source `Named` to any sink `Named` (though erroring if connecting to `CircuitName` or `CircuitTopName`). If the source/sink is a `ModuleName` as opposed to a `ComponentName`, then it is assumed that the connection is to stop at the port level. The source and the sinks can then be an arbitrary mix of modules or components and `WiringTransform` does the right thing based on what it's given.

It would seem like the latter approach of generic connectivity is the way to go, but there's also a case for keeping `WiringTransform` simple and sharp.

seldridge:
Hold off on the review. There's an obvious issue that the tests aren't catching that prevents correct Verilog generation. There's a namespace collision that crops up for bundles in sources that add ports, e.g., [similar to this]( https://github.com/freechipsproject/firrtl/blob/master/src/test/scala/firrtlTests/WiringTests.scala#L153).

seldridge:
@azidar, @jackkoenig: This should _now_ be good to review. Likely low priority as I think I'm the only one that is banging on the Wiring Transform. I'll get an associated sink/source annotator into `chisel3` once this is in, assuming it's in.

src/main/scala/firrtl/graph/EulerTour.scala
```diff
@@ -0,0 +1,224 @@
...
+/** Euler Tour companion object */
+object EulerTour {
+  /** Create an Euler Tour of a `DiGraph[T]` */
+  def apply[T](diGraph: DiGraph[T], start: T): EulerTour[Seq[T]] = {
+    val r = mutable.Map[Seq[T], Int]()
+    val e = mutable.ArrayBuffer[Seq[T]]()
+    val h = mutable.ArrayBuffer[Int]()
+
+    def tour(u: T, parent: Seq[T] = Seq(), height: Int = 0): Unit = {
+      val id = parent :+ u
```
jackkoenig: I know that premature optimization is the root of all evil, but `Seq()` returns a `List` which is obviously not the best thing to append to. it would be better if this could either be prepending or just use a Vector

since tour is only called with default arguments once and is nested function, perhaps ditch the default args?

src/main/scala/firrtl/graph/EulerTour.scala
```diff
@@ -0,0 +1,224 @@
...
+      if (!r.contains(id)) { r(id) = e.size }
```
jackkoenig: Seems like a good place for [getOrElseUpdate](http://www.scala-lang.org/api/current/scala/collection/mutable/Map.html#groupBy[K](f:A=>K):scala.collection.immutable.Map[K,Repr])

src/main/scala/firrtl/graph/EulerTour.scala
```diff
@@ -0,0 +1,224 @@
...
+      e += id
+      h += height
+      diGraph.getEdges(id.last).toSeq.flatMap( v => {
```
jackkoenig: May I ask why the edges need to be converted to a Seq and why a flatMap? I think this should just be a foreach on the edges.

src/main/scala/firrtl/graph/EulerTour.scala
```diff
@@ -0,0 +1,224 @@
...
+  }
+  private lazy val st = constructSparseTable(a)
+
+  /** Precompute all possible RMQs for an array of size `n where each
+    * entry in the range is different from the last by only +-1
+    */
+  private def constructTableLookups(n: Int): Array[Array[Array[Int]]] = {
+    def sortSeqSeq[T <: Int](x: Seq[T], y: Seq[T]): Boolean = {
+      if (x(0) != y(0)) { x(0) < y(0)                }
+      else              { sortSeqSeq(x.tail, y.tail) }
```
jackkoenig: I think this could be written in a single line as:

```scala
if (x(0) != y(0)) x(0) < y(0) else sortSeqSeq(x.tail, y.tail)
```
I guess I don't particularly like using curly braces on if else blocks with a single statement each

src/main/scala/firrtl/passes/wiring/Wiring.scala
```diff
@@ -10,157 +10,159 @@ import firrtl.Mappers._
...
+            addPort = Some((portNames(pm), DecWire)),
+            cons = (meta(pm).cons :+ (to, from)).distinct
+          )
+          meta(cm) = meta(cm).copy(
+            addPort = Some((portNames(cm), DecInput))
+          )
+        case _ =>
+          throw new WiringException(s"""Unexpectedly short path from LCA to sink (possibly due to source being a submodule of a sink?):
+  - LCA: ${lca.map(_.name)}
+  - sink: ${sink.map(_.name)}""")
```
jackkoenig: I would prefer these tabbed over a bit with some tactical use of `|` and stripMargin

src/main/scala/firrtl/graph/EulerTour.scala
```diff
@@ -0,0 +1,224 @@
...
+        tmp(base)(size) = min
+        assert(min >= base)
+        min
+      }
+    }
+
+    for (i <- (0 to x.size - 1);
+      j <- (0 to math.ceil(lg(x.size)).toInt - 1);
+      if i + (1 << j) - 1 < x.size) yield {
+      tableRecursive(i, j)
```
jackkoenig: I think different tabbing here would be helpful, maybe: 
```scala
   for (i <- (0 to x.size - 1);
        j <- (0 to math.ceil(lg(x.size)).toInt - 1);
        if i + (1 << j) - 1 < x.size) yield {
      tableRecursive(i, j)
```
Also I might be wrong but I don't think the yield is necessary? The for loop isn't returning anything

src/main/scala/firrtl/graph/EulerTour.scala
jackkoenig: Similar idea with other if else blocks

src/main/scala/firrtl/passes/wiring/WiringUtils.scala
```diff
@@ -18,6 +21,19 @@ case object DecInput extends DecKind
...
+  addPort: Option[(String, DecKind)] = None,
+  cons: Seq[(String, String)] = Seq.empty) {
+
+  override def toString: String = serialize("")
+
+  def serialize(tab: String): String = s"""
+$tab addPort: $addPort
+$tab cons: $cons
+"""
+}
```
jackkoenig: Similarly would like tactical use of tabbing and stripMargin

src/main/scala/firrtl/passes/wiring/WiringUtils.scala
```diff
@@ -87,9 +100,9 @@ object WiringUtils {
   /** Counts the number of instances of a module declared under a top module
     */
   def countInstances(childrenMap: ChildrenMap, top: String, module: String): Int = {
-    if(top == module) 1
-    else childrenMap(top).foldLeft(0) { case (count, (i, child)) =>
-      count + countInstances(childrenMap, child, module)
+    if (top == module) { 1 }
+    else { childrenMap(top).foldLeft(0) { case (count, (i, child)) =>
+      count + countInstances(childrenMap, child, module) }
     }
```
jackkoenig: What is this curly brace closing? Seems weird that it's lined up with the else but isn't closing the else?

src/main/scala/firrtl/analyses/Netlist.scala
```diff
@@ -80,5 +82,15 @@ class InstanceGraph(c: Circuit) {
...
-}
+  /** An `[[EulerTour]]` representation of the `[[DiGraph]]` */
+  lazy val tour = EulerTour(graph, trueTopInstance)
 
+  /** Finds the lowest common ancestor instances for two module names in
+    * a design
+    */
+  def lowestCommonAncestor(moduleA: Seq[WDefInstance],
+    moduleB: Seq[WDefInstance]):
+      Seq[WDefInstance] = {
```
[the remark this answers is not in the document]
seldridge: I'm all for adopting a consistent style. Some of these weird formats are primarily a function of me using an unconfigured emacs scala-mode (which will indent however it does by default) and trying to stay under 80 columns. The alignment of parameters makes it a lot more readable. I'll update throughout.

src/main/scala/firrtl/graph/EulerTour.scala
seldridge: Premature optimization of this class is likely a good thing. Any performance bugs here usually get amplified (as you either don't use any WiringTransforms or you use a lot of them, e.g., scan chain insertion). 

Yes, I don't think the defaults are needed here.

src/main/scala/firrtl/graph/EulerTour.scala
seldridge: This is likely just an inefficient approach. I'm usually testing things on the REPL to get the behavior that I want. Usually, I'm trying to `map` everything and don't fallback to the obvious `foreach`. That would simplify things and avoid the conversion cost.

src/main/scala/firrtl/passes/wiring/WiringUtils.scala
seldridge: Should be the added `{` for the `else`. Adding all these isn't doing anything... However, I realize that this method isn't actually used anymore. I'm trying to aggressively remove deprecated portions of WiringUtils that are relying on Lineage/ChildrenMap. I'll rip this whole thing out. That _may_ have implications for others that are depending on portions of WiringUtils existing...

src/main/scala/firrtl/passes/wiring/Wiring.scala
```diff
@@ -10,157 +10,160 @@ import firrtl.Mappers._
...
+    *   - Adds a pin to each sink module
+    *   - Punches ports up from source signals to the LCA of each sink
+    *   - Punches ports down from LCAs to each sink module
+    *   - Wires sources up to LCA, sinks down from LCA, and across each LCA
+    *
+    * @param c The circuit to modify
+    * @param wi A WiringInfo object containing source, sink, component, and pin information
+    * @return The modified circuit
+    *
+    * @throws WiringException if a sink is equidistant to two sources
```
jackkoenig: Please correct me if I am wrong but it feels like this refactor is intending to do all wirings in one fell swoop, but it doesn't look like that's happening. This exception for example mentions a sink being equidistant to two sources, but given that we call `wire` for each `WiringInfo`, how can there be two sources for a single sink? Don't we recreate all of these data structures for each wiring?

src/main/scala/firrtl/passes/wiring/WiringUtils.scala
```diff
@@ -84,86 +97,132 @@ object WiringUtils {
...
-      ((l: Lineage) => l match {
-        case Lineage(name, _, _, _, _, true, _, _, _) => //SinkParent
-          val tos = l.children.filter { case (i, c) => (c.sinkParent || c.sink) && !c.sourceParent } map { case (i, c) => s"$i.${portNames(c.name)}" }
-          val from = s"${portNames(name)}"
-          l.copy(cons = l.cons ++ tos.map(t => (t, from)))
-        case l => l
-      })
-    )
-    funs.foldLeft(lin)((l, fun) => fun(l))
+  def namedName(n: Named): String = {
```
jackkoenig: Since this is extracting the name of the module, I think `moduleName` might be a better name than `namedName`?

src/main/scala/firrtl/passes/wiring/Wiring.scala
```diff
@@ -10,157 +10,160 @@ import firrtl.Mappers._
...
+    // Split out wiring info
+    val (ComponentName(compName, ModuleName(source,_)), sinks, pin) = (wi.source, wi.sinks, wi.pin)
+    val iGraph = new InstanceGraph(c)
 
     // Create valid port names for wiring that have no name conflicts
     val portNames = c.modules.foldLeft(Map.empty[String, String]) { (map, m) =>
       map + (m.name -> {
         val ns = Namespace(m)
-        if(sinks.contains(m.name)) ns.newName(pin)
+        if(sinks.filter(namedName(_) == m.name).nonEmpty) ns.newName(pin)
```
jackkoenig: `.filter(...)` followed by `.nonEmpty` is better expressed as just `.exists(...)`

src/main/scala/firrtl/passes/wiring/Wiring.scala
```diff
@@ -10,157 +10,160 @@ import firrtl.Mappers._
...
-        throw new WiringException(s"Module ${wi.top} does not contain all instances of $m.")
-    }
+    // Split out wiring info
+    val (ComponentName(compName, ModuleName(source,_)), sinks, pin) = (wi.source, wi.sinks, wi.pin)
+    val iGraph = new InstanceGraph(c)
 
     // Create valid port names for wiring that have no name conflicts
     val portNames = c.modules.foldLeft(Map.empty[String, String]) { (map, m) =>
       map + (m.name -> {
         val ns = Namespace(m)
```
jackkoenig: I know you didn't write this code, but if we're trying to make wiring fast, this Namespace call requires walking the entire module, which we do here for every single module. Combine this with my observation above that we have to call wire for every single wiring we're trying to do... I think this gets expensive.

src/main/scala/firrtl/passes/wiring/Wiring.scala
```diff
@@ -10,157 +10,160 @@ import firrtl.Mappers._
...
-    val map = pointToLineage(withThings)
-
     // Obtain the source component type
     val sourceComponentType = getType(c, source, compName)
+    val sinkComponents = sinks
+      .foldLeft(Map.empty[String, String]) { (map, m) =>
+        map ++ { m match {
+          case ComponentName(c, ModuleName(m, _)) => Map(m -> c)
+          case _ => Map()
+        }}}
```
jackkoenig: I think this could be better written as just
```scala
val sinkComponents = sinks.collect { case ComponentName(c, ModuleName(m, _)) => m -> c }.toMap
```

src/main/scala/firrtl/passes/wiring/WiringUtils.scala
```diff
@@ -84,86 +97,132 @@ object WiringUtils {
...
+      .withDefaultValue(false)
+
+    i.fullHierarchy.keys
+      .filter { case WDefInstance(_, _, module, _) => module == source }
+      .map    { k => i.fullHierarchy(k)
+        .map  { l => queue.enqueue(l)
+                     owners(l) = Seq(l)                                }}
+
+    while (queue.nonEmpty) {
+      val u = queue.dequeue
```
jackkoenig: I think this is better as just
```scala
for (u <- queue) { ...
```

src/main/scala/firrtl/passes/wiring/WiringUtils.scala
```diff
@@ -84,86 +97,132 @@ object WiringUtils {
...
+    val indent = "  "
+
+    val owners = new mutable.HashMap[Seq[WDefInstance], Seq[Seq[WDefInstance]]]
+      .withDefaultValue(Seq())
+    val queue = new mutable.Queue[Seq[WDefInstance]]
+    val visited = new mutable.HashMap[Seq[WDefInstance], Boolean]
+      .withDefaultValue(false)
+
+    i.fullHierarchy.keys
+      .filter { case WDefInstance(_, _, module, _) => module == source }
```
jackkoenig: Since there's only 1 source, might this be better as a find?

src/main/scala/firrtl/passes/wiring/Wiring.scala
seldridge: This is currently doing all wirings for a given `Source Component -> Set(Sink Component)` mapping in one fell swoop. I would much prefer if this was doing _all_ wirings for all wiring annotations at once (I intended that specifically to be a future addition). That's a little more complicated, but not overly so. That will improve the performance, and justify the whole Euler Tour nonsense better, as you only pay that construction cost once for all wiring annotations (unless you have passes that generate new wiring annotations and call the wiring transform...).

This problem of a "sink being equidistant to two sources" comes up because components are tied to a module and not an instance. Hence, some weird things can happen once you start instantiating multiple copies of the same source module. 

Specifically, you can create a circuit like the following test where you have two sources at an equivalent level in the hierarchy that are supposed to wire to a single sink that is at a higher level in the hierarchy. 
  - Equidistant problem:  https://github.com/seldridge/firrtl/blob/2d7f86f14f7270a43cdeed0dd0c33e1a0d00fc4a/src/test/scala/firrtlTests/WiringTests.scala#L480
Here, there's no clear way to determine what the user intended ("Wire these modules", but the user has specified a hierarchy that doesn't make sense). I do not expect most people to hit this, but it can come up if you were trying to, e.g., wire a rocket core to an L2. This would work fine for a single core, but then you'd run into a problem if you instantiated two cores. I'd say that if you really want this type of behavior, you need to be using diplomacy or some library that can figure out what you instantiated and then generate the logic you need.

Alternatively, this could be represented as "Unable to determine unique mapping of source to sink".

Seem reasonable?

Alternatively, the much more frequent case is the following where the sink is encased neatly under the source (though this is really just a special case of it not being equidistant). Here, it's unambiguous what the user meant
  - No Equidistant Problem: https://github.com/seldridge/firrtl/blob/2d7f86f14f7270a43cdeed0dd0c33e1a0d00fc4a/src/test/scala/firrtlTests/WiringTests.scala#L435

seldridge:
Thanks for all the feedback, Jack. I'll get these changes in.

In response to your question about the sinks under sources... Nothing precludes this working. I just didn't include the logic for it. There's a special case of needing to check when you see the path to the LCA as being a single module (i.e., you are the LCA). This is already correctly handled for the source being the LCA (https://github.com/seldridge/firrtl/blob/2d7f86f14f7270a43cdeed0dd0c33e1a0d00fc4a/src/main/scala/firrtl/passes/wiring/Wiring.scala#L124), but I didn't get around to doing it for the sink case.

The old WiringTransform did not handled this at all, but that seemed like much more of a quick placeholder for something more robust.

I'm fine on holding off on getting the full version in there. Once this is exposed publicly (arguably via the Chisel3 PR) I can expect people trying all sorts of ridiculous things and this needing to get fixed quickly. I'll make a note of this via a checklist above.

src/main/scala/firrtl/passes/wiring/Wiring.scala
jackkoenig: I think it is totally reasonable to do all wirings in one fell swoop in a future PR. Thank you for the clarifications! A++ for having tests to illustrate!

jackkoenig:
I mainly want source under sink because I badly want that feature! That being said, I'm okay with merging this PR without that feature, but it definitely needs to be supported before any exposed Chisel API uses it.

seldridge:
Alright some updates:
  - I added both source under sink with a test
  - I applied @jackkoenig changes, excepting a few things
  - I modified this to do all the wirings at once
  - I have a performance kludge that makes the (hopefully) common case of single-sources fast

I did some performance analysis to try and figure out what was going on as this was running for a lot longer than I anticipated (using a higher-level transform that I have that uses the new WiringTransform)... The current problem point is when you need to assign ownership. This requires a BFS seeded with the sources that fans out to find all the sinks. This has to be run for each set of `WiringInfo` even with processing lumped together. Either I'm doing the BFS inefficiently or there are some improvements that can be made in place... 

I opted for one of these improvements where if there is only a single source instance, it short circuits and assigns ownership of all sinks to the single source. There are further optimizations that could potentially be done, but I don't know of a clean way to get around the worst case `O(n)` for multiple source instances.

src/main/scala/firrtl/passes/wiring/WiringUtils.scala
seldridge: Multiple instances of a source module, so punting on this.

src/main/scala/firrtl/passes/wiring/WiringUtils.scala
seldridge: Updated in a newer version. Not fantastic still, but should be more legible.

src/main/scala/firrtl/passes/wiring/WiringUtils.scala
```diff
@@ -84,86 +97,132 @@ object WiringUtils {
...
+    // this should fail is if a sink is equidistant to two sources.
+    sinkInsts.map { s =>
+      if (!owners.contains(s) || owners(s).size > 1) {
+        throw new WiringException(s"Unable to determine source mapping for sink '${s.map(_.name)}'") }
+    }
+
+    owners
+      .filter { case (k, v) => sinkInsts.contains(k) }
+      .map    { case (k, v) => (k, v.flatten)        }
+      .toMap
```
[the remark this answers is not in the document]
seldridge: Thanks! Yes, this does look better.

src/main/scala/firrtl/passes/wiring/WiringUtils.scala
seldridge: This makes much more sense. Changed to `getModuleName`.

seldridge:
@jackkoenig: good to go for another look. 

Much less hurriedly than the description above, the performance bottleneck is in determining owners, specifically in the BFS. I've muddled around with that to try and improve it (mostly just by using the correct data structure for the operations at hand). See: [this section](https://github.com/seldridge/firrtl/blob/58a9563a4aa86e405e268159304afd99c2d660bb/src/main/scala/firrtl/passes/wiring/WiringUtils.scala#L153). However, there is likely a way to further improve this.

jackkoenig:
Apologies for the delay, thanks for the bugfix! I've slated this for the 1.1 release so I'll re-review and merge soon.

src/main/scala/firrtl/passes/wiring/WiringTransform.scala
```diff
@@ -11,86 +11,87 @@ import scala.collection.mutable
...
+    ToWorkingIR,
+    InferTypes,
+    ExpandConnects,
+    Uniquify,
+    ExpandWhens,
+    CheckInitialization,
+    InferTypes,
+    ResolveKinds,
+    ResolveGenders
+  )
```
jackkoenig: Rather than including a mishmash of transforms that essentially perform HighForm -> MidForm should we instead just make this transformation return HighForm?

seldridge: Seems reasonable to me as this has ballooned from four Passes to all the passes... Just need to play musical chairs with the test output FIRRTL orderings...

jackkoenig:
@seldridge Can you help me understand why this needs to run lowering again? Why does it need ExpandWhens, last connect semantics? I guess it uses bulk connect for aggregates and thus needs ExpandConnects?

seldridge:
`ExpandWhens` was specifically to resolve last connect semantics. `ExpandConnects` was to, as you infer, get the bulk connects converted back over. That should not be needed if it targets HighForm, though. That then puts it soundly in the category of needing to run High->Mid lowering.

I think that makes sense. Do note that most of that plethora of passes was me just trying to get it to work with my instrumentation passes and not DCE the additions.

jackkoenig:
Yeah that makes sense, the ability to emit high form and then re-lower it is an important feature of Firrtl anyway.

I think in the future if we expand reliance on the transform we might consider making it emit clean MidForm without requiring relowering but that's strictly a performance optimization and I don't think necessary right now.

jackkoenig:
@seldridge So what do you think the status of this is? Do you think it's ready or is there some more work to do?

seldridge:
Always more work to do. However, it's strictly an improved superset of the old version. I have some trepidation as the stricter emission of WVoid was what started some last minute changes here and seem to have broken (exposed incorrect FIRRTL generation on my part) in my instrumentation passes.

I think the former point is enough to warrant shipping, but I would feel more comfortable if I had the current set of patches tested with my instrumentation.

Maybe bump to 1.1.1? I'm fine with going forward and am prepared to help maintain this.

seldridge:
__Update__: This is continuing to work seamlessly with my instrumentation passes. With that verified to work on my end (and all the normal tests passing) I'd consider this ready and good to merge.

So, I'm of the opinion of merging this for the 1.1.0 window. Actions following:
  - [ ] Open an issue to improve WiringTransform performance by emitting MidForm directly
  - [ ] Open an issue to update the YAML format for the 1.1.1 window
  - [ ] Pull in BoringUtils for the Chisel 3.1.1 window (or maybe 3.1.0?)

But, I defer to you all.

jackkoenig:
Thank you for all of your hard work on this @seldridge, I'm ready to merge. As I reviewed it I can help maintain so between the two of us I'm not too worried.

PR #649, opened 2017-08-31
Added option to emit final annotations

azidar:
Added option to emit final annotations to a file.

### DIFF
--- a/src/main/scala/firrtl/ExecutionOptionsManager.scala
+++ b/src/main/scala/firrtl/ExecutionOptionsManager.scala
@@ -301,7 +301,7 @@ trait HasFirrtlOptions {
 
   parser.opt[String]("output-annotation-file")
     .abbr("foaf")
-    .valueName ("<output-anno-file>>")
+    .valueName ("<output-anno-file>")
     .foreach { x =>
       firrtlOptions = firrtlOptions.copy(outputAnnotationFileName = x)
     }.text {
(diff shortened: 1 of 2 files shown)


src/main/scala/firrtl/ExecutionOptionsManager.scala
```diff
@@ -298,6 +299,15 @@ trait HasFirrtlOptions {
       "use this to force appending annotation file to annotations being passed in through optionsManager"
     }
 
+  parser.opt[String]("output-annotation-file")
+    .abbr("foaf")
```
edwardcwang: What does foaf mean?

src/main/scala/firrtl/ExecutionOptionsManager.scala
```diff
@@ -283,8 +284,8 @@ trait HasFirrtlOptions {
 
   parser.opt[String]("annotation-file")
     .abbr("faf")
-    .valueName ("<output>").
-    foreach { x =>
+    .valueName ("<input-anno-file>>")
```
edwardcwang: Is this intended to have an extra right bracket?

azidar:
Re freechipsproject/chisel3#566, i think that was for dumping annotations before FIRRTL compilation, while this dumps annotations after FIRRTL compilation. I'll make the small changes and merge.

src/main/scala/firrtl/ExecutionOptionsManager.scala
azidar: foaf = firrtl output annotation file, its like the previous style for shortened commandline flags

src/main/scala/firrtl/ExecutionOptionsManager.scala
azidar: nope, i'll fix it.

edwardcwang:
Sounds good; LGTM.

Issue #545, opened 2017-04-25
Are graph traversals deterministic?


donggyukim:
DiGraph uses `Set` to contain edges:
https://github.com/ucb-bar/firrtl/blob/master/src/main/scala/firrtl/graph/DiGraph.scala#L83

Thus, it's not clear to me `getEdges` returns a set with a deterministic order:
https://github.com/ucb-bar/firrtl/blob/master/src/main/scala/firrtl/graph/DiGraph.scala#L93

donggyukim:
I find #529, so close it.

Issue #529, opened 2017-04-07
Make Deterministic DiGraph


jackkoenig:
Non-deterministic DiGraph is fine for checking for combinational loops, but any time graph functionality is desired for code generation, it needs to be deterministic. This determinism is maintained throughout the Firrtl compiler by using LinkedHashMaps.

Protip, there's a [Scala bug](https://issues.scala-lang.org/browse/SI-9594) that we have been [burned by](https://github.com/ucb-bar/firrtl/pull/303) ([MULTIPLE TIMES](https://github.com/ucb-bar/rocket-chip/pull/603)) where if you iterate on the keys of a LinkedHashMap, the iteration order is **no longer deterministic**. Thus we should always iterate on the keys and values together.

jackkoenig:
Done in #653

PR #653, opened 2017-09-07
Make digraph methods deterministic

albert-magyar:
There are three requirements for making digraph methods deterministic (#529, #545):

1. Use `LinkedHashMap{Set,Map}` exclusively
2. Hide the `mutable.{Set,Map}`-type references to prevent direct mutation in unpredictable ways
3. Use only plain iteration ordering of `LinkedHashMap`s, and avoid using any built-in methods that perform an operation on a `{Set,Map}` and return a collection (`keySet`, etc).

This PR also addresses the API of adding edges to add three flavors that either throw, add vertices, or fail when attempting to add edges between vertices not currently in the graph. This may not be the best solution and/or naming, but it reflects existing use cases and simplifies pass-writers code while making intent clearer.

### DIFF
--- a/src/main/scala/firrtl/graph/DiGraph.scala
+++ b/src/main/scala/firrtl/graph/DiGraph.scala
@@ -1,111 +1,70 @@
 package firrtl.graph
 
-import scala.collection.immutable.{Set, Map, HashSet, HashMap}
+import scala.collection.{Set, Map}
 import scala.collection.mutable
-import scala.collection.mutable.MultiMap
-
-/** Represents common behavior of all directed graphs */
-trait DiGraphLike[T] {
-  /** Check whether the graph contains vertex v */
-  def contains(v: T): Boolean
-
-  /** Get all vertices in the graph
-    * @return a Set[T] of all vertices in the graph
-    */
-  def getVertices: collection.Set[T]
-
-  /** Get all edges of a node
-    * @param v the specified node
-    * @return a Set[T] of all vertices that v has edges to
-    */
-  def getEdges(v: T): collection.Set[T]
-}
-
-/** A class to represent a mutable directed graph with nodes of type T
-  * 
-  * @constructor Create a new graph with the provided edge data
-  * @param edges a mutable.MultiMap[T,T] of edge data
-  * 
-  * For the edge data MultiMap, the values associated with each vertex
-  * u in the graph are the vertices with inedges from u
-  */
-class MutableDiGraph[T](
-  private[graph] val edgeData: MultiMap[T,T] =
-    new mutable.HashMap[T, mutable.Set[T]] with MultiMap[T, T]) extends DiGraphLike[T] {
-
-  // Inherited methods from DiGraphLike
-  def contains(v: T) = edgeData.contains(v)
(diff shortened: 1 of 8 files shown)


jackkoenig:
Why "DO NOT MERGE"?

albert-magyar:
I wanted someone to comment on the changes to the API (types, edge adding variants) first. I figured I'd remove that label after some feedback.

Go back to bed!

src/main/scala/firrtl/graph/DiGraph.scala
```diff
@@ -340,3 +306,39 @@ class DiGraph[T] (val edges: Map[T, Set[T]]) extends DiGraphLike[T] {
...
+    */
+  def addPairWithEdge(u: T, v: T): Unit = {
+    edges.getOrElseUpdate(v, new LinkedHashSet[T])
+    edges.getOrElseUpdate(u, new LinkedHashSet[T]) += v
+  }
+
+  /** Add edge (u,v) to the graph if and only if both u and v are in
+    * the graph prior to calling addEdgeIfValid.
+    */
+  def addEdgeIfValid(u: T, v: T): Unit = {
```
jackkoenig: Should this perhaps return Boolean to indicate if it was successful?

src/main/scala/firrtl/graph/DiGraph.scala
```diff
@@ -1,102 +1,59 @@
...
-    new DiGraph((mdg.edgeData mapValues { _.toSet }).toMap[T, Set[T]])
-
-  /** Create a DiGraph from a MultiMap[T] of edge data */
-  def apply[T](edgeData: MultiMap[T,T]): DiGraph[T] =
-    new DiGraph((edgeData mapValues { _.toSet }).toMap[T, Set[T]])
+  def apply[T](mdg: MutableDiGraph[T]): DiGraph[T] = mdg
 
   /** Create a DiGraph from a Map[T,Set[T]] of edge data */
-  def apply[T](edgeData: Map[T,Set[T]]) = new DiGraph(edgeData)
+  def apply[T](edgeData: Map[T,Set[T]]) = {
```
jackkoenig: If the edgeData already is a LinkedHashMap, would it be reasonable to shortcut to just creating the DiGraph? Could be another apply method.

I'm writing code to use this I imagine for anyone who cares about determinism, this will be a common use case?

albert-magyar: The factory method you're proposing is dangerous because it creates an 'immutable' digraph that is backed by a mutable collection that the user can modify.

Personally, I would use it if it existed, but I don't know the convention around that kind of mutation hole.

jackkoenig: Ah good point, so I guess the right pattern here is to create a Mutable DiGraph and mutate that rather than creating the original LinkedHashMap?

jackkoenig: Then create an immutable DiGraph from the mutable one?

albert-magyar: That was the original intent

Issue #654, opened 2017-09-07
Instance Graph Merges Structurally Equivalent Instances


seldridge:
### Problem
Instance Graph merges multiple instances with the same `WDefInstance(info, name, module, tpe)` signature. This stems from the fact that `WDefInstance` (a case class) is using structural comparison when being added to a `Set` (which is arguably how it should work for checks of `WIR`/`IR` case classes!).

### Example

This is a simplified version of the [`"Wiring from r to X"` test of `WiringTests.scala`](https://github.com/freechipsproject/firrtl/blob/master/src/test/scala/firrtlTests/WiringTests.scala#L36).

__Circuit with two instances of `D` named `d`__
```firrtl
circuit Top :
  module Top :
    input clock: Clock
    inst a of A
    a.clock <= clock
  module A :
    input clock: Clock
    inst b of B
    b.clock <= clock
    inst d of D      @ 'd of D'
    d.clock <= clock
  module B :
    input clock: Clock
    inst c of C
    c.clock <= clock
    inst d of D      @ 'd of D' conflicts with above
    d.clock <= clock
  module C :
    input clock: Clock
    reg r: UInt<5>, clock
  module D :
    input clock: Clock
```

This produces the following edges:
```
[info]   (c,C) -> Set()
[info]   (a,A) -> Set((d,D), (b,B))
[info]   (d,D) -> Set()
[info]   (Top,Top) -> Set((a,A))
[info]   (b,B) -> Set((d,D), (c,C))
```

Both `a` and `b` are connected to `d`. While not incorrect, this breaks how I expected Instance Graph to work and gives you something between an Instance Graph and a Module Graph.

__Circuit with two instances of `D` with different names (`d1` and `d`)__
```
circuit Top :
  module Top :
    input clock: Clock
    inst a of A
    a.clock <= clock
  module A :
    input clock: Clock
    inst b of B
    b.clock <= clock
    inst d1 of D     @ renaming d to d1 produces the expected result
    d1.clock <= clock
  module B :
    input clock: Clock
    inst c of C
    c.clock <= clock
    inst d of D
    d.clock <= clock
  module C :
    input clock: Clock
    reg r: UInt<5>, clock
  module D :
    input clock: Clock
```

This produces the following edges:
```
[info]   (c,C) -> Set()
[info]   (a,A) -> Set((d1,D), (b,B))
[info]   (d1,D) -> Set()
[info]   (d,D) -> Set()
[...]

PR #655, opened 2017-09-12
Make pathsInDAG walk all possible paths

seldridge:
This is gated on clarification of the intent of `pathsInDAG`.

This changes `pathsInDAG` to walk every possible path from a root to all reachable nodes. The old version, defining visited-ness by instance definition, would only return paths for all nodes that have unique instance definitions. 

I _think_ this is what it's supposed to be doing (based on the comment), but I need the above-mentioned clarification of intent.

See: https://github.com/freechipsproject/firrtl/issues/654#issuecomment-328889277.

### DIFF
--- a/src/main/scala/firrtl/graph/DiGraph.scala
+++ b/src/main/scala/firrtl/graph/DiGraph.scala
@@ -273,16 +273,16 @@ class DiGraph[T] (val edges: Map[T, Set[T]]) extends DiGraphLike[T] {
     // paths(v) holds the set of paths from start to v
     val paths = new mutable.HashMap[T,mutable.Set[Seq[T]]] with mutable.MultiMap[T,Seq[T]]
     val queue = new mutable.Queue[T]
-    val visited = new mutable.HashSet[Seq[T]]
+    val reachable = reachableFrom(start)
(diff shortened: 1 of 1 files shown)


albert-magyar:
You are right, this is a bug! Thanks for finding this. The intent is to have it enumerate all paths, even in non-trees.

However, the bug is not that each node is visited only once -- visiting each node once is sufficient to get every path. The problem is that the queue order must be a valid linearization; I was sloppy and used a greedier ordering without thinking properly. Unfortunately, the `visited` logic produces a valid linearization for some non-tree graphs, which hid the bug in my (limited) testing of `pathsInDAG` on non-trees.

I updated the PR to incorporate that change (visiting only once, but using the correct linearization order) since it will dramatically reduce the asymptotic worst-case runtime. Would you mind testing my fix?

seldridge:
This is working for me! (Tested as producing the expected hierarchy for my incomplete wiring transform experiments). 

Thanks for fixing the issue with the asymptotic complexity. My approach seemed inefficient, but I couldn't properly describe "an instance should only be visited once it's parents have been visited", i.e., topologically sorted, until you pointed this out. The hard refresher on graph theory continues...

Fine to merge on my end.

PR #661, opened 2017-09-21
Fix problem where wrong verilog file is used.

chick:
When calling verilator in a subdirectory like ./test_run_dir/...
verilator will read the verilog file from the current working directory
if there is a file there with the right name.  This fix specifies
the full absolute path of the verilog file intended.

### DIFF
--- a/src/main/scala/firrtl/util/BackendCompilationUtilities.scala
+++ b/src/main/scala/firrtl/util/BackendCompilationUtilities.scala
@@ -102,7 +102,7 @@ trait BackendCompilationUtilities {
 
     val command = Seq(
       "verilator",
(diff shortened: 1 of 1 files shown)


ucbjrl:
Unfortunately, this appears to break some tests in my environment and I don't see the travis test results here. Were they successful?

jackkoenig:
It did pass Travis

chick:
@ucbjrl , the more I think about it I think this could present a **spaces-in-the-path** type problem.  I should probably quote the path or escape the spaces.

ucbjrl:
Yes, xcode tools (stealth) update required agreeing to Apple's license before they would function. The notice was in the test run output. All good now.

PR #663, opened 2017-09-22
Fix string lit

jackkoenig:
*Note:* This PR should be squashed and merged, the history isn't helpful

This PR changes the implementation of StringLit from Array[Byte] to String.

The original motivation was to undo the mistake of using Array[Byte] instead of just the built-in String. At a bare minimum, using Array suggested that StringLits were intended to be mutable whereas the rest of the AST is immutable (whoops). I suspected it would improve performance, but when I originally implemented this, it didn't help at all. Mildly disappointed, I shelved it for a little while.

Now, comparing to master, there is a modest performance improvement:

```
master
[info] BenchmarkVerilogCompiler.measureRob     avgt    5  11.065 ± 0.155   s/op
[info] BenchmarkVerilogCompiler.measureRocket  avgt    5  16.690 ± 1.081   s/op


fix-string-lit
[info] BenchmarkVerilogCompiler.measureRob     avgt    5  10.931 ± 0.327   s/op
[info] BenchmarkVerilogCompiler.measureRocket  avgt    5  13.347 ± 1.320   s/op
```

**But that's not close to the most exciting part**. I started experimenting with monitoring the GC behavior of Firrtl, and it seems that this has a *massive* impact on memory use.

To try this out yourself, add the following to firrtl's build.sbt and run with `-ll info`:
```scala
fork in run := true

javaOptions ++= Seq(
  "-XX:+PrintGC",
  "-verbose:gc"
)
```

I don't 100% understand what's going on, but basically there are two types of GC runs, runs that only look at the young generation (eden or short-lived objects), and full GC runs that also look at longer-term objects. If you run this on master with regress/rocket.fir or other large designs, the parser has many GC runs, both young and full, but then you will not see another full GC run for the rest of the compiler. This is somewhat expected because we use immutable datastructures and replace so much of the circuit so frequently.

However, if you use this branch, you will eventually see a full GC, and then the memory use drops substantially.

For regress/rocket.fir: Size after GC  550M -> 240M, total heap size 1.8G -> 1.5G
For a large rocket-chip design: Size after GC  7.2G -> 2.2G, total heap size 16G -> 12G
(These numbers are from the heap debugging prints, they aren't the amount used by the process but they're certainly related)


jackkoenig:
I think there are further improvements to be made by changing FileInfo from using a single String to a String, Int and Int for the common case of @[filename, line#:column#]. We can intern the filenames since they should repeat a lot. Of course we can keep the FileInfo with just a String around since some people may rely on it and it's technically allowed, but the case emitted by Chisel could use even less memory.

PR #664, opened 2017-09-26
Reorder vector size nesting to fix multi-dimensional vecs

albert-magyar:
**Note: there is probably a chisel "bug" counteracting this "bug," so this is definitely not ready for primetime**

The following is currently illegal, causing the compiler to complain that the index `2` is too big for the range of a 2-element vector.
```
circuit Top :
  module Top :
    input clock : Clock
    output out : UInt<1>
    reg r_vec : UInt<1>[4][2], clock
    out <= r_vec[UInt<2>(2)][UInt<1>(1)]
```
This arises because the order in which index notation is considered is not compatible between vec declarations and access expressions. There are two ways to fix it (fixing the declarations or fixing the expressions), and this fix assumes that the access expression from the above block should be resolved to the following AST, as it currently does in master:

```
    SubAccess
    |       |
SubAccess   1
|       |
Ref     2
```

The counterintuitive, but correct implication is that the vec should have type `Vec(Vec(_, 1), 2)`. This PR modifies the grammar to cause that to happen, by using a single production to iterate through the declared sizes of multi-dimensional vecs, followed by a `foldRight` to produce the right IR structure.

### DIFF
--- a/src/main/antlr4/FIRRTL.g4
+++ b/src/main/antlr4/FIRRTL.g4
@@ -51,6 +51,10 @@ dir
   | 'output'
   ;
 
+idx
+  : '[' intLit ']'
+  ;
+
 type 
   : 'UInt' ('<' intLit '>')?
   | 'SInt' ('<' intLit '>')?
@@ -58,7 +62,7 @@ type
   | 'Clock'
   | 'Analog' ('<' intLit '>')?
   | '{' field* '}'        // Bundle
-  | type '[' intLit ']'   // Vector
+  | type idx+   // Vector
   ;
 
 field
--- a/src/main/scala/firrtl/Visitor.scala
+++ b/src/main/scala/firrtl/Visitor.scala
@@ -133,7 +133,8 @@ class Visitor(infoMode: InfoMode) extends FIRRTLBaseVisitor[FirrtlNode] {
           else AnalogType(UnknownWidth)
           case "{" => BundleType(ctx.field.asScala.map(visitField))
         }
-      case typeContext: TypeContext => new VectorType(visitType(ctx.`type`), string2Int(ctx.intLit(0).getText))
+      case typeContext: TypeContext =>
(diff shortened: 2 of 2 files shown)


jackkoenig:
To add some food for thought (because I have been struggling to wrap my head around this), here is some example Chisel code and what it maps to:

```scala
  val vecA = Wire(Vec(2, Vec(4, UInt(32.W))))
  vecA(1)(3) := "haa".U
```
```
    wire vecA : UInt<32>[4][2]
    vecA[1][3] <= UInt<8>("h0aa")
```
```scala
  class MyBundle extends Bundle { val foo = Vec(4, UInt(32.W)) }
  val vecB = Wire(Vec(2, new MyBundle))
  vecB(1).foo(3) := "hbb".U
```
```
    wire vecB : {foo : UInt<32>[4]}[2]
    vecB[1].foo[3] <= UInt<8>("h0bb")
```
Now the Chisel code makes sense in that it matches the behavior of Scala, eg.
```scala
scala> val xs = Seq.tabulate(2) { i =>
     |   Seq.tabulate(4) { j => (i + 1)*(j + 2) }
     | }
xs: Seq[Seq[Int]] = List(List(2, 3, 4, 5), List(4, 6, 8, 10))

scala> println(xs(1)(3))
10
```
However, the generated Firrtl code differs from what one would expect when you look at equivalent C code:
```c
#include <stdio.h>

typedef struct {
  int foo[4];
} foo_t ;

int main(void) {
  int arr1[4][2];
  foo_t arr2[2];
  arr1[3][1] = 4;
  arr2[1].foo[3] = 5;
  printf("%d %d\n", arr1[3][1], arr2[1].foo[3]);
  return 0;
}
```

In summary: The Chisel API looks and acts like one would expect in Scala. The confusing bit is that it maps to type definitions that look like C but are indexed like Scala, which is what I believe @albert-magyar is trying to address. As he said, this does require a change to Chisel emission in addition to this PR.

seldridge:
Won't fix per @azidar comment.

PR #670, opened 2017-10-02
Add Yosys formal equivalence checking to Travis regressions

jackkoenig:
This adds a helper script for running equivalence checks using Yosys. For now it only checks BOOM's Rob because there are namespace issues with circuits with multiple modules. The check only runs on PRs and only if the emitted Verilog is actually different.

I have run the script locally and it seems to work well, but I can't test it on actual Travis without a PR. I'm going to see if creating a PR against this branch that trivially changes Verilog emission will trigger it.

### DIFF
--- a/.install_yosys.sh
+++ b/.install_yosys.sh
@@ -1,0 +1,11 @@
+set -e
+# Install Yosys (https://github.com/cliffordwolf/yosys)
+if [ ! -f $INSTALL_DIR/bin/yosys ]; then
+  mkdir -p $INSTALL_DIR
+  git clone https://github.com/cliffordwolf/yosys.git
+  cd yosys
+  git pull
+  git checkout yosys-0.7
+  make
+  make PREFIX=$INSTALL_DIR install
+fi
--- a/.run_formal_checks.sh
+++ b/.run_formal_checks.sh
@@ -1,0 +1,17 @@
+set -e
+# Run formal check only for PRs
+if [ $TRAVIS_PULL_REQUEST = "false" ]; then
+    echo "Not a pull request, no formal check"
+    exit 0
+elif [[ $TRAVIS_COMMIT_MESSAGE == *"[skip formal checks]"* ]]; then
+    echo "Commit message says to skip formal checks"
+    exit 0
+else
+    # $TRAVIS_BRANCH is branch targeted by PR
+    # Travis does a shallow clone, checkout PR target so that we have it
+    # THen return to previous branch so HEAD points to the commit we're testing
+    git remote set-branches origin $TRAVIS_BRANCH && git fetch
+    git checkout $TRAVIS_BRANCH
+    git checkout -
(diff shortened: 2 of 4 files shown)


aswaterman:
This is pretty cool.  Once we sort out the namespace issues, does it make sense to rely exclusively on LEC for Firrtl PRs, instead of Verilog simulation?

jackkoenig:
I think utilizing LEC to help verify even more of the codebase is the right way forward (along with more focesed unit tests for code coverage). I’m not sure about completely ditching simulation since we at least want sanity checks that our emitted code works with verilator; especially in circumstances where the LEC checks might fail because of undefined behavior or we do actually tweek the formal semantics slightly. I think we’ll just have to see how effective and robust this test is.

jackkoenig:
I have a PR against this branch to test the LEC checking, currently failing but once it passes I think this PR is good to go.

The other PR https://github.com/freechipsproject/firrtl/pull/672

yunsup:
This is great!  For a split second, I thought Yosys was reading FIRRTL directly, and thought that would've been super cool.  :)

azidar:
Another thing we could do in the mean time is inline everything for rocket.fir. Not sure how long it would take though.

jackkoenig:
Alright after fighting with git and Travis for a while I think this PR is ready. To show that it worked, I created to PRs against this branch: one that should pass LEQ and one that should fail.

- This PR does not change code generation at all - [Against master, my script found the emitted Verilog files were identical so no reason to run LEQ](https://travis-ci.org/freechipsproject/firrtl/jobs/282128241#L873)
- #672 just added a space to emission - [Yosys showed the emitted Verilog is equivalent](https://travis-ci.org/freechipsproject/firrtl/jobs/282128663#L871)
- #673 emits all `<=` as just `<` - [Yosys caught it and Travis failed](https://travis-ci.org/freechipsproject/firrtl/jobs/282133205#L872)

Additionally, all branch Travis runs skipped LEQ as expected because this test requires branches to compare. I plan on adding more LEQ tests in the future to test the compiler more directly.

jackkoenig:
To summarize what this PR actually does:  For PRs **only**, check that the Verilog generated from regress/Rob.fir by the PR and the branch the is formally equivalent. This is intended to detect bugs introduced by a PR. If a PR is supposed to change the formal semantics, include "[skip formal checks]" in the commit message.

[Possible] future uses of LEQ in Firrtl testing:
- Show that passes that notionally do not change the circuit actually do not change the circuit (like optimization passes)
- Show that passes that DO change the circuit are equivalent to the original circuit under certain circumstances
  - eg. FAME-1 should be the same so long as all stall signals are false
- Include printfs and assertions in LEQ
- Use LEQ as a more robust correctness definition for some current tests that use incredibly brittle string comparison (small enough unit tests like this are *probably* okay)
- Prove that certain Chisel/Firrtl circuits are the same as equivalent Verilog

PR #671, opened 2017-10-02
Discussion about firrtl graph representation

dt27182:
There was mention of a firrtl graph representation developed for some firrtl pass at UCB. We also had to make one for a pass at Google. I would be great if we can develop off a common implementation. 

I have attached the node definitions for our implementation in this pull request. Could some one point me to the Berkeley implementation? Hopefully this can get a discussion started on converging to a common version.


albert-magyar:
src/main/scala/firrtl/graph/DiGraph.scala

seldridge:
Also, there's the old WiringUtils [`Lineage`](https://github.com/freechipsproject/firrtl/blob/master/src/main/scala/firrtl/passes/wiring/WiringUtils.scala#L21) that is basically deprecated in favor of [`InstanceGraph`](https://github.com/freechipsproject/firrtl/blob/master/src/main/scala/firrtl/analyses/Netlist.scala#L12) which relies on [`DiGraph`](https://github.com/freechipsproject/firrtl/blob/master/src/main/scala/firrtl/graph/DiGraph.scala#L72).

Note, I generally view there being 3 different representations of a FIRRTL graph:
  * `Module` level (FIRRTL AST)
  * `DefInstance` level (this is an `InstanceGraph`, which is a `DiGraph` of instance definitions, which is not guaranteed to be a tree)
  * Actual instances ([`pathsInDag`](https://github.com/freechipsproject/firrtl/blob/master/src/main/scala/firrtl/graph/DiGraph.scala#L262) method of `DiGraph` which is guaranteed to be a tree)

FWIW, I've found these to be sufficient abstractions.

dt27182:
Thanks for pointing out the existing graph code! After looking through the DiGraph and InstanceGraph code you guys pointed out, it seems that the graph representation in this PR tries to solve a sightly different problem.

This graph definition more or less has a node type for each firrtl ast node type and has explicitly named pointers to neighbor nodes. For example, the ConnectGraphNode class has explicitly named "loc" and "expr" node pointers. 

This was necessary because we wanted to do a circuit modification on the graph representation of the circuit and then turn it back into an firrtl ast tree. The explicitly named neighbor pointers are necessary to turn the modified graph back into an ast tree. 

I have added our transform to this commit. It's purpose is to remove a set of user specified sub module instances from a top module, put them in an new wrapper module definition, and then put an instance of the new wrapper module in place of the old sub module instances inside the top module.

I think the graph representation in this PR would be useful for any transform that needs to modify the circuit in graph form and then turn it back into the AST form. It's really a thing orthogonal to the DiGraph.

dt27182:
Please review this PR for merging into firrtl. I've removed the actual transform using the IR nodes for now because it uses some google infrastructure that will some some time to port over, but it would be good to get the graph definition merged in so that future passes can use it.

seldridge:
One thing that isn't totally clear to me is if it would make more sense to handle the example transform you reference as directly operating on the AST.

My perhaps incorrect understanding is that the AST is supposed to be left intact and modified via pure functions (passes/transforms). Auxiliary data structures may help track information internal to this transform, but are not supposed to replace the AST as the circuit representation.

It would seem like this is really proposing a parallel representation of the AST as a node graph. This would then seem to beg the question of why use the AST at all? This is then much closer to the way Verilog::Perl works where the Verilog AST is parsed to a graph, you modify the graph, then emit the new AST. This is entirely valid, but I'm not sure if this is the model that FIRRTL wants to use.

So, I can see this as being useful as an alternative representation, but this may result in some unintended consequences going forward as it seems to want to redefine what a pass/transform is (to something which operates on a node graph with transformations to/from a node graph representation on either end).

Apologies, as I'm likely missing something here...

dt27182:
@seldridge 

Right I intended this as an alternate representation of the circuit. It would exist in parallel with the AST representation, not as a replacement for it. I think both representations make some transforms easier and some transforms harder to do. 

My transform needed the graph representation of the circuit because I wanted to extract a set of circuit elements (wires, regs, etc) out of the top level circuit and replace them with a module instance. All of the top level signals driving the set of circuit elements now need to be connected to the inputs of the module instance and all of the signals being driven by the set of circuit elements now need to be connected to the outputs of the module instance. 

The AST representation doesn't allow me to easily find and modify these source/sink relationships, so I had to transform it into a graph representation where each circuit element has pointers to its source and sink nodes. We could augment the AST representation with hash tables storing these source/sink relationships, but I think that's more difficult to work with. 

There are various graph representations we could consider, but I do think that for this type of transformation where arbitrary circuit elements are being replaced, some type of graph representation is needed. It would be very helpful to have an official graph representation and functions to convert between the graph and AST representations as writing that code is actually quite time consuming.

As for the definition of passes, I think it would be fine to require that passes take in an AST and return an AST. If the user needs to use the graph representation in the passes, they are required to transform it back into an AST.

In summary, I think some types of transforms need to work on some sort of graph representation of the circuit, whether that is the AST augmented with some hashmaps or a more formal graph structure. Providing a formal graph structure along with functions to covert between it and and AST would save a lot of repeated ad hoc coding and in my opinion make it easier to work with the graph form.

albert-magyar:
My thoughts may be kind of similar to Schuyler's.

As far as I can tell, this is a set of parallel classes that effectively re-implement the AST with the addition of two analyses.
1.) Track the expressions referring to a node
2.) Add upward edges to from child to parent in expression trees

These are certainly useful analyses, but what happens when a third analysis is needed? If this parallel set of classes includes it, it increases the amount of responsibility for every pass using them, given the desire to formalize this as a more complete representation. If it doesn't, the pass writer has to fall back to alternatives like hash tables, anyway.

I think factoring out individual analyses and providing supporting code for that particular analysis (finding references, linking to parents) can actually decrease boilerplate over wrapping a select few analyses into the IR. Certain analyses (type for one) are already directly linked with the AST nodes, but those are special, since they actually correspond to syntactic constructs in some cases.

Just as an implementation detail, wouldn't this be possible by adding mix-in traits or by just extending the existing AST type hierarchy? That would cut a lot of the boilerplate.

dt27182:
So I guess the two possible implementations are:
1. A adjacency list graph structure implemented as classes (this PR)
2. The standard firrtl AST augmented by 3+ maps (expression node to parent node map, reference node to named node map, named node to reference node map). Probably need even more maps because ports aren't actually nodes in the firrtl AST and they needed to be manipulated as graph nodes in the pass I was writing.

There are pros and cons of both implementations.
Adjacency list graph structure implemented as classes:
Pros:
1. Easy to do graph traversal and easy to modify graph node connectivity.
2. Easy to keep graph structure consistent when modifying graph.
Cons:
1. The contents of 3+ analysis have to be bundled as one unit. User passes need to worry about artifacts of all 3+ analysis even if they only need one.

Firrtl AST augmented by maps:
Pros:
The 3+ analysis can be used separately by user passes. User passes can use only the analysis they need.
Cons:
1. Graph traversal is harder and more verbose because user has to look up neighbor pointers in 4+ separate data structures. Same for modifying graph node connectivity.
2. Harder to keep graph structure consistent because it's spread across 4+ separate data structures.

If you look at the transform I checked in https://github.com/freechipsproject/firrtl/commit/28f0273927fe9687184b954ac7bbe76abd1256ab, the graph manipulation is already very long even after I have spent a lot of effort factoring commonly used manipulations into helper functions. If the firrtl AST augmented by maps approach is used, this would become even more verbose and difficult to maintain.

From my experience of having to actually write this type of graph manipulation pass, the ease of writing the manipulation code is a more important concern than the concern about bundling the logic of the 3+ analysis into one unit. Also, no one is forcing user passes to use this graph representation. If they only needed one of the analysis, they are free to write that analysis as it's own function and check it into repo no?

seldridge:
Thanks for all the clarifications, here. I think I'm understanding better... 

I still have some confusion in that I've written (non-public) passes like this that add new instances internal to modules, cut existing wires, and splice in the ports of the new modules. Roughly, these all look like:
  1. Analyze the modules to figure out all the new `case class Modifications[...]` that need to be applied in terms of defines, connections, deletions, renames (right hand side `WRef` replacements for all `Connect`/`PartialConnect`), and whatever else which is packaged up and keyed by module name `Map[String, Modifications]`. Any new modules will need to be applied as Circuit-level modifications (a built up `Seq[DefModule]`). I can't really take credit for this as it's learned from the way that Adam wrote the original WiringTransform... Note: this analysis may have to be built up incrementally by looking at a module multiple times.
  2. Go through all the modules and apply your modifications
  3. Add your new modules to the circuit

This has truly proven to be sufficient for what I'm doing and I think it's equivalent to the type of this Google pass. However, this was highly, highly counterintuitive to me until I started to look heavily at the structure of the Wiring Transform and some pieces of what Strober/MIDAS do.

Part of my confusion is that I don't see the need as being something inherent to your type of transform, but stemming from the chosen transform implementation. If this is the case, then this falls into the category of an auxiliary representation of the AST used for some Google passes.

__Towards this former point, I could be entirely wrong!__ I need to look at your referenced pass in more detail as I've only been operating on your initial description of it.

This has absolutely nothing to say about _secondary effects_ such as usability, accesibility of writing new passes, not having to work directly with the AST, etc. and _tertiary effects_ of broadening adoption, paving the way for Googles passes to be upstreamed, avoiding a hypothetical fork, etc. Secondary effects I'm a horrible judge of (as my modus operandi is to see how Berkeley does it and then think of a minimal diff on that) and tertiary effects are way above my pay grade... :smile:

The broader question (and PR title!) moves towards, "What is a sufficient set of abstractions for writing transforms?" I've (as usual) drunk the koolaid and am fine with things as they are, but you and other people who have dealt with this a lot (extrapolating from Angie's emoji...) seem to think that the current representations are woefully insufficient.

albert-magyar:
I think we're going to have to agree to disagree on the ease of maintaining consistency, since I think the tricky part of updates is forcing consistency, which this graph does not do by itself. Calling `parents(node) = x` seems just as easy as calling `node.parent = x`; the hard part is remembering all the steps to maintain a consistent structure, which should be wrapped in a helper function regardless. One advantage of the AST + analyses approach with respect to consistency is that the AST is the single source of truth.

Zooming out a bit, is the intention to add public passes to FIRRTL that use this representation?

dt27182:
@seldridge 
Ok I see, thanks for pointing out a concrete example of how to do this type of firrtl transform with the firrtl AST. My viewpoint is biased as well since I only have experience with my implementation. I'll have to take a look at the wiring transform and see how that works. I think this will help me understand your viewpoint.

@albert-magyar 
I completely agree that the hard part of maintaining graph consistency is remembering all the steps to maintain a consistent structure, which should be wrapped in a helper function. However, some of these helper functions will have to depend on maps from multiple analysis. For example a helper function that does replaces one named node with another will need to modify both the reference node to named node map and the named node to reference node map. 

So if a user pass uses this helper function, it is forced to deal with the artifacts both analysis. If the user pass uses these helper functions, then they will have to deal with the same con of using the formalized graph structure, which is being forced to deal with artifacts of multiple analysis. So it really seems to me that the firrtl AST + analysis maps + helper functions approach is roughly equivalent to the formalized graph + helper functions approach from a user pass perspective.

I think we'll have to agree to disagree on which approach is easier to implement and maintain.

At a higher level, the intention is to upstream Google passes that use this implementation. We are just trying to give technical contributions back to the project, as we had discussed previously.

albert-magyar:
The graph is cyclic, so that's a bad name :)

jackkoenig:
Good point, unfortunately `firrtl.dcgir` just doesn't have the same ring to it 😞

azidar:
This has been a very interesting discussion, thanks everyone for sharing their thoughts.

I believe I understand both points, and I agree that we currently need a better solution to make these types of transformations easier.

I also certainly want to encourage Google to PR more code, and given this is the key building block of all of your code, I feel its definitely important to accept as is so that @dt27182 can continue moving forward.

That aside, I do agree with @albert-magyar that we should make it easier to write analyses. I also wrote all the above before @jackkoenig responded, so basically, I concur with @jackkoenig.

albert-magyar:
I agree, if the goal is to upstream passes that use it, it's a good idea to pave the way!

dt27182:
I've put the code inside a firrtl.altIR package. Please review!

jackkoenig:
Sorry! Looks good! I will approve but lets hold off on merging until 1.0.0 is published!

PR #682, opened 2017-11-02
Emit source locators as comments in emitted Verilog

jackkoenig:
This simply emits source locators as a comment at the end of the associated line in the Verilog. I might've missed a couple of things, but I think it hits most cases.

Future work (that I'm working on):
- Preserve source locators through ExpandWhens
- Improve instance connections (currently we have lots of extra assigns from instance ports, WDefInstanceConector provides a mechanism to remove these extra assigns and should also support Info)

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -233,6 +233,10 @@ class VerilogEmitter extends SeqTransform with Emitter {
       case (i: Int) => w write i.toString
       case (i: Long) => w write i.toString
       case (i: BigInt) => w write i.toString
+      case (i: Info) => i match {
+        case NoInfo => // Do nothing
+        case ix => w.write(s" //$ix")
+      }
       case (s: Seq[Any]) =>
         s foreach (emit(_, top + 1))
         if (top == 0) w write "\n"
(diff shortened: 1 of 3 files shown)


jackkoenig:
Also I noticed the default info-mode is "append" (which generates source locators for each line of Firrtl and appends it to any parsed) as opposed to what I would expect the default to be "use". Anyone have thoughts on this?

edwardcwang:
Any tests (even if just quickly see what the output would look like)?

jackkoenig:
Yeah I'll add some. Testing the emitted Verilog is notoriously brittle but I'll add some tests (as well as bring in some source locator propagation tests I've been working on)

colinschmidt:
Does use generate firrtl line numbers if there are none? I think sometimes append can be useful for later stages even if use does the nice thing.

jackkoenig:
use does not generate if there are none, we could add a new style "use-else-gen" 😃

jackkoenig:
Now introducing `search`, a test function for searching for certain subtrees inside of a Firrtl AST (using partial functions so that you can use case statements and wildcards)

jackkoenig:
@edwardcwang Done. I've also added some scalatest helpers that leads to some slick test writing in my opinion, check out InfoSpec!

PR #684, opened 2017-11-10
Remove wires, replacing them with nodes

jackkoenig:
This isn't quite ready to merge, needs a few things

- [x] https://github.com/freechipsproject/firrtl/pull/653 must be merged first
- [x] Need to fix Flatten Transform test
- [x] This breaks the no-check-comb-loop test. This means no-check-comb-loop would not prevent an error if the combinational loop is within a single module. I think this is reasonable since the benefits outweigh the costs but it requires thought.
- [x] Fix Travis formal check because it currently requires 1:1 for all nodes in the compared designs

For the latter especially the real question is if we should ban wires from low firrtl. We don't technically have to, but not having wires makes *a lot* of things easier. Since combinational loops within a module this has implications for the no-check-comb-loops flag.


albert-magyar:
How does no-check-comb-loop work?

jackkoenig:
no-check-comb-loop just disables the check for combinational loops. Currently I think everything just kind of works with this disabled even though we generally assume no combinational loops. It is obviously not possible to replace a combinational loop using wires with flow-forward node logic so the question is how to deal with this.

To solve this in the short-term, we could just abort this transform if there's a combinational loop. Currently, this would just result in worse QoR from optimization passes which isn't that big of a deal. Ultimately I want to make low firrtl not allow wires which would simplify transforms quite a bit. To do that, we either need to accept that this is going to be a problem for people with false combinational loops (which currently easily can come from using Mux1H in Chisel for example), or we need to break these false combinational loops. One simple workaround could be to just insert a passthrough module to "break" the loop from the flow-forward logic perspective.

albert-magyar:
So it breaks no-comb-loop-check in a way totally unrelated to the combinational loop checker (which isn't even run), since it just fails to replace wires with nodes?

jackkoenig:
Yeah that's right. It only breaks if the combinational loop is within a module.

jackkoenig:
I think what I'm going to do is just abort removing wires if there's a combinational loop and do the work of banning wires from low firrtl in a future PR.

chick:
@jackkoenig as a passing note, the BitReduction pass I have been working on changes all the low firrtl nodes with wires (based on a suggestion from @azidar). Maybe this is all fine, that effectively I'm bumping my ingested low Firrtl to middle Firrtl and I can bump back down at the end.  But still seems worth mentioning.

jackkoenig:
@chick Thanks for the heads up. Replacing nodes with wires is perfectly fine and can make it easier to do certain things (like take advantage of last connect semantics to inject logic). I do believe that ultimately LowFirrtl should mandate no wires. If we do add that restriction there's nothing wrong with introducing them and then removing them again. We can make that decision later though, I'm not going to try to ban wires with this PR.

jackkoenig:
I finally got Travis to pass!

In summary, I improved the yosys check to be more robust, but then Travis would kill the check because it took too long. I have not yet figured out how to get Travis to not kill the check (I know about and tried travis_wait, but it seems that there's per-process stuff here and travis_wait goes on the wrapper script, not the yosys process so it gets killed anyway). Instead, I changed it from checking the Rob to checking a few smaller pieces from rocket that should nevertheless give similar if not better "coverage" of firrtl features.

jackkoenig:
@azidar so what do you think about merging this? It both removes wires and refactors the formal testing, I *could* try to split them up but it could get pretty gnarly

jackkoenig:
@azidar So it actually doesn't reorder as badly as you would assume since ExpandWhens does something similar with leftover logic followed by connections from when blocks followed by simulation constructs. It does a pretty good job of keeping non-dependent but nearby logic together. If you think about groups of nodes as "basic blocks" and connections to wires as "branches" it seems to move blocks of logic around but keeps the order inside intact (even if it doesn't have to). You can try out this PR vs. master on any of the code in regress and do a diff (vimdiff helps) to see what I mean.

jackkoenig:
I was able to clean the commit history up substantially, I'm going to merge this once CI passes

PR #692, opened 2017-12-01
Make Register Update Flattening a Transform and Delete Dangling Nodes

### LINKED ISSUE #681: Register Mux Tree formation leaves dead code in emitted Verilog
Consider:

```
circuit Test :
  module Test :
    input clock : Clock
    input cond : UInt<1>
    input x : UInt<8>
    output z : UInt<8>
    reg r : UInt<8>, clock
    when cond :
      r <= x
    z <= r
```

This results in (with randomization logic removed)
```verilog
module Test(
  input        clock,
  input        cond,
  input  [7:0] x,
  output [7:0] z
);
  reg [7:0] r;
  wire [7:0] _GEN_0;
  assign _GEN_0 = cond ? x : r;
  assign z = r;
  always @(posedge clock) begin
    if (cond) begin
      r <= x;
    end
  end
endmodule
```

Note the dangling `_GEN_0`.

jackkoenig:
Fixes #681 

This pulls the register update flattening out of the actual Emitter logic into a separate transform (that is run in the VerilogEmitter). It follows running that transform with DeadCodeElimination to elimination any dangling logic.

This currently fails 1 test, [the test for supporting registers with no ports](https://github.com/freechipsproject/firrtl/blob/40dda493a277f721306d428ee967dcb670813275/src/test/scala/firrtlTests/VerilogEmitterTests.scala#L123). The 2nd run of DeadCodeElimination fails because the first run deletes the only module. I'm thinking this test is nonsensical but maybe DCE should be careful not to delete the top module even if it has no ports?

I'm also wondering if the two runs of DCE are necessary, we could just run it once but it seems weird to do that in the Verilog compiler only rather than in the transform LowFirrtlOptimization. That being said, it really seems to me that the mux tree flattening should be part of the Verilog Emitter's set of transforms since it is really only needed for Verilog emission. Maybe there's some way to restructure transform ordering to make this cleaner...


jackkoenig:
Ensure `-no-dce` opt prevents both DCEs from running

azidar:
Is this ready for review?

src/main/scala/firrtl/transforms/FlattenRegUpdate.scala
```diff
@@ -0,0 +1,97 @@
...
+        case _: IsInvalid => throwInternalError
+        case _ => // Do nothing
+      }
+      stmt
+    }
+    mod.map(onStmt)
+    netlist
+  }
+
+  private def flattenReg(mod: Module): Module = {
```
azidar: Why is this private? Can you move this to a companion object and add scaladoc describing the behavior of the function?

src/main/scala/firrtl/transforms/FlattenRegUpdate.scala
```diff
@@ -0,0 +1,97 @@
...
+  *
+  * This transform flattens register updates into a single expression on the rhs of connection to
+  * the register
+  */
+// TODO Preserve source locators
+class FlattenRegUpdate extends Transform {
+  def inputForm = MidForm
+  def outputForm = MidForm
+
+  type Netlist = mutable.HashMap[WrappedExpression, Expression]
```
azidar: I'd move this to a companion object.

src/main/scala/firrtl/transforms/FlattenRegUpdate.scala
```diff
@@ -0,0 +1,97 @@
...
+
+  private def buildNetlist(mod: Module): Netlist = {
```
azidar: This too.

src/test/scala/firrtlTests/VerilogEmitterTests.scala
```diff
@@ -120,14 +120,4 @@ class VerilogEmitterSpec extends FirrtlFlatSpec {
       lines should contain (c)
     }
   }
-  "The Verilog Emitter" should "support Modules with no ports" in {
```
azidar: Why does this not work?

src/main/scala/firrtl/transforms/FlattenRegUpdate.scala
jackkoenig: Yeah I can do that

jackkoenig:
I've addressed the feedback!

src/main/scala/firrtl/transforms/FlattenRegUpdate.scala
```diff
@@ -0,0 +1,117 @@
...
+import firrtl.ir._
+import firrtl.Mappers._
+import firrtl.Utils._
+
+import scala.collection.mutable
+
+object FlattenRegUpdate {
+
+  /** Mapping from references to the [[Expression]]s that drive them */
+  type Netlist = mutable.HashMap[WrappedExpression, Expression]
```
azidar: We really need to formalize this into an analysis.

jackkoenig: Agreed, it pops up all over the place

PR #696, opened 2017-12-13
Fix bug in ConstProp where module dependency edges were dropped

jackkoenig:
This resulted in parent modules sometimes being constant proppagated
before a child module. If the child module has a constant driving one of
its outputs, the parent module would thus not see the constant. This
resulted in strange unstable constant propagation behavior where
sometimes constant outputs would not propagate.

Also add test illustrating why this occurs with uses of InstanceGraph

We could consider making transformNodes in DiGraph combine the edge sets as I clearly expected it to do. see the example in InstanceGraphTests as an illustration of what went wrong in ConstProp.

### DIFF
--- a/src/main/scala/firrtl/graph/DiGraph.scala
+++ b/src/main/scala/firrtl/graph/DiGraph.scala
@@ -302,7 +302,8 @@ class DiGraph[T] private[graph] (private[graph] val edges: LinkedHashMap[T, Link
     * @return a transformed DiGraph[Q]
     */
   def transformNodes[Q](f: (T) => Q): DiGraph[Q] = {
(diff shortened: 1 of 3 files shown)


jackkoenig:
This is only a "bug" insofar as instability in emitted Verilog should be considered incorrect behavior. There is no correctness bug here.

jackkoenig:
@albert-magyar and I agree this is actually a bug in DiGraph, will push a test and fix to this PR

albert-magyar:
I fixed `transformNodes` and made the test actually test the new, correct behavior.

jackkoenig:
Fixed a comment and added a test of the new behavior to DiGraph tests

PR #697, opened 2017-12-15
Change default source locator to "use" instead of "append"

jackkoenig:
I think for most users, the Firrtl source locators are not useful. I have told people about source locators in the emitted Verilog and gotten the response "Yeah, but it's just the Firrtl source locators, I need the Scala". The option to use append is still there on the command-line, but in general I think just using the Chisel ones is more useful.

### DIFF
--- a/src/main/scala/firrtl/ExecutionOptionsManager.scala
+++ b/src/main/scala/firrtl/ExecutionOptionsManager.scala
@@ -172,7 +172,7 @@ case class FirrtlExecutionOptions(
     inputFileNameOverride:  String = "",
     outputFileNameOverride: String = "",
     compilerName:           String = "verilog",
-    infoModeName:           String = "append",
+    infoModeName:           String = "use",
     inferRW:                Seq[String] = Seq.empty,
     firrtlSource:           Option[String] = None,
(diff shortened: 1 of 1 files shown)


edwardcwang:
How does it impact the usecase of Chisel -> .fir -> run some firrtl transforms separately (e.g. via command line)? Do both the Chisel and FIRRTL source locators show up?

jackkoenig:
In that case, it would still only be the Chisel source locators. With the current behavior on master, you would get 3 source locators, Chisel, Firrtl 1, and Firrtl 2.

edwardcwang:
Does it make sense to change it so that the default workflow for Chisel -> firrtl uses only the Chisel source locators, but if you run firrtl from the command it will use "append" and also include the firrtl source locators?

My concern is that if some extra transform gets run it will be harder to diagnose than if the firrtl locators were also included from the extra transform.

jackkoenig:
I think I'll just have rocket-chip change its use of Firrtl, this doesn't need to change

colinschmidt:
Wait doesn't append mean they get both? Why do they complain about having both?

PR #704, opened 2017-12-18
Refactor Annotations Driver API

jackkoenig:
This is the first part of the annotations refactor

Summary of changes:
- Add Driver.dramaticWarning for warning about features that will be removed
- Remove force-append-anno-file option (making it default behavior to append annotations from anno file)
  - This was basically required when using annotation files because otherwise using any command-line option that created annotations would prevent the annotation file from being used
- Fix emission of annotation file to actually use YAML
- Change Driver.loadAnnotations to return annotations instead of mutating firrtlOptions
  - Should this be done differently? Perhaps deprecate old function and just write a new one? Is anyone else using this function?
- Support multiple annotation files passed via -faf
- Deprecate (via loud warning) implicit annotation file from sharing name with design
- Deprecate overriding annotation filename (which can't be done from commandline anymore)
- Always include BlackBoxTargetDirAnno
  - (unclear why this was only included if there was at least one other anno?)

TODO
 - [x] This currently warns if you have an annotation file with the topName even if you include it explicitly with faf, worse still it loads the file twice. Need to fix this before merging

### DIFF
--- a/src/main/scala/firrtl/Driver.scala
+++ b/src/main/scala/firrtl/Driver.scala
@@ -196,7 +196,7 @@ object Driver {
         case OneFilePerModule(dirName) =>
           val emittedModules = finalState.emittedComponents collect { case x: EmittedModule => x }
           if (emittedModules.isEmpty) throwInternalError // There should be something
-          emittedModules.foreach { case module =>
+          emittedModules.foreach { module =>
             val filename = optionsManager.getBuildFileName(firrtlConfig.outputSuffix, s"$dirName/${module.name}")
             val outputFile = new java.io.PrintWriter(filename)
             outputFile.write(module.value)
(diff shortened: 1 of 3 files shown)


jackkoenig:
@chick I think the most debatable part of this PR is that it deprecates the "implicit" annotation file API (ie. given a topName, the file `$topName.anno` implicitly being read in). I am not fixed in wanting to remove it, I'll explain why I decided to below but I'm open to other ideas:

The annotations refactor needs to change the serialization format of annotations, specifically by adding type tags (aka type hints). This resulted in my changing the format from yaml to json because the tool support is just much better*. To ensure this breaks things as little as possible, I'm disambiguating with the old `.anno` style by using the suffix `.anno.json`. I thought it a little weird to have firrtl magically look for both `$topName.anno` and `$topName.anno.json` and implicitly combine these annotations with those given explicitly via `-faf`. In any case I wanted to support multiple annotation files because json files are a pain to combine.

*We *could* support yaml although it'd have to be a different yaml format than what we currently use

jackkoenig:
I integrated your changes into the appropriate commits

PR #706, opened 2017-12-19
Make submodule inputs void in ExpandWhens

### LINKED ISSUE #705: latest FIRRTL generates incorrect code in rocket-chip
See https://github.com/freechipsproject/rocket-chip/issues/1160

jackkoenig:
Fixes #705 

The ports of submodules were not voided in ExpandWhens which means it does not error if they are not properly initialized. Worse still, since they don't have a default value in the netlist, the logic in ExpandWhens just picks the first connection and ignores all the rest.

I've added a test for each of the above cases.

### DIFF
--- a/src/main/scala/firrtl/passes/ExpandWhens.scala
+++ b/src/main/scala/firrtl/passes/ExpandWhens.scala
@@ -103,11 +103,15 @@ object ExpandWhens extends Pass {
                       defaults: Defaults,
                       p: Expression)
                       (s: Statement): Statement = s match {
+        case stmt @ (_: DefNode | EmptyStmt) => stmt
         case w: DefWire =>
           netlist ++= (getFemaleRefs(w.name, w.tpe, BIGENDER) map (ref => we(ref) -> WVoid))
           w
         case w: DefMemory =>
           netlist ++= (getFemaleRefs(w.name, MemPortUtils.memType(w), MALE) map (ref => we(ref) -> WVoid))
+          w
+        case w: WDefInstance =>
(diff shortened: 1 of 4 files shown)


src/main/scala/firrtl/passes/ExpandWhens.scala
```diff
@@ -103,12 +103,16 @@ object ExpandWhens extends Pass {
...
                       (s: Statement): Statement = s match {
+        case stmt @ (_: DefNode | EmptyStmt) => stmt
         case w: DefWire =>
           netlist ++= (getFemaleRefs(w.name, w.tpe, BIGENDER) map (ref => we(ref) -> WVoid))
           w
         case w: DefMemory =>
           netlist ++= (getFemaleRefs(w.name, MemPortUtils.memType(w), MALE) map (ref => we(ref) -> WVoid))
           w
+        case w: WDefInstance =>
+          netlist ++= (getFemaleRefs(w.name, w.tpe, MALE).map(ref => we(ref) -> WVoid))
```
colinschmidt: `MALE` of `getFemaleRefs` here means to get all the inputs to the submodule?

colinschmidt: Clarified offline

jackkoenig: That's correct, similar to how it's used on the `DefMemory` above.

jackkoenig:
Due to https://github.com/freechipsproject/chisel3/issues/746 this currently doesn't work with rocket-chip, I am trying to get this fixed tonight

PR #710, opened 2017-12-20
Verify shl/shr amount is > 0.

### LINKED ISSUE #527: shl(x, -1) should issue an error
Instead, it generates the Verilog ```x << -1```.  (I wasn't sure what that meant, but my reading of the Verilog-2005 spec is that it is equivalent to ```x << 32'hFFFFFFFF```, since the -1 is taken as an unsigned number modulo its range (2^32, since it's an integer)).

ucbjrl:
Fixes #527

### DIFF
--- a/src/main/scala/firrtl/passes/Passes.scala
+++ b/src/main/scala/firrtl/passes/Passes.scala
@@ -198,6 +198,15 @@ object Legalize extends Pass {
       e
     }
   }
+  private def legalizeShiftLeft(e: DoPrim): Expression = {
+    require(e.op == Shl)
+    val amount = e.consts.head.toInt
+    if (amount < 0) {
+      error(s"Argument $amount < 0 for Primop Shift Left")
+     } else {
+      e
+    }
+  }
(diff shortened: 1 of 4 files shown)


src/main/scala/firrtl/passes/Passes.scala
```diff
@@ -198,6 +198,15 @@ object Legalize extends Pass {
...
   }
+  private def legalizeShiftLeft(e: DoPrim): Expression = {
+    require(e.op == Shl)
+    val amount = e.consts.head.toInt
+    if (amount < 0) {
+      error(s"Argument $amount < 0 for Primop Shift Left")
+     } else {
+      e
+    }
+  }
```
jackkoenig: I think this check should go in CheckHighForm. Legalize (admittedly poorly named) is intended to include constant prop-like logic that is required for Verilog emission (as opposed to just an optimization).

ucbjrl:
Should shr() have a similar check?

jackkoenig:
Yeah it should, mind adding it?

ucbjrl:
Will do.

PR #712, opened 2017-12-20
Make ExpandWhens implicitly invalidate submodule inputs (in 1.0.x)

jackkoenig:
This is a modified form of
https://github.com/freechipsproject/firrtl/pull/706 to preserve the
bugfix but preserve the [buggy] 1.0.x API that firrtl was not properly
enforcing initialization requirements for submodule inputs

The problem with #706 is that because Firrtl wasn't properly enforcing initialization requirements on submodules, code that used to work no longer does. This alternative version still fixes the incorrect Verilog generation bug but does not enforce the initialization requirement.

### DIFF
--- a/src/main/scala/firrtl/passes/ExpandWhens.scala
+++ b/src/main/scala/firrtl/passes/ExpandWhens.scala
@@ -103,11 +103,15 @@ object ExpandWhens extends Pass {
                       defaults: Defaults,
                       p: Expression)
                       (s: Statement): Statement = s match {
+        case stmt @ (_: DefNode | EmptyStmt) => stmt
         case w: DefWire =>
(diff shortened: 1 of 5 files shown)


jackkoenig:
The only real differences between this and #706 are that it uses `WInvalid` instead of `WVoid` on the submodule ports in ExpandWhens and the `CheckInitializationSpec` test does not actually expect an exception to be thrown.

jackkoenig:
I updated this to print a warning about invalidating submodule ports

azidar:
So is the idea that for 1.1.x, we will change the API to not automatically invalidating instance ports (e.g. using WVoid instead of WInvalid)?

azidar:
I'll approve it assuming this is what you are doing.

ucbjrl:
This has been implemented in 1.0.1 (and 1.0.x) without the unnecessary change to `src/test/scala/firrtlTests/FlattenTests.scala`.

Issue #693, opened 2017-12-07
Issue in converting to Verilog.


kevinkim06:
Hello,

I'm very new to FIRRTL. I'm just trying to convert the following FIRRTL to Verilog, but it fails with error 
> line 8:18 mismatched input '(' expecting DEDENT

```
circuit mod_example:
  module mod_example:
    input  indata: UInt<16>
    input  dem: UInt<16>
    output outdata: UInt<16>

    outdata <= mod(indata, dem)
```

Command I used is

> ./utils/bin/firrtl -tn top -i mod_example.fir -o mod_example.v

Could you tell me what's wrong with my FIRRTL? I'll appreciate.

Thank you.

jackkoenig:
It appears that the spec is quite out of date. Instead of `mod`, try `rem`.

kevinkim06:
I see. Thank you.

PR #713, opened 2017-12-24
Spec erroneously says mod instead of rem.

grebe:
Was browsing through old notifications and saw #693.


grebe:
I do not have the powers to merge

PR #714, opened 2017-12-26
Update ISSUE_TEMPLATE.md

azidar:
Added request for users to supply the FIRRTL directly which causes the bug.


edwardcwang:
Maybe "provide the FIRRTL file or a relevant fragment if the file is large" since firrtl circuits can often be very large (many MBs)?

azidar:
Good point, I'll update.

PR #735, opened 2018-01-30
Fix Bugs in Constant Propagation

jackkoenig:
In particular: 
Two instances of the same module could collide in counting the number of
instances of each Module. This could lead to constants being propagated
on inputs when it is incorrect to do so.

h/t @donggyukim for not only figuring out there was a problem, but providing a reduced test case

I also fixed another bug I came across although it is minor


donggyukim:
I confirm it fixes errors in my side.

Issue #700, opened 2017-12-15
LSb larger than MSb in bits does not error


jackkoenig:
The assignment just gets removed.

```
circuit bar :
  module bar :
    input in : UInt<8>
    output foo : UInt
    foo <= bits(in, 3, 4)
```

PR #738, opened 2018-02-07
Add high form check that MSB >= LSB

seldridge:
Adds a high form check for #700 (some low hanging fruit that I didn't see anybody working on for 1.1.0).

I'm assuming that `CheckHighForm` is the right place to be doing this. Alternatively, the check can be moved to `CheckWidths`. I'm unclear on what makes the most sense here...

### DIFF
--- a/src/main/scala/firrtl/passes/Checks.scala
+++ b/src/main/scala/firrtl/passes/Checks.scala
@@ -51,6 +51,8 @@ object CheckHighForm extends Pass {
     s"$info: A single module must be named $name.")
   class NegArgException(info: Info, mname: String, op: String, value: Int) extends PassException(
     s"$info: [module $mname] Primop $op argument $value < 0.")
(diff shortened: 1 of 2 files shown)


src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -51,6 +51,8 @@ object CheckHighForm extends Pass {
     s"$info: A single module must be named $name.")
   class NegArgException(info: Info, mname: String, op: String, value: Int) extends PassException(
     s"$info: [module $mname] Primop $op argument $value < 0.")
+  class LsbLargerThanMsb(info: Info, mname: String, op: String, lsb: Int, msb: Int) extends PassException(
```
kevintownsend: Should this be LsbLargerThanMsbException? It seems like the convention to and the "Exception" suffix to classes that extend Exception.

seldridge: Good catch, thanks. New PR inbound...

PR #740, opened 2018-02-08
[name change] Use LsbLargerThanMsbException

seldridge:
- Rename `LsbLargerThanMsb` to `LsbLargerThanMsbException` (@kevintownsend)


src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -51,7 +51,7 @@ object CheckHighForm extends Pass {
     s"$info: A single module must be named $name.")
   class NegArgException(info: Info, mname: String, op: String, value: Int) extends PassException(
     s"$info: [module $mname] Primop $op argument $value < 0.")
-  class LsbLargerThanMsb(info: Info, mname: String, op: String, lsb: Int, msb: Int) extends PassException(
+  class LsbLargerThanMsbException(info: Info, mname: String, op: String, lsb: Int, msb: Int) extends PassException(
```
kevintownsend: Nit: If the line limit is 100 characters you are over. Line break just before the "extends" keyword.

src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -90,7 +90,8 @@ object CheckHighForm extends Pass {
           correctNum(Option(1), 2)
           val (msb, lsb) = (e.consts(0).toInt, e.consts(1).toInt)
           if (lsb > msb) {
-            errors.append(new LsbLargerThanMsb(info, mname, e.op.toString, lsb, msb))
+            errors.append(
```
kevintownsend: Nit: If the line limit is 100 characters you don't need a line break.

src/main/scala/firrtl/passes/Checks.scala
jackkoenig: I think 120 is the normal Scala-style limit, I've been moving toward 100 but the official limit remains 120

src/main/scala/firrtl/passes/Checks.scala
seldridge: I'm in agreement. This is my internal >80 alarm going off. I need to setup per-project settings...

PR #743, opened 2018-02-13
Stop DiGraph.BFS reporting its root

### LINKED ISSUE #742: DiGraph.BFS doesn't treat the root node as visited
* **Type of issue**
  - [x] Bug report
  - [ ] Feature request
  - [ ] Other enhancement

* **If the current behavior is a bug, please provide the steps to reproduce the problem:**

I would expect that `DiGraph.BFS` when run on a cyclic graph should not return its root. The following test case will fail:

```scala
class DiGraphTests extends FirrtlFlatSpec {

  val cyclicGraph = DiGraph(Map(
    "a" -> Set("b","c"),
    "b" -> Set("d"),
    "c" -> Set("d"),
    "d" -> Set("a")))

  "a cyclic graph" should "not report its root node during a BFS" in {
    cyclicGraph.BFS("a").keys should not contain "a"
  }
}
```

The current BFS for this is `Map(b -> a, c -> a, d -> b, a -> d)`.

I'd expect to see `Map(b -> a, c -> a, d -> b)`.

The BFS should be reporting visited nodes, not visited edges.

* **Impact**
  - [ ] no functional change
  - [x] API modification (__technically__)
  - [ ] unknown

* **Development Phase**
  - [ ] request
  - [x] proposal

I'll add a PR with a simple fix momentarily.

seldridge:
Proposal to fix #742.

### DIFF
--- a/src/main/scala/firrtl/graph/DiGraph.scala
+++ b/src/main/scala/firrtl/graph/DiGraph.scala
@@ -115,7 +115,7 @@ class DiGraph[T] private[graph] (private[graph] val edges: LinkedHashMap[T, Link
     while (!queue.isEmpty) {
       val u = queue.dequeue
       for (v <- getEdges(u)) {
-        if (!prev.contains(v)) {
+        if (!prev.contains(v) && v != root) {
           prev(v) = u
           queue.enqueue(v)
         }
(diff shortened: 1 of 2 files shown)


seldridge:
Closing due to violating original intent of the `BFS` method. See: https://github.com/freechipsproject/firrtl/issues/742#issuecomment-365141176

PR #744, opened 2018-02-13
Add graph summation "+" to DiGraph

seldridge:
I'm working through some circuit transformations and am coming across some operations which I'd like to have for graphs. This is the first. This takes two graphs and creates a new graph that includes all vertices and edges as defined by both (a mathematical union).

If this makes sense, this can either be accepted directly or broadened to include more operations that would make sense to have for `DiGraph`.

Comments:
  - I think this is the correct terminology (summation)
  - This is perhaps more intuitive as `++`, but that may imply "graph union" which is generally something different
  - This may be unintuitive for the user as applying `+` to two `MutableDiGraph` will return a `DiGraph`


albert-magyar:
This looks reasonable to me, but I'm not sure whether it preserves the fact that the current `DiGraph` API avoids breaking deterministic ordering to make it more useful in cases where it generates hardware. Have you thought about how this could be mapped onto `LinkedHashMaps` and `LinkedHashSet`.

seldridge:
You're entirely right. This would introduce non-determinism (and after the big non-determinism refactor this would be a step backwards as is).

I pushed a commit that should be deterministic using `LinkedHashSet` and exploiting the fact that the `map` iterator should be ordered (or, doing things in the same way that you did for the determinism refactor).

I'm now testing this for the two graph summation tests with a method that compares graph ordering as opposed to usual `Map` structural equality. In order to test this, I needed to change the example test graphs to use ordered maps/sets to make the test deterministic.

albert-magyar:
Cool! I do think that `+` is a useful addition, thanks for doing this.

One thing to note about determinism tests, which you may already know: unlike other languages where hash data structures _always_ produce distinct orderings to avoid algorithmic complexity attacks, Scala `HashMap`s are non-deterministic only because they don't go out of their way to preserve ordering under resizing or GC. Therefore, you often need a pretty sizable test to observe non-determinism.

seldridge:
That also makes sense about the non-determinism. What I'm talking about there is incorrect, then. I was attempting to come up with a test case that would fail for the incorrect fix and pass for the correct fix. What I was capturing was the unorderedness of Map/Set vs. the linked variants. That's likely unnecessary and it just clutters up the test case. For the sake of simplicity, I'll remove it.

azidar:
Is this ready for review? If so, @albert-magyar can you review it? Thanks!

PR #745, opened 2018-02-20
Add Metadata type for holding annotations and renames

jackkoenig:
Move those fields from CircuitState to Metadata

Here's Metadata, PR-ed against the refactor annotations branch

Some things get kind of annoying to express, like adding annotations to the ones already there:
```scala
val newAnnos = ...
// Before
state.copy(annotations = state.annotations ++ newAnnos)
// After
state.copy(metadata = state.metadata.copy(annotations = state.metadata.annotations ++ newAnnos))
```

Thoughts?


src/main/scala/firrtl/Compiler.scala
```diff
@@ -105,20 +105,28 @@ object AnnotationSeq {
...
+/** Metadata associated with a Firrtl circuit
+  *
+  * @param annotations The current collection of [[firrtl.annotations.Annotation Annotation]]
+  * @param renames A map of [[firrtl.annotations.Named Named]] things that have been renamed.
+  *        Generally only a return value from [[Transform]]s
+  */
+case class Metadata(annotations: Seq[Annotation], renames: Option[RenameMap])
+object Metadata {
+  def apply(annotations: Seq[Annotation]) = new Metadata(annotations, None)
+}
```
azidar: Can you add a 
```scala
def add(newAnnos: Seq[Annotation]): Metadata = apply(annotations ++ newAnnos)
```

src/main/scala/firrtl/Compiler.scala
```diff
@@ -105,20 +105,28 @@ object AnnotationSeq {
...
-  * @param renames A map of [[firrtl.annotations.Named Named]] things that have been renamed.
-  *   Generally only a return value from [[Transform]]s
+  * @param metadata Holds auxiliary information related to the circuit
   */
 case class CircuitState(
     circuit: Circuit,
     form: CircuitForm,
-    annotations: AnnotationSeq,
-    renames: Option[RenameMap]) {
+    metadata: Metadata) {
```
azidar: What do you think about pulling form into Metadata? I think it makes sense, as its actually metadata on the circuit.

src/main/scala/firrtl/passes/Uniquify.scala
```diff
@@ -370,7 +370,7 @@ object Uniquify extends Transform {
 
     sinfo = c.info
     val result = Circuit(c.info, c.modules map uniquifyPorts(renames) map uniquifyModule(renames), c.main)
-    CircuitState(result, outputForm, state.annotations, Some(renames))
+    CircuitState(result, outputForm, state.metadata.copy(renames = Some(renames)))
```
azidar: I'd add another `def add(newRenames)` to simplify this syntax.

src/main/scala/firrtl/transforms/DeadCodeElimination.scala
```diff
@@ -316,17 +316,17 @@ class DeadCodeElimination extends Transform {
     // Preserve original module order
     val newCircuit = c.copy(modules = c.modules.flatMap(m => modulesxMap.get(m.name)))
 
-    state.copy(circuit = newCircuit, renames = Some(renames))
+    state.copy(circuit = newCircuit, metadata = state.metadata.copy(renames = Some(renames)))
```
azidar: we could add `def annotations: Seq[Annotation] = metadata.annotations` to CircuitState, which would simplify this syntax as well.

jackkoenig:
We decided against this approach, instead we're going to make some changes to CircuitState.

PR #746, opened 2018-02-23
Adding the firrtl proto.

kevintownsend:
This is just the initial upload. This needs to be built by sbt or whatever the build system will be.

Primary for better space performance. However, other reasons exist. Protos seem like a good choice for a multi-language machine readable IR.

### DIFF
--- a/src/main/proto/firrtl.proto
+++ b/src/main/proto/firrtl.proto
@@ -1,8 +1,9 @@
 syntax = "proto3";
 
-package hardware.chips.chisel3;
+package firrtl;
 
-option java_package = "hardware.chips.chisel3.firrtl";
+option java_package = "firrtl";
+option java_outer_classname = "FirrtlProtos";
 
 message Firrtl {
   repeated Circuit circuit = 1;
@@ -148,46 +149,6 @@ message Firrtl {
       Expression en = 4;
     }
 
-    message InlineVerilog {
-      // Required.
-      repeated InlineVerilogPiece pieces = 1;
-
-      message InlineVerilogPiece {
-        oneof contents {
-          string literal = 1;
-          Expression expression = 2;
-        }
-      }
-    }
-
-    message Assert {
-      // Required.
-      Expression clk = 1;
-      // Required.
-      Expression reset = 2;
-      // Required.
-      Expression condition = 3;
-      // Required.
-      string label = 4;
-      // Required.
-      bytes message = 5;
-      repeated Expression message_arg = 6;
-      // Required.
-      string error_level = 7;
-      // Required.
-      bool ignore_unknowns = 8;
-    }
-
-    message Cover {
-
-      // Required.
-      Expression condition = 1;
-      // Required.
-      string label = 2;
-      Expression clock = 3;
-      Expression reset = 4;
-    }
-
     message Skip {
       // Empty
     }
@@ -245,9 +206,6 @@ message Firrtl {
       When when = 7;
       Stop stop = 8;
       Printf printf = 10;
-      InlineVerilog inline_verilog = 11;
-      Assert assert = 12;
-      Cover cover = 13;
       Skip skip = 14;
       Connect connect = 15;
       PartialConnect partial_connect = 16;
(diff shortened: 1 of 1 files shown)


src/main/proto/firrtl.proto
```diff
@@ -0,0 +1,451 @@
...
+      // Required.
+      bytes value = 1;
+      repeated Expression arg = 2;
+      // Required.
+      Expression clk = 3;
+      // Required.
+      Expression en = 4;
+    }
+
+    message InlineVerilog {
```
azidar: InlineVerilog, Assert, and Cover haven't been added yet, so perhaps just comment them out with a TODO? Or add an issue to discuss these new constructs, and reference the issue in the comment?

src/main/proto/firrtl.proto
```diff
@@ -0,0 +1,451 @@
+syntax = "proto3";
+
+package hardware.chips.chisel3;
```
azidar: Should change package, maybe to "proto"? Not sure...

src/main/proto/firrtl.proto
```diff
@@ -0,0 +1,451 @@
...
+
+option java_package = "hardware.chips.chisel3.firrtl";
```
azidar: Change this too?

src/main/proto/firrtl.proto
```diff
@@ -0,0 +1,451 @@
...
+    // Required.
+    string name = 1;
+  }
+
+  message Circuit {
+    repeated Module module = 1;
+    repeated Top top = 2;
+  }
+
+  message Module {
```
azidar: If we want to match the scala ir, then Module -> DefModule, and UserModule -> Module.

src/main/proto/firrtl.proto
```diff
@@ -0,0 +1,451 @@
...
+
+    // Required.
+    oneof module {
+      ExternalModule external_module = 1;
+      UserModule user_module = 2;
+    }
+  }
+
+  message Statement {
+    message Wire {
```
azidar: Similarly, Wire -> DefWire, Register -> DefRegister, etc.

src/main/proto/firrtl.proto
```diff
@@ -0,0 +1,451 @@
...
+
+    message SIntType {
+      Width width = 1;
+    }
+
+    message ClockType {
+      // Empty.
+    }
+
+    message RecordType {
```
azidar: RecordType -> BundleType

src/main/proto/firrtl.proto
```diff
@@ -0,0 +1,451 @@
...
+    message Mux {
+      // Required.
+      Expression condition = 1;
+      // Required.
+      Expression t_value = 2;
+      // Required.
+      Expression f_value = 3;
+    }
+
+    message ExtractBits {
```
azidar: Not sure why this is separate from PrimOp?

kevintownsend: Fine by me. I am not sure what was behind that decision. Similarly Mux, SubIndex, SubAccess could be changed to.

src/main/proto/firrtl.proto
kevintownsend: Done.

src/main/proto/firrtl.proto
kevintownsend: Done.

src/main/proto/firrtl.proto
kevintownsend: Done.

src/main/proto/firrtl.proto
kevintownsend: Deleting for now.

kevintownsend:
Can you be explicit about what you want with a "serialization/deserialization test, as well as integrating with Driver."? Ideally I would like to keep PRs small and manageable.

src/main/proto/firrtl.proto
azidar: Oh, it stands for "Definition", but for better or for worse, its the name of the corresponding firrtl ir node (in firrtl/ir/IR.scala).

azidar:
With the PR, I'd like to see the code actually running in a simple test. That means adding a test in src/test/proto, demonstrating the serialization/deserialization of test into the proto representation, and back.

PR #750, opened 2018-03-02
Fix annotation deserialization of component subfields

jackkoenig:
Also make InvalidAnnotationFileException extend FIRRTLException for better
error reporting

I'm honestly amazed this hasn't come up before (unless I introduced it in the annotation refactor), but I don't care enough to do any archeology.


jackkoenig:
TLDR using tokenize here would result in a match error on component subfields.

Issue #504, opened 2017-03-21
stale black_box_verilog_files.f may cause test failures


ucbjrl:
`BlackBoxSourceHelper.execute()` creates a `black_box_verilog_files.f` in the target directory iff there are `BlackBoxResource` or `BlackBoxInline` annotations.
`chisel3.iotesters.verilogToVCS()` unconditionally adds a `-f black_box_verilog_files.f` to the vcsFlags if that file exists.
This can cause tests following a black box test to fail when the file(s) mentioned in `black_box_verilog_files.f` no longer exist.
```
[info] [0.000] Elaborating design...
[info] [0.158] Done elaborating.
verilator --cc doohickey.v -f .../test_run_dir/black_box_verilog_files.f --assert -Wno-fatal -Wno-WIDTH -Wno-STMTDLY --trace -O1 --top-module doohickey +define+TOP_TYPE=Vdoohickey +define+PRINTF_COND=!doohickey.reset +define+STOP_COND=!doohickey.reset -CFLAGS -Wno-undefined-bool-conversion -O1 -DTOP_TYPE=Vdoohickey -DVL_USER_FINISH -include Vdoohickey.h -Mdir .../test_run_dir --exe .../test_run_dir/doohickey-harness.cpp
%Error: Cannot find file containing library module: .../test_run_dir/BlackBoxTest.v
%Error: This may be because there's no search path specified with -I<dir>.
%Error: Looked in:
%Error:       .../test_run_dir/BlackBoxTest.v
%Error:       .../test_run_dir/BlackBoxTest.v.v
%Error:       .../test_run_dir/BlackBoxTest.v.sv
%Error:       .../test_run_dir/.../test_run_dir/BlackBoxTest.v
%Error:       .../test_run_dir/.../test_run_dir/BlackBoxTest.v.v
%Error:       .../test_run_dir/.../test_run_dir/BlackBoxTest.v.sv
%Error: Exiting due to 9 error(s)
[...]

PR #758, opened 2018-03-12
Clear out stale files when makeTargetDir is called. Chisel3 Issue #504

chick:
Addresses Chisel3 Issue #647 where stale file cause
unpredictable behavior

A whole bunch of tests played games by creating the work directory
manually then putting resource files in, then calling Driver.execute
which called makeTargetDir which removed the resource just loaded.
This kind of custom directory management requires the new
preserveTargetDir to be set to true

added new common option "preserveTargetDir"
getBuildFileName was calling makeTargetDir, this seems excessive and
caused lot's of problems because target dir was cleared each time.

Added tests to DriverSpec that demonstrate default clearing and how
to disable that.

### DIFF
--- a/src/main/scala/firrtl/Driver.scala
+++ b/src/main/scala/firrtl/Driver.scala
@@ -4,10 +4,11 @@ package firrtl
 
 import scala.collection._
 import scala.io.Source
-import scala.sys.process.{BasicIO,stringSeqToProcess}
-import scala.util.{Try, Success, Failure}
+import scala.sys.process.{BasicIO, stringSeqToProcess}
+import scala.util.{Failure, Success, Try}
 import scala.util.control.ControlThrowable
 import java.io.{File, FileNotFoundException}
+import java.nio.file.Files
 
 import net.jcazevedo.moultingyaml._
 import logger.Logger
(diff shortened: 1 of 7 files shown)


chick:
Good:
* Should fix reported problem of stale bad verilog files corrupting builds
* Removing all files in a work directory is the default behavior
* Remove attempted work directory creation every time build related path generated.

To Review carefully:
* Custom file injection into work directories, may not work by default.
* **Important:** Deletion of files creates possibilities for injection attacks
  * I tried to cover obvious cases of not deleting `/` or the project current directory.
  * You may wonder what caused me to add the protection against current directory deletion.

azidar:
Looks like this fails the tests:
`Error: delete directory /home/travis/build/freechipsproject/firrtl/chisel3/test_run_dir/WithResetTest/201803120615425335124662382660362 will not delete absolute paths`

azidar:
Why delete the target directory, and not just all files within it?

edwardcwang:
May have to be careful if targetDir is `.` or something weird like that.

azidar:
Maybe instead, have an option to specify the file extension to delete. So you could say
```--remove .v --remove .cpp```
and it would recursively delete those types of files it sees?

chick:
I think this needs a quick dev discussion because:
* Inconsistent work directory creation  make it hard to find a place to implement this.
* While sensible, requiring a flag to turn this on falls short of goal of protecting developers from the in-obvious affects of stale verilog hanging around.
* Removing directories is risky, user typographical errors can delete wrong things.
* restricting this to specific suffixes may be better

azidar:
Is this still relevant?

chick:
This is stale, and the implementation did not really resolve the driving issue. The driving issue has been closed earlier

PR #760, opened 2018-03-14
Pass up annotations in return value from Driver.execute

chick:
Backward compatible with existing usage.
Adds CircuitState to FirrtlExecutionSuccess, but
that member is not part of the unapply.
"To a single file per module if OneFilePerModule is specified"
test shows example of getting access to annotations

### DIFF
--- a/src/test/scala/firrtlTests/DriverSpec.scala
+++ b/src/test/scala/firrtlTests/DriverSpec.scala
@@ -355,7 +355,12 @@ class DriverSpec extends FreeSpec with Matchers with BackendCompilationUtilities
                                                  emitOneFilePerModule = true)
         }
 
-        firrtl.Driver.execute(manager)
+        firrtl.Driver.execute(manager) match {
+          case success: FirrtlExecutionSuccess =>
+            success.circuitState.annotations.length should be > (0)
+          case _ =>
+
+        }
 
         for (name <- expectedOutputFileNames) {
           val file = new File(name)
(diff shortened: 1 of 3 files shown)


ducky64:
Looks usable, though @azidar might want to weigh in.

What are the semantics in the different cases a FirrtlExecutionSuccess could be generated? Will the result differ based on the output form? What does it mean when Verilog output is requested?

Longer-term (API design wise), can we not have strings (unless the target specifically is Verilog code)? And especially not string-types (`"high", "middle", "low", "verilog", or "sverilog"`), unless the firrtl compiler is actually internally stringly-typed in that regard?

src/test/scala/firrtlTests/DriverSpec.scala
```diff
@@ -355,7 +355,12 @@ class DriverSpec extends FreeSpec with Matchers with BackendCompilationUtilities
                                                  emitOneFilePerModule = true)
         }
 
-        firrtl.Driver.execute(manager)
+        firrtl.Driver.execute(manager) match {
+          case success: FirrtlExecutionSuccess =>
+            success.circuitState.annotations.length should be > (0)
+          case _ =>
```
kevintownsend: The test should fail if this reached right?

azidar:
You need to fix for the test. Why don't you add another unapply with three arguments?

src/test/scala/firrtlTests/DriverSpec.scala
chick: The firrtl passes leave a number of annotations in the CircuitState, I think there 4 or 5 laying around in here.

chick:
@azidar You can't have overloaded functions (int this case unapply) with the same arguments but different return results.

PR #763, opened 2018-03-19
Masks for zero-width fields of mems should be width zero.

### LINKED ISSUE #666: Mems with 0-width entries can't be assigned to
The following firrtl crashes the compiler:
```
circuit BadMem :
  module BadMem :
    input clk  : Clock
    input addr : UInt<4>
    input in   : {0: UInt<10>, 1: UInt<0>}

    cmem ram : {0: UInt<10>, 1: UInt<0>}[16]
    infer mport ramin = ram[addr], clk

    ramin.0 <= in.0
    ramin.1 <= in.1
```

I get the error `java.util.NoSuchElementException: key not found: ram_1`. I think this should be legal firrtl.

FWIW, if you switch in to an output and flip the assignments, everything works fine.

grebe:
Closes #666.

### DIFF
--- a/src/main/scala/firrtl/passes/memlib/MemUtils.scala
+++ b/src/main/scala/firrtl/passes/memlib/MemUtils.scala
@@ -46,6 +46,7 @@ object createMask {
   def apply(dt: Type): Type = dt match {
     case t: VectorType => VectorType(apply(t.tpe), t.size)
     case t: BundleType => BundleType(t.fields map (f => f copy (tpe=apply(f.tpe))))
+    case GroundType(w) if w == IntWidth(0) => UIntType(IntWidth(0))
     case t: GroundType => BoolType
   }
 }
--- a/src/test/resources/features/ZeroWidthMem.fir
+++ b/src/test/resources/features/ZeroWidthMem.fir
@@ -1,0 +1,28 @@
+; See LICENSE for license details.
+circuit ZeroWidthMem :
+  module ZeroWidthMem :
+    input clock : Clock
+    input reset : UInt<1>
+    input waddr : UInt<4>
+    input in   : {0: UInt<10>, 1: UInt<0>}
+    input raddr : UInt<4>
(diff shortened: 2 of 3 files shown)


src/main/scala/firrtl/passes/memlib/MemUtils.scala
```diff
@@ -46,6 +46,7 @@ object createMask {
   def apply(dt: Type): Type = dt match {
     case t: VectorType => VectorType(apply(t.tpe), t.size)
     case t: BundleType => BundleType(t.fields map (f => f copy (tpe=apply(f.tpe))))
+    case GroundType(w) if w == IntWidth(0) => UIntType(IntWidth(0))
```
grebe: Is there a nicer syntax for this?

chick: How about
```case GroundType(IntWidth(0)) => UIntType(IntWidth(0))```
?

grebe: You'd think so, but the unapply for intwidth wants a BigInt, and `GroundType(IntWidth(BigInt(0)))` gave me some other weird error that I didn't really understand (maybe `BigInt` has its own unapply doing something weird?).

azidar: Yeah, it doesn't work. What you can do is somewhere define:
```val zeroW = IntWidth(0)```
and then you can do:
```case GroundType(`zeroW`) => ...```

azidar:
Either merge it when you are ready, or if you don't have write access, ping me and I'll do it.

src/main/scala/firrtl/passes/memlib/MemUtils.scala
grebe: Ah, good point- I rarely use the backticks, but this is a good example where it would be good. I think it's a wash and I'll just leave the existing syntax. Thanks!

PR #765, opened 2018-03-19
Annotations as Options Replacement, Immutable Option Parsing

### LINKED ISSUE #485: Funny Behavior when topName isn't specified

```scala
    val input =
      """circuit Top :
        |  module Top :
        |    input foo : UInt<32>
        |    output bar : UInt<32>
        |    bar <= foo
        """.stripMargin
                                         firrtlSource = Some(input))
    }
```

### LINKED ISSUE #552: Approach for Providing Parameters to Custom FIRRTL Passes
Does there exist a method for communicating parameters to a custom FIRRTL pass or are these expected to be provided entirely by annotations?

My knee-jerk reaction is to add support for some command line argument string that is blindly given to all custom passes (or to the pass which preceded it). However, I think this can be equivalently done entirely with annotations.

Is there an advised approach here?

### LINKED ISSUE #558: Add Support for Custom Command-Line Options
It would be really cool if transforms could register their own command-line options--at least boolean and single argument string options. I find that directory names and filenames are super common arguments to transformations and it would be great to be able to do this on the command-line.

I think the API for this is pretty simple: override a function that registers your commandline options. Any transformations passed to firrtl via the `-fct` option will have this function called (and possibly any transforms instantiated via an annotation present in the annotation file). Firrtl would obviously check for no option collisions, although it would be cool if transformations could also *share* an option.

### LINKED ISSUE #683: Command-line option class-log-level should be allowed more than once
I think it would be better if I could write:
`-cll firrtl.transforms.CheckCombLoops:Info -cll firrtl.transforms.DeadCodeElimination:Info` instead of having to write `-cll firrtl.transforms.CheckCombLoops:Info,firrtl.transforms.DeadCodeElimination:Info`

### LINKED ISSUE #764: [RFC] Annotations subsume ExecutionOptions?

* **Type of issue**
  - Other enhancement

* **Impact**

* **Development Phase**
  - proposal

seldridge:
#### Overview
Annotations and command line options should be fungible. A user calling the FIRRTL compiler should be able to interact with the compiler in the same way that an earlier frontend that produces annotations, but does not interact with a `Driver` (and vice versa). This does the following high level refactors:
  - The old mutable scopt parsing is replaced with immutable parsing
  - All command line options are parsed into an `AnnotationSeq`

The latter change enables immutable option parsing to be split across disjoint traits while using the same parser (e.g., `HasFirrtlOptions`, `HasChiselExecutionOptions` both parse using a shared `OptionParser[AnnotationSeq]`). The drivers or transforms that consume command line options are then expected to extract what they need from the command-line created `AnnotationSeq`, deleting what is specific to them. This then stages for a logical refactor of `Driver` subclassing `Transform`.

#### API changes
  - Removes methods for mutating options (e.g., `setTopName`)
  - A warning (as opposed to an error) results if the implicit annotation file is not found
  - All custom transforms must be globally visible (no anonymous transforms)
  - All custom transforms will be constructed (specific transform _instances_ cannot be run)
  - An `ExecutionOptionsManager` can only be constructed using `(String, Array[String])`
  - Case classes for driver options (e.g., `FirrtlOptions`) use `Option[T]` to store optional data, as opposed to empty strings
  - `CommonOptions` no longer exists and has been rolled into `FirrtlExecutionOptions`

#### API Deprecations
  - `FirrtlExecutionOptions.{getInputFileName, getOutputConfig, getTargetFile}` (moved to `HasFirrtlOptions`)

#### API additions
  - Annotation files may include annotation files via `InputAnnotationFileAnnotation` with automatic include guard checking 
  - Adds `Driver.execute(args: String)`
  - `HasFirrtlOptions.{getInputFileName, getOutputConfig, getTargetFile}`

#### Miscellaneous changes
  - Checks that no long or short options conflict with each other (with tests)
  - Bump scopt from 3.6.0 to 3.7.0 as we need some new methods to extract information to run the checks in the preceding bullet
  - Factors out `HasFirrtlOptions` from ExecutionOptionsManager.scala into FirrtlExecutionOptions.scala
  - Options vals (`commonOptions`, `firrtlOptions`) are necessarily lazily evaluated
 - Lots of documentation

#### Issues
- Fixes #764 
- Fixes #485 
- Fixes #683 
- Fixes #552 
- Fixes #558 

#### Todo
- [x] Align with Chisel3
  - [ ] Remove `[skip chisel tests]` (I'm not sure this can be merged without this...)
- [x] Tests
  - [x] Multiple Annotations/Transforms try to claim the same option
  - [x] Check that we catch a recursive annotation file import
- [x] Read annotation files directly during option parsing as opposed to delaying this until the Driver
- [x] Documentation
- [x] Cleanup
- ~[ ] All FIRRTL options should be deleted after use~ (I'm opting to leave this in for now---it's unclear to me if other transforms need these)
- [x] Integrate @chick toy-options registration
- [x] Remove usage of `ExecutionOptionsManager with X with Y with Z` as a parameter
- ~[ ] Seal `Annotation`s that should not be emitted by `Transform`s, e.g., `RunFirrtlTransform`~
    - I don't think that we can reasonably assert what a user will want to do here. Also, there are complications with `RunFirrtlTransformAnnotation` as Chisel's `Driver` uses this. Similar logic applies for "Driver-only" annotations, like `TopNameAnnotation`. A transform could reasonably change this. There's a lot of power here, but I think the better approach would be to roll this into adding checks for weirdness at the form levels.


jackkoenig:
We should also think about possible command-line option collisions that could occur if two custom transforms (likely in different projects), try to claim the same option. I'm sure scopt errors when this occurs, but I wonder if there's some way we can prevent this from happening. Would it be possible to enforce that custom transforms have a prefix on their option or something? This might be really hard or onerous to do, but maybe it's possible!

seldridge:
Good thought, @jackkoenig. While I think that this will be unlikely for some time, it's a huge issue... 

Surprisingly, scopt _does not_ error for this... All that appears to be happening is the last callback is the one that wins. This, unfortunately, appears to happen entirely silently...

I think that the only way to utterly disambiguate things is to use the full class path. That seems reasonable to me, even if it's verbose. It's just a question of whether this should be enforced and how it should be enforced.

seldridge:
Scala 2.12.4 seems to be bumping up against Travis memory limits. The closets that I've gotten to replicating is that I can locally run myself out of metaspace if I repeatedly run `tests` in the same `sbt` session. Not sure if this is what I'm doing or something else. The problems with 2.12.4 clearly seem to start with the `ClassFind` commits in the patch set.

jackkoenig:
Did you remove the class-finder stuff due to 2.12.4 issues with Travis memory limits?

seldridge:
@jackkoenig: No, not motivated by that (should have clarified in the commit). After discussing this last week, the consensus was that this was not the right approach for the following two reasons:
  - It's way too slow generally if you have to look through all classes. Theoretically, this should work and be fast to find options in arbitrary libraries pulled in from a Maven repo. However, `ClassFinder` churning through every possible class searching for any transform that may provide options is abhorrently slow. I was skirting around this by only looking inside FIRRTL. 
  - There's got to be a better way than to inspect everything. Reflection can very quickly find anything that  mixes in a `sealed trait` via `knownDirectSubclasses`, but that's not suitable here either as the traits can be anywhere.

Summarily, everyone thought that there should exist a cleaner solution and that this PR should be separated from that. I'm working to get this more towards only dealing with refactoring the options manager to provide:
1. Immutability of option parsing
2. Representation of all options as annotations (i.e., the parser is an `OptionParser[AnnotationSeq]`)

jackkoenig:
@seldridge Ah, sorry I missed the meeting--had a big paper push going on.

I'm totally down with automatically finding such transforms to a separate PR since this it is a fairly separate piece of work. It also relates to potentially improving annotation deserialization and some way of automatically instantiating transforms based on the presence of certain annotations (ie. moving that functionality from Chisel to FIRRTL).

Just for my own understanding, **how** slow is it? I wanted to use ClassFinder for finding classes that extend Annotation, but if it's really that slow then I guess it won't work. As you alluded to above, maybe it will be fast enough if we can be judicious in which classes we search through (wg. explicitly exclude sbt, scala, and specific plugins)...

seldridge:
It's actually not that bad, but it was an annoying shock from not using it.

It takes about 2 seconds to do the class path search with `ClassFinder`. That is searching everything locally (including ~/.ivy2). It speeds up a little once things are cached (e.g., when running a test suite).

Now that I think about it, this may not be all that bad from a usability perspective. It's basically the same as `import tensorflow`. This is assumedly linear with respect to the number of classes, however... Larger projects would get a larger hit.

jackkoenig:
I see. Well we can take a look at a later time! This PR looks to be shaping up well!

src/main/scala/firrtl/ExecutionOptionsManager.scala
```diff
@@ -2,584 +2,60 @@
...
-    Some((arg.emitType, arg.emitted))
+sealed class AugmentedOptionParser(name: String) extends OptionParser[AnnotationSeq](name) {
+  var terminateOnExit = true
+  override def terminate(exitState: Either[String, Unit]): Unit =
+    if (terminateOnExit) sys.exit(0)
+
+  /* Check that no long or short option declarations are duplicated */
+  override def parse(args: Seq[String], init: AnnotationSeq): Option[AnnotationSeq] = {
+    val longDups = options.map(_.name).groupBy(identity).collect{ case (k, v) if v.size > 1 && k != "" => k }
+    val shortDups = options.map(_.shortOpt).flatten.groupBy(identity).collect{ case (k, v) if v.size > 1 => k }
```
jackkoenig: `.map(...).flatten` => `.flatMap(...)`

Also I didn't know you could do `.groupBy(identity)` cool! I usually do `.groupBy(x => x)` but I like `groupBy(identity)` better

src/main/scala/firrtl/FirrtlExecutionOptions.scala
```diff
@@ -0,0 +1,752 @@
...
+import net.jcazevedo.moultingyaml._
+import firrtl.annotations.AnnotationYamlProtocol._
+import scala.util.{Try, Failure}
+
+/** Indicates that a subclass is an [[annotations.Annotation]] with an
+  * option consummable by [[HasFirrtlExecutionOptions]]
+  *
+  * This must be mixed into a subclass of [[annotations.Annotation]]
+  */
+trait FirrtlOption { self: Annotation => }
```
jackkoenig: I think the use of a type here merely for partitioning things below that will then be fed into a massive match statement anyway is a little weird. I think I might be neater if `FirrtlOption` has a virtual method `consume(opt: FirrtlExecutionOptions): FirrtlExecutionOptions ` or something that performs the work in the match below, then just call that method instead of matching.

src/main/scala/firrtl/FirrtlExecutionOptions.scala
```diff
@@ -0,0 +1,752 @@
...
+  * @param value compiler name
+  */
+case class CompilerNameAnnotation(value: String = "verilog") extends SingleStringAnnotation with FirrtlOption
+
+/** Holds the unambiguous class name of a [[Transform]] to run
+  *  - will be append to [[FirrtlExecutionOptions.customTransforms]]
+  *  - set with `-fct/--custom-transforms`
+  * @param value the full class name of the transform
+  */
+case class RunFirrtlTransformAnnotation(value: String) extends SingleStringAnnotation with FirrtlOption
```
jackkoenig: @azidar and I previously talked about adding this (I think I even PRed it at one point). One issue we were concerned about was the misconception that a transform could add this annotation to the AnnotationSeq during normal execution. The writer of the transform is obviously expected the transform the specified in the RunFirrtlTransformAnnotation would be run but it wouldn't be. 

That being said, maybe we can solve that issue with a `VerifyAnnotations` check or something that would check all kinds of things including that there are no annotations like this, throwing a warning or error if they're found. We could also just document that and assume that when people see their transform not being run they will check the ScalaDoc or whatever and realize that's not how it works.

src/main/scala/firrtl/FirrtlExecutionOptions.scala
jackkoenig: Also would it be possible to seal this trait or make it `private[firrtl]` since users can't just arbitrarily extend this to get something injected into the FirrtlOptions`?

src/main/scala/firrtl/FirrtlExecutionOptions.scala
```diff
@@ -0,0 +1,752 @@
...
+    .abbr("lcn")
+    .action( (x, c) => c :+ LogClassNamesAnnotation )
+    .maxOccurs(1)
+    .text(s"shows class names and log level in logging output, useful for target --class-log-level")
+
+  parser.arg[String]("<arg>...")
+    .unbounded()
+    .optional()
+    .action( (x, c) => c :+ ProgramArgsAnnotation(x) )
+    .text("optional unbounded args")
```
jackkoenig: What are these used for?

src/main/scala/firrtl/FirrtlExecutionOptions.scala
```diff
@@ -0,0 +1,752 @@
...
+  emitOneFilePerModule:     Boolean                     = false) {
+
+  /** Return the info mode */
+  def infoMode(): Parser.InfoMode = {
+    infoModeName match {
+      case "use"    => Parser.UseInfo
+      case "ignore" => Parser.IgnoreInfo
+      case "gen"    => Parser.GenInfo(inputFileNameOverride.getOrElse("[not defined]"))
+      case "append" => Parser.AppendInfo(inputFileNameOverride.getOrElse("[not defined]"))
+      case _        => Parser.UseInfo
```
jackkoenig: Was this like this before? We don't want people to be able to give "boijwfoij" and just get `UseInfo` right?

src/main/scala/firrtl/FirrtlExecutionOptions.scala
```diff
@@ -0,0 +1,752 @@
...
+  /** Determine the target directory with the following precedence (highest to lowest):
+    *  - Explicitly from the user-specified `--target-dir`
+    *  - Implicitly from the default of [[FirrtlExecutionOptions.targetDirName]]
+    *
+    * @param annotations input annotations to extract targetDir from
+    * @return the target directory
+    * @note this is safe to use before [[HasFirrtlExecutionOptions.firrtlOptions]] is set
+    */
+  def targetDir(annotations: Seq[Annotation]): String = annotations
+    .reverse
```
jackkoenig: Why is this reversing?

src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -77,7 +77,14 @@ object JsonProtocol {
...
       case FileInput(file) =>
-        Failure(new InvalidAnnotationFileException(file, e))
+        /* This occurs during scopt parsing, however, scopt will only print the
+         * first error message and does not properly deal with nested
+         * exceptions. Hence, the better processing below of wrapping the
+         * annotation with an [[InvalidAnnotationFileException]] is
+         * currently disabled for improved user understanding of the
+         * error. */
+        /* Failure(new InvalidAnnotationFileException(file, e)) */
+        Failure(e)
```
jackkoenig: So this means the JSON exception will still be reported, it just won't indicate what the Annotation filename is? Not ideal but I guess it could be worse. Would be nice to get this behavior back but I wouldn't block this PR over it

src/main/scala/firrtl/Compiler.scala
```diff
@@ -478,3 +479,6 @@ trait Compiler extends LazyLogging {
 
 }
 
+trait ProvidesOptions {
+  def provideOptions: (OptionParser[AnnotationSeq]) => Unit
```
jackkoenig: Is there anyway this could be immutable? Is it mutable because that's what the ExecutionOptions are using anyway?

jackkoenig: Also is there any reason it's `def provideOptions: (OptionParser[AnnotationSeq]) => Unit` instead of just `def provideOptions(parser: OptionParser[AnnotationSeq]: Unit`

build.sbt
```diff
@@ -140,3 +140,5 @@ scalacOptions in Compile in doc ++= Seq(
   "-doc-title", name.value,
   "-doc-root-content", baseDirectory.value+"/root-doc.txt"
 ) ++ scalacOptionsVersion(scalaVersion.value)
+
+Test / fork := true
```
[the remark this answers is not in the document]
seldridge: This sneaked in there... I can remove it. This is more of a convenience (it's not an issue with CI, etc.). If I run the tests repeatedly in the same sbt shell, eventually I run myself out of memory and have to restart the shell. This avoids that. It does, technically, slow things down slightly compared to not using it. I should be able to do this with a local sbt configuration and not change this here.

src/main/scala/firrtl/FirrtlExecutionOptions.scala
seldridge: I see what you're saying here. Ugh... this gets very meta very quickly...

Using this in a transform would be an anti-pattern. If a transform needs to be composed of a sequence of transforms, then it should be calling that transform (or it's assumedly associated pass) directly as opposed to using this to run a transform.

`VerifyAnnotations` makes sense. We can talk about this... I've been thinking that we may want to divide the annotations into (at least) two groups. Certain annotations are allowed to be added at specific times. Specifically, I was toying with a division of "annotations that can be added by command lines" and "annotations which can show up in a file". I was thinking about this with the  `InputAnnotationFileAnnotation` and considering disallowing that from adding command line options annotations. That seems like a recipe for disaster as the interactions there are not clear, things can get overwritten, etc. (and it avoids having to deal with input annotation files including other input annotation files).

An alternative, which I'm not in favor of, is having the compiler try to clean up any new `RunFirrtlTransformAnnotation` by scheduling new transforms during lowering. I'm against this as it makes it very non-deterministic how the compiler is going to be run and there's a clear good pattern of having a transform schedule other transforms internally.

src/main/scala/firrtl/FirrtlExecutionOptions.scala
seldridge: I'm not sure... they were there, so I didn't delete them (and I did not poke through the logs to try and figure it out). I highly, highly expect that these are some vestigial feature that nobody is using. I can look into it and throw a deprecation warning if they're used. Otherwise, I'm fine to get rid of it and deal with the consequences later...

src/main/scala/firrtl/FirrtlExecutionOptions.scala
seldridge: Apparently it was! However, there was a validation check during option parsing to make sure that `infoModeName` was one of the allowable strings. That same check is still in place, however, if the user passes in this option via an annotation file, then that check won't happen. This needs to be fixed, or the check needs to be moved somewhere else.

src/main/scala/firrtl/FirrtlExecutionOptions.scala
seldridge: This requires some more thought... but, I was intending that the last `TargetDirAnnotation` should win. However, this conflicts with how the validation of the command line options is done. You can't specify `--target-dir` more than once. It would seem logical that this same check should be applied to the associated annotation.

src/main/scala/firrtl/FirrtlExecutionOptions.scala
seldridge: One addendum, while not nearly as critical, `TargetDirAnnotation` has the same oddity, though I doubt a user would be doing this with a Transform. If a transform adds an annotation that is clearly some top-level construct like `TargetDirAnnotation`, how should we handle it? I'm very much in favor of doing nothing and possibly barfing. The command line options are, through this transform, intended to be immutable. Any options-derived annotations similarly are disallowed from being emitted by Transforms. Summarily, `FirrtlOption` really, really needs to be `sealed`.

src/main/scala/firrtl/annotations/JsonProtocol.scala
seldridge: I need to investigate this and get a small example provided... (I can't remember exactly.)

src/main/scala/firrtl/Compiler.scala
seldridge: I may need some clarification on your first question about how this is mutable. Are you saying that this is able to mutably keep adding options to the parser if we're not disciplined and only call it once?

I'll have to check on the second point. It seems like I would have tried that first and there may have been a problem. However, it's possibly just over-engineered. I can test it.

src/main/scala/firrtl/FirrtlExecutionOptions.scala
chick: These are used by the Chisel Tutorial so that command line execution scripts can pass program arguments to the various main functions in the project. A user might similarly want to use this function to write bash scripts that invoke mains.

seldridge: Thanks for the necessary clarification! "Used by the tutorial" is about as far as possible from "vestigial feature"...

seldridge:
While getting an example of this working with Treadle, I noticed that this breaks some previous guarantees about how an `ExecutionOptionsManager` behaves. I wanted to provide some documentation on what is going on here and solicit some feedback on the pattern that this mandates for any concrete subclasses of `ExecutionOptionsManager`.

Consider the case of Treadle's REPL. A user is expected to interact with this as follows:
1. Start it up without any arguments: `./treadle.sh`
2. Load a FIRRTL circuit via `load`
3. Interact with it, e.g., `poke`, `peek`

Currently (freechipsproject/firrtl#765) , any `ExecutionOptionsManager` that mixes in `HasFirrtlExecutionOptions` gets access to `lazy val firrtlOptions: FirrtlExecutionOptions`. This is lazily evaluated from available annotations with all annotation sanity checks running at this time. One of these checks requires that a FIRRTL source (`-i/--input-file/InputFileAnnotation` or `--firrtl-source/FirrtlSourceAnnotation`) is available.

This creates a pattern where a partially complete, but invalid `ExecutionOptionsManager` can be created, but it must not be observed (use of any methods or reading of it's options lazy val). Otherwise, the options will lazily evaluate and everything will barf. More specifically, this is the state of Treadle's REPL after step 1 (above) and before step 2 finishes because no FIRRTL source exists.

__Is this a sane pattern or is this overly complicating things?__

seldridge:
Thinking about this a bit more, the problem is that all these lazy vals can throw Exceptions if they fail validation.

There's no feedback to a developer interacting with an options manager that this lazy val isn't safe to interact with. The correct approach would seem to be to switch to one of the standard Scala approaches for this problem: `Option`, `Either`, or `Try`. Alternatively, we could leave this with a comment and indicate that this can throw exceptions (meaning if you care about safely interacting with it, you should be using `try/catch` or just be careful). 

@jackkoenig: you may have an opinion here?

src/main/scala/firrtl/options/ExecutionOptionsManager.scala
```diff
@@ -33,12 +34,53 @@ class ExecutionOptionsManager(
...
+      .parse(args, annotations)
+      .getOrElse(throw new FIRRTLException("Failed to parse command line options"))
+  }
+
+  /* A hidden option that exposes what libraries and transforms are
+   * registered. This is purely for debugging. */
+  parser.opt[Unit]("show-registrations")
+    .hidden()
+    .action{ (_, c) =>
+      println(s"""|Registered Libraries:
```
jackkoenig: println or should this be logger.info? logger.warn?

seldridge: I think this is safe to leave in as it is. I could wrap this in a logger, but this is really information that I want to get to the command line for debugging before a logger may exist and I want this to happen even if a logger doesn't exist or I have the incorrect log level. 

I.e., I think I can wrap the scopt `parse` call with a logger, but I'd want to pass in an explicit log level which would seem to violate the motivation for having a logger at all.

src/main/scala/firrtl/options/OptionParser.scala
```diff
@@ -0,0 +1,28 @@
...
+
+/** A modified [[scopt.OptionParser]] with mutable termination and additional checks
+  *
+  * @param name the name of the application
+  */
+trait DuplicateHandling extends OptionParser[AnnotationSeq] {
+  /* Check that no long or short option declarations are duplicated */
+  override def parse(args: Seq[String], init: AnnotationSeq): Option[AnnotationSeq] = {
+    val longDups = options.map(_.name).groupBy(identity).collect{ case (k, v) if v.size > 1 && k != "" => k }
+    val shortDups = options.map(_.shortOpt).flatten.groupBy(identity).collect{ case (k, v) if v.size > 1 => k }
```
ucbjrl: This seems to be responsible for:
```bash
An exception or error caused a run to abort: scopt.OptionDef.shortOpt()Lscala/Option; 
java.lang.NoSuchMethodError: scopt.OptionDef.shortOpt()Lscala/Option;
	at firrtl.options.DuplicateHandling$$anonfun$5.apply(OptionParser.scala:22)
	at firrtl.options.DuplicateHandling$$anonfun$5.apply(OptionParser.scala:22)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
	at scala.collection.immutable.List.foreach(List.scala:392)
	at scala.collection.generic.TraversableForwarder$class.foreach(TraversableForwarder.scala:35)
	at scala.collection.mutable.ListBuffer.foreach(ListBuffer.scala:45)
	at scala.collection.TraversableLike$class.map(TraversableLike.scala:234)
	at scala.collection.AbstractTraversable.map(Traversable.scala:104)
	at firrtl.options.DuplicateHandling$class.parse(OptionParser.scala:22)
	at firrtl.options.ExecutionOptionsManager$$anon$1.parse(ExecutionOptionsManager.scala:20)
	at firrtl.options.ExecutionOptionsManager.parse(ExecutionOptionsManager.scala:68)
	at firrtl_interpreter.Driver$.execute(Driver.scala:33)
	at firrtl_interpreter.DriverSpec$$anonfun$1$$anonfun$apply$mcV$sp$1.apply$mcV$sp(DriverSpec.scala:21)
	at firrtl_interpreter.DriverSpec$$anonfun$1$$anonfun$apply$mcV$sp$1.apply(DriverSpec.scala:10)
	at firrtl_interpreter.DriverSpec$$anonfun$1$$anonfun$apply$mcV$sp$1.apply(DriverSpec.scala:10)
```

ucbjrl: False alarm. This was due to firrtl-interpreter using an old version of scopt.

seldridge: Possibly a stale scopt version? That option showed up in 3.7.0.

src/main/scala/firrtl/annotations/Annotation.scala
```diff
@@ -28,6 +28,37 @@ trait NoTargetAnnotation extends Annotation {
...
+  /** Do the conversion to an [[Annotation]] that can be serialized to JSON. This will check that the conversion does not
+    * have the [[Unserializable]] trait.
+    */
+  final def _toJsonSerializable: Annotation = this.toJsonSerializable match {
+    case _: Unserializable =>
+      throw new AnnotationException(s"Annotation '${this.getClass.getName}' cannot be serialized to JSON")
+    case a => a
+  }
+}
+
```
seldridge: @chick: Here's what I came up with for using `Annotation`s for internal tool data movement of things which we can't serialize with json4s (currently). The idea here is that you mix this into an `Annotation` and before this is serialized to JSON, it will be converted to an annotation that _is serializable_. This is damn ugly as I wasn't able to think of a way to do this with type checking. Maybe there's a way with proper variances, but nothing was jumping out at me. This _should enable_ something (which I'll add to Chisel) of a `ChiselCircuitGeneratorAnnotation(dut: () => RawModule)` that will serialize to a `ChiselCircuitAnnotation(circuit: Circuit)` but can be used for Chisel testers.

@jackkoenig: This is to address a narrow problem of passing annotations into the `Driver` of Chisel testers. It needs a `() => MultiIOModule` to pass to a `PeekPokeTester` and this is information that can't be contained in a serializable annotation without completely rewriting how Chisel testers works. Can you think of any way to add type safety here?

seldridge:
Ping @jackkoenig, @chick: I have a separate branch with this PR rebased and refactored massively. I could PR this as either sequential commits or about 9 PRs (possibly more...). Note that only one of those 9 PRs would have an actual functional impact on FIRRTL (the one that migrates to the "options are annotations" pattern). Everything else is either small patches or adding the framework to enable "options are annotations".

I haven't tested that every commit builds, but they are supposed to (or that's the intention). I'll spoon feed them to Travis to get those :heavy_check_mark:s.

Would it be a good idea to close this and reopen? Any thoughts on separate PRs or one PR with a chain of sequential, reasoned patches?

See: https://github.com/freechipsproject/firrtl/compare/master...seldridge:issue-764-refactor

FWIW, if anyone else is trying to do this type of rebasing, instead of fighting with rebasing a dumpster :fire: of 77 commits with endless merge conflicts, I had luck with applying the equivalent patch onto a new branch and then repeatedly rebasing to break out logical items.

seldridge:
This makes sense to me. I'm opting for the series of PRs approach. I'll re-open if people feel differently.

jackkoenig:
I think the series of PRs sounds great, thanks!

PR #766, opened 2018-03-19
GroupModule Transform

azidar:
This pass takes a GroupAnnotation of multiple signals in a module, and will group them into their own module (and then instantiate and connect to the new module).

Reviewer Checklist:
- [x] Revert my reverts of sbt bumps
- [x] Remove WireHelper stub
- [x] Discuss orderedLinearization, probably remove it
- [x] Add checks for DefInstance and WDefInstanceConnector
- [x] Add blacklist to documentation in DiGraph
- [x] Add spaces around the @
- [x] Add spaces around |


colinschmidt:
This works on my rather large design with no issues 👍

src/test/scala/firrtlTests/transforms/GroupComponentsSpec.scala
```diff
@@ -0,0 +1,290 @@
...
+        |  module $top :
+        |    input clk: Clock
+        |    input data: UInt<16>
+        |    output out: UInt<16>
+        |    reg r: UInt<16>, clk
+        |    r <= data
+        |    out <= r
+      """.stripMargin
+    val groups = Seq(
+      GroupAnnotation(Seq(topComp("r")), "MyReg", "rInst", Some("_OUT"), Some("_IN"))
```
edwardcwang: Say if I have module A which instantiates modules C1 and C2 and also (unrelated) module B which instantiates modules C1 and C2 as well - does this pass handle that in one pass, or would I just use 2 separate GroupAnnotations and try to de-dup later?

src/main/scala/firrtl/analyses/InstanceGraph.scala
```diff
@@ -70,14 +61,14 @@ class InstanceGraph(c: Circuit) {
...
-  lazy val fullHierarchy = graph.pathsInDAG(trueTopInstance)
+  lazy val fullHierarchy: collection.Map[WDefInstance,Seq[Seq[WDefInstance]]] = graph.pathsInDAG(trueTopInstance)
 
   /** Finds the absolute paths (each represented by a Seq of instances
     * representing the chain of hierarchy) of all instances of a
     * particular module.
     *
     * @param module the name of the selected module
-    * @return a Seq[Seq[WDefInstance]] of absolute instance paths
+    * @return a Seq[ Seq[WDefInstance] ] of absolute instance paths
```
edwardcwang: Curiosity: scaladoc doesn't like not having the spaces?

azidar: So double square-brackets indicate a type in scaladoc, so the end of this type would be recognized as a weird trailing bracket. I put spaces in to just avoid it messing up. I'm sure the correct fix.

src/test/scala/firrtlTests/transforms/GroupComponentsSpec.scala
azidar: Sorry which module has the components to group, in this example?

edwardcwang: Sorry - my comment was more general and not tied to this specific example. I guess I'm just asking in the case of grouping the same thing across distinct modules.

azidar: Grouping can only occur within a module, and the name of that module will not change. So, any all instances of the old module become instances of the now-split module.

src/main/scala/firrtl/analyses/InstanceGraph.scala
```diff
@@ -94,4 +85,30 @@ class InstanceGraph(c: Circuit) {
...
+    * @return
+    */
+  def collectInstances(insts: mutable.Set[WDefInstance])
+                      (s: Statement): Statement = s match {
+    case i: WDefInstance =>
+      insts += i
+      i
+    case _ =>
+      s map collectInstances(insts)
+  }
```
jackkoenig: I think this should include a check for DefInstance and WDefInstanceConnector and should error if it sees either

src/main/scala/firrtl/graph/DiGraph.scala
```diff
@@ -108,14 +107,22 @@ class DiGraph[T] private[graph] (private[graph] val edges: LinkedHashMap[T, Link
     * @return a Map[T,T] from each visited node to its predecessor in the
     * traversal
     */
-  def BFS(root: T): Map[T,T] = {
-    val prev = new LinkedHashMap[T,T]
+  def BFS(root: T): Map[T,T] = BFS(root, Set.empty[T])
+
+  /** Performs breadth-first search on the directed graph, with a blacklist of nodes
+    *
+    * @param root the start node
```
jackkoenig: Include blacklist in the ScalaDoc

src/main/scala/firrtl/transforms/GroupComponents.scala
```diff
@@ -0,0 +1,340 @@
...
+    * It is non-directioned; there is an edge from source to sink and from sink to souce
+    * @param m module to compute connectivity
+    * @return a bi-directional representation of component connectivity
+    */
+  def getComponentConnectivity(m: Module): MutableDiGraph[String] = {
+    val bidirGraph = new MutableDiGraph[String]
+    val simNamespace = Namespace()
+    val simulations = new mutable.HashMap[String, Statement]
+    def onExpr(sink: WRef)(e: Expression): Expression = e match {
+      case w@WRef(name, _, _, _) =>
```
jackkoenig: Minor nit but I much prefer spaces around the `@`, and less important

azidar:
Things to clean up for review:
- [ ] Add checks for DefInstance and WDefInstanceConnector
- [ ] Add blacklist to documentation in DiGraph

src/main/scala/firrtl/transforms/GroupComponents.scala
```diff
@@ -0,0 +1,340 @@
...
+          groupStatements(group) += r mapExpr inGroupFixExps(group, topStmts)
+          Block(topStmts)
+        case c: Connect if byNode(getWRef(c.loc).name) != "" =>
+          // Sink is in a group
+          val topStmts = mutable.ArrayBuffer[Statement]()
+          val group = byNode(getWRef(c.loc).name)
+          groupStatements(group) += Connect(c.info, c.loc, inGroupFixExps(group, topStmts)(c.expr))
+          Block(topStmts)
+        // TODO Attach if all are in a group?
+        case _: IsDeclaration|_: Connect|_: Attach =>
```
jackkoenig: I also prefer spaces around `|` here 🤷‍♀️

src/main/scala/firrtl/analyses/InstanceGraph.scala
jackkoenig: I'm also not a huge fan of public APIs that aren't side-effect free

src/main/scala/firrtl/graph/DiGraph.scala
```diff
@@ -85,7 +84,7 @@ class DiGraph[T] private[graph] (private[graph] val edges: LinkedHashMap[T, Link
       if (unmarked.contains(n)) {
         tempMarked += n
         unmarked -= n
-        for (m <- getEdges(n)) {
+        for (m <- getEdges(n).toSeq.reverse) {
```
jackkoenig: Why is this being reversed?

src/main/scala/firrtl/transforms/Dedup.scala
```diff
@@ -18,134 +19,301 @@ case class NoDedupAnnotation(target: ModuleName) extends SingleTargetAnnotation[
...
+  }
+}
+
+/**
+  * Utility functions for [[DedupModules]]
+  */
+object DedupModules {
+  /**
+    * Change's a module's internal signal names, types, infos, and modules.
+    * Populates renameMap.
```
jackkoenig: Where's the renameMap being populated?

src/main/scala/firrtl/transforms/Dedup.scala
```diff
@@ -18,134 +19,301 @@ case class NoDedupAnnotation(target: ModuleName) extends SingleTargetAnnotation[
...
+              case Field("rdata", _, _) => true
+              case _ => false
+            }))
+        val newDataType = if(index.nonEmpty) {
+          //If index nonempty, then there exists a port
+          val i = index.get
+          newType.asInstanceOf[BundleType].fields.head.tpe.asInstanceOf[BundleType].fields(i).tpe
+        } else {
+          //If index is empty, this mem has no ports, and so we don't need to record the dataType
+          retype(d.name + ";&*^$")(d.dataType)
```
jackkoenig: What is this magical string `";&*^$"`?

src/main/scala/firrtl/transforms/Dedup.scala
```diff
@@ -18,134 +19,301 @@ case class NoDedupAnnotation(target: ModuleName) extends SingleTargetAnnotation[
...
+      case d: DefMemory =>
+        val oldType = MemPortUtils.memType(d)
+        val newType = retype(d.name)(oldType)
+        val index = oldType
+          .asInstanceOf[BundleType].fields.headOption
+          .map(_.tpe.asInstanceOf[BundleType].fields.indexWhere(
+            {
+              case Field("data", _, _) => true
+              case Field("wdata", _, _) => true
+              case Field("rdata", _, _) => true
```
jackkoenig: minor nit but these could be combined into:
```scala
case Field("data" | "wdata" | "rdata", _,_) => true
```
That's a matter of taste though

src/main/scala/firrtl/transforms/Dedup.scala
```diff
@@ -18,134 +19,301 @@ case class NoDedupAnnotation(target: ModuleName) extends SingleTargetAnnotation[
...
+    * Turns a module into a name-agnostic module
+    * @param module module to change
+    * @return name-agnostic module
+    */
+  def agnostify(module: DefModule, name2tag: mutable.HashMap[String, String], tag2name: mutable.HashMap[String, String]): DefModule = {
+    val namespace = Namespace()
+    val nameMap = mutable.HashMap[String, String]()
+    val typeMap = mutable.HashMap[String, Type]()
+    def rename(name: String): String = {
+      if(nameMap.contains(name)) nameMap(name) else {
```
jackkoenig: This is a minor thing but canonical Scala style calls for a space after `if`, eg. `if (nameMap.contains(name))...`

src/main/scala/firrtl/transforms/Dedup.scala
```diff
@@ -18,134 +19,301 @@ case class NoDedupAnnotation(target: ModuleName) extends SingleTargetAnnotation[
...
+        name2tag(originalModule.name) = originalModule.name
+        tag2name(originalModule.name) = originalModule.name
+        //templateModules += originalModule.name
+      } else { // Try to dedup
+
+        // Build name-agnostic module
+        val agnosticModule = DedupModules.agnostify(originalModule, name2tag, tag2name)
+
+        // Build tag
+        val tag = md5Hash(agnosticModule match {
```
jackkoenig: Why do we need this secure hash for the tag? Can't we just use `.hashCode` or even just the serialized String of the Module?

src/main/scala/firrtl/transforms/Dedup.scala
```diff
@@ -18,134 +19,301 @@ case class NoDedupAnnotation(target: ModuleName) extends SingleTargetAnnotation[
...
+    tag2all.foreach { case (tag, all) => tag2name(tag) = all.reduce(order)}
+
+    // Create map from original to dedup name
+    val name2name = name2tag.map{case (name, tag) => name -> tag2name(tag)}
+
+    // Build Remap for modules with deduped module references
+    val tag2module = tag2name.map { case (tag, name) => tag -> DedupModules.dedupInstances(name, moduleMap, name2name, renameMap) }
+
+    // Build map from original name to corresponding deduped module
+    val name2module = name2tag.map {case (name, tag) => name -> tag2module(tag)}
```
jackkoenig: Would appreciate consistent spacing around the curly braces, all 3 of these use different spacing. I generally prefer `collection.map { case (...) => }` although @albert-magyar might be slowly converting me to `collection.map({ case (...) => })`, jury is still out on that one.

src/main/scala/firrtl/transforms/GroupComponents.scala
```diff
@@ -0,0 +1,340 @@
...
+  */
+class GroupComponents extends firrtl.Transform {
+  type MSet[T] = mutable.Set[T]
+
+  def inputForm: CircuitForm = MidForm
+  def outputForm: CircuitForm = MidForm
+
+  override def execute(state: CircuitState): CircuitState = {
+    val groups = state.annotations.collect {case g: GroupAnnotation => g}
+    val module2group = groups.groupBy{_.currentModule}
```
jackkoenig: I think regular parentheses would look better here

src/main/scala/firrtl/transforms/GroupComponents.scala
```diff
@@ -0,0 +1,340 @@
...
+    }
+    val cs = state.copy(circuit = state.circuit.copy(modules = newModules))
+    val csx = InferTypes.execute(cs)
+    csx
+  }
+
+  def groupModule(m: Module, groups: Seq[GroupAnnotation], mnamespace: Namespace): Seq[Module] = {
+    val namespace = Namespace(m)
+    val groupRoots = groups.map(_.components.map(_.name))
+    val totalSum = groupRoots.foldLeft(0){(total, set) => total + set.size}
```
jackkoenig: could just be 
```scala 
val totalSum = groupRoots.map(_.size).sum
```

src/main/scala/firrtl/transforms/GroupComponents.scala
```diff
@@ -0,0 +1,340 @@
...
+    val union = groupRoots.foldLeft(Set.empty[String]){(all, set) => all.union(set.toSet)}
+
+    require(groupRoots.forall{_.forall{namespace.contains}}, "All names should be in this module")
+    require(totalSum == union.size, "No name can be in more than one group")
+    require(groupRoots.forall(_.nonEmpty), "All groupRoots must by non-empty")
+
+
+    // Order of groups, according to their label. The label is the first root in the group
+    val labelOrder = groups.collect{case g:GroupAnnotation => g.components.head.name}
```
jackkoenig: I think this is better spaced as:
```scala
val labelOrder = groups.collect { case g: GroupAnnotation => g.components.head.name }
```

src/main/scala/firrtl/transforms/GroupComponents.scala
```diff
@@ -0,0 +1,340 @@
...
+    }
+    bidirGraph
+  }
+}
+
+/**
+  * Splits a module into multiple modules by grouping its components via [[GroupAnnotation]]'s
+  * Tries to deduplicate the resulting circuit
+  */
+class GroupAndDedup extends GroupComponents {
```
jackkoenig: Rather than this inheriting from GroupComponents, I think it should just be a different Transform that calls GroupComponents.

src/main/scala/firrtl/transforms/Dedup.scala
```diff
@@ -18,134 +19,301 @@ case class NoDedupAnnotation(target: ModuleName) extends SingleTargetAnnotation[
...
-  def execute(state: CircuitState): CircuitState = {
-    val noDedups = state.annotations.collect { case NoDedupAnnotation(ModuleName(m, c)) => m }
-    val (newC, renameMap) = run(state.circuit, noDedups)
-    state.copy(circuit = newC, renames = Some(renameMap))
+  def getAffectedExpressions(root: Expression): Seq[Expression] = {
+    val all = mutable.ArrayBuffer[Expression]()
+
+    def onExp(expr: Expression): Unit = {
+      expr.tpe match {
+        case g: GroundType =>
```
jackkoenig: I think use an `_` here since you don't care about the `GroundType`, ie.
```scala
case _: GroundType =>
```

src/test/scala/firrtlTests/transforms/GroupComponentsSpec.scala
```diff
@@ -0,0 +1,290 @@
...
+         |    output out: UInt<8>
+         |    out <= UInt(2)
+         |  module PassThrough :
+         |    input in: UInt
+         |    output out: UInt
+         |    out <= in
+      """.stripMargin
+    val groups = Seq(
+      GroupAnnotation(Seq(topComp("c1a"), topComp("c2a"), topComp("asum")), "A", "cA", Some("_OUT"), Some("_IN")),
+      GroupAnnotation(Seq(topComp("c1b"), topComp("c2b"), topComp("bsum")), "B", "cB", Some("_OUT"), Some("_IN"))
```
jackkoenig: If the suffixes are allowed to be `None`, I'd like to see an example with that.

src/main/scala/firrtl/transforms/GroupComponents.scala
```diff
@@ -0,0 +1,340 @@
...
+        case other => other map onStmt
+      }
+    }
+
+
+    // Build datastructures
+    val newTopBody = Block(labelOrder.map(g => WDefInstance(NoInfo, label2instance(g), label2module(g), UnknownType)) ++ Seq(onStmt(m.body)))
+    val finalTopBody = Block(Utils.squashEmpty(newTopBody).asInstanceOf[Block].stmts.distinct)
+
+    val newModules = labelOrder.filter(_ != "") map { group =>
```
jackkoenig: Why is this filtering for empty labels? What does it mean if they're empty?

src/main/scala/firrtl/transforms/GroupComponents.scala
```diff
@@ -0,0 +1,340 @@
...
+  * Specifies a group of components, within a module, to pull out into their own module
+  * Components that are only connected to a group's components will also be included
+  *
+  * @param components components in this group
+  * @param newModule suggested name of the new module
+  * @param newInstance suggested name of the instance of the new module
+  * @param outputSuffix suggested suffix of any output ports of the new module
+  * @param inputSuffix suggested suffix of any input ports of the new module
+  */
+case class GroupAnnotation(components: Seq[ComponentName], newModule: String, newInstance: String, outputSuffix: Option[String] = None, inputSuffix: Option[String] = None) extends Annotation {
```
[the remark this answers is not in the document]
azidar: A unique name is always generated, regardless of whether a prefix is specified or not.

src/test/scala/firrtlTests/transforms/GroupComponentsSpec.scala
azidar: An example is beneath, line 253.

src/main/scala/firrtl/transforms/GroupComponents.scala
azidar: The empty label indicates the original module's group. This means that it won't be grouped with any of the specified groups, but will remain in the original module.

azidar: I will add some comments to clarify.

PR #771, opened 2018-03-22
Bad anno file

jackkoenig:
A suite of improvements to the error reporting for bad Annotation files

### DIFF
--- a/src/main/scala/firrtl/annotations/JsonProtocol.scala
+++ b/src/main/scala/firrtl/annotations/JsonProtocol.scala
@@ -13,18 +13,6 @@ import firrtl.ir._
 import firrtl.Utils.error
 
 object JsonProtocol {
-
-  // Helper for error messages
-  private def prettifyJsonInput(in: JsonInput): String = {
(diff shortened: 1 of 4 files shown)


src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -66,19 +54,30 @@ object JsonProtocol {
   def deserialize(in: JsonInput): Seq[Annotation] = deserializeTry(in).get
 
   def deserializeTry(in: JsonInput): Try[Seq[Annotation]] = Try({
-    def throwError() = throw new InvalidAnnotationFileException(prettifyJsonInput(in))
     val parsed = parse(in)
     val annos = parsed match {
       case JArray(objs) => objs
-      case _ => throwError()
+      case _ => throw new InvalidAnnotationJSONException("Unknown Error")
```
chick: Might it provide a clue to the problem if you'd use a var here and then add it to the Exception.
`case x => throw new InvalidAnnotationJSONException("Unexpected value in parsing '$x'")

PR #772, opened 2018-03-22
Const prop improvement

jackkoenig:
Propagate connections to references, see the test for an example

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,10 +34,10 @@ jobs:
         - bash .run_chisel_tests.sh
     - stage: test
       script:
-        - travis_wait ./.run_formal_checks.sh RocketCore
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -34,10 +34,10 @@ jobs:
         - bash .run_chisel_tests.sh
     - stage: test
       script:
-        - ./.run_formal_checks.sh RocketCore
+        - travis_wait ./.run_formal_checks.sh RocketCore
```
azidar: what does this do?

jackkoenig: It's supposed to increase the time before travis kills a test due to "timeout". It seems that this is a bigger problem with Yosys not being able to handle this change so I'm looking into it.

PR #774, opened 2018-03-23
Make WiringTransform remove its used annotations

seldridge:
This modifies the WiringTransform to actually delete its annotations (if they wind up being used). Without this, the WiringTransform could be run multiple times with the same annotations (e.g., if some other Transform runs the WiringTransform directly).

### DIFF
--- a/src/main/scala/firrtl/passes/wiring/WiringTransform.scala
+++ b/src/main/scala/firrtl/passes/wiring/WiringTransform.scala
@@ -69,7 +69,7 @@ class WiringTransform extends Transform {
             val wis = sources.foldLeft(Seq[WiringInfo]()) { case (seq, (pin, source)) =>
               seq :+ WiringInfo(source, sinks(pin), pin)
             }
-            val annosx = AnnotationSeq((state.annotations.toSeq.toSet -- annos).toSeq)
+            val annosx = state.annotations.filterNot(annos.toSet.contains)
             transforms(wis)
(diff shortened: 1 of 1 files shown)


src/main/scala/firrtl/passes/wiring/WiringTransform.scala
```diff
@@ -69,7 +69,10 @@ class WiringTransform extends Transform {
             val wis = sources.foldLeft(Seq[WiringInfo]()) { case (seq, (pin, source)) =>
               seq :+ WiringInfo(source, sinks(pin), pin)
             }
-            transforms(wis).foldLeft(state) { (in, xform) => xform.runTransform(in) }
+            val annosx = AnnotationSeq((state.annotations.toSeq.toSet -- annos).toSeq)
```
jackkoenig: While it's probably not ideal that things are this way, many transforms that consume annotations do things in an order sensitive to the ordering of annotations. In the interest of keeping things as deterministic as possible, I think it'd be better to generate annosx with something like:
```scala
val annoSet = annos.toSet // We could probably put this in the anonymous function below, JVM should figure it out
val annosx = annotations.filterNot(annoSet.contains)
```
Less important but you shouldn't need to call `.toSeq` or `AnnotationSeq.apply` because of implicit conversions between `Seq[Annotation]` and `AnnotationSeq`. If you prefer the code to be more explicit though, that's fine with me.

seldridge: Makes sense to me!

PR #775, opened 2018-03-27
Enhance RenameMap to support circuit renaming

jackkoenig:
Also dynamically enforce some type safety

### DIFF
--- a/src/main/scala/firrtl/Compiler.scala
+++ b/src/main/scala/firrtl/Compiler.scala
@@ -76,7 +76,8 @@ final class RenameMap private () {
   def get(key: Named): Option[Seq[Named]] = key match {
     case c: ComponentName => this.get(c)
     case m: ModuleName => this.get(m)
-    case c: CircuitName => this.get(c)
+    // The CircuitName version returns Option[CircuitName]
+    case c: CircuitName => this.get(c).map(Seq(_))
   }
 
   // Mutable helpers
--- a/src/main/scala/firrtl/Compiler.scala
+++ b/src/main/scala/firrtl/Compiler.scala
@@ -27,41 +27,60 @@ object RenameMap {
 // TODO This should probably be refactored into immutable and mutable versions
 final class RenameMap private () {
   private val underlying = mutable.HashMap[Named, Seq[Named]]()
+
+  /** Get renames of a [[CircuitName]]
+    * @note A [[CircuitName]] can only be renamed to one-or-more [[CircuitName]]s
+    */
+  def get(key: CircuitName): Option[Seq[CircuitName]] = underlying.get(key).map {
+    names => names.map {
+      case c: CircuitName => c
+      case other => error(s"Unsupported rename of $key to $other!")
+    }
+  }
+  /** Get renames of a [[ModuleName]]
+    * @note A [[ModuleName]] can only be renamed to one-or-more [[ModuleName]]s
+    */
+  def get(key: ModuleName): Option[Seq[ModuleName]] = {
+    def nestedRename(m: ModuleName): Option[Seq[ModuleName]] =
+      this.get(m.circuit).map { circuits =>
+        circuits.map(cname => ModuleName(m.name, cname))
+      }
+    underlying.get(key) match {
+      case Some(names) => Some(names.flatMap {
+        case m: ModuleName =>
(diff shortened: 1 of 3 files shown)


jackkoenig:
The fails are spurious, but unfortunately this change does seem to have a pretty large performance impact. It makes sense, every single ComponentName lookup requires 3 gets from the map, not to mention function call overhead. I'm looking into ways to measuring the performance and improving it.

jackkoenig:
After further evaluation, the performance impact is actually quite small (my test case was to blame for how bad I perceived the impact to be). I think this PR is good and ready to merge.

colinschmidt:
Cool. I wanted this along time ago and couldn't even work around it within the rename map see https://github.com/ucb-bar/barstools/commit/70805f137c3f2403b8f34530dbca26cba72bf37b

src/main/scala/firrtl/Compiler.scala
```diff
@@ -27,42 +27,61 @@ object RenameMap {
...
-    * This is analogous to get on standard Scala collection Maps
-    * None indicates the key was not renamed
-    * Empty indicates the name was deleted
+
+  /** Get renames of a [[CircuitName]]
+    * @note A [[CircuitName]] can only be renamed to one-or-more [[CircuitName]]s
     */
-  // TODO Is there a better way to express this?
-  def get(key: Named): Option[Seq[Named]] = {
+  def get(key: CircuitName): Option[Seq[CircuitName]] = underlying.get(key).map {
```
colinschmidt: How do you rename it to multiple `CircuitName`s? Is that splitting the circuit into two circuits? Does the rest of firrtl support that?

src/main/scala/firrtl/Compiler.scala
```diff
@@ -27,42 +27,61 @@ object RenameMap {
...
+    names => names.map {
+      case c: CircuitName => c
+      case other => error(s"Unsupported rename of $key to $other!")
```
colinschmidt: Maybe say `Unsupported Circuit rename...`

src/test/scala/firrtlTests/RenameMapSpec.scala
```diff
@@ -71,4 +75,55 @@ class RenameMapSpec extends FirrtlFlatSpec {
     renames.rename(foo, bar)
     renames.get(foo) should be (Some(Seq(barB)))
   }
+
+  it should "rename modules if their circuit is renamed" in {
+    val renames = RenameMap()
+    renames.rename(cir, cir2)
+    renames.get(modA) should be (Some(Seq(modA2)))
```
colinschmidt: Shouldn't you also check that modB renamed?

src/test/scala/firrtlTests/RenameMapSpec.scala
```diff
@@ -71,4 +75,55 @@ class RenameMapSpec extends FirrtlFlatSpec {
...
+  }
+
+  it should "rename components if their circuit is renamed" in {
+    val renames = RenameMap()
+    renames.rename(cir, cir2)
+    renames.get(foo) should be (Some(Seq(foo2)))
```
colinschmidt: Can we also check fooB for renaming

src/test/scala/firrtlTests/RenameMapSpec.scala
jackkoenig: In this case, the only thing in the rename map is the circuit rename. Thus checking for modA and modB is identical.

src/test/scala/firrtlTests/RenameMapSpec.scala
jackkoenig: Same as above

src/test/scala/firrtlTests/RenameMapSpec.scala
```diff
@@ -88,42 +90,27 @@ class RenameMapSpec extends FirrtlFlatSpec {
     renames.get(foo) should be (Some(Seq(foo2)))
   }
 
-  it should "error if a component is renamed to a module or circuit" in {
-    a [FIRRTLException] shouldBe thrownBy {
+  // Renaming from to all of the tos at once should error
```
colinschmidt: Renaming from to each of the tos one at a time should error

PR #776, opened 2018-03-27
Replace unconnected registers with 0 in Constant Propagation

jackkoenig:
Formerly done in RemoveValidIf, but since it's more of a constprop anyway and the situation is frequently created there, I moved it. This fixes a class of common linting issues with FIRRTL-generated Verilog.

### DIFF
--- a/src/main/scala/firrtl/transforms/ConstantPropagation.scala
+++ b/src/main/scala/firrtl/transforms/ConstantPropagation.scala
@@ -378,6 +378,8 @@ class ConstantPropagation extends Transform {
             nodeMap(lname) = constPropExpression(nodeMap, instMap, constSubOutputs)(pad(fval, ltpe))
           case Mux(_, tval: Literal, fval: WRef, _) if weq(lref, fval) =>
             nodeMap(lname) = constPropExpression(nodeMap, instMap, constSubOutputs)(pad(tval, ltpe))
+          case WRef(`lname`, _,_,_) => // If a register is connected to itself, propagate zero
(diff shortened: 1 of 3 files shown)


src/main/scala/firrtl/transforms/ConstantPropagation.scala
```diff
@@ -378,6 +378,8 @@ class ConstantPropagation extends Transform {
             nodeMap(lname) = constPropExpression(nodeMap, instMap, constSubOutputs)(pad(fval, ltpe))
           case Mux(_, tval: Literal, fval: WRef, _) if weq(lref, fval) =>
             nodeMap(lname) = constPropExpression(nodeMap, instMap, constSubOutputs)(pad(tval, ltpe))
+          case WRef(`lname`, _,_,_) => // If a register is connected to itself, propagate zero
```
chick: Why are the back ticks here?

jackkoenig: Backticks mean check if the value of the field is equal to the variable in backticks rather than binding a new variable name, ie.
```scala
case WRef(`lname`, _,_,_) => // note lname is a variable name bound above
// Is the same as:
case WRef(rname, _,_,_) if lname == rname =>
```

chick: Thanks, I just looked it up.  I've never used that feature.

jackkoenig: I discovered it a few months ago but haven't had many opportunities to use it yet! I'm a big fan.

PR #778, opened 2018-03-29
CyclicException identifies a problem node.

chick:
Needed for special handling in Treadle.
Small refactor that allows users of DiGraph#linearize
to return the first node found in a cycle.
Fixed RemoveWiresTransfrom to handle this.
Added test to show usage of this feature.

If there's a better more canonical way of doing this please let me know.


albert-magyar:
Seems reasonable, LGTM

PR #782, opened 2018-04-04
Support infoMode for Strings

edwardcwang:
It seemed like an API hole that I couldn't use infoMode with a string but had to manually create an iterator first.


edwardcwang:
Okay. I'd imagine the `parse(string)` case is slightly more common, so kept the default there.

edwardcwang:
Arrrgh. Will take another stab at it later.

azidar:
@ucbjrl will look into

ucbjrl:
@edwardcwang could you sign off on this if you feel it satisfies your use case?

edwardcwang:
LGTM!

PR #783, opened 2018-04-05
Add JNI backend support

ucbjrl:
Add code to generate a shared library suitable for loading into the JVM. This assumes the target system supports `bash` and `make` (i.e., msys/mingw/cygwin on Windows systems).

Perhaps it's time to move this into chisel, where the resources required to actually build something live.

### DIFF
--- a/src/main/scala/firrtl/util/BackendCompilationUtilities.scala
+++ b/src/main/scala/firrtl/util/BackendCompilationUtilities.scala
@@ -98,6 +98,17 @@ object BackendCompilationUtilities {
     }
     Seq(includeDir.toPath) ++ (subDirs sortWith sortByOSSubDir)
   }
+
+  /** A function to generate Posix compatible paths.
+    * @param path - a string possibly containing Windows path delimiters,
+    * @return - a string with '/' delimiters.
+    */
+  def getPosixCompatibleAbsolutePath(path: String): String = {
+    if (osVersion == Windows)
+      path.replace('\\', '/')
+    else
+      path
+  }
 }
 
 trait BackendCompilationUtilities {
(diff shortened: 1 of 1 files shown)


ducky64:
What are the full effects on Windows users? I don't like requiring some kind of shim (like cygwin), unless it's optional (with a fallback to the old behavior) or if it was required by other things already (Verilator tends not to work with native Windows; Windows compilation of C++ code is a hot mess in general)

ucbjrl:
Windows users need to
 - install MSys2/Mingw64
 - obtain a copy of fsync.c (http://ab-initio.mit.edu/octave-Faddeeva/gnulib/lib/fsync.c), compile it, and turn it into a library (installing it in `C:/msys64/mingw64/x86_64-w64-mingw32/local` for example)

Then
```bash
export CPPFLAGS=-IC:/msys64/mingw64/x86_64-w64-mingw32/local/include
export LIBS="-LC:/msys64/mingw64/x86_64-w64-mingw32/local/lib -lfsync -lmingwex -lws2_32 -lpsapi"
cd .../chisel3
sbt test
```
In order to use the JNI backend (Verilator code loaded into the JVM), Windows users would need to
 - clone https://github.com/dlfcn-win32/dlfcn-win32.git and build its library (installing it in `C:/msys64/mingw64/x86_64-w64-mingw32/local` for example)
- add `-ldl` to `LIBS`

Both the `fsync` and `dlfcn-win32` dependencies could be replaced with native solutions.

ducky64:
Cool, I just wanted to make sure that additional dependency requirements aren't introduced where they might not be needed.

azidar:
I do think this should move out of FIRRTL into Chisel, but I suppose that could be a separate PR. I would like to see some tests running this, which would require this to be moved. How difficult would it be to move to chisel?

ucbjrl:
It looks like the only references to `BackendCompilationUtilities` (outside of the file itself), are in tests. I think it would be relatively easy to move this to Chisel. I'll put together PR's to do this.

ucbjrl:
Moving `BackendCompilationUtilities` to Chisel would require moving a significant number of tests (to avoid cyclic dependencies between FIRRTL and Chisel). We should discuss this at the next chisel-dev meeting.

PR #785, opened 2018-04-10
Make DiGraph.linearize be iterative instead of recursive

jackkoenig:
Now insanely long chains will no longer cause a stack overflow. 

~~This does change the ordering in RemoveWires a bit, using a breadth-first traversal rather than a depth-first traversal.~~

Edit: The below is no longer relevent

For example:
```
circuit foo :
  module foo :
    input x : UInt<8>
    output z : UInt
    node a = add(x, UInt(1))
    node b = add(a, UInt(2))
    node c = add(b, UInt(3))
    node d = add(a, UInt(4))
    z <= and(c, d)
```
This ordering used to be maintained, but now it swaps the order of `c` and `d` because `d` depends directly on `a` whereas `c` depends on `b` which depends on `a`.
```
circuit foo :
  module foo : 
    input x : UInt<8> 
    output z : UInt<11> 
  
    node a = add(x, UInt<1>("h1")) 
    node b = add(a, UInt<2>("h2")) 
    node d = add(a, UInt<3>("h4")) 
    node c = add(b, UInt<2>("h3"))
    z <= and(c, d)
```

### DIFF
--- a/src/main/scala/firrtl/graph/DiGraph.scala
+++ b/src/main/scala/firrtl/graph/DiGraph.scala
@@ -71,42 +71,42 @@ class DiGraph[T] private[graph] (private[graph] val edges: LinkedHashMap[T, Link
     * traversal
     */
   def linearize: Seq[T] = {
-    // Kahn's algorithm
-    // Calculate all in degrees, linked because it affects initial toVisit order
-    val inDegree = mutable.LinkedHashMap[T, Int]()
-    for ((u, vs) <- edges) {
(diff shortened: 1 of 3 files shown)


jackkoenig:
In the interest of not changing code generation without a better reason than this, I changed it back to the old recursive algorithm explicitly transformed to be iterative.

albert-magyar:
LGTM, this is more or less how I've dealt with this issue in other places where ordering changes would be bad.

PR #786, opened 2018-04-10
Cleaning up BlackBoxSourceHelper

hcook:
The implementation had a few quirks, all addressed by this PR:
* It copied over the resource file once-per-instance, not once-per-module-annotation
* The resulting file list listed files once-per-instance, not once-per-file
* The file list contained a `-v` prefix per line, which is not standard for `.f` files AFAIK
* The file list file name assumed that all resources are verilog (they might be `.cc`, `.h`, `.tab` etc)
* The file list file name scala constant used a capital letter while not being a class

### DIFF
--- a/src/main/scala/firrtl/transforms/BlackBoxSourceHelper.scala
+++ b/src/main/scala/firrtl/transforms/BlackBoxSourceHelper.scala
@@ -113,8 +113,7 @@ object BlackBoxSourceHelper {
 
   def writeFileList(files: Set[File], targetDir: File) {
     if (files.nonEmpty) {
-      val text = files.map { fileName => s"-v $fileName" }.mkString("\n")
-      writeTextToFile(text, new File(targetDir, fileListName))
+      writeTextToFile(files.mkString("\n"), new File(targetDir, fileListName))
     }
   }
 
--- a/src/main/scala/firrtl/transforms/BlackBoxSourceHelper.scala
+++ b/src/main/scala/firrtl/transforms/BlackBoxSourceHelper.scala
@@ -67,22 +67,28 @@ class BlackBoxSourceHelper extends firrtl.Transform {
     */
   override def execute(state: CircuitState): CircuitState = {
     val (annos, targetDir) = collectAnnos(state.annotations)
-    val fileList = annos.foldLeft(List.empty[String]) {
-      case (fileList, anno) => anno match {
-        case BlackBoxResourceAnno(_, resourceId) =>
-          val name = resourceId.split("/").last
-          val outFile = new File(targetDir, name)
-          BlackBoxSourceHelper.copyResourceToFile(resourceId,outFile)
-          outFile.getAbsolutePath +: fileList
-        case BlackBoxInlineAnno(_, name, text) =>
-          val outFile = new File(targetDir, name)
-          val writer = new PrintWriter(outFile)
-          writer.write(text)
-          writer.close()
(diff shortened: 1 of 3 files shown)


src/main/scala/firrtl/transforms/BlackBoxSourceHelper.scala
```diff
@@ -67,40 +67,33 @@ class BlackBoxSourceHelper extends firrtl.Transform {
...
-          outFile.getAbsolutePath +: fileList
-        case _ => throwInternalError()
-      }
+
+    val resourceFiles: Set[File] = annos.collect {
+      case BlackBoxResourceAnno(_, resourceId) =>
+        val name = resourceId.split("/").last
+        val outFile = new File(targetDir, name)
+        (resourceId, outFile)
+    }.toSet[(String, File)].map { case (res, file) =>
```
jackkoenig: I actually think it would be better to use `.distinct` here instead of changing to a `Set` just for determinism in the ordering

src/main/scala/firrtl/transforms/BlackBoxSourceHelper.scala
```diff
@@ -73,7 +73,7 @@ class BlackBoxSourceHelper extends firrtl.Transform {
         val name = resourceId.split("/").last
         val outFile = new File(targetDir, name)
         (resourceId, outFile)
-    }.toSet[(String, File)].map { case (res, file) =>
+    }.map { case (res, file) =>
```
jackkoenig: Now it won't remove duplicates, right? Need `.distinct`

src/main/scala/firrtl/util/BackendCompilationUtilities.scala
```diff
@@ -91,7 +91,7 @@ trait BackendCompilationUtilities {
     val topModule = dutFile
 
     val blackBoxVerilogList = {
-      val list_file = new File(dir, firrtl.transforms.BlackBoxSourceHelper.FileListName)
+      val list_file = new File(dir, firrtl.transforms.BlackBoxSourceHelper.fileListName)
```
chick: Isn't the preferred style to have **constants** in companion objects start with upper case.

jackkoenig: Upon further evaluation, it is: https://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods

That being said, this isn't really a constant since it depends on dir

ucbjrl:
Put a note about this change into the release notes for the release containing this commit - 997fb0a.

jackkoenig:
Is this included in RC2?

ucbjrl:
No. It didn't have a 1.1.0 milestone. And I would have had to add the recent fix to chisel-testers in order to get all the release tests to pass.

jackkoenig:
Yeah that makes sense, I was just a little confused what release you were referring to in your previous comment, or do you just mean abstractly whichever released ends up including it?

ucbjrl:
To see what commits are included in a release, click on the release tag in the left-hand pane of the appropriate release ([`v1.1.0-RC2`](https://github.com/freechipsproject/firrtl/tree/v1.1.0-RC2) on the [frirrtl releases page](https://github.com/freechipsproject/firrtl/releases)), then click on the [`1,253 commits`](https://github.com/freechipsproject/firrtl/commits/v1.1.0-RC2) link.

PR #787, opened 2018-04-11
Fix bug in Constant Propagation for registers propped to zero

jackkoenig:
It wasn't properly padding the width of the constant zero.
Also add a test that shows the buggy behavior.


jackkoenig:
afaik this hasn't manifested, I was just poking around the code and noticed this issue

PR #788, opened 2018-04-13
Fix pathological behavior of Namespace for name collisions

jackkoenig:
When I wrote the bad code in question, I assumed the number of names being created would never be that bad. With some enhanced profiling, I saw that something like 15% of the compile time for the Rob was going to this function, doh!

### DIFF
--- a/src/main/scala/firrtl/Namespace.scala
+++ b/src/main/scala/firrtl/Namespace.scala
@@ -11,7 +11,8 @@ class Namespace private {
   private val tempNamePrefix: String = "_GEN"
   // Begin with a tempNamePrefix in namespace so we always have a number suffix
   private val namespace = mutable.HashSet[String](tempNamePrefix)
-  private var tempN = 0
+  // Memoize where we were on a given prefix
+  private val indices = mutable.HashMap[String, Int]()
 
   def tryName(value: String): Boolean = {
     val unused = !contains(value)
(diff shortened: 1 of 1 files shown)


jackkoenig:
This improves performance on Rob and a dual-core rocket-chip config by a decent amount:
```
master
[info] BenchmarkVerilogCompiler.measureRob       ss   10  13.782 ± 2.543   s/op
[info] BenchmarkVerilogCompiler.measureRocket    ss   10  29.050 ± 5.140   s/op

This PR
[info] BenchmarkVerilogCompiler.measureRob       ss   10  10.885 ± 2.791   s/op
[info] BenchmarkVerilogCompiler.measureRocket    ss   10  24.615 ± 3.496   s/op
```

~20% faster on Rob, ~15% faster on rocket-chip

src/main/scala/firrtl/Namespace.scala
```diff
@@ -21,23 +22,23 @@ class Namespace private {
...
+    // First try, just check
+    if (tryName(value)) value
+    else {
+      var idx = indices.getOrElse(value, 0)
+      var str = value
+      do {
+        str = s"${value}_$idx"
+        idx += 1
+      }
+      while (!(tryName(str)))
```
colinschmidt: Is this the right style? I always put these on the same line as closing brackets

jackkoenig: Yeah I usually put it on the same line as well, I just goofed

jackkoenig:
It does trace runtime for memory usage. I was careful to ensure that it only stores indices when there is a collision. We could add a threshold for when it should start storing the index (once there are 4 or 10 or something). I don't have a good memory profiling setup so I'll look into that and report back.

jackkoenig:
I don't have a great way to evaluate memory use, but using `/usr/bin/time -l` on my MBP I measured the max heap use and got:

|   | Rob | dual-core rocket |
| --- | --- | --- |
| master | 1627.77 +/- 29.29 MB | 2237.27 +/- 28.55 MB |
| this PR | 1509.31 +/- 63.19 MB | 2245.83 +/- 35.59 MB |

Somehow this PR *reduces* the total memory size on Rob (no change on dual core rocket). I don't understand how, but the results are pretty consistent across many runs 🤷‍♀️ 

In any case, I think this is good to go @colinschmidt

jackkoenig:
h/t @aswaterman for pointing out that during the former pathological behavior, the linear search involved creating lots of strings which will far outpace the memory used by the `indices` map. Thus this should be an improvement on performance and a wash or mild improvement for memory use.

PR #790, opened 2018-04-13
Remove infinitely recursive function

jackkoenig:
h/t @sdtwigg

It's unused (obviously or things wouldn't work). It just shouldn't be there.


azidar:
Lol that's an old function.... I think it was supposed to call the times function below, so the argument ordering didn't matter.

PR #792, opened 2018-04-20
Adds tests that show VerilogMemDelays crashing.

chick:
@jackkoenig I came across this testing Treadle.
I created a VerilogMemDelaySpec that illustrates the problem
Test may be redundant


jackkoenig:
We should certainly have tests for VerilogMemDelays actually correctly working, although integration tests with SeqReadMems and any other higher read or write latency does that to some extent

chick:
I removed one of the redundant tests.

PR #794, opened 2018-04-24
Mechanism to stop verilator from generating VCD file Chisel Issue #808

chick:
Add optional argument to verilogToCpp to provide access to option.
New option suppressVerilatorVCD can be used to turn off verilator trace mode.
Provide simple test of this functionality.


jackkoenig:
I see that FIRRTL constructs the command-line used in Verilator, but unless I'm missing something, the only place this new FirrtlOptions field is read is within `verilogToCpp`. Why not make `suppressVerilatorVCD` a Boolean argument to `verilogToCpp` and then instead add it as a field to the `TesterOptionsManager` where all of the other backend specific options are?

PR #798, opened 2018-05-01
TopWiring Transform

alonamid:
A transform that punches out annotated wires hierarchically to the top level. This transform is complimentary to wiring utils, since it handles the case of punching out through multiple instantiations of modules.
The transform takes in target wires through annotations, and a prefix annotation to determine the name of the new top-level ports.
The transform can also take in an optional function parameter (through an annotation) to generate custom output files based on the new ports that were punched out (useful for future features).
@azidar @jackkoenig

### DIFF
--- a/src/main/scala/firrtl/analyses/InstanceGraph.scala
+++ b/src/main/scala/firrtl/analyses/InstanceGraph.scala
@@ -18,7 +18,7 @@ import firrtl.Mappers._
   */
 class InstanceGraph(c: Circuit) {
 
-  private val moduleMap = c.modules.map({m => (m.name,m) }).toMap
+  val moduleMap = c.modules.map({m => (m.name,m) }).toMap
   private val instantiated = new mutable.HashSet[String]
   private val childInstances =
     new mutable.HashMap[String,mutable.Set[WDefInstance]]
@@ -93,6 +93,14 @@ class InstanceGraph(c: Circuit) {
   def moduleOrder: Seq[DefModule] = {
     graph.transformNodes(_.module).linearize.map(moduleMap(_))
   }
+
+  
+  /** Given a circuit, returns a map from module name to children
+     * instance/module definitions
+     */
+  def getChildrenInstances: scala.collection.Map[String,mutable.Set[WDefInstance]] = childInstances 
+
+
 }
 
 object InstanceGraph {
--- a/src/main/scala/firrtl/transforms/TopWiring.scala
+++ b/src/main/scala/firrtl/transforms/TopWiring.scala
@@ -16,17 +16,13 @@ import collection.mutable
 import firrtl.passes.wiring._
 
 /** Annotation for optional output files **/
-case class TopWiringOutputFilesAnnotation(dirName: String, outputFunction: (String,String,Seq[((ComponentName, Type, Boolean, Seq[String]), Int)], CircuitState) => CircuitState) extends NoTargetAnnotation
+case class TopWiringOutputFilesAnnotation(dirName: String, outputFunction: (String,Seq[((ComponentName, Type, Boolean, Seq[String],String), Int)], CircuitState) => CircuitState) extends NoTargetAnnotation
 
-/** Annotation for a prefix that will be added to all the port names that are punched out to the top level **/
-case class TopWiringPrefixAnnotation(prefix: String) extends NoTargetAnnotation
-
-/** Annotation for indicating component to be wired */
-case class TopWiringAnnotation(target: ComponentName) extends
+/** Annotation for indicating component to be wired, and what prefix to add to the ports that are generated */
+case class TopWiringAnnotation(target: ComponentName, prefix: String) extends
     SingleTargetAnnotation[ComponentName] {
   def duplicate(n: ComponentName) = this.copy(target = n)
 }
-
 
 
 /** Punch out annotated ports out to the toplevel of the circuit.
@@ -42,32 +38,32 @@ class TopWiringTransform extends Transform {
  
   /** Get the names of the targets that need to be wired */
 
-  private def getSourceNames(state: CircuitState): Seq[ComponentName] = {
-    state.annotations.collect { case TopWiringAnnotation(srcname) => srcname }
+  private def getSourceNames(state: CircuitState): Map[ComponentName, String] = {
+    state.annotations.collect { case TopWiringAnnotation(srcname,prefix) => (srcname -> prefix) }.toMap.withDefaultValue("")
   }
 
 
   private def getSourceModNames(state: CircuitState): Seq[String] = {
-    state.annotations.collect { case TopWiringAnnotation(ComponentName(_,ModuleName(srcmodname, _))) => srcmodname }
+    state.annotations.collect { case TopWiringAnnotation(ComponentName(_,ModuleName(srcmodname, _)),_) => srcmodname }
   }
 
 
 
-  def getSourceTypes(sourceList: Seq[ComponentName], sourceMap: mutable.Map[String, Seq[(ComponentName, Type, Boolean, InstPath)]], currentmodule: ModuleName, state: CircuitState)(s: Statement): Statement = s match {
+  def getSourceTypes(sourceList: Map[ComponentName, String], sourceMap: mutable.Map[String, Seq[(ComponentName, Type, Boolean, InstPath, String)]], currentmodule: ModuleName, state: CircuitState)(s: Statement): Statement = s match {
     // If target wire, add name and size to to sourceMap
     case w: IsDeclaration =>
-      if (sourceList.contains(ComponentName(w.name, currentmodule))) {
-          val (isport, tpe) = w match {
-            case d: DefWire => (false, d.tpe)
-            case d: DefNode => (false, d.value.tpe)
-            case d: DefRegister => (false, d.tpe)
-            case d: Port => (true, d.tpe)
+      if (sourceList.keys.toSeq.contains(ComponentName(w.name, currentmodule))) {
+          val (isport, tpe, prefix) = w match {
+            case d: DefWire => (false, d.tpe, sourceList(ComponentName(w.name,currentmodule)))
+            case d: DefNode => (false, d.value.tpe, sourceList(ComponentName(w.name,currentmodule)))
+            case d: DefRegister => (false, d.tpe, sourceList(ComponentName(w.name,currentmodule)))
+            case d: Port => (true, d.tpe, sourceList(ComponentName(w.name,currentmodule)))
             case _ => throw new Exception(s"Cannot wire this type of declaration! ${w.serialize}")
           }
           val name = w.name
           sourceMap.get(currentmodule.name) match {
-            case Some(xs:Seq[(ComponentName, Type, Boolean, InstPath)]) => sourceMap.update(currentmodule.name, xs :+ (ComponentName(w.name,currentmodule), tpe, isport ,Seq[String](w.name)))
-            case None => sourceMap(currentmodule.name) = Seq((ComponentName(w.name,currentmodule), tpe, isport ,Seq[String](w.name)))
+            case Some(xs:Seq[(ComponentName, Type, Boolean, InstPath, String)]) => sourceMap.update(currentmodule.name, xs :+ (ComponentName(w.name,currentmodule), tpe, isport ,Seq[String](w.name), prefix))
+            case None => sourceMap(currentmodule.name) = Seq((ComponentName(w.name,currentmodule), tpe, isport ,Seq[String](w.name), prefix))
           }
           //sourceMap(currentmodule.name) +: (ComponentName(w.name,currentmodule), tpe, isport ,Seq[String](currentmodule.name))
       }
@@ -79,18 +75,18 @@ class TopWiringTransform extends Transform {
 
 
 
-  def getSourceTypesPorts(sourceList: Seq[ComponentName], sourceMap: mutable.Map[String, Seq[(ComponentName, Type, Boolean, InstPath)]], currentmodule: ModuleName, state: CircuitState)(s: Port): CircuitState = s match {
+  def getSourceTypesPorts(sourceList: Map[ComponentName, String], sourceMap: mutable.Map[String, Seq[(ComponentName, Type, Boolean, InstPath, String)]], currentmodule: ModuleName, state: CircuitState)(s: Port): CircuitState = s match {
     // If target port, add name and size to to sourceMap
     case w: IsDeclaration =>
-      if (sourceList.contains(ComponentName(w.name, currentmodule))) {
-          val (isport, tpe) = w match {
-            case d: Port => (true, d.tpe)
+      if (sourceList.keys.toSeq.contains(ComponentName(w.name, currentmodule))) {
+          val (isport, tpe, prefix) = w match {
+            case d: Port => (true, d.tpe, sourceList(ComponentName(w.name,currentmodule)))
             case _ => throw new Exception(s"Cannot wire this type of declaration! ${w.serialize}")
           }
           val name = w.name
           sourceMap.get(currentmodule.name) match {
-            case Some(xs:Seq[(ComponentName, Type, Boolean, InstPath)]) => sourceMap.update(currentmodule.name, xs :+ (ComponentName(w.name,currentmodule), tpe, isport ,Seq[String](w.name)))
-            case None => sourceMap(currentmodule.name) = Seq((ComponentName(w.name,currentmodule), tpe, isport ,Seq[String](w.name)))
+            case Some(xs:Seq[(ComponentName, Type, Boolean, InstPath, String)]) => sourceMap.update(currentmodule.name, xs :+ (ComponentName(w.name,currentmodule), tpe, isport ,Seq[String](w.name), prefix))
+            case None => sourceMap(currentmodule.name) = Seq((ComponentName(w.name,currentmodule), tpe, isport ,Seq[String](w.name), prefix))
           }
           //sourceMap(currentmodule.name) +: (ComponentName(w.name,currentmodule), tpe, isport ,Seq[String](currentmodule.name))
       }
@@ -109,7 +105,7 @@ class TopWiringTransform extends Transform {
     *
     * These paths are relative but cross module (they refer down through instance hierarchy)
     */
-  private def getSourcesMap(state: CircuitState): Map[String, Seq[(ComponentName, Type, Boolean, InstPath)]] = {
+  private def getSourcesMap(state: CircuitState): Map[String, Seq[(ComponentName, Type, Boolean, InstPath, String)]] = {
     val sSourcesModNames = getSourceModNames(state)
     val sSourcesNames = getSourceNames(state)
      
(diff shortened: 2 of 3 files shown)


alonamid:
@seldridge I originally tried implementing this using the WiringTransform. The problem was with the case in which there were signals/wires with the same name in different modules instantiated multiple time across the hierarchy (and perhaps the same module instantiated multiple times). I want to know exactly where each signal is coming from (so I can perform some actions at the top), and I could not determine the WiringTransform's behavior for that case.

seldridge:
I think you're bumping into https://github.com/freechipsproject/firrtl/issues/773...

Specifically, the following will fail with #799 , though it should be supported:
```scala
    val sinks = Seq(ModuleName("Top", CircuitName("Top")))
    val sourceA = ComponentName("r", ModuleName("A", CircuitName("Top")))
    val sas = Seq(WiringInfo(sourceA, sinks, "pin"))
    val input =
      """|circuit Top :
         |  module Top :
         |    input clock: Clock
         |    inst a1 of A
         |    a1.clock <= clock
         |    inst a2 of A
         |    a2.clock <= clock
         |  module A :
         |    input clock: Clock
         |    reg r: UInt<5>, clock
         |""".stripMargin
    val c = passes.foldLeft(parse(input)) {
      (c: Circuit, p: Pass) => p.run(c)
    }
    val wiringPass = new Wiring(sas)
    val retC = wiringPass.run(c)
```

This will fail with: `[firrtl.passes.wiring.WiringException: Unable to determine source mapping for sink 'List(Top)'`. The wiring transform has a built-in assumption (my lack of foresight) that you always wanted to wire one source to one or more sinks. The sources and sinks may be instantiated multiple times. The mapping from sources to sinks is then implicit based on distance---a sink is "owned" by its nearest source. If a sink doesn't have an owner, then it fails with this error.

Wiring to the top, while obviously needed, was never considered (again, this is my fault). However, that's a subset of multiple source, single sink wiring. This is generally useful and needs to be supported in the wiring transform, e.g., connect all these performance counters to my blackbox. So, #773 has to be cleared, I just have not taken a stab at it just yet. I would much prefer a full solution for #773 as opposed to a separate transform.

src/main/scala/firrtl/transforms/TopWiring.scala
```diff
@@ -0,0 +1,247 @@
...
+    * For Modules that contain or are in the parent hierarchy to modules containing target wires
+    * 1. Add ports for each target wire this module is parent to
+    * 2. Connect these ports to ports of instances that are parents to some number of target wires
+    */
+  // TODO Make code more clear
+  private def onModule(prefix: String, sources: Map[String, Seq[(ComponentName, Type, Boolean, InstPath)]])
+                      (module: DefModule): DefModule = {
+    sources.get(module.name) match {
+      case Some(p) => 
+        val newPorts = p.map{ case (ComponentName(cname,_), tpe, _ , path) => Port(NoInfo, prefix + path.mkString("_"), Output, tpe)}
```
seldridge: I like the use of `InstPath` to determine the name (you get an understanding of where the signal is going), however, this can still collide with other modules components, yes? You can use a `Namespace` here to make sure that the name is valid and then either get a new one or fail if you really wanted that name.

src/main/scala/firrtl/transforms/TopWiring.scala
```diff
@@ -0,0 +1,247 @@
...
+import firrtl.graph._
+
+import java.io._
+import scala.io.Source
+import collection.mutable
+
+import firrtl.passes.wiring._
+
+/** Annotation for optional output files **/
+case class TopWiringOutputFilesAnnotation(dirname: String, outputfunction: (String,String,Seq[((ComponentName, Type, Boolean, Seq[String]), Int)], CircuitState) => CircuitState) extends NoTargetAnnotation
```
seldridge: I'd be inclined to use a case class for the inputs to the `outputfunction` for the sake of clarity (also, I think `dirName` and `outputFunction` for Scala naming conventions).

src/main/scala/firrtl/transforms/TopWiring.scala
```diff
@@ -0,0 +1,247 @@
...
+          }
+     }
+     testOutputFile.close()
+     state 
+  }
+
+
+  def execute(state: CircuitState): CircuitState = {
+
+    lazy val outputTuple: Option[(String, (String,String,Seq[((ComponentName, Type, Boolean, InstPath), Int)], CircuitState) => CircuitState)] = state.annotations.collectFirst { case TopWiringOutputFilesAnnotation(td,of) => (td, of) }
```
seldridge: Why the lazy vals here? 

Also, you're using `collectFirst` here, but wouldn't you want to be able to punch multiple disjoint sources to the top with different `TopWiringOutputFilesAnnotation`? Ideally, I would expect that all of this is either keyed on a value or packaged up into one larger annotation (forthcoming multiple target annotations would simplify this...).

src/main/scala/firrtl/transforms/TopWiring.scala
```diff
@@ -0,0 +1,247 @@
...
+  /** Create a map of Module name to target wires under this module
+    *
+    * These paths are relative but cross module (they refer down through instance hierarchy)
+    */
+  private def getSourcesMap(state: CircuitState): Map[String, Seq[(ComponentName, Type, Boolean, InstPath)]] = {
+    val sSourcesModNames = getSourceModNames(state)
+    val sSourcesNames = getSourceNames(state)
+     
+
+    val cMap = WiringUtils.getChildrenMap(state.circuit).toMap
```
seldridge: `getChildrenMap` should've been deprecated in favor of the DiGraph/InstanceGraph APIs. The only thing still using that is `ClockList` and that's overdue to be upgraded. Sorry, this should've been marked as deprecated long ago... Fixing that in #801.

src/main/scala/firrtl/transforms/TopWiring.scala
```diff
@@ -0,0 +1,247 @@
...
+    //val digraph = DiGraph(cMap.mapValues(v => v.map(_._2).toSet))
+    //val topSort = digraph.linearize.reverse
+    //previous 2 line replaced by the following single line
+    val topSort = (new firrtl.analyses.InstanceGraph(state.circuit)).moduleOrder.reverse
+
+    // Map of component name to relative instance paths that result in a debug wire
+    val sourcemods: mutable.Map[String, Seq[(ComponentName, Type, Boolean, InstPath)]] =
+      mutable.Map(sSourcesModNames.map(_ -> Seq()): _*)
+
+    state.circuit.modules.foreach { m => m map getSourceTypes(sSourcesNames, sourcemods, ModuleName(m.name, CircuitName(state.circuit.main)) , state) }
```
seldridge: `m => m map` -> `_ map`. Same thing in the next line.

src/main/scala/firrtl/transforms/TopWiring.scala
```diff
@@ -0,0 +1,247 @@
...
+     }
+     testOutputFile.close()
+     state 
+  }
+
+
+  def execute(state: CircuitState): CircuitState = {
+
+    lazy val outputTuple: Option[(String, (String,String,Seq[((ComponentName, Type, Boolean, InstPath), Int)], CircuitState) => CircuitState)] = state.annotations.collectFirst { case TopWiringOutputFilesAnnotation(td,of) => (td, of) }
+    lazy val prfix: Option[String] = state.annotations.collectFirst { case TopWiringPrefixAnnotation(pr) => pr }
```
seldridge: Same issue with `collectFirst` here. How are you going to deal with multiple prefixes for different sources that you want to wire out?

src/main/scala/firrtl/transforms/TopWiring.scala
```diff
@@ -0,0 +1,247 @@
+// See LICENSE for license details.
+package firrtl.transform
+package TopWiring
+
+import firrtl._
+import firrtl.ir._
+import firrtl.passes._
+import firrtl.annotations._
+import firrtl.Mappers._
+import firrtl.graph._
```
seldridge: Probably best to avoid wildcard imports for new code, [per the Chisel3 style guide](https://github.com/freechipsproject/chisel3/blob/master/doc/style.md#imports).

src/main/scala/firrtl/transforms/TopWiring.scala
```diff
@@ -0,0 +1,247 @@
...
+  def getSourceTypes(sourceList: Seq[ComponentName], sourceMap: mutable.Map[String, Seq[(ComponentName, Type, Boolean, InstPath)]], currentmodule: ModuleName, state: CircuitState)(s: Statement): Statement = s match {
+    // If target wire, add name and size to to sourceMap
+    case w: IsDeclaration =>
+      if (sourceList.contains(ComponentName(w.name, currentmodule))) {
+          val (isport, tpe) = w match {
+            case d: DefWire => (false, d.tpe)
+            case d: DefNode => (false, d.value.tpe)
+            case d: DefRegister => (false, d.tpe)
+            case d: Port => (true, d.tpe)
+            case _ => sys.error(s"Cannot wire this type of declaration! ${w.serialize}")
```
seldridge: Any reason why not just use an exception?

src/main/scala/firrtl/transforms/TopWiring.scala
```diff
@@ -0,0 +1,247 @@
...
+    sources.get(module.name) match {
+      case Some(p) => 
+        val newPorts = p.map{ case (ComponentName(cname,_), tpe, _ , path) => Port(NoInfo, prefix + path.mkString("_"), Output, tpe)}
+        
+
+        // Add connections to Module
+        module match {
+          case m: Module =>
+            val connections: Seq[Connect] = p.map { case (ComponentName(cname,_), _, _ , path) =>
+                val modRef = WRef(prefix + path.mkString("_"))
```
seldridge: Same problem as above related to use of a Namespace. This can collide if you aren't careful.

src/test/scala/firrtlTests/transforms/TopWiringTest.scala
```diff
@@ -0,0 +1,167 @@
...
+
+   def TopWiringDummyOutputFilesFunction(prefix: String, dir: String, mapping: Seq[((ComponentName, Type, Boolean, Seq[String]), Int)], state: CircuitState): CircuitState = {
+     state
+   }
+
+   def TopWiringTestOutputFilesFunction(prefix: String, dir: String, mapping: Seq[((ComponentName, Type, Boolean, Seq[String]), Int)], state: CircuitState): CircuitState = {
+     val testOutputFile = new PrintWriter(new File(dir, "TopWiringOutputTest.txt" ))
+     mapping map {
+          case ((_, tpe, _, path), index) => {
+            val portwidth = tpe match { case GroundType(IntWidth(w)) => w }
```
seldridge: This will error out if you try to match on a non-GroundType. Is this guaranteed to never happen?

src/main/scala/firrtl/transforms/TopWiring.scala
alonamid: The alternative here (as you said) is to add all of these properties (prefix, output directory, output function) into the wire annotation, and carry the information through the sources and mapping data-structure. I was planning to run this transform only once, so I guess that's the reason I wrote it this way.

alonamid:
I can fix the stylistic issues, and merge all the annotation into a one big annotation. However, the question here is whether you want to pull this, or whether you want to wait for the complete wiring transform. Given that I originally tried to generate the files I need using the wiring transform, I don't mind either (I already have code written to wrap the wiring transform to generate my relevant output files from the time I assumed it was able to handle multiple instantiations). For me, it's just a matter of it handling multiple instantiations.

src/main/scala/firrtl/transforms/TopWiring.scala
alonamid: Aren't instance names unique? if an instance name is unique, than the InstPath must be unique as well, isn't it? Can you give me an example of a collision?

src/main/scala/firrtl/transforms/TopWiring.scala
alonamid: What would be the correct API then? I could not find an exact equivalent in DiGraph/InstanceGraph

src/test/scala/firrtlTests/transforms/TopWiringTest.scala
alonamid: Does midForm have non-GroundTypes?

seldridge:
I've got this scheduled to talk about tomorrow during the dev meeting. I'm taking a look at fixing (hopefully once and for all) the wiring transform to support arbitrarily complex operations that involve instance-to-instance connections. This may take some time, though. Let's see what @azidar thinks here...

What may be beneficial is if you can help flesh out #773 and make sure that what is proposed covers your usage case (and potentially others that you may have been thinking about).

src/test/scala/firrtlTests/transforms/TopWiringTest.scala
seldridge: MidForm can have aggregates. `LowerTypes` is what removes those.

src/main/scala/firrtl/transforms/TopWiring.scala
seldridge: Instance names are guaranteed to be unique. However, there's nothing preventing a user from defining an existing wire/port/reg in the design with one of these names. In your first test case, you create the top-level port `output topwiring_a1_b1_c1_x: UInt<1>`. However, the user could have created this signal already defined `input topwiring_a1_b1_c1_x: Clock`. This _shouldn't_ happen, but the user should be able to name (or Chisel/earlier FIRRTL passes) should be able to name signals whatever they want without having to worry about this transform generating conflicting names.

src/main/scala/firrtl/transforms/TopWiring.scala
seldridge: There won't be an exact equivalent (and I'd have to actually think of what you want here), but the genericity of DiGraph/InstanceGraph should give you all the primitives you need here. It won't be the exact same API, but `InstanceGraph` usually takes care of all of these things. 

ChildrenMap should be the edges for all WDefInstance nodes that share the same module name.

seldridge:
Update: what we can do with this is get it cleaned up and merge it. Once a full solution for #773 is done, then we'll deprecate this and transparently switch to the wiring transform as the backend.

src/test/scala/firrtlTests/transforms/TopWiringTest.scala
```diff
@@ -0,0 +1,278 @@
...
+}
+import firrtl.transform.TopWiring._
+
+
+/**
+ * Tests TopWiring transformation
+ */
+class TopWiringTests extends LowTransformSpec {
+
+   def TopWiringDummyOutputFilesFunction(dir: String, mapping: Seq[((ComponentName, Type, Boolean, Seq[String], String), Int)], state: CircuitState): CircuitState = {
```
seldridge: Begin with  lowercase for a function.

src/test/scala/firrtlTests/transforms/TopWiringTest.scala
```diff
@@ -0,0 +1,278 @@
...
+     state
+   }
+
+   def TopWiringTestOutputFilesFunction(dir: String, mapping: Seq[((ComponentName, Type, Boolean, Seq[String], String), Int)], state: CircuitState): CircuitState = {
```
seldridge: Lowercase for a function.

seldridge: I like this whole thing and I think it would make a lot of sense to make this the fixed behavior for top wiring.

src/main/scala/firrtl/transforms/TopWiring.scala
```diff
@@ -0,0 +1,238 @@
...
+    val instGraph = new firrtl.analyses.InstanceGraph(state.circuit)
+    val cMap = instGraph.getChildrenInstances.map{ case (m, wdis) => (m -> wdis.map{ case wdi => (wdi.name, wdi.module) }.toSeq) }.toMap
+    val topSort = instGraph.moduleOrder.reverse
+
+    // Map of component name to relative instance paths that result in a debug wire
+    val sourcemods: mutable.Map[String, Seq[(ComponentName, Type, Boolean, InstPath, String)]] =
+      mutable.Map(sSourcesModNames.map(_ -> Seq()): _*)
+
+    state.circuit.modules.foreach { m => m map getSourceTypes(sSourcesNames, sourcemods, ModuleName(m.name, CircuitName(state.circuit.main)) , state) }
+    state.circuit.modules.foreach { m => m.ports.foreach { p => Seq(p) map getSourceTypesPorts(sSourcesNames, sourcemods, ModuleName(m.name, CircuitName(state.circuit.main)) , state) }}
```
seldridge: You may consider putting these in a more "Scala-like" format of separating the method invocations on separate lines (you'll need to do something to get this under 120 characters). Also, usually there's no need to use `{}` when `()` will suffice.

src/main/scala/firrtl/analyses/InstanceGraph.scala
```diff
@@ -18,7 +18,7 @@ import firrtl.Mappers._
   */
 class InstanceGraph(c: Circuit) {
 
-  private val moduleMap = c.modules.map({m => (m.name,m) }).toMap
```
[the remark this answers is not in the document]
alonamid: I originally wanted to use it, but then decided not to. It still seems to be a useful public API

src/main/scala/firrtl/transforms/TopWiring.scala
```diff
@@ -0,0 +1,238 @@
...
+import firrtl.annotations._
+import firrtl.Mappers._
+import firrtl.graph._
+
+import java.io._
+import scala.io.Source
+import collection.mutable
+
+/** Annotation for optional output files **/
+case class TopWiringOutputFilesAnnotation(dirName: String, outputFunction: (String,Seq[((ComponentName, Type, Boolean, Seq[String],String), Int)], CircuitState) => CircuitState) extends NoTargetAnnotation
```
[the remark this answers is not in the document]
alonamid: My assumption is that other passes will wrap/use this transform. My specific use-case requires to output some files with information about the wires that were punched to the top (for that file to be included in a Vivado flow). Hence, I'm assuming I will have a wrapper pass that has that outputFunction which will generate my requires files. I'm not assuming that this annotation will be passed down from chisel.

alonamid:
I'm not sure I understand why you say this has to be in MidForm. My original use-case actually had it using LowForm (The reason I had it as MidForm here was because I was blending code from several sources). I change the transform the LowForm, and I it seems to at least pass the basic tests. I currently prefer not dealing with connecting aggregates, and it seems like LowForm is a better solution here rather than an aggregate exception. (especially given that my output file use-case would prefer to know about independent wires rather than aggregates).
Other than that, I hope I addressed most of your cosmetics / docs / testing requests

seldridge:
Thanks for addressing the feedback, @alonamid.

After thinking about it more, I'm wrong -- the use of `MidForm` is an implementation choice. The only major benefit that I can think of is that you get cheap `AggregateType` connections. _Annotations get automatically renamed, so there's no issue of aliasing (this is what I was thinking of above, but it's a non-issue)_. @azidar, you likely have some thoughts here as you wrote the original `WiringTransform` using `MidForm`?

Anyway, I think you can implement it either way and the benefits are subjective. `LowForm` :+1:. No problem there.

I _think_ that this will work for the generic case of wiring signals all the way to the top and should have positive implications for all of the following:
  - https://github.com/freechipsproject/rocket-chip/issues/1233, https://github.com/seldridge/rocket-rocc-examples/issues/9
  - https://github.com/freechipsproject/rocket-chip/issues/1379
  - https://stackoverflow.com/questions/41679713/how-to-retrieve-rocket-core-module-parameters-from-top-module-in-fpga-zynq-repos

lgtm

src/main/scala/firrtl/analyses/InstanceGraph.scala
```diff
@@ -93,6 +93,14 @@ class InstanceGraph(c: Circuit) {
   def moduleOrder: Seq[DefModule] = {
     graph.transformNodes(_.module).linearize.map(moduleMap(_))
   }
+
+  
+  /** Given a circuit, returns a map from module name to children
+     * instance/module definitions
+     */
+  def getChildrenInstances: scala.collection.Map[String,mutable.Set[WDefInstance]] = childInstances
```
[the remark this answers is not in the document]
azidar: Yup, I think its a very useful function.

src/main/scala/firrtl/transforms/TopWiring.scala
```diff
@@ -0,0 +1,269 @@
...
+
+  /** Create a map of Module name to target wires under this module
+    *
+    * These paths are relative but cross module (they refer down through instance hierarchy)
+    */
+  private def getSourcesMap(state: CircuitState): Map[String,Seq[(ComponentName, Type, Boolean, InstPath, String)]] = {
+    val sSourcesModNames = getSourceModNames(state)
+    val sSourcesNames = getSourceNames(state)
+    val instGraph = new firrtl.analyses.InstanceGraph(state.circuit)
+    val cMap = instGraph.getChildrenInstances.map{ case (m, wdis) =>
```
jackkoenig: `getChildrenInstances` returns a `scala.collection.Map[String,mutable.Set[WDefInstance]]`.
Below, in line 148 `val seqChildren: Seq[(ComponentName,Type,Boolean,InstPath,String)] = cMap(mod.name).flatMap {`
values from `cMap` are iterated on and used to generate Seqs that are returned for use later. This *could* result in nondeterminism

PR #799, opened 2018-05-01
Enable top ports for singly-instantiated sources in WiringTransform 

seldridge:
Practically, this enables functionality in the WiringTransform to wire signals all the way to the top. 

If the sink module is the lowest common ancestor, then the type of connection (a wire or a port) is a function of whether or not we're wiring to a component or a module. If the later, then the implied functionality is that we want a port on this module. This already happens when the sink module is not the LCA, but not for this case.

While not explicitly enabling blind "wire this signal to the top", this should provide the functionality for it.

@alonamid: I _think_ that this will give you all the functionality you need for your top annotations. I'm hoping that this will enable not having to maintain two wiring transforms.

See: https://github.com/freechipsproject/firrtl/pull/798

### DIFF
--- a/src/main/scala/firrtl/passes/wiring/Wiring.scala
+++ b/src/main/scala/firrtl/passes/wiring/Wiring.scala
@@ -107,16 +107,17 @@ class Wiring(wiSeq: Seq[WiringInfo]) extends Pass {
         // Case where the sink is the LCA
         case Seq(WDefInstance(_,_,pm,_)) =>
           // Case where the source is also the LCA
+          val tpe = if (sinkComponents.contains(pm)) DecWire else DecOutput
           if (source.drop(lca.size).isEmpty) {
             meta(pm) = meta(pm).copy (
-              addPortOrWire = Some((portNames(pm), DecWire))
+              addPortOrWire = Some((portNames(pm), tpe))
             )
           } else {
             val WDefInstance(_,ci,cm,_) = source.drop(lca.size).head
             val to = s"${portNames(pm)}"
             val from = s"$ci.${portNames(cm)}"
             meta(pm) = meta(pm).copy(
-              addPortOrWire = Some((portNames(pm), DecWire)),
+              addPortOrWire = Some((portNames(pm), tpe)),
               cons = (meta(pm).cons :+ (to, from)).distinct
             )
           }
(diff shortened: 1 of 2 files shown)


seldridge:
After further review, this enables wiring _one source_ all the way to the top. This will not work if you have multiple sources (or multiply instantiated sources) wiring to the same sink.

Nonetheless, I think this is actually the correct behavior here.

For circuit included in the test, trying to wire source component `a` of module `A` to sink module `Top` has the following behavior. This is the initial circuit:
```
circuit Top :
  module Top :
    input clock : Clock
  
    inst a of A
    a.clock <= clock

  module A :
    input clock : Clock
  
    reg a : UInt<5>, clock with :
      reset => (UInt<1>("h0"), a)
```

Behavior before showing this wire to `Top` and create a wire connection:
```
circuit Top :
  module Top :
    input clock : Clock
  
    wire pin : UInt<5>
    inst a of A
    a.clock <= clock
    pin <= a.a_0

  module A :
    input clock : Clock
    output a_0 : UInt<5>
  
    reg a : UInt<5>, clock with :
      reset => (UInt<1>("h0"), a)
    a_0 <= a
```

Behavior with this PR showing this wire to `Top` and create an output port:
```
circuit Top :
  module Top :
    input clock : Clock
    output pin : UInt<5>
  
    inst a of A
    a.clock <= clock
    pin <= a.a_0

  module A :
    input clock : Clock
    output a_0 : UInt<5>
  
    reg a : UInt<5>, clock with :
      reset => (UInt<1>("h0"), a)
    a_0 <= a
```

azidar:
So to clarify, this lets you wire a single source to a single top-level pin, but not multiply-instantiated modules wiring separately to the top? I believe that is the major use-case that @alonamid is handling? Is there a way to incorporate some of #798 into WiringTransform to enable that use-case?

seldridge:
This will wire one or more singly instantiated modules to a top-level pin. The pin, by luck, shows up as a flattened aggregate (e.g., `pin`, `pin_0`, `pin_1`). Multiply instantiated submodules will fail, however. This PR is really a bug fix to correct what I think was previously incorrect behavior not to deal with arbitrary top wiring.

>  I believe that [(top wiring)] is the major use-case that @alonamid is handling?

Yes, it is. I think the wiring transform should be modified to handle this use-case, though. Ideally, I would like one wiring primitive backed by one wiring transform. More complex transforms emit wiring transform primitives that the wiring transform cleans up, like for `ReplSeqMem`.

> Is there a way to incorporate some of #798 into WiringTransform to enable that use-case?

Some ideas, yes. @alonamid provides perfect, hierarchical wiring names (which was badly needed). The algorithm is, however, specific to the use case of wiring to the top. This won't handle other multi-source situations (e.g., connecting scattered performance counters to a RoCC that go up and down in the hierarchy). Granted, it isn't designed to!

seldridge:
With the existence of the `TopWiringTransform` and the slow pace of a generic wiring refactor for multi-source/single-sink support, I'm going to go ahead and close this.

PR #800, opened 2018-05-02
Replace truncating add and sub with addw/subw

jackkoenig:
Replaces old VerilogWrap which didn't work with split expressions and was
actually buggy anyway. This functionality reduces unnecessary intermediates in
emitted Verilog.

### DIFF
--- a/src/main/scala/firrtl/transforms/ReplaceTruncatingArithmetic.scala
+++ b/src/main/scala/firrtl/transforms/ReplaceTruncatingArithmetic.scala
@@ -1,0 +1,53 @@
+package firrtl
+package transforms
+
+import firrtl.ir._
+import firrtl.Mappers._
+import firrtl.PrimOps._
+import firrtl.WrappedExpression._
+
+import scala.collection.mutable
+
+/** Replaces non-expanding arithmetic
+  *
+  * This replaces FIRRTL primops with ops that are not legal FIRRTL. They are useful for emission to
+  * languages that support non-expanding arithmetic (like Verilog)
+  */
+object ReplaceTruncatingArithmetic {
+
+  /** Mapping from references to the [[Expression]]s that drive them */
+  type Netlist = mutable.HashMap[WrappedExpression, Expression]
+
+  val SeqBIOne = Seq(BigInt(1))
+
+  def onExpr(netlist: Netlist)(expr: Expression): Expression = 
+    expr.map(onExpr(netlist)) match {
+      case orig @ DoPrim(Tail, Seq(e), SeqBIOne, tailtpe) =>
+        netlist.getOrElse(we(e), e) match {
+          case DoPrim(Add, args, cs, _) => DoPrim(Addw, args, cs, tailtpe)
+          case DoPrim(Sub, args, cs, _) => DoPrim(Subw, args, cs, tailtpe)
+          case _ => orig // Not a candidate
+        }
+      case other => other // Not a candidate
+    }
+
+  def onStmt(netlist: Netlist)(stmt: Statement): Statement =
+    stmt.map(onStmt(netlist)).map(onExpr(netlist)) match {
+      case node @ DefNode(_, name, value) =>
+        netlist(we(WRef(name))) = value
+        node
+      case other => other
+    }
+  def onMod(mod: DefModule): DefModule = mod.map(onStmt(new Netlist))
+}
+
+class ReplaceTruncatingArithmetic extends Transform {
+  def inputForm = LowForm
+  def outputForm = LowForm
+
+  def execute(state: CircuitState): CircuitState = {
+    val modulesx = state.circuit.modules.map(ReplaceTruncatingArithmetic.onMod(_))
+    state.copy(circuit = state.circuit.copy(modules = modulesx))
+  }
+}
+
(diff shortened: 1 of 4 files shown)


src/main/scala/firrtl/transforms/ReplaceTruncatingArithmetic.scala
```diff
@@ -0,0 +1,53 @@
...
+
+/** Replaces non-expanding arithmetic
+  *
+  * This replaces FIRRTL primops with ops that are not legal FIRRTL. They are useful for emission to
+  * languages that support non-expanding arithmetic (like Verilog)
+  */
+object ReplaceTruncatingArithmetic {
+
+  /** Mapping from references to the [[Expression]]s that drive them */
+  type Netlist = mutable.HashMap[WrappedExpression, Expression]
```
azidar: Is there a reason this is WrappedExpression instead of Memoized?

src/main/scala/firrtl/transforms/ReplaceTruncatingArithmetic.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  val SeqBIOne = Seq(BigInt(1))
+
+  def onExpr(netlist: Netlist)(expr: Expression): Expression =
```
azidar: Since these are public functions (that may be useful outside this transform), perhaps make the names more clear? Or maybe users should infer the behavior from its context, in that its in a `ReplaceTruncatingArithmetic` object. Food for thought. I do think its worthwhile to add scaladoc which says it looks at netlist, but doesn't update it.

src/main/scala/firrtl/transforms/ReplaceTruncatingArithmetic.scala
```diff
@@ -0,0 +1,53 @@
...
+      case orig @ DoPrim(Tail, Seq(e), SeqBIOne, tailtpe) =>
+        netlist.getOrElse(we(e), e) match {
+          case DoPrim(Add, args, cs, _) => DoPrim(Addw, args, cs, tailtpe)
+          case DoPrim(Sub, args, cs, _) => DoPrim(Subw, args, cs, tailtpe)
+          case _ => orig // Not a candidate
+        }
+      case other => other // Not a candidate
+    }
+
+  def onStmt(netlist: Netlist)(stmt: Statement): Statement =
```
azidar: Similarly, consider changing this name.

src/main/scala/firrtl/transforms/ReplaceTruncatingArithmetic.scala
```diff
@@ -0,0 +1,53 @@
...
+      case other => other
+    }
+  def onMod(mod: DefModule): DefModule = mod.map(onStmt(new Netlist))
+}
+
+class ReplaceTruncatingArithmetic extends Transform {
+  def inputForm = LowForm
+  def outputForm = LowForm
+
+  def execute(state: CircuitState): CircuitState = {
```
azidar: Where should we put documentation for a transform? The object? The class? The execute function?

src/main/scala/firrtl/transforms/ReplaceTruncatingArithmetic.scala
jackkoenig: It's WrappedExpression so that WRefs that aren't fully filled out (like just filling out the name) will map to the same value, but it possibly should be `MemoizedHash[WrappedExpression]`

src/main/scala/firrtl/transforms/ReplaceTruncatingArithmetic.scala
jackkoenig: Yeah agreed. I'm not super keen on exposing functions that mutate their arguments (like `onStmt` does), but you previously (correctly) pointed out that they may be useful elsewhere.

I'll write Scaladoc

src/main/scala/firrtl/transforms/ReplaceTruncatingArithmetic.scala
jackkoenig: Great question. I just checked and the default link for something (like looking at [List](https://www.scala-lang.org/api/current/scala/collection/immutable/List.html)) is the class. I should move the Scaladoc from the object to the class. Also in general I think that the primary documentation (like with examples) should be on the class with Scaladoc on methods being supplementary to that.

src/main/scala/firrtl/transforms/ReplaceTruncatingArithmetic.scala
jackkoenig: Do you have any suggestions on better names for these? I'm wondering if the short names are okay by the virtue of the very specific enclosing object name `ReplaceTruncatingArithmetic`. In fact, I'd prefer to call them `apply` but then I'd have to use `mapExpr`, `mapStmt` etc. (although that might honestly be better).

azidar: That is an interesting idea, I do think using mapExpr would clearer to users.

jackkoenig:
I know you already approved, but I added some more ScalaDoc. I stuck with `onExpr` and `onStmt` instead of `apply` because you can't overload functions with multiple argument lists :(. I'm not sure what more specific name I should give them given the name of the object.

I know you already approved, but feel free to take another look at the documentation changes.

PR #803, opened 2018-05-04
Proposal for EdgeDataDiGraph with views to amortize graph construction

albert-magyar:


azidar:
This looks good, I think I understand what you were saying now. I think what I was saying is there may be some generic edge-labelled algorithms which are useful and take, as a parameter, functions consuming edge-label data. These algorithms/functions might complex/useful enough to bake into the `EdgeLabeledDiGraph` class.

chick:
Ping @albert-magyar @azidar Any thoughts on when people can start playing with this. Joe Matthews at CMU thinks it may be advantageous for his obfuscation work.

PR #804, opened 2018-05-05
Add RestructureMuxes transform for better code generation

jackkoenig:
This transforms nested mux expressions that have a common input. For example:

```
r <= mux(a, mux(b, s, r), r)
```
becomes
```
r <= mux(and(a, b), s, r)
```

This is especially important for register updates where the default is always the register itself. This reduces the line count of the RocketCore.fir in regress by 4.5%

### DIFF
--- a/src/main/scala/firrtl/transforms/RestructureMuxes.scala
+++ b/src/main/scala/firrtl/transforms/RestructureMuxes.scala
@@ -1,0 +1,91 @@
+// See LICENSE for license details.
+
+package firrtl
+package transforms
+
+import firrtl.ir._
+import firrtl.Mappers._
+import firrtl.WrappedExpression._
+import firrtl.Utils._
+
+import scala.collection.mutable
+
+object RestructureMuxes {
+
+  private def AND(e1: Expression, e2: Expression): DoPrim = DoPrim(PrimOps.And, Seq(e1, e2), Nil, BoolType)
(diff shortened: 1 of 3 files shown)


jackkoenig:
Glancing at the Verilog, I need to check that this doesn't hurt synthesis results

aswaterman:
That’s what I was about to say. I think this is a nice improvement if it doesn’t hurt QoR. (I’m optimistic it won’t.)

jackkoenig:
It looks like this does **not** hurt synthesis results. Users should keep an eye though when bumping.

jackkoenig:
This can cause stack overflows in certain Vec dynamic access situations, we should hold off on merging until this is fixed (and synthesis results are checked again regarding this)

src/main/scala/firrtl/transforms/RestructureMuxes.scala
```diff
@@ -0,0 +1,91 @@
...
+
+  private def AND(e1: Expression, e2: Expression): DoPrim = DoPrim(PrimOps.And, Seq(e1, e2), Nil, BoolType)
+  private def NOT(e1: Expression): DoPrim = DoPrim(PrimOps.Not, Seq(e1), Nil, BoolType)
+
+  /** Mapping from references to the [[Expression]]s that drive them */
+  type Netlist = mutable.HashMap[WrappedExpression, Expression]
+
+  /** Restructures nested mux expressions where the two muxes have a common input
+    *
+    * This function is recursive
```
albert-magyar, line 23: This function is not recursive

PR #805, opened 2018-05-07
Use attach to connect analogs when grouping

colinschmidt:
This fixes a bug I ran into when your grouped module contains analog ports.


colinschmidt:
Sorry was swamped when I submitted this. I have some more PRs/FRs to make for grouping so I'll be adding tests for those.

jackkoenig:
Yeah I realized that and thats why I went ahead and merged :)

I look forward to the future PRs!

PR #808, opened 2018-05-11
Don't use bash to determine command availability - fixes #807

### LINKED ISSUE #807: Running Chisel stack on Windows tries to use bash to test for programs existing
# This is a bug report

* To reproduce try to run Chisel on windows through SBT or IntelliJ
* Firrtl uses bash to test for existence of programs like VCS and verilator but bash is not typically present on windows systems
* This is important because many new chisel users use windows based systems
* Change firrtl to test for programs directly and to fail gracefully when the program is not present
* Users should be able to use other simulators (such as the interpreter) even if other programs are not present
* Found during training developers to use IntelliJ

ucbjrl:


jackkoenig:
@ucbjrl Do you think this is reasonable for 1.1.1?

ucbjrl:
I do.

PR #809, opened 2018-05-12
Combinational Dependency Annotation

azidar:
Simple annotation that reports which top inputs are combinationally connected to top outputs.


src/main/scala/firrtl/transforms/CheckCombLoops.scala
```diff
@@ -207,8 +223,14 @@ class CheckCombLoops extends Transform {
         errors.append(new CombLoopException(m.info, m.name, expandedCycle))
       }
     }
+    val MN = ModuleName(c.main, CircuitName(c.main))
```
ducky64: interesting variable name style

src/main/scala/firrtl/transforms/CheckCombLoops.scala
```diff
@@ -207,8 +223,14 @@ class CheckCombLoops extends Transform {
...
+    val annos = simplifiedModuleGraphs(c.main).getEdgeMap.toSeq.filter( _._2.nonEmpty ).map { edge =>
```
jackkoenig: I think this would be a little more clear (and remove a 2 unnecessary traversals) as:
```scala
val annos = simplifiedModuleGraphs(c.main).getEdgeMap.collect { case (from, tos) if tos.nonEmpty => ...
```

src/main/scala/firrtl/transforms/CheckCombLoops.scala
jackkoenig: I would change this to `mn`

src/main/scala/firrtl/transforms/CheckCombLoops.scala
```diff
@@ -24,6 +24,22 @@ object CheckCombLoops {
 
 case object DontCheckCombLoopsAnnotation extends NoTargetAnnotation
 
+case class CombinationalPath(sink: ComponentName, sources: Seq[ComponentName]) extends Annotation {
+  override def update(renames: RenameMap): Seq[Annotation] = {
+    val newSources = sources.flatMap { s =>
+      renames.get(s) match {
+        case None => Seq(s)
+        case Some(seq) => seq.collect{case c: ComponentName => c}
+      }
```
jackkoenig: Is the collect necessary? `RenameMap.get` with a `ComponentName` returns an `Optino[Seq[ComponentName]]`: https://github.com/freechipsproject/firrtl/blob/396ee7ca63eb8a9e201dcdea965cbfc3e9d36783/src/main/scala/firrtl/Compiler.scala#L56

jackkoenig: If that's the case, then you could just do `renames.getOrElse(s, Seq(s))`

PR #811, opened 2018-05-15
Deprecate SingleStringAnnotation

jackkoenig:
It does not provide anything over NoTargetAnnotation. Its existence
suggests some significance so removing it for clarity.


jackkoenig:
I do see what you mean about `TargetDirAnnotation` extending `NoTargetAnnotation`. A bit of an unfortunately name choice.

PR #814, opened 2018-05-22
Add Circuit as option to FirrtlOptions

jackkoenig:
This allows providing the circuit as an option FIRRTL rather than having to parse. This will enable direct Chisel IR -> CHIRRTL in memory to bypass the ANTLR parser which has terrible memory use on large designs.

This does have a minor API change in that it firrtlSource would just silently override inputFileNameOverride and here I made that an error, so I'm not sure if it can be in 1.1.1. I'm also wondering if functions like `getCircuit` and `getAnnotations` should be defined on the FirrtlOptions? I'm not sure how this relates to https://github.com/freechipsproject/firrtl/pull/765, thoughts @seldridge?

### DIFF
--- a/src/main/scala/firrtl/Driver.scala
+++ b/src/main/scala/firrtl/Driver.scala
@@ -158,14 +158,14 @@ object Driver {
   def getCircuit(optionsManager: ExecutionOptionsManager with HasFirrtlOptions): Try[ir.Circuit] = {
     val firrtlConfig = optionsManager.firrtlOptions
     Try {
-      // Check that only "override" is used
+      // Check that only one "override" is used
       val circuitSources = Map(
         "firrtlSource" -> firrtlConfig.firrtlSource.isDefined,
         "firrtlCircuit" -> firrtlConfig.firrtlCircuit.isDefined,
         "inputFileNameOverride" -> firrtlConfig.inputFileNameOverride.nonEmpty)
       if (circuitSources.values.count(x => x) > 1) {
         val msg = circuitSources.collect { case (s, true) => s }.mkString(" and ") +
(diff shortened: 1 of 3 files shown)


seldridge:
This may have consequences for #765 and #764 (and the direction of those is still open to debate...). Those disallow direct creation of the `FirrtlExecutionOptions` case class. That would be defined only inside the `Driver` of a utility. The only representation of options would be an `Array[String]`, an `AnnotationSeq`, or both (where `String` args will be added to the existing `AnnotationSeq`).

Obviously `Array[String]` cannot handle a `Circuit` argument, however, can the circuit be efficiently encoded in an annotation and avoid parsing? I.e., is the following tractable?
```scala
case class FirrtlCircuitAnnotation(circuit: Circuit) extends NoTargetAnnotation
```

If it can (and associated JSON serialization/deserialization doesn't need parsing), then the thrust of this should be fine.

jackkoenig:
Yeah I see. I don't think converting it to use such an Annotation is a problem at all. Is the plan on #765 still to convert to TypeClasses and views?

seldridge:
Yes, that's still the working plan. I've held off on updating #765 to use that until there was some consensus. Developer meeting consensus so far (Chick/Jim) has been positive on the `TypeClass`/view pattern.

No rush there. The only useful information to consider is how that approach aligns with or violates intended uses of `FirrtlOptions`/`ChiselOptions`/`TreadleOptions`/etc., like what you propose here.

Yeah, doesn't look like there's any problem here. (So, some additional evidence that the #764 pattern at least encompasses some new intended usage cases.) :+1:

PR #815, opened 2018-05-23
Add utilities for UInt and SInt literals

jackkoenig:
Also minor cleanup to literal construction in Visitor

Minor but useful stuff.

I'm also thinking about making width of Literal Expressions be IntWidth, does it make sense for them ever to be Unknown? It fails in InferWidths if they are Unknown (as you can see in this diff, the Visitor was filling in the widths if they aren't specified)


src/main/scala/firrtl/ir/IR.scala
```diff
@@ -159,13 +159,28 @@ case class UIntLiteral(value: BigInt, width: Width) extends Literal {
...
   def mapType(f: Type => Type): Expression = this
   def mapWidth(f: Width => Width): Expression = SIntLiteral(value, f(width))
 }
+object SIntLiteral {
+  // To calculate bitwidth of negative number,
+  //  1) negate number and subtract one to get the maximum positive value.
+  //  2) get bitwidth of max positive number
+  //  3) add one to account for the signed representation
+  def minWidth(value: BigInt): Width = {
+    val w = if (value < 0) (value.abs - BigInt(1)).bitLength + 1 else value.bitLength + 1
```
aswaterman: This is basically a reimplementation of `bitLength` -- it is equivalent to

    val w = 1 + value.bitLength

jackkoenig: Haha yeah I just copied over very old code from the Visitor, whoops

src/main/scala/firrtl/passes/RemoveAccesses.scala
```diff
@@ -54,7 +54,7 @@ object RemoveAccesses extends Pass {
       ls.zipWithIndex map {case (l, i) =>
         val c = (i / stride) % wrap
         val basex = l.base
-        val guardx = AND(l.guard,EQV(uint(c),e.index))
+        val guardx = AND(l.guard,EQV(UIntLiteral(c),e.index))
```
chick: This is fine but I feel compelled to voice my distaste for the lower case L as a variable name

src/main/scala/firrtl/Visitor.scala
```diff
@@ -284,42 +284,31 @@ class Visitor(infoMode: InfoMode) extends FIRRTLBaseVisitor[FirrtlNode] {
...
-          // This could be better
-          val (width, value) =
-            if (ctx.getChildCount > 4)
-              (IntWidth(string2BigInt(ctx.intLit(0).getText)), string2BigInt(ctx.intLit(1).getText))
-            else {
-              val bigint = string2BigInt(ctx.intLit(0).getText)
-              (IntWidth(BigInt(scala.math.max(bigint.bitLength, 1))), bigint)
-            }
-          UIntLiteral(value, width)
+          if (ctx.getChildCount > 4) {
```
chick: I'm OCD today. 4 seems a bit of a magic number, I guess it means here that there are enough fields in the ctx to contain a width. Not worth refactoring.

src/main/scala/firrtl/passes/RemoveAccesses.scala
jackkoenig: I totally agree, obviously this PR was not trying to do general style improvements, but I think some effort in that direction would be a good idea. Maybe I'll work on getting IntelliJ up and running

src/main/scala/firrtl/Visitor.scala
jackkoenig: Hahaha yeah, basically "does it have a width". Hopefully this old code will be bypassed soon enough.

PR #818, opened 2018-06-06
Attach through wire

jackkoenig:
This change enables the pattern of attaching a module port to an instance port via a wire, eg.
```scala
attach(io.analog, analogWire)
attach(analogWire, io.analog)
```
(You can also use bulk connects)

While this "works" before this PR, it emits Verilog that does not work in Verilator. By constant propagating the wire away, we can emit Verilog like that shown in the test.

### DIFF
--- a/src/main/scala/firrtl/transforms/ConstantPropagation.scala
+++ b/src/main/scala/firrtl/transforms/ConstantPropagation.scala
@@ -397,7 +397,7 @@ class ConstantPropagation extends Transform {
         case Connect(info, lhs, rref @ WRef(rname, _, NodeKind, _)) if !dontTouches.contains(rname) =>
           Connect(info, lhs, nodeMap(rname))
         // If an Attach has at least 1 port, any wires are redundant and can be removed
-        case Attach(info, exprs) if exprs.map(kind(_)).contains(PortKind) =>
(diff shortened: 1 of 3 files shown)


src/main/scala/firrtl/transforms/ConstantPropagation.scala
```diff
@@ -396,6 +396,9 @@ class ConstantPropagation extends Transform {
         // Propagate connections to references
         case Connect(info, lhs, rref @ WRef(rname, _, NodeKind, _)) if !dontTouches.contains(rname) =>
           Connect(info, lhs, nodeMap(rname))
+        // If an Attach has at least 1 port, any wires are redundant and can be removed
+        case Attach(info, exprs) if exprs.map(kind(_)).contains(PortKind) =>
```
chick: Seems like here either:
* use exists in the if to do less work
* just go ahead and do the filterNot, seems cheaper than creating a new list with map then seeing if it contains something.

jackkoenig: Yeah I agree with the exists approach. It prevents creation of an unnecessary collection and will early out instead of doing at least 1 full traversal. Will push an update

I don't think the latter suggestion will work since the `filterNot` is on WireKind whereas we're only doing that filter if we find a `PortKind`

PR #820, opened 2018-06-09
Allow escaped single quotes in RawParams

ducky64:
If, for instance, your FPGA tools (COUGH iCEcube2 COUGH) happens to be inexplicably sensitive to the particular formatting of parameters, so you couldn't just plop down an IntParam and call it a day, but needed `6'b0110_01`. Or if you just wanted to pass through binary parameters from Chisel through FIRRTL without needing to convert them to hex or decimal, since Scala doesn't support binary literals.

The escaping framework isn't completely general (and I don't know if this is the right way), though it seems to be no worse than how `\"` is handled. (I know, really low bar, suggestions on how to improve are welcome though)

FIRRTL part of the fix for freechipsproject/chisel3#830

### DIFF
--- a/src/main/scala/firrtl/Visitor.scala
+++ b/src/main/scala/firrtl/Visitor.scala
@@ -92,7 +92,7 @@ class Visitor(infoMode: InfoMode) extends FIRRTLBaseVisitor[FirrtlNode] {
       case (int, null, null, null) => IntParam(name, string2BigInt(int.getText))
       case (null, str, null, null) => StringParam(name, visitStringLit(str))
       case (null, null, dbl, null) => DoubleParam(name, dbl.getText.toDouble)
-      case (null, null, null, raw) => RawStringParam(name, raw.getText.tail.init) // Remove "\'"s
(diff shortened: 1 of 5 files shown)


src/main/scala/firrtl/Visitor.scala
```diff
@@ -92,7 +92,7 @@ class Visitor(infoMode: InfoMode) extends FIRRTLBaseVisitor[FirrtlNode] {
       case (int, null, null, null) => IntParam(name, string2BigInt(int.getText))
       case (null, str, null, null) => StringParam(name, visitStringLit(str))
       case (null, null, dbl, null) => DoubleParam(name, dbl.getText.toDouble)
-      case (null, null, null, raw) => RawStringParam(name, raw.getText.tail.init) // Remove "\'"s
+      case (null, null, null, raw) => RawStringParam(name, raw.getText.tail.init.replace("\\'", "'")) // Remove "\'"s
```
jackkoenig: Need to re-escape in `RawStringParam.serialize` when emitting to FIRRTL

ducky64:
Added re-escape in serialize. Derrrrrp. Fixed, with tests.

Issue #749, opened 2018-03-02
In low Firrtl second clock referenced before declared.


chick:
This repository's issues are reserved for feature requests and bug reports.

* **Bug Report**
The chisel3 MultiClockSpec ClockDividerTest is converted to low Firrtl, the firrtl is re-arranged so that clock2 is referenced before it is declared.

* **How to reproduce the problem:**

Here is the Chirrtl.
```
;buildInfoPackage: chisel3, version: 3.1-SNAPSHOT, scalaVersion: 2.11.12, sbtVersion: 1.0.4
circuit ClockDividerTest : 
  module ClockDividerTest : 
    input clock : Clock
    input reset : UInt<1>
    output io : {}
    
    reg cDiv : UInt<1>, clock with : (reset => (reset, UInt<1>("h01"))) @[MultiClockSpec.scala 12:21]
    node _T_8 = eq(cDiv, UInt<1>("h00")) @[MultiClockSpec.scala 13:11]
    cDiv <= _T_8 @[MultiClockSpec.scala 13:8]
    node clock2 = asClock(cDiv) @[MultiClockSpec.scala 14:21]
    reg reg1 : UInt<8>, clock with : (reset => (reset, UInt<8>("h00"))) @[MultiClockSpec.scala 16:21]
    node _T_12 = add(reg1, UInt<1>("h01")) @[MultiClockSpec.scala 17:16]
    node _T_13 = tail(_T_12, 1) @[MultiClockSpec.scala 17:16]
    reg1 <= _T_13 @[MultiClockSpec.scala 17:8]
    reg reg2 : UInt<8>, clock2 with : (reset => (reset, UInt<8>("h00"))) @[MultiClockSpec.scala 18:41]
    node _T_17 = add(reg2, UInt<1>("h01")) @[MultiClockSpec.scala 19:16]
    node _T_18 = tail(_T_17, 1) @[MultiClockSpec.scala 19:16]
    reg2 <= _T_18 @[MultiClockSpec.scala 19:8]
    node _T_20 = lt(reg1, UInt<4>("h0a")) @[MultiClockSpec.scala 21:14]
    when _T_20 : @[MultiClockSpec.scala 21:22]
      node _T_22 = div(reg1, UInt<2>("h02")) @[MultiClockSpec.scala 22:26]
      node _T_23 = eq(reg2, _T_22) @[MultiClockSpec.scala 22:17]
[...]

PR #823, opened 2018-06-13
Resolve register clock dependencies in RemoveWires

seldridge:
Candidate fix for #749.

This modifies `getOrderedNodes` to incorporate register clock dependencies. There should still be a problem with reset dependencies (which I can clean up), but I think this is enough to unblock @chick and @ducky64.

@jackkoenig: thoughts?

### DIFF
--- a/src/main/scala/firrtl/transforms/RemoveWires.scala
+++ b/src/main/scala/firrtl/transforms/RemoveWires.scala
@@ -41,7 +41,7 @@ class RemoveWires extends Transform {
   // Transform netlist into DefNodes
   private def getOrderedNodes(
     netlist: mutable.LinkedHashMap[WrappedExpression, (Expression, Info)],
-    regInfo: Map[WrappedExpression, DefRegister]): Try[Seq[Statement]] = {
+    regInfo: mutable.Map[WrappedExpression, DefRegister]): Try[Seq[Statement]] = {
     val digraph = new MutableDiGraph[WrappedExpression]
     for ((sink, (expr, _)) <- netlist) {
       digraph.addVertex(sink)
@@ -116,7 +116,7 @@ class RemoveWires extends Transform {
     m match {
       case mod @ Module(info, name, ports, body) =>
         onStmt(body)
-        getOrderedNodes(netlist, regInfo.toMap) match {
+        getOrderedNodes(netlist, regInfo) match {
           case Success(logic) =>
             Module(info, name, ports, Block(decls ++ logic ++ otherStmts))
           // If we hit a CyclicException, just abort removing wires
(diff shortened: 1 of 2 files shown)


src/main/scala/firrtl/transforms/RemoveWires.scala
```diff
@@ -40,7 +40,8 @@ class RemoveWires extends Transform {
 
   // Transform netlist into DefNodes
   private def getOrderedNodes(
-      netlist: mutable.LinkedHashMap[WrappedExpression, (Expression, Info)]): Try[Seq[DefNode]] = {
+    netlist: mutable.LinkedHashMap[WrappedExpression, (Expression, Info)],
+    regInfo: Map[WrappedExpression, DefRegister]): Try[Seq[Statement]] = {
```
jackkoenig: I think just make it a `mutable.Map` (or `collection.Map`) and don't bother with calling `.toMap` below

ducky64:
The multiclock designs compile fine through FIRRTL now, though the tests fail so simulation doesn't seem to do the right thing. Though this could be a treadle issue (@chick), or an issue with the test. It appears that when there is a register update operation in the divided clock domain, the register output is always a zero.

chick:
When I tried to build this locally I got
```
[error] /Volumes/UCB-BAR/seldridge-firrtl/src/main/scala/firrtl/Compiler.scala:34:67: type mismatch;
[error]  found   : firrtl.annotations.CircuitName
[error]  required: firrtl.Named
[error]   def get(key: CircuitName): Option[CircuitName] = underlying.get(key).map {
[error]                                                                   ^
[error] /Volumes/UCB-BAR/seldridge-firrtl/src/main/scala/firrtl/Compiler.scala:43:18: value circuit is not a member of firrtl.ModuleName
[error]       this.get(m.circuit).map(cname => Seq(ModuleName(m.name, cname)))
[error]                  ^
```
And many other errors.  Hmm, what am I missing here?

seldridge:
@chick: What are you running specifically? Just `sbt test` on a clone of my issue-749 branch?

ducky64:
Yes, using `sbt test` on Chisel3 on the testers2 branch with the clock crossing test.

chick:
@seldridge I was just trying test:compile in a clone of your firrtl branch in the issue-749 branch

seldridge:
Will take a look at this now, @ducky64, and see if I can dig anything up.

@chick, hmmm... that's odd. I'm trying this locally with the following and things compile.
```bash
git clone git@github.com:seldridge/firrtl -b issue-749 firrtl-issue-749
cd firrtl-issue-749/
sbt test:compile
```

I confused as to where it's getting `firrtl.Named` and `firrtl.ModuleName` as those shouldn't exist per the package structure; it should be `firrtl.annotations.{Named, ModuleName}`.

chick:
I have no idea what was going on. git said I was on a clean branch, but I blew it all away and started from scratch.  Proceeding to test my use case

chick:
This seems to have fixed the problem on the ClockCrossing example @ducky64 is working on.  Now there is another problem but the firrtl ordering problem seems fixed.

jackkoenig:
Great work, thanks @seldridge!

PR #825, opened 2018-06-14
Fix TopWiringTests use of /tmp.

ucbjrl:
Relying on /tmp as a place for test output will fail on multiuser systems and may fail if multiple instances of tests are running for the same user.


ucbjrl:
Hmmm. Travis claims
```bash
$ yosys -V
yosys: command not found
The command "yosys -V" exited with 127.
```

jackkoenig:
That's so bizarre because other tests executed `yosys -V` just fine. Look spurious... I'll keep rerunning it to see if it'll pass

Issue #509, opened 2017-03-22
InferReadWrite fails with MatchError if annotation circuit name is not the top-level


edwardcwang:
FIRRTL commit 13fc27f35e85026c002e644b61c32268bd258d78

With a circuit like

```
module x:
  <...>
  module foo:
    <...>
  module bar:
    <...>
```

Running something like `firrtl.Driver -i input.fir -o input.v -X verilog --infer-rw foo` causes a MatchError, since the circuit main is "x" whereas the annotation that the command line interface creates has a circuit name of "foo", so it causes a MatchError.

Workaround is to just change the command-line interface call or hack the pass to run regardless of the circuit name.

Proposed solution: create another "InferReadWriteAnnotationModule(moduleName)" to handle this case, and change the command-line and/or barstools to use this.

https://github.com/ucb-bar/firrtl/blob/master/src/main/scala/firrtl/passes/memlib/InferReadWrite.scala#L163

azidar:
Will be fixed with annotation refactor

jackkoenig:
Fixed by https://github.com/freechipsproject/firrtl/pull/721

PR #829, opened 2018-06-20
Remove unneeded [String] argument to --infer-rw

seldridge:
After the annotations refactor, `--infer-rw`/`-firw` no longer takes an argument. However, a vestigial `[String]` argument remained. This removes that.

### DIFF
--- a/src/main/scala/firrtl/ExecutionOptionsManager.scala
+++ b/src/main/scala/firrtl/ExecutionOptionsManager.scala
@@ -406,9 +406,8 @@ trait HasFirrtlOptions {
       """Inline one or more module (comma separated, no spaces) module looks like "MyModule" or "MyModule.myinstance"""
     }
 
-  parser.opt[String]("infer-rw")
+  parser.opt[Unit]("infer-rw")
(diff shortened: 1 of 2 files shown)


seldridge:
@edwardcwang: interestingly, I believe that #509 was definitively fixed by the merged annotations refacor as Adam mentions. The vestigial `[String]` argument (whatever it was) would get treated as a program argument (which is still wrong, but harmless).

seldridge:
@edwardcwang: That would likely depend on whether or not this is a performance bottleneck. I'm in a bad position to empirically answer that, but you or @jackkoenig may have better information from some of your larger designs...

jackkoenig:
I think that could be a useful feature but I personally haven't seen a need for it. Generally people just want read-write ports (and a more important thing to do is make specifying them in Chisel actually be preserved to FIRRTL which I don't believe it is).

Anyway this is great, since it does change the command-line API I'm tagging it 1.2

seldridge:
1.2.0 makes sense as this would break Rocket without a (trivial) PR.

PR #830, opened 2018-06-20
Use "_" for Verilog keyword disambiguation (not $)

### LINKED ISSUE #729: Firrtl generates verilog with odd variable names (trailing '$')
* **Type of issue**
  - [x] Bug report
  - [ ] Feature request
  - [ ] Other enhancement

Firrtl occasionally emits verilog signals with a trailing '$'.  I'm not sure if this is properly a bug, but it's unclear why this occurs, and some commercial verilog parsers don't appreciate the syntax.

I was able to reproduce this on current rocket-chip (8799508).  After building DefaultConfig, you can see examples at line 174566 (`wire [63:0] logic$;` in `AMOALU`) or line 168431 (`wire  default$;` in `PMPChecker`).  I've also observed `reg reg$`, although this doesn't seem to appear in DefaultConfig.

seldridge:
To avoid issues with buggy parsers that get confused with the legal use of `$` in emitted Verilog names, this switches to the safer `_`. This is the candidate fix proposed by @aswaterman in #729.

Fixes #729


jackkoenig:
I support this although I would like to hear @azidar's input.

It is probably beyond the scope of this PR but I would like to see Inline instances follow the Uniquify pattern such that if **any** of the names it's introducing would collide, it adds suffixes *to the prefix* rather than just that individual signal. That being said it's just my preference to keep structure when disambiguating collisions, others may disagree and rather that a single name change leading to a collision only change the name of the colliding signal rather than all of the signals from an inlined module getting changed 🤷‍♀️

jackkoenig:
This change needs to make sure that InlineInstances doesn't introduce collisions, `$` is just assumed to not collide in the current implementation. Integration with Uniquify could handle this...

azidar:
I'd like to see a new test case to show the collision, something like the following, where inlining `foo` would conflict with `wire foo_a`.

```
module Foo:
  wire a: UInt

module Bar:
  wire foo_a: UInt
  inst foo of Foo
```

@jackkoenig how would integration with Uniquify help this? I liked the solution proposed by @seldridge in #729 with an optional pass that uses Uniquify.

jackkoenig:
Uniquify checks if lowering the names of leaf nodes in aggregates would collide with anything, you could treat instances that are being inlined as if they are aggregates being lowered and thus use a consistent "uniquification" algorithm

aswaterman:
Now that I've started to make use of this pass, I also have to support `_` over `$` on aesthetic/text-editor-related grounds

edwardcwang:
One idea briefly discussed at the dev meeting last Friday was to make the choice of keyword disambiguation character an option/function/parameter with a default (of presumably `_` or `$`).

aswaterman:
I'm having trouble seeing a compelling reason to expose the choice of separator.  Supporting that seems to encourage going down the road of using dedicated separators for each flattening purpose, which doesn't compose.  Better, I think, to use the same flattening character everywhere, and ferret out whatever bugs that exposes.

seldridge:
Finally revisiting this. One approach would be to make the delimiter an argument private to firrtl or passes. That would avoid having to rewrite ClockList for this PR (which I think, but have not verified, is broken for designs with signals that use `$`). Basically two PRs, one to swing `$` to `_` and a second to make clocklist not use inlined names as encoding hierarchy.

seldridge:
Sounds good @aswaterman. Working through this exposed some fundamental, though straightforward to fix, problems. Inlining is working (though failing Chisel tests, likely due to explicit checks on output circuits---I haven't looked into this, but I will). I'm working my way through the Verilog renamer.

@jackkoenig: Inlining now follows the Uniquify pattern and I'm intending to use this for backend keyword name mangling. This produces very readable results. Relatedly, the Uniquify pattern is actually insanely useful, much more useful than `Namespace` as that really is only suitable for name generation of temporaries. There's likely some unification of these two or a better API to wrap the Uniquify machinery in the future.

FWIW, this may be broken up into two PRs. One for inlining and one for backend keyword mangling.

PR #831, opened 2018-06-21
Provide a ProcessLogger() to capture all output in isCommandAvailable().

ucbjrl:


jackkoenig:
What does this do?

ucbjrl:
`isCommandAvailable()` tries to run a command (i.e., "vcs" or "verilator") to see if that command is available. The previous version assumed that passing `None` as the 3rd argument to `BasicIO()` was sufficient to suppress `stderr`. That proved to be incorrect and passing `None` causes `stderr` to be output normally.

This is particularly obnoxious since running `verilator` without any arguments causes it to display its help message on `stderr` and return a non-zero exit code. The fix for that particular problem is to run `verilator --version`, but in any case, we shouldn't be inadvertently displaying `stderr`.

The fix is to provide a `ProcessLogger()` that captures (and ignores) both `stderr` and `stdout`.

I put together a test for this, but it requires using `System.setOut()` which doesn't play well in a multi-threaded environment.

jackkoenig:
I see, well, looks good!

PR #832, opened 2018-06-22
Protobuf

jackkoenig:
At long last, here is actual protobuf serialization and deserialization.

It's a lot of code because translating to and from the ProtoBuf generated classes is just repetitive and verbose.

Some decisions I made and the justification
* No support for WIR
    * I see it as a follow on, this is enough to get Chisel -> FIRRTL via protobuf working which is my main goal
* Using Java Protobuf instead of ScalaPB
    * ScalaPB provides nice Scala-y datastructures from the protobuf. It doesn't provide much if anything over the default Java implementation because we're just converting to and from it anyway (note this code would be just as verbose). The Java implementation is supported by the protobuf team so I didn't see sufficient reason to use ScalaPB

I still need to add support for protobuf input and output but I figured I should PR this now since it is ready for overall code review and this is enough for some performance benchmarking (which I haven't done yet).

### DIFF
--- a/src/main/proto/firrtl.proto
+++ b/src/main/proto/firrtl.proto
@@ -34,6 +34,7 @@ message Firrtl {
     oneof source_info {
       None none = 1;
       Position position = 2;
+      string text = 3;
     }
   }
 
--- a/src/main/proto/firrtl.proto
+++ b/src/main/proto/firrtl.proto
@@ -38,6 +38,11 @@ message Firrtl {
     }
   }
 
+  message BigInt {
+    // 2's complement binary representation
+    bytes value = 1;
+  }
+
   message Top {
     // Required.
     string name = 1;
@@ -50,9 +55,20 @@ message Firrtl {
 
   message Module {
     message ExternalModule {
+      message Parameter {
+        string id = 1;
+        oneof value {
+          BigInt integer = 2;
+          double double = 3;
+          string string = 4;
+          string raw_string = 5;
+        }
+      }
       // Required.
       string id = 1;
       repeated Port port = 2;
+      string defined_name = 3;
+      repeated Parameter parameter = 4;
     }
 
     message UserModule {
@@ -199,6 +215,10 @@ message Firrtl {
       Expression expression = 5;
     }
 
+    message Attach {
+      repeated Expression expression = 1;
+    }
+
     // Required.
     oneof statement {
       Wire wire = 1;
@@ -215,6 +235,7 @@ message Firrtl {
       PartialConnect partial_connect = 16;
       IsInvalid is_invalid = 17;
       MemoryPort memory_port = 18;
+      Attach attach = 20;
     }
 
     SourceInfo source_info = 19;
@@ -260,6 +281,15 @@ message Firrtl {
       uint32 size = 2;
     }
 
+    message FixedType {
+      Width width = 1;
+      Width point = 2;
+    }
+
+    message AnalogType {
+      Width width = 3;
+    }
+
     // Required.
     oneof type {
       UIntType uint_type = 2;
@@ -267,6 +297,8 @@ message Firrtl {
       ClockType clock_type = 4;
       BundleType bundle_type = 5;
       VectorType vector_type = 6;
+      FixedType fixed_type = 7;
+      AnalogType analog_type = 8;
     }
   }
 
@@ -308,6 +340,12 @@ message Firrtl {
       // Required.
       IntegerLiteral value = 1;
       Width width = 2;
+    }
+
+    message FixedLiteral {
+      BigInt value = 1;
+      Width width = 2;
+      Width point = 3;
     }
 
     message ValidIf {
@@ -380,6 +418,13 @@ message Firrtl {
         OP_AS_UINT = 28;
         OP_AS_SINT = 29;
         OP_EXTRACT_BITS = 30;
+        OP_AS_CLOCK = 31;
+        OP_AS_FIXED_POINT = 32;
+        OP_AND_REDUCE = 33;
+        OP_OR_REDUCE = 34;
+        OP_SHIFT_BINARY_POINT_LEFT = 35;
+        OP_SHIFT_BINARY_POINT_RIGHT = 36;
+        OP_SET_BINARY_POINT = 37;
       }
 
       // Required.
@@ -388,11 +433,14 @@ message Firrtl {
       repeated IntegerLiteral const = 3;
     }
 
+    reserved 5;
+
     // Required.
     oneof expression {
       Reference reference = 1;
       UIntLiteral uint_literal = 2;
       SIntLiteral sint_literal = 3;
+      FixedLiteral fixed_literal = 11;
       ValidIf valid_if = 4;
       //ExtractBits extract_bits = 5;
       Mux mux = 6;
--- a/src/main/proto/firrtl.proto
+++ b/src/main/proto/firrtl.proto
@@ -100,6 +100,7 @@ message Firrtl {
       uint32 read_latency = 5;
       repeated string reader_id = 6;
       repeated string writer_id = 7;
+      repeated string readwriter_id = 8;
     }
 
     message CMemory {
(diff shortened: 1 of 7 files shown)


jackkoenig:
Preliminary results but for a 420MB `.fir` file (460MB in `.pb`)
```
Parser
~90 seconds
51.5 GB

Protobuf
~7 seconds
13GB
```
Those memory use numbers include the FIRRTL AST which is ~4.5GB

src/main/proto/firrtl.proto
```diff
@@ -34,9 +34,15 @@ message Firrtl {
     oneof source_info {
```
jackkoenig: @sdtwigg if you have a chance, I would appreciate your looking at this, especially as it relates to possible integration with your representation (although I believe you mentioned y'all are using proto2 so maybe that just won't work?)

jackkoenig:
@chick Something I was wondering specifically is what do you think about the change to the command-line API? It used to be if you gave an input file that didn't end in `.fir`, it would add `.fir` to the end. I changed it to no longer do that, using "fileNameOverride" exactly as given for the filename and then treating `.pb` files as protobuf and anything else as firrtl files (to remain similar to how it used to work).

This is similar to how tools like g++ work where it will treat a `.c` as C, `.cpp` as c++, but if it doesn't recognize the extension, it actually just errors (to be precise, if it doesn't recognize the extension, it passes it to the linker which will then error if it doesn't recognize the extension).

chick:
@jackkoenig This sounds fine to me. As things move forward some little convenience items like that can't always work the way they did before. I'd like to think no too many people rely on this and that those that have used this side effect have not built up too many scripts or Makefiles that rely on it.

jackkoenig:
I've also added an Apache 2.0 license for my contributions as an employee of SiFive. I'm following the pattern used in rocket-chip where the files I'm creating and modifying also get the notification.

We should probably clean up the current license as it's like the BSD 3-clause but phrased in an atypical order (and it also misspells license as "licence"), and rename it to LICENSE.Berkeley in the root directory as done in rocket-chip. However, I think that's a separate, self-contained PR.

seldridge:
@jackkoenig: possibly out of scope of this PR, but I'm commenting since you reference the license here...

Are _contributions_ licensed individually or are files jointly licensed once touched? Do other organizations have the ability to choose how their contributions are licensed (BSD vs. Apache v2)? I expect that most non-academic organizations would prefer Apache due to the patent clause. That then brings up the question of how contributions from external organizations are licensed by default.

I expect that this whole situation could be remedied by bulk migrating to Apache v2 or clarifying with a DCO, etc (e.g., all contributions must be signed and will be under the Apache v2 license).

jackkoenig:
@seldridge I think it is beyond the scope of this PR, but I am the one who brought it up 😬 

I'm going to remove all of that from this PR, but I think proper discussion of licensing and contributions is long overdue. I think it's relevant to the Free Chips Porject at large although there are probably small differences between how different projects are handled (for example, [GCC clarifies that you can compile non-GPL programs](https://gcc.gnu.org/onlinedocs/libstdc++/manual/license.html), since we are not copy-left we don't have quite that big of a problem, but we might want to clarify that you don't need to include the Chisel and FIRRTL copyright notices in the product of Chisel+FIRRTL compilation).

In any case, I think this is a discussion for the Chisel Governance meeting and not for a PR, my apologies for bringing it up.

seldridge:
All good, Jack! I'm in total agreement with your comments: this is long overdue, an issue to be brought up with Chisel governance, and can be a model for how the Free Chips Project (or whatever the final name is) will operate. Obviously, if there's a business need and/or some specific worry on the SiFive side about the need to get a patent clause in there then it makes sense to do it now. Otherwise, I assume it can wait for a broader discussion.

I think this only hit me oddly as Chisel/FIRRTL is much more decoupled from SiFive than Rocket Chip (i.e., I fully expect SiFive to run the licensing of Rocket Chip however they want and to protect themselves there especially since Rocket Chip is, at this point, a fully SiFive artifact).

Yes, the notion of this generating circuits should likely be clarified, as well. I have no idea how that works or if there's some motivation in investigating how the patent clause would extend to artifacts produced.

jackkoenig:
There's nothing specific, this was just a large contribution and got me thinking about licensing issues. As you said, I think most non-academics organizations (like SiFive) would be more comfortable with an Apache license. In any case this issue can wait until we can talk about this in a governance context and come to a consensus!

jackkoenig:
I think this is ready to go. @chick has looked at the bulk of it. The only big change since then is the is the addition of the `ToProto.writeToStream*` methods. I'm ready to merge this, and hope people will take a/another look.

`ToProto.writeToStreamFast` is perhaps the main thing that could be contentious so I should explain. The normal ProtoBuf serialization looks something like FIRRTL -> ProtoBuf -> serialized, the intermediate ProtoBuf data structure is about the same size as the FIRRTL representation. For small designs this is irrelevant, but for large designs it's a substantial increase to memory use. Because the intermediate structure is large, it survives several small GCs and ultimately gets promoted to the Tenured heap (which is expensive). I wrote `ToProto.writeToStreamFast` to improve serialization to do the FIRRTL -> ProtoBuf -> serialized on a Module at a time instead of the whole circuit. This allows the intermediate ProtoBuf objects to be garbage collected during serialization while in the Young generation rather than afterwards while [frequently] Tenured.

That doesn't explain the "deconstructed" Circuit function arguments. Fundamentally, moving from a custom written Emitter in Chisel to multi-step serialization is going to be slower. When I used just the naive way of translating the whole circuit one step at a time, the slowdown for Chisel serialization wasn't great. For Chisel, ProtoBuf serialization has an additional step: ChiselIR -> FIRRTL -> ProtoBuf -> serialized. By deconstructing the Circuit and accepting `() => DefModule`, the intermediate FIRRTL objects can be garbage collected in the same way as the ProtoBuf objects. While ProtoBuf serialization is still slower than the existing Chisel emitter, `ToProto.writeToStreamFast` still has a nice benefit in serialization times:

I'll let FastProto be Chisel serialization using `writeToStreamFast` and SlowProto be serialization using naive full Circuit -> FIRRTL -> ProtoBuf -> serialized

| Config | Chisel Elaboration | Chisel Emitter | SlowProto | FastProto |
| ---- | ---- | ----- | ---- | ----- |
| Medium | 10 s | 0.85 s | 1.9 s | 1.4 s |
| Large | 100 s | 5.6 s | 16 s | 7.2 s |
| Huge | 1060s | 40 s |  | 19s |

Small (<5%) performance hit for most Chisel designs, small improvement for huge designs, and a nice gain for FIRRTL:

These result are for full compilation (Parser was/is memory bottleneck):

| Config | w/ Parser Time | w/ Parser Memory | w/ Proto Time | w/ Proto Memory |
| ---- | ---- | ----- | ---- | ----- |
| Medium | 55 s | 800 MB | 50 s | 700 MB |
| Large | 280 s | 7 GB | 260 s | 4 GB |
| Huge | 422 s | 40 GB | 400 s | 15 GB |

chick:
I've done another once over and it still looks good to me. I few nits and comments
* IntelliJ did not seem to know how to get started with this branch, but it cleared itself up when ran sbt and did `clean` and `test:compile`
* FromProto style violation
  * Line 143 def convert does not declare it's return type
* ToProto style violations
  * Lines 135, 140, 145 Suspicious shadowing of variable padding
  * In same section as above, `tpe` is set in pattern match several times but not used
There's a few more nits that IntelliJ's checker complains about but I don't think are necessarily what we consider bad style (though it would be nice to get theirs matched up with ours)
Tests all passed.
Otherwise, as before, LGTM

jackkoenig:
Alright I've integrated @chick's/IntelliJ's suggestions (I even used IntelliJ to do it!)

As @chick noted, this adds an additional code generation task and unfortunately IntelliJ seems to struggle with them. You either need to run compile on the command-line before trying to build with IntelliJ or you can run the code gen tasks manually:
```
$ sbt
> protobufGenerate
> antlr4:antlr4Generate
```
Obnoxiously, `protobufGenerate` deletes the `src_managed` directory so it needs to be run first, otherwise it'll delete the ANTLR-generated files.

jackkoenig:
Barring objection I'm going to merge this once Travis passes

PR #833, opened 2018-06-25
Support for load memory annotations in chisel

chick:
This PR
 * Delays the BlackBoxSourceHelper transformation to the Emitter stage of the VerilogCompiler
   * remove from VerilogCompiler
   * move to VerilogEmitter
 * Changes the verilog emitter to allow programmatic access to the verilog module declaration
   * Creating a bindable module requires headers to match
 * Provides a unit test that shows how to generate a bindable module.


chick:
@jackkoenig the changes here enable what I think I need to complete implementation of [LoadMemory RFC #835](https://github.com/freechipsproject/chisel3/issues/835). This seemed the least invasive but I am open to other approaches.

azidar:
Interesting approach. I think I understand why you need `VerilogRenderer`, though its worth thinking about which functions and datastructures should be private. Maybe we can talk about this in person?

PR #310, opened 2016-09-21
Fix clock connections in InferReadWrite

donggyukim:
* SHOULD BE MERGED AFTER #306 AND CHANGE BASE TO MASTER

This fixes #307, and should be merged quickly due to ucb-bar/chisel3#288.

PR #834, opened 2018-06-26
Improve code generation for smem wmode and [w]mask ports

aswaterman:
A common case for these port-enables is

    wen = valid & write
    ren = valid & !write

which the RW-port transform currently turns into

    en = (valid & write) | (valid & !write)
    wmode = valid & write

because it proved `wen` and `ren` are mutually exclusive via `write`.

Synthesis tools can trivially optimize `en` to `valid`, so that's not a problem, but the wmode field can't be optimized if going into a black box. This PR instead sets `wmode` to whatever node was used to prove mutual exclusion, which is always a simpler expression.  In this case:

    en = (valid & write) | (valid & !write)
    wmode = write

The RemoveCHIRRTL changes in this PR effect an analogous improvement for the mask/wmask ports.


aswaterman:
It looks like this defeats the hack in ResolveMaskGranularity that removes write masks if they're equal to wmode (since wmode has now gotten simpler).  Fixing the suboptimal code generation for write masks was next on my list, anyway, so I'll try to do it as part of this PR.

aswaterman:
For my test cases, @jackkoenig's improvement to RemoveCHIRRTL addresses both the ResolveMaskGranularity issue I just mentioned, along with the more general write-mask code generation deficiency I alluded to.

jackkoenig:
This is failing LEC. I would think with behavioral memories this would be logically equivalent... will investigate today.

jackkoenig:
Running formality locally to see if there's a real issue I get:
```
********************************* Verification Results *********************************
Verification SUCCEEDED
----------------------
 Reference design: ref:/WORK/ICache
 Implementation design: imp:/WORK/ICache
 137394 Passing compare points
----------------------------------------------------------------------------------------
Matched Compare Points     BBPin    Loop   BBNet     Cut    Port     DFF     LAT   TOTAL
----------------------------------------------------------------------------------------
Passing (equivalent)           0       0       0       0     313  137081       0  137394
Failing (not equivalent)       0       0       0       0       0       0       0       0
Not Compared
  Unread                       0       0       0       0       0       6       0       6
****************************************************************************************
```
```
********************************* Verification Results *********************************
Verification SUCCEEDED
----------------------
 Reference design: ref:/WORK/FPU
 Implementation design: imp:/WORK/FPU
 4641 Passing compare points
----------------------------------------------------------------------------------------
Matched Compare Points     BBPin    Loop   BBNet     Cut    Port     DFF     LAT   TOTAL
----------------------------------------------------------------------------------------
Passing (equivalent)           0       0       0       0     236    4405       0    4641
Failing (not equivalent)       0       0       0       0       0       0       0       0
Not Compared
  Constant reg                                                        36       0      36
  Unread                       0       0       0       0       0     164       0     164
****************************************************************************************
```
```
********************************* Verification Results *********************************
Verification SUCCEEDED
----------------------
 Reference design: ref:/WORK/RocketCore
 Implementation design: imp:/WORK/RocketCore
 5911 Passing compare points
----------------------------------------------------------------------------------------
Matched Compare Points     BBPin    Loop   BBNet     Cut    Port     DFF     LAT   TOTAL
----------------------------------------------------------------------------------------
Passing (equivalent)           0       0       0       0    1476    4435       0    5911
Failing (not equivalent)       0       0       0       0       0       0       0       0
Not Compared
  Constant reg                                                        64       0      64
  Unread                       0       0       0       0       0     257       0     257
****************************************************************************************
```
I'm going to wave the formal checks since this they pass with Formality

src/main/scala/firrtl/passes/RemoveCHIRRTL.scala
```diff
@@ -82,12 +82,10 @@ object RemoveCHIRRTL extends Transform {
...
       )
-      def set_write(vec: Seq[MPort], data: String, mask: String) = vec flatMap {r =>
+      def set_write(vec: Seq[MPort], data: String, mask: String) = vec flatMap { r =>
         val tmask = createMask(sx.tpe)
-        IsInvalid(sx.info, SubField(SubField(Reference(sx.name, ut), r.name, ut), data, tdata)) +:
-             (create_exps(SubField(SubField(Reference(sx.name, ut), r.name, ut), mask, tmask))
-               map (Connect(sx.info, _, zero))
-             )
+        val portRef = SubField(Reference(sx.name, ut), r.name, ut)
+        Seq(IsInvalid(sx.info, SubField(portRef, data, tdata)), IsInvalid(sx.info, SubField(portRef, mask, tmask)))
```
azidar: To solidify my understanding, you are invalidating the mask instead of zeroing it because the write enable takes care of any when statements between the memory and write port declaration?

src/main/scala/firrtl/passes/memlib/ResolveMaskGranularity.scala
```diff
@@ -65,7 +65,8 @@ object AnalysisUtils {
       if (nodeWidth == extractionWidth) getOrigin(connects)(args.head) else e
     case DoPrim((PrimOps.AsUInt | PrimOps.AsSInt | PrimOps.AsClock), args, _, _) => 
       getOrigin(connects)(args.head)
-    case ValidIf(cond, value, ClockType) => getOrigin(connects)(value)
+    // It is a correct optimization to treat ValidIf as a connection
+    case ValidIf(cond, value, _) => getOrigin(connects)(value)
```
azidar: Why the change of requirement for ClockType?

jackkoenig: Good question, so I traced the origin of this line to https://github.com/freechipsproject/firrtl/pull/310 where it was added to ensure we can optimize past ValidIfs on clocks. In doing this PR, I realized we need to optimize past ValidIfs for enable signals as well which are Bools. It's just a minor generalization.

src/main/scala/firrtl/passes/RemoveCHIRRTL.scala
jackkoenig: Exactly, the write enable signal is inferred as the conjunction of when conditions between memory declaration and port declaration. In order to not duplicate that condition on the mask, this PR is inferring the mask as the conjunction of when conditions between declaration of the port and the actual connection to the subfield or subindex of the write port.

PR #835, opened 2018-06-27
Remove #0.002 delay to initial rand assignments

rmac-sifive:
This delay no longer appears to be required, and is introducing issues to simulation behavior

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -631,12 +631,6 @@ class VerilogEmitter extends SeqTransform with Emitter {
           emit(Seq("`ifdef RANDOMIZE"))
           emit(Seq("  integer initvar;"))
           emit(Seq("  initial begin"))
-          // This enables test benches to set the random values at time 0.001,
-          //  then start the simulation later
-          // Verilator does not support delay statements, so they are omitted.
-          emit(Seq("    `ifndef verilator"))
-          emit(Seq("      #0.002 begin end"))
-          emit(Seq("    `endif"))
           for (x <- initials) emit(Seq(tab, x))
           emit(Seq("  end"))
           emit(Seq("`endif // RANDOMIZE"))


jackkoenig:
I traced the delay to this commit from @yunsup from Chisel2 https://github.com/ucb-bar/chisel2-deprecated/commit/2b97ddf8d829e0b505e9c9689d0518b40b450ae7 (interesting historical tidbit, that commit is from the same day that I met Yunsup in person during visit weekend, perhaps +/- a day).

The comment was added by @azidar in https://github.com/freechipsproject/firrtl/pull/149 (rebased and merged in https://github.com/freechipsproject/firrtl/pull/164).

I don't know why the delay is there. It doesn't seem necessary to me but if @yunsup or @azidar know of compelling reasons to keep it, please advise.

aswaterman:
I can’t remember what race condition the delay was there to resolve, but it could well have been a workaround for a simulator issue that’s no longer an issue. I take it you visually confirmed that random initializarion is still happening correctly in VCS?

rmac-sifive:
@aswaterman, yes, I visually confirmed in DVE that removing the delay results in random initialization happening correctly. In fact, the motivation for this change is that I'm seeing the random initialization happen _incorrectly_ when the delay is present

aswaterman:
cool, no objections here, but I'll check with @yunsup to see if he recalls the original motivation

azidar:
Yeah, when I added Chisel 2's delay to FIRRTL, I do remember _a_ discussion, but I have no idea what it was about. If ya'll come to a consensus either way about it, its fine with me.

donggyukim:
Can anyone confirm it's ok with (the latest version of) VCS? My simulation of rocketchip didn't work without delays.

aswaterman:
Sorry, I've forgotten to revisit this issue.

The resolution is that these delays are still necessary, and indeed not sufficient -- the hardwired constant 0.002 might be less than the time precision, thus rounded down to zero, just as though no delay were used at all.  I fixed that limitation with https://github.com/freechipsproject/firrtl/commit/25a7ca977a3b38e5f4996ff0c7fa378582bcc12c

So, I think we should close this issue.

jackkoenig:
superseded by https://github.com/freechipsproject/firrtl/pull/872, although if we come up with a safer way to do this that doesn't run the risk of being rounded down to 0, we should do that

PR #841, opened 2018-07-05
Demonstrate using Eta to build a simple compiler transform

jwiegley:
This PR re-implements the `tutorial.lesson1.AnalyzeCircuit` compiler transform (under the name `tutoral.lesson3.AnalyzeCircuit`) using the [Eta](https://eta-lang.org/) implementation of the Haskell language to both interact with Scala, and to implement the "meat" of the analysis without recourse to the Scala library (for example, module summation is done using the Haskell data structures `Data.Set` and `Data.Map`).

Many thanks are due to @puffnfresh and @rahulmutt from the Eta development team, who helped with both advice and rapid fixes in `eta` and `sbt-eta` to get this working.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,7 @@ test_run_dir
 .cache
 .history
 .lib/
+dist/
 dist/*
 target/
 lib_managed/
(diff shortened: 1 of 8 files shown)


jwiegley:
Note that this was developed using the current `master` branch of `eta`, and `sbt-eta`. It will not work with the release versions.

albert-magyar:
Closing this just from non-activity; I think the repo and CI hurdles for contributors were a bit less stable when this was open, so feel free to open again if this is still active.

jwiegley:
It is no longer active, thank you!

PR #843, opened 2018-07-06
Make InstanceGraph have defined iteration order

jackkoenig:
DiGraph is deterministic but the construction of them via InstanceGraph is not. This PR makes it so.

I'm not a fan of explicitly returning mutable data structures, but as we all know, Scala kind of bites us on this by making the only immutable Map with defined iteration order very slow (ListMap).

`SeqMap` may exist in Scala 2.13, and may even have a reasonable implementation with `VectorMap`: https://github.com/scala/scala/pull/6854. But as long as we're supporting Scala 2.12 we'll be stuck with this weird issue. We could copy the current implementation of VectorMap/LinkedMap (it's BSD licensed) https://github.com/mdedetrich/linked-map and use it within the compiler rather than returning LinkedHashMaps. I don't know, what do y'all think?

### DIFF
--- a/src/main/scala/firrtl/analyses/InstanceGraph.scala
+++ b/src/main/scala/firrtl/analyses/InstanceGraph.scala
@@ -19,20 +19,19 @@ import firrtl.Mappers._
 class InstanceGraph(c: Circuit) {
 
   val moduleMap = c.modules.map({m => (m.name,m) }).toMap
-  private val instantiated = new mutable.HashSet[String]
+  private val instantiated = new mutable.LinkedHashSet[String]
   private val childInstances =
-    new mutable.HashMap[String,mutable.Set[WDefInstance]]
+    new mutable.LinkedHashMap[String, mutable.LinkedHashSet[WDefInstance]]
   for (m <- c.modules) {
-    childInstances(m.name) = new mutable.HashSet[WDefInstance]
+    childInstances(m.name) = new mutable.LinkedHashSet[WDefInstance]
     m map InstanceGraph.collectInstances(childInstances(m.name))
     instantiated ++= childInstances(m.name).map(i => i.module)
   }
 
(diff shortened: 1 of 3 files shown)


albert-magyar:
`grep -r "[^d]HashMap" ./src`

jackkoenig:
Terrifying. @albert-magyar What do you think about including an implementation of VectorMap and then we could use it all over in lieu of HashMap?

albert-magyar:
We could definitely do something. The tricky thing is that we have 4 cases:

1. Mutated, iteration order used to generate hardware
2. Mutated, iteration order not used to generate hardware
3. Constant, iteration order used to generate hardware
4. Constant, iteration order not used to generate hardware

And then people use things in slightly different ways and the categories bleed into each other :(

mdedetrich:
@jackkoenig Just chiming in here

The intention of https://github.com/mdedetrich/linked-map is its going to be a polyfill for SeqMap/Vector map for Scala 2.12 and lower, so that you still have complete source code compatibility across Scala versions minor one proviso (which is that I can't make `LinkMap` extend `SeqMap` since its an external library, maybe there is some hacky way to do this)

I haven't finalized worked on https://github.com/mdedetrich/linked-map because I am waiting for the `SeqMap` design to be finalised (I for example have to rename `LinkedMap` to `SeqMap`)

jackkoenig:
@mdedetrich Thank you for your consideration!

That all makes sense! Is your plan with https://github.com/mdedetrich/linked-map to publish it or is it something we should consider including it as a submodule? I would love to use it here as a polyfill!

Thank you for your work on this!

mdedetrich:
@jackkoenig I will publish it, but only when the design is finalized. Its supposed to be ultra stable, so I will only make a release when everything is good and the performance of the implementation is adequate

PR #846, opened 2018-07-10
InferWidths: improve performance

### LINKED ISSUE #842: InferWidths is pathologically slow on a 138 KB circuit
**Type of issue**: bug report (performance)

A 138 KB circuit is taking 64 minutes to run through FIRRTL to generate Verilog from Chirrtl.

Profiling indicates that InferWidths is taking most of the time, with `InferWidths.collectMinMax` taking most of the time.

Chisel3 takes <1 minute to elaborate to FIRRTL, and before this RTL was ported to Chisel3, Chisel2 was able to elaborate to Verilog in <1 minute, so this is a serious performance regression from a user's perspective.

Reproduced on firrtl master (revision ceac36d), locally and on a-machines.

[FIRRTL file](https://gist.github.com/edwardcwang/b13bd38d4b5aea5b879861a16c57649e) (138 KB)

---
`firrtl -i Core-test.fir -o Core-test-out -X verilog` - `Total FIRRTL Compile Time: 6766360.7 ms` (112.8 min)
`firrtl -i Core-test.fir -o Core-test-out -X verilog --no-check-comb-loops --no-dce` - `Total FIRRTL Compile Time: 880962.3 ms` (14.7 min)

edwardcwang:
On circuits with large numbers of width inferences, prepend to a linked list instead of appending and having to make a copy.

Thanks @azidar for the catch

Close #842

### DIFF
--- a/src/main/scala/firrtl/passes/InferWidths.scala
+++ b/src/main/scala/firrtl/passes/InferWidths.scala
@@ -41,12 +41,12 @@ object InferWidths extends Pass {
       case wx => wx
     }
     def collectMinMax(w: Width): Width = w map collectMinMax match {
-      case MinWidth(args) => MinWidth(unique((args.foldLeft(Seq[Width]())) {
-        case (res, wxx: MinWidth) => wxx.args ++ res
+      case MinWidth(args) => MinWidth(unique(args.foldLeft(List[Width]()) {
+        case (res, wxx: MinWidth) => wxx.args ++: res
         case (res, wxx) => wxx +: res
       }))
-      case MaxWidth(args) => MaxWidth(unique((args.foldLeft(Seq[Width]())) {
-        case (res, wxx: MaxWidth) => wxx.args ++ res
+      case MaxWidth(args) => MaxWidth(unique(args.foldLeft(List[Width]()) {
+        case (res, wxx: MaxWidth) => wxx.args ++: res
         case (res, wxx) => wxx +: res
       }))
       case wx => wx
--- a/src/main/scala/firrtl/passes/InferWidths.scala
+++ b/src/main/scala/firrtl/passes/InferWidths.scala
@@ -42,12 +42,12 @@ object InferWidths extends Pass {
     }
     def collectMinMax(w: Width): Width = w map collectMinMax match {
       case MinWidth(args) => MinWidth(unique((args.foldLeft(Seq[Width]())) {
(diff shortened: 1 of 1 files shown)


src/main/scala/firrtl/passes/InferWidths.scala
```diff
@@ -42,12 +42,12 @@ object InferWidths extends Pass {
     }
     def collectMinMax(w: Width): Width = w map collectMinMax match {
       case MinWidth(args) => MinWidth(unique((args.foldLeft(Seq[Width]())) {
```
jackkoenig: All over the place we tend to use `Seq(...)` and `Seq.empty` to create sequences when I think we should be a more explicit and call out what data structure we're instantiating. Since the appending was causing such massive problems, can you also change it to `List[Width]()` (which is what `Seq[Width]()` creates) just to be a little more obvious?

edwardcwang: Done. Interesting comparison of `List` vs `Vector`: https://docs.scala-lang.org/overviews/collections/performance-characteristics.html

src/main/scala/firrtl/passes/InferWidths.scala
```diff
@@ -41,13 +41,13 @@ object InferWidths extends Pass {
       case wx => wx
     }
     def collectMinMax(w: Width): Width = w map collectMinMax match {
-      case MinWidth(args) => MinWidth(unique((args.foldLeft(Seq[Width]())) {
-        case (res, wxx: MinWidth) => res ++ wxx.args
-        case (res, wxx) => res :+ wxx
+      case MinWidth(args) => MinWidth(unique((args.foldLeft(List[Width]())) {
+        case (res, wxx: MinWidth) => wxx.args.toList ++ res
```
jackkoenig: I'm actually not certain if this matters in practice (eg. compile *may* do the right thing automatically, but you could do `wxx.args ++: res` so that the append operator returns the type of the right collection rather than the left collection like `++` does. This potentially saves a traversal and construction of a List from the args of `wxx`

edwardcwang: Done.

src/main/scala/firrtl/passes/InferWidths.scala
jackkoenig: See also http://www.lihaoyi.com/post/BenchmarkingScalaCollections.html

PR #848, opened 2018-07-13
Make Scala 2.12.4 the default.

ucbjrl:


build.sbt
```diff
@@ -16,9 +16,9 @@ name := "firrtl"
 
 version := "1.2-SNAPSHOT"
 
-scalaVersion := "2.11.12"
+scalaVersion := "2.12.6"
 
-crossScalaVersions := Seq("2.11.12", "2.12.4")
+crossScalaVersions := Seq("2.12.6", "2.11.12")
```
jackkoenig: Does this order matter?

ucbjrl: It doesn't matter for `sbt`, but for `mill`, we use the first entry as the default, and since I'm bumping the 2.12 version and changing that line anyway ...

jackkoenig: Ah I see, thanks for the clarification!

ucbjrl:
Could someone with some Travis expertise take a look at the Travis failures?

jackkoenig:
I looked at it a little and have no idea what's going on...

This might require looking into reproducing the Travis build environment via it's Docker image to get more information.

jackkoenig:
I'm trying out debugging using the Travis docker image. I've also emailed Travis support to ask for them to turn on debug-mode for the FIRRTL repo which allows sshing into a failed test to poke around and see what's wrong. They're on European time so hopefully tomorrow we'll be able to debug better.

ucbjrl:
Thanks Jack. I had a thought as I was drifting off to sleep last night (possibly triggered by having to get protobuf support working under mill). `com.github.os72.protocjar` has to determine the current platform to execute the appropriate protobuf compiler. I wonder if the fact that it's running in a container is confusing it. I would expect some diagnostic to that effect and we should have seen this when you first started working on protobuf support, so it's probably not related. Just a thought.

More importantly, good luck.

ucbjrl:
Just make 2.12.4 the default. Leave the 2.12.6 migration to another time.

PR #849, opened 2018-07-17
Constant prop add

albertchen-sifive:
- added yosys miter testers
- added transforms to rename modules for yosys miter tests
- added MinimumVerilog Compiler for equivalence testing
- added FoldADD object to propagate constant addition/addition by zero
- rename FoldLogicalOp to FoldCommutativeOp


src/main/scala/firrtl/transforms/ConstantPropagation.scala
```diff
@@ -44,7 +44,19 @@ class ConstantPropagation extends Transform {
     }
   }
 
-  object FoldAND extends FoldLogicalOp {
+  object FoldADD extends FoldCommutativeOp {
+    def fold(c1: Literal, c2: Literal) = (c1, c2) match {
+      case (_: UIntLiteral, _: UIntLiteral) => UIntLiteral(c1.value + c2.value, (c1.width max c2.width) + IntWidth(1))
```
aswaterman: Is addition defined to be non-expanding when one of the arguments has width 0?  i.e. should the width be something like

    if (c1.width == 0) c2.width
    else if (c2.width == 0) c1.width
    else (c1.width + c2.width) + 1

jackkoenig: I'm not sure why `simplify` is padded but `fold` is not. @aswaterman wrote the code 2 years ago, do you happen to remember? https://github.com/freechipsproject/firrtl/commit/359021cd1753cd8ad6e5315da6ef2638c5000323

aswaterman: Just because it was easy to make all the `fold` cases return the correct width, but that seemed too onerous for all the `simplify` cases.

aswaterman: (I don't actually remember; that's just my best guess)

src/test/scala/firrtlTests/FirrtlSpec.scala
```diff
@@ -5,22 +5,129 @@ package firrtlTests
...
+      val newState = state.copy(circuit = state.circuit.copy(modules = modulesx, main = newTopName))
+      newState
+    }
+  }
+
+  /** Create a namespace with this circuit
+    *
+    * namespace is used by RenameModules to get unique names
+    */
+  private class GetNamespace(newTopName: String) extends Transform {
```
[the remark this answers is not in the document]
albertchen-sifive: Sounds good, I'll change this. I did it this way because I didn't know you could return the namespace as an annotation and renaming it in GetNamespace was just easier at the time.

src/test/resources/constant_propagation_tests/WidthsAddUInt.fir
```diff
@@ -0,0 +1,13 @@
+circuit WidthsAddUInt :
```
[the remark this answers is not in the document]
albertchen-sifive: I can change this. It was just easier to edit separate files for the firrtl tests.

src/main/scala/firrtl/transforms/ConstantPropagation.scala
albertchen-sifive: Do we need to address this since we can't create zero-width literals?

azidar: For the record, my feeling is that adding a zero-width wire should increase the width. Semantically its weird for constant prop to affect the width of a wire, unless that is properly handled (which could be a future optimization).

albertchen-sifive:
@azidar pushed changes

PR #853, opened 2018-07-23
Fix equivalence tests

albertchen-sifive:
- use `equiv_simple`/`equiv_induct` instead of `miter` for Yosys equivalence checking, because equiv commands handle uninitialized registers better
- make CommonSubexpressionElimination a `Transform` instead of a `Pass`
- add `RemoveValidIf` pass to `MinimumVerilogCompiler`
- add equivalence tests for RocketCore, FPU, ICache, and Rob


src/main/scala/firrtl/transforms/CommonSubexpressionElimination.scala
```diff
@@ -46,4 +44,9 @@ object CommonSubexpressionElimination extends Pass {
     }
     Circuit(c.info, modulesx, c.main)
   }
+
+  def inputForm: LowForm.type = LowForm
+  def outputForm: LowForm.type = LowForm
```
jackkoenig: I forgot to comment on this on the previous PR, but it's kind of weird to define a method type as the type of a singleton case object. I know IntelliJ is probably complaining about this but we usually don't put an explicit type here, or just put `CircuitForm`

albertchen-sifive:
Are the tests are timing out? They pass when I run them locally.

ucbjrl:
@albertchen-sifive, if this PR is still relevant/desirable, could you resolve the conflicts?

albertchen-sifive:
sorry for not maintaining this branch after you all went through the trouble of reviewing it. I've rebased it and dropped the commit that changed ConstantPropagation from a pass to transform since that caused a bunch of conflicts and isn't necessary for this PR.

Ran a few of them locally, and the equivalence test seem to be passing except for the ConstantPropagation ones. I think this is still relevant, but I don't know how desirable this is, and not sure whether it is worth debugging this further.

ucbjrl:
Thanks @albertchen-sifive, we'll discuss this during Monday's chisel-dev meeting.

albert-magyar:
One thing to note is that this does have binary-incompatible changes, so it should go in 1.3.0 or 1.4.0.

I'm marking this as 1.3.0 for now so that we don't forget to look at it. Feel free to change that, though.

azidar:
@albert-magyar will take a quick look.

albert-magyar:
I was able to minimize the LEC failures from the FPU to a (very) simple testcase, and I can see that this PR is failing because it's catching a **real LEC failure**!

This PR not only refactors how LEC is run in Yosys, but it also adds a suite of new Scala tests that rely on LEC and turn specific transforms on to compare with base `mverilog` output. Due to the issue fixed in #1519 where `PadWidths` is necessary in `mverilog`, this LEC flow relies on buggy code.

The reason this correctness issue is detectable as a LEC issue is (hilariously) that `ConstantPropagation` mutates the logic cone in such a way that it doesn't rely on the missing explicit sign-extension.

albert-magyar:
The good news: the true LEC issue I was able to diagnose was fixed in #1519.

The bad news: LEC still doesn't pass. At this time I am still trying to figure out if it is a spurious or real failure.

I did bump the branch further and add a simpler regression LEC test with the new `regress/ops` file.

azidar:
Deprecate and add new methods, to enable upstreaming in 1.3.X.

albert-magyar:
@azidar per the discussion in the last meeting, I updated this to only run the compiler-path-comparing LEC tests on the simple `Ops` design, which prevents the runtime and false-negative issues we were seeing with the others.

If someone wants to take another pass at reviewing this, that would be great. I changed the milestone to 1.4.0 since it's an API change, and since there's little reason to backport it.

PR #855, opened 2018-07-28
Use LinkedHashSet in propagateAnnotations

albertchen-sifive:
- use `LinkedHashSet` in `propagateAnnotations` to preserve order of annotations
- add a test case to test for annotation order preservation

### DIFF
--- a/src/main/scala/firrtl/Compiler.scala
+++ b/src/main/scala/firrtl/Compiler.scala
@@ -283,10 +283,8 @@ abstract class Transform extends LazyLogging {
       resAnno: AnnotationSeq,
       renameOpt: Option[RenameMap]): AnnotationSeq = {
     val newAnnotations = {
-      val inSet = new mutable.LinkedHashSet[Annotation]
-      inSet ++= inAnno
-      val resSet = new mutable.LinkedHashSet[Annotation]
-      resSet ++= resAnno
+      val inSet = mutable.LinkedHashSet() ++ inAnno
+      val resSet = mutable.LinkedHashSet() ++ resAnno
       val deleted = (inSet -- resSet).map {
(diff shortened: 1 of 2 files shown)


src/main/scala/firrtl/Compiler.scala
```diff
@@ -283,8 +283,10 @@ abstract class Transform extends LazyLogging {
       resAnno: AnnotationSeq,
       renameOpt: Option[RenameMap]): AnnotationSeq = {
     val newAnnotations = {
-      val inSet = inAnno.toSet
-      val resSet = resAnno.toSet
+      val inSet = new mutable.LinkedHashSet[Annotation]
```
jackkoenig: This isn't a huge deal, but I think it's more idiomatic to write:
```scala
val inSet = mutable.LinkedHashSet() ++ inAnno
```

seldridge:
This makes sense overall to me. 

One thought: while annotation order may affect the output circuit structure, it should never affect, nor be important in, output circuit function. Or: any transform that depends on annotation order is inherently broken. That being said, I'm in favor of a sane annotation order producing expected ouput circuits.

I believe this PR was to resolve an issue with an exact circuit output check that differed from what would be expected if annotations weren't getting moved around?

albertchen-sifive:
@seldridge yes, `GroupComponentsSpec` from #857

jackkoenig:
I agree with @seldrige that output should not depend on Annotation order. Unfortunately this happens a lot anyway so it's good to reduce the impact. We should think about how to prevent these issues (maybe a little testing infrastructure that randomly permutes the AnnotationSeq and checks that the output is unchanged). In any case, barring objection, I'm going to merge this.

PR #858, opened 2018-07-30
Fix Travis

jackkoenig:
It looks like Travis is currently using a really old pointer to SBT that is now broken which is why all of our builds are failing. This PR is intended to fix that based on discussion https://github.com/travis-ci/travis-ci/issues/9816

I'm PRing this so it will run, initially this should fail but I'll push to this as I try ways of fixing it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,6 @@ jobs:
         - bash .install_yosys.sh
         - yosys -V
         # Workaround for https://github.com/travis-ci/travis-ci/issues/9816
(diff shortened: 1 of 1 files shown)


jackkoenig:
The real fix appears to have been clearing the cache and deleting .ivy2 and .sbt and letting that get put into the new cache. If we merge this to master then merge master into other PRs, I *think* things will use the new working cache and will work

jackkoenig:
@azidar @ucbjrl @chick  ready for approval

Issue #598, opened 2017-06-01
Assignment of init random values for register


jchang0:
Migrated from 
https://github.com/ucb-bar/chisel3/issues/622

Clifford noted that the way random numbers are generated when number of bits > 32 is not very random:

_GEN_0 = {2{$random}};

As this will produce one 32-bit random value and then repeat them. Better to do:
_GEN_0 = {$random, $random};

In fact, he pointed out that $random isn't very good at all:
http://svn.clifford.at/handicraft/2014/vlogrnd/bad_random.v

I'm not sure what the solution for that is, I suppose the work around is probably different for each simulator.

cliffordwolf:
> I'm not sure what the solution for that is, I suppose the work around is probably different for each simulator.

A big issue with `$random` is that the standard actually defines which RNG an implementation must use, and it's not a very good one. `$urandom` does not have that limitation and implementations that support it usually provide a better RNG for `$urandom`, but it's a SystemVerilog feature, so it's not available everywhere.

Simply running the results of `$random` though a simple hash function, and/or only using a few of the bits returned by the function can already improve the randomness drastically. For example taking the values returned by `$random` modulo a large prime and then only using the lower 16 bits:

    _GEN_0[15:0] = $random % 371064217;
    _GEN_0[31:16] = $random % 371064217;
    _GEN_0[47:32] = $random % 371064217;
    _GEN_0[63:48] = $random % 371064217;

> In fact, he pointed out that $random isn't very good at all:
http://svn.clifford.at/handicraft/2014/vlogrnd/bad_random.v

A few words explanation: This example demonstrates that whenever the bits `[5:0]` returned by `$random` are zero, the bits `[27:24]` will be zero as well. A good RNG should of course return bits that are all independent of each other.

cliffordwolf:
> For example taking the values returned by $random modulo a large prime and then only using the lower 16 bits.

Yeahh.. about that: Don't do it! I've been running that through dieharder and arguably the results are even worse than for just $random.

But the following seems to work pretty well. So far it passed the first 80 tests in dieharder and failed none. I'll post an update when dieharder has finished.

	function [31:0] myrand;
		input [31:0] v1;
		input [31:0] v2;
		begin
			myrand = v1;
			myrand = myrand ^ (myrand << 13);
			myrand = myrand ^ (myrand >> 17);
			myrand = myrand ^ (myrand << 5);
			myrand = myrand ^ v2;
			myrand = myrand ^ (myrand << 13);
[...]

PR #859, opened 2018-07-31
allowing overrides to $random

debs-sifive:
**Purpose:** Provides macros for overriding default `$random` behavior.
- After some discussion, putting `$urandom` directly into FIRRTL seems ill fitting, since `$urandom` is SystemVerilog specific.

**Files changed:** `src/main/scala/firrtl/Emitter.scala`

@jackkoenig 

## New macros

**`` `RANDOM``:** Can be overwritten, for example with `$urandom`.

**`` `INIT_RANDOM``:** Will be called on the `initial` block of every module before random function calls. Can be specified to run code such as `$srandom` calls.


jackkoenig:
At long last, a fix for https://github.com/freechipsproject/firrtl/issues/598!

edwardcwang:
Do we still care about the ability of the user to specify a custom random function?

jackkoenig:
Yes, `$random` is awful. We had been using a sed script to replace `$random` with `$urandom` but it turns out that we really need to seed the random calls so that they're not identical for different instances of the same module. I suspect other people have done similarly hacky things so this provides a reasonable way to provide better randomization

edwardcwang:
Yes, in addition to switching to `$urandom`, what about the ability for users to put in custom random functions (apart from `$random` / `$urandom`)?

jackkoenig:
Yeah probably, perhaps instead of
```scala
        emit(Seq("`else"))
        emit(Seq("  `define RANDOM $urandom"))
```
It could be like
```scala
        emit(Seq("`else"))
        emit(Seq("  `ifndef RANDOM")
        emit(Seq("    `define RANDOM $urandom"))
        emit(Seq("   `endif"))
```
You might also then want to make the seeding be setable with ifdefs as well.

PR #862, opened 2018-08-01
Ignore tests requiring firrtlEquivalenceTest() if yosys isn't available.

ucbjrl:
Users may not have `yosys` installed. Ignore `ConstantPropagationEquivalenceSpec` tests using `firrtlEquivalenceTest()` if `yosys` is not available.


jackkoenig:
Is it possible to add the assume to `firrtlEquivalenceTest`? It might require some trickery but it would be cool because then we couldn't mess it up by forgetting to add that assumption.

Also is there a way to make this less easy to mess up. Imagine if we were to port our CI from Travis to Circle but forget to install yosys, `sbt test` would then pass for these tests even when in that case we want it to error if yosys isn't installed. Any thoughts?

ucbjrl:
@ducky64 had a similar response to my PR to tag chisel tests requiring a backend. Perhaps it's time to delve into the `ScalaTest` machinery.

This shouldn't be an issue on a move to CircleCI, since we run the tests inside a docker container that has `yosys` installed - required by `.circleci/run_formal_checks.sh`.

`sbt` does report ignored tests. Anything diligent about running tests, should pay attention to that.

jackkoenig:
Yeah I agree we're unlikely to mess this up (especially because of `run_formal_checks`). However, as a general pattern and for future maintainability, I'm a little concerned about tests passing due to an incomplete setup.

ucbjrl:
Just to be clear, these tests don't pass if there's no `yosys`; they're CANCELED (not IGNORED as I reported earlier).

If testing FIRRTL requires `yosys`, then that should be made explicit. We should have a `testPrepSpec` that verifies that for any external tools required for testing, the correct version is installed.

jackkoenig:
`CANCELED` still gives a green `[success]` at the end of `sbt test` though, I'm happy to have them cancelled but as you suggested, we should have a `TestingEnvironmentSpec` or something that checks for Verilator, Yosys, and any dependencies of theirs (Verilator needs a c++ compiler for example)

jackkoenig:
* Don't want `sbt test` to pass if someone is missing `verilator`, `yosys`, or future dependencies
* It would be fine to have an additional option to `sbt test` to make it skip those tests (like `sbt test --skip-nonScalaTests?`
  * There's some feature for this (probably)

jackkoenig:
Also problematic, `yosysEquiv` can fail with the exact same input.

azidar:
Decision - make yosys required for all tests to pass.

PR #865, opened 2018-08-03
Instance Annotations

azidar:
Target Inheritance (green are case classes, blue are traits, strikethrough are deprecated):
screen shot 2018-09-12 at 2 38 03 pm

Major Code Changes/Features Added:
- Added Target trait as replacement for Named
- Added TargetToken as token in building Target
- Added GenericTarget as a catch-all Target
- Added CircuitTarget, ModuleTarget, ReferenceTarget, and InstanceTarget
- Added ResolvePaths annotation
- Added EliminateTargetPaths (and helper class DuplicationHelper)
- Updated Dedup to work with instance annotations
- Updated RenameMap to work with instance annotations
- DCE & ConstantProp extend ResolveAnnotationPaths

Ongoing Discussion:
- [x] Annotation now extends Product, and has `getTargets` API (**Ok, makes sense**)
- [x] RenameMap.get now calculates rename without the Option[Seq], which might be slower? Hard to tell. (**Probably ok**)
- [x] RenameMap.get is ~O(N^3), where N is the hierarchy depth of Target - I think its impossible to calculate in linear time (**ok, maybe check if there are hierarchy/module/instance name changes**)
- [x] Semantics of renaming - EliminateTargetPaths renames instances to modules (**make this legal**)
- [x] Semantics of deletion - if Dedup maps two modules+their annotations to the same module/annotations, do the prior annotations get a DeletedAnnotation, or are they more 'renamed' and thus aren't 'deleted' (**delete them, but print to logger**)
- [x] New API: Added resolvePaths, resolvePathsOf, getAnnotationsOf to CircuitState for easy per-transform resolution of instance annotations
- [x] New API: Added prepare to Transform, allowing transforms to modify input CircuitState before execute
- [x] New API: Added ResolvedAnnotationPaths (requiring implementing ```val annotationClasses```)which allows transforms to automatically resolve instance annotations prior to `execute`

TODO:
- [x] Remove IRLookup, as is unrelated to this PR
- [x] Check BrittleAnnotation (which might be unnecessary to this PR)
- [x] Refactor "Component" to "Target"?
- [x] Restrict targets to be renamed only to pathless targets
- [x] Check completeness of Target prior to renaming it to/from
- [x] Check validity of Target prior to eliminating its path
- [x] Add filtering for long paths to avoid long runtimes
- [x] Fix scaladoc comments
- [x] Clean up code style
- [x] Add better exceptions for Target-related code
- [ ] Write more Target Tests
- [x] Write more Dedup Tests
- [x] Write more Rename Tests
- [x] Write more EliminateTargetPaths Tests
- [X] Support mapping refs to modules
- [x] Log 'deleted' annotations
- [X] Check if there are hierarchy/module/instance name changes

Old Discussion:

Regarding RenameMap.get, it appears to be between O(N^3) and O(N^4), but caching reduces this significantly so that I don't think its an issue any more:
screen shot 2018-08-31 at 11 59 24 am


src/main/scala/firrtl/annotations/Target.scala
```diff
@@ -0,0 +1,342 @@
...
+
+import firrtl.ir.Expression
+import AnnotationUtils.{toExp, validComponentName, validModuleName}
+import TargetToken._
+
+import scala.collection.mutable
+
+
+/**
+  * Refers to something in a FIRRTL [[firrtl.ir.Circuit]]. Used for Annotation targets.
```
jackkoenig: This is a fairly minor thing, and ScalaDoc does not mandate a style, but I think it is more common to have the explanation start on the first line of the comment, ie.
```scala
/** Here is a short description
   * === This line often left blank ===
   * Here starts longer descriptions
   * ...
   */
```
Starting on the second line follows JavaDoc style, and for that reason is probably the default in IntelliJ

azidar: Is the `*` of subsequent lines supposed to be lined up with the second `*` on the first line?

src/main/scala/firrtl/Compiler.scala
```diff
@@ -26,60 +29,104 @@ object RenameMap {
...
-        case other => error(s"Unsupported Component rename of $key to $other")
-      })
-      case None => nestedRename(key)
+      //println("  " * set.size + key.serialize)
+
+      set += key
+      val getter = recursiveGet(set, cache, errors)(_)
+      // First, check whole key
+      // If it matches, then
+      val ret = checked.flatMap {
```
seldridge: Can you walk me through this? I get most of the case classes, but some get confusing.

src/main/scala/firrtl/Compiler.scala
```diff
@@ -26,60 +29,104 @@ object RenameMap {
...
+          // Check ref(i)
+          val deep = if(c.path.size == 1) c.module.get else c.path.dropRight(1).last._2.value
+          val renamedInstance = getter(c.copy(reference = c.reference.dropRight(2) :+ Ref(instance.value))).map{ x =>
+            require(x.notPath.size == 1)
+            val newRef = x.reference.last.value.toString
+            x.copy(reference = x.reference.dropRight(1) ++ Seq(Instance(newRef), of))
+          }
+          // Check of(m)
+          val renamedModule = getter(c.copy(module = Some(of.value), reference = Nil))
+          val renamedOfModule = if(renamedModule.size == 1) {
```
seldridge: What is this checking for? What does it mean if a module is renamed to more than one thing?

azidar:
@seldridge @jackkoenig its ready to be reviewed!

src/main/scala/firrtl/Utils.scala
```diff
@@ -177,7 +177,7 @@ object Utils extends LazyLogging {
     error("Internal Error! %sPlease file an issue at https://github.com/ucb-bar/firrtl/issues".format(string), throwable)
   }
 
-  private[firrtl] def time[R](block: => R): (Double, R) = {
+  /*private[firrtl]*/ def time[R](block: => R): (Double, R) = {
```
edwardcwang: Hmm? (i.e. why comment as opposed to just remove)

src/main/scala/firrtl/Compiler.scala
```diff
@@ -145,11 +373,26 @@ case class CircuitState(
...
       throw new FIRRTLException(s"No EmittedCircuit found! Did you delete any annotations?\n$deletedAnnotations")
   }
+
   /** Helper function for extracting emitted components from annotations */
   def emittedComponents: Seq[EmittedComponent] =
     annotations.collect { case emitted: EmittedAnnotation[_] => emitted.value }
   def deletedAnnotations: Seq[Annotation] =
     annotations.collect { case anno: DeletedAnnotation => anno }
+
+  def resolvePaths(targets: Seq[CompleteTarget]): CircuitState = {
```
edwardcwang: Missing scaladocs?

src/main/scala/firrtl/Compiler.scala
```diff
@@ -145,11 +373,26 @@ case class CircuitState(
...
+    val newCS = new EliminateTargetPaths().runTransform(this.copy(annotations = ResolvePaths(targets) +: annotations ))
+    newCS.copy(form = form)
+  }
+
+  def resolvePathsOf(annoClasses: Class[_]*): CircuitState = {
```
edwardcwang: Missing scaladocs?

src/main/scala/firrtl/Compiler.scala
```diff
@@ -145,11 +373,26 @@ case class CircuitState(
...
+    val targets = getAnnotationsOf(annoClasses:_*).flatMap(_.getTargets)
+    if(targets.nonEmpty) resolvePaths(targets.flatMap{_.getComplete}) else this
+  }
+
+  def getAnnotationsOf(annoClasses: Class[_]*): AnnotationSeq = {
```
edwardcwang: Missing scaladocs?

src/main/scala/firrtl/Compiler.scala
```diff
@@ -246,6 +489,8 @@ abstract class Transform extends LazyLogging {
     state.annotations.collect { case a: LegacyAnnotation if a.transform == this.getClass => a }
   }
 
+  def prepare(state: CircuitState): CircuitState = state
```
edwardcwang: Missing scaladocs?

src/main/scala/firrtl/Compiler.scala
```diff
@@ -321,6 +579,16 @@ abstract class SeqTransform extends Transform with SeqTransformBased {
   }
 }
 
+trait ResolvedAnnotationPaths {
```
edwardcwang: Missing scaladocs?

src/main/scala/firrtl/annotations/Annotation.scala
```diff
@@ -18,6 +22,15 @@ trait Annotation {
     * @note In [[logger.LogLevel.Debug]] this is called on every Annotation after every Transform
     */
   def serialize: String = this.toString
+
+  private def extractComponents(ls: scala.collection.Traversable[_]): Seq[Target] = {
```
edwardcwang: Missing scaladocs?

src/main/scala/firrtl/annotations/Annotation.scala
```diff
@@ -18,6 +22,15 @@ trait Annotation {
...
+    ls.collect {
+      case c: Target => Seq(c)
+      case ls: scala.collection.Traversable[_] => extractComponents(ls)
+    }.foldRight(Seq.empty[Target])((seq, c) => c ++ seq)
+  }
+
+  def getTargets: Seq[Target] = extractComponents(productIterator.toSeq)
```
edwardcwang: Missing scaladocs?

src/main/scala/firrtl/annotations/AnnotationUtils.scala
```diff
@@ -64,13 +64,25 @@ object AnnotationUtils {
     case Array(c, m, x) => ComponentName(x, ModuleName(m, CircuitName(c)))
   }
 
+  def toSubComponents(s: String): Seq[TargetToken] = {
```
edwardcwang: Missing scaladocs?

src/main/scala/firrtl/annotations/Target.scala
```diff
@@ -0,0 +1,626 @@
...
+
+}
+
+object Target {
+
+  def apply(circuitOpt: Option[String], moduleOpt: Option[String], reference: Seq[TargetToken]) = GenericTarget(circuitOpt, moduleOpt, reference)
+
+  def unapply(t: Target): Option[(Option[String], Option[String], Seq[TargetToken])] = Some((t.circuitOpt, t.moduleOpt, t.tokens))
+
+  case class NamedException(message: String) extends Exception(message)
```
edwardcwang: Missing scaladocs? Also is this meant to be an internal exception?

src/main/scala/firrtl/annotations/TargetToken.scala
```diff
@@ -0,0 +1,37 @@
+package firrtl.annotations
+
+sealed trait TargetToken {
```
edwardcwang: Missing scaladocs?

src/main/scala/firrtl/annotations/TargetToken.scala
```diff
@@ -0,0 +1,37 @@
...
+  def is(keywords: String*): Boolean = {
+    keywords.map { kw =>
+      val lastClass = this.getClass
+      lastClass == TargetToken.keyword2targettoken(kw)("0").getClass
+    }.reduce(_ || _)
+  }
+  //override def toString = s"/$keyword@$value"
+}
+
+case object TargetToken {
```
edwardcwang: Missing scaladocs?

src/main/scala/firrtl/annotations/TargetToken.scala
```diff
@@ -0,0 +1,37 @@
...
+sealed trait TargetToken {
+  def keyword: String
+  def value: Any
+  def is(keywords: String*): Boolean = {
+    keywords.map { kw =>
+      val lastClass = this.getClass
+      lastClass == TargetToken.keyword2targettoken(kw)("0").getClass
+    }.reduce(_ || _)
+  }
+  //override def toString = s"/$keyword@$value"
```
edwardcwang: Is there a reason this is commented out?

src/main/scala/firrtl/annotations/Target.scala
```diff
@@ -0,0 +1,626 @@
...
+
+  implicit def convertCircuitTarget2CircuitName(c: CircuitTarget): CircuitName = c.toNamed
+  implicit def convertModuleTarget2ModuleName(c: ModuleTarget): ModuleName = c.toNamed
+  implicit def convertLocalComponent2ComponentName(c: IsComponent): ComponentName = c.toNamed
+  implicit def convertTarget2Named(c: Target): Named = c.toNamed
+  implicit def convertCircuitName2CircuitTarget(c: CircuitName): CircuitTarget = CircuitTarget(c.name)
+  implicit def convertModuleName2ModuleTarget(c: ModuleName): ModuleTarget = ModuleTarget(c.circuit.name, c.name)
+  implicit def convertComponentName2IsLocalComponent(c: ComponentName): ReferenceTarget = {
```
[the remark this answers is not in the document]
azidar: So no, a `ComponentName` cannot be an `InstanceTarget`, but it can be a `ReferenceTarget` that refers to an instance. The name though is relevant to the encapsulating module, not the ofmodule. This means that if you were referencing the port of an instance, it would reference `i.port` not the port inside the ofmodule of `i`.

src/main/scala/firrtl/annotations/Target.scala
```diff
@@ -0,0 +1,626 @@
...
+    toTargetTokens(c.name).toList match {
+      case Ref(r) :: components => ReferenceTarget(c.module.circuit.name, c.module.name, Nil, r, components)
+      case other => throw NamedException(s"Cannot convert $c into [[LocalReferenceTarget]]: $other")
```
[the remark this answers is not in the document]
azidar: Yeah, you are right, but still going to make it complete just in case other stuff changed and that assumption no longer applies.

src/main/scala/firrtl/annotations/Target.scala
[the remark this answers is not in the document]
seldridge: This agrees with my understanding. Thanks for the explanation.

src/main/scala/firrtl/annotations/Target.scala
```diff
@@ -0,0 +1,626 @@
...
+    * @param token
+    * @return
+    */
+  def add(token: TargetToken): GenericTarget = {
+    token match {
+      case _: Instance  => requireLast(true, "inst", "of")
+      case _: OfModule  => requireLast(false, "inst")
+      case _: Ref       => requireLast(true, "inst", "of")
+      case _: Field     => requireLast(true, "ref", "[]", ".", "clock", "init", "reset")
+      case _: Index     => requireLast(true, "ref", "[]", ".", "clock", "init", "reset")
```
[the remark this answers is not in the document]
azidar: Good catch, but although its impossible, but I think i'd prefer checking this later when seeing if a target is legal. This way this construction is the same as using ReferenceTarget methods. But good catch.

src/main/scala/firrtl/annotations/analysis/DuplicationHelper.scala
```diff
@@ -0,0 +1,141 @@
...
+    * @param t An instance-resolved component
+    */
+  def expandHierarchy(t: IsMember): Unit = {
+    val path = t match {
+      case m: IsModule => m.asPath
+      case _ => t.path
+    }
+    path.reverse.tails.map { p => p.reverse }.toSeq.foreach { p =>
+      duplicate(t.module, p)
+    }
```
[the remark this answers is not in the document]
azidar: The double reverse is ok because otherwise you have to do a dropRight (or something equivalent) which is an O(N) operation each recursive call.

src/main/scala/firrtl/Compiler.scala
```diff
@@ -4,113 +4,341 @@ package firrtl
...
+
+        /** If t is an InstanceTarget (has a path) but has no references:
+          * 1) Check whether the instance has been renamed (asReference)
+          * 2) Check whether the ofModule of the instance has been renamed (only 1:1 renaming is ok)
+          */
+        case t: InstanceTarget =>
+          getter(t.asReference).map {
+            case t2:InstanceTarget => t2
+            case t2@ReferenceTarget(c, m, p, r, Nil) =>
+              val t3 = InstanceTarget(c, m, p, r, t.ofModule)
```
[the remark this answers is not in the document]
azidar: I can't use a copy because its going from a ReferenceTarget to an InstanceTarget

src/main/scala/firrtl/Compiler.scala
```diff
@@ -4,113 +4,341 @@ package firrtl
...
   def apply() = new RenameMap
 }
+
 /** Map old names to new names
   *
   * Transforms that modify names should return a [[RenameMap]] with the [[CircuitState]]
   * These are mutable datastructures for convenience
   */
 // TODO This should probably be refactored into immutable and mutable versions
 final class RenameMap private () {
```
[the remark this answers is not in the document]
azidar: Agreed, we can do this in a future PR.

src/main/scala/firrtl/annotations/Target.scala
```diff
@@ -0,0 +1,626 @@
...
+  /** @return The circuit of this target */
+  def circuit: String
+
+  /** @return The [[CircuitTarget]] of this target's circuit */
+  def circuitTarget: CircuitTarget = CircuitTarget(circuitOpt.get)
+
+  def getComplete: Option[CompleteTarget] = Some(this)
+
+  /** Checks whether the component tokens has no instance/ofModule subcomponents */
+  def isLocal: Boolean
```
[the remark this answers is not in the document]
azidar: Basically, yeah. Also I'm not crazy on the name, but I spent some more time trying to think of a new name but can't, so I guess we are keeping it.

src/main/scala/firrtl/Compiler.scala
```diff
@@ -4,113 +4,341 @@ package firrtl
...
+          */
+        case t: InstanceTarget =>
+          getter(t.asReference).map {
+            case t2:InstanceTarget => t2
+            case t2@ReferenceTarget(c, m, p, r, Nil) =>
+              val t3 = InstanceTarget(c, m, p, r, t.ofModule)
+              val ofModuleTarget = t3.ofModuleTarget
+              getter(ofModuleTarget) match {
+                case Seq(ModuleTarget(newCircuit, newOf)) if newCircuit == t3.circuit => t3.copy(ofModule = newOf)
+                case Seq(`ofModuleTarget`) => t3
```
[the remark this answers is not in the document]
azidar: Yup you are right, it will never trigger because it catches on the earlier case. I wrote a test just to check as well.

src/main/scala/firrtl/annotations/AnnotationUtils.scala
```diff
@@ -64,13 +64,29 @@ object AnnotationUtils {
...
+
+
   /** Given a serialized component/subcomponent reference, subindex, subaccess,
    *  or subfield, return the corresponding IR expression.
    *  E.g. "foo.bar" becomes SubField(Reference("foo", UnknownType), "bar", UnknownType)
    */
   def toExp(s: String): Expression = {
     def parse(tokens: Seq[String]): Expression = {
-      val DecPattern = """([1-9]\d*)""".r
+      val DecPattern = """([0-9]\d*)""".r
```
jackkoenig: Actually I think it was right before, it's saying the numbers can't start with a bunch of leading zeros, if that should be legal, then you should just change it to
```scala
val DecPattern = """(\d+)""".r
```

src/main/scala/firrtl/annotations/Target.scala
azidar: This is a user-facing error, so I don't believe it is an internal exception?

azidar:
So prior to merging, I'd like to get @davidbiancolin and @jackkoenig to approve the PR, just to make sure that it doesn't break MIDAS/SiFive code. Otherwise I believe I have addressed all of the previous reviewer comments.

src/main/scala/firrtl/transforms/Dedup.scala
```diff
@@ -28,62 +29,78 @@ class DedupModules extends Transform {
...
+    A_ -> A
+    Annos: A_, A
+    Should not dedup
+
+    Depends on Annotation type. If A_ and A share the exact same annotation, then is it ok to dedup?
+    Wire(A_/a, A_/b)
+    Wire(A/a, A/b) ok
+
+    Wire(A_/a, A_/b:B_/c)
+    Wire(A/a, A/b:B/c) ok
```
edwardcwang: Is this just a suggestion/TODO or is it already done?

davidbiancolin:
Confirming that as of 86f8ae8, this does not break FireSim/MIDAS.

jackkoenig:
One remaining issue I'm seeing. I get errors during json4s deserialization for the below. At first I thought it had to do with still using "Named" stuff but it looks like it errors either way with `DontTouchAnnotation`, so I don't know what the deal is. Any chance you could take a look @azidar?
```scala
import firrtl._
import firrtl.annotations._
import firrtl.transforms._

object Test extends App {
  //val anno = DontTouchAnnotation(ComponentName("d", ModuleName("m", CircuitName("c"))))
  val target = ReferenceTarget("c", "m", Seq(), "d", Seq())
  val anno = DontTouchAnnotation(target)
  val ser = JsonProtocol.serialize(List(anno))
  println(ser)                                                                                                                                                                    
  val anno2 = JsonProtocol.deserialize(ser)
  println(anno2)
}
```

azidar:
Good catch @jackkoenig, i need to update the JSON serializer/deserializer, I'm surprised the existing firrtl tests didn't catch this.

azidar:
Ok, I think its ready to merge (assuming @jackkoenig's test case is the last thing to check off)

src/main/scala/firrtl/annotations/Target.scala
```diff
@@ -0,0 +1,652 @@
...
+    * @return
+    */
+  def isOnly(seq: Seq[TargetToken], keywords:String*): Boolean = {
+    seq.map(_.is(keywords:_*)).foldLeft(false)(_ || _) && keywords.nonEmpty
+  }
+
+  /** @return [[Target]] from human-readable serialization */
+  def deserialize(s: String): Target = {
+    val regex = """(?=[~|>/:.\[@])"""
+    s.split(regex).foldLeft(GenericTarget(None, None, Nil)) { (t, tokenString) =>
```
jackkoenig: In practice, this probably won't matter, but `.split` returns `Array[String]` upon which `.tail` is expensive.

Using a regex to split also might be inefficient in practice, might be worth using `.split(separators: Array[Char]`

```scala
    val seps = Array('~', '|', '>', '/', ':', '.', '\\', '[', '@')
    s.split(seps).toList.foldLeft(GenericTarget(None, None, Nil)) { (t, tokenString) =>
```

src/main/scala/firrtl/annotations/Target.scala
```diff
@@ -111,6 +120,29 @@ object Target {
...
+          if(value == "???") t else t.copy(moduleOpt = Some(value))
+        case '/' => t.add(Instance(value))
+        case ':' => t.add(OfModule(value))
+        case '>' => t.add(Ref(value))
+        case '.' => t.add(Field(value))
+        case '[' if value.dropRight(1).toInt >= 0 => t.add(Index(value.dropRight(1).toInt))
+        case '@' if value == "clock" => t.add(Clock)
+        case '@' if value == "init" => t.add(Init)
+        case '@' if value == "reset" => t.add(Reset)
+        case other => throw NamedException(s"Cannot deserialize Target: $s")
```
jackkoenig: I think a suite of tests showing how this looks would be nice, I was a little surprised to see
`~circuit|module>component` (not that I'm complaining, it would just be nice to see the format laid out in examples)

src/main/scala/firrtl/annotations/Target.scala
```diff
@@ -111,6 +120,29 @@ object Target {
...
+        case '|' if t.moduleOpt.isEmpty && t.tokens.isEmpty =>
+          if(value == "???") t else t.copy(moduleOpt = Some(value))
+        case '/' => t.add(Instance(value))
+        case ':' => t.add(OfModule(value))
+        case '>' => t.add(Ref(value))
+        case '.' => t.add(Field(value))
+        case '[' if value.dropRight(1).toInt >= 0 => t.add(Index(value.dropRight(1).toInt))
+        case '@' if value == "clock" => t.add(Clock)
+        case '@' if value == "init" => t.add(Init)
+        case '@' if value == "reset" => t.add(Reset)
```
jackkoenig: This one is probably okay, but I think we should benchmark to be sure:

`.add` involves appending to Lists which could get expensive for deeply nested aggregates. For anything < 5 it's fast because those Lists are optimized.

I don't know if this is worth "fixing" without benchmarks showing it's a problem though because it might be kind of a pain to make fast.

src/main/scala/firrtl/Compiler.scala
```diff
@@ -178,7 +465,7 @@ sealed abstract class CircuitForm(private val value: Int) extends Ordered[Circui
   *
   * See [[CDefMemory]] and [[CDefMPort]]
   */
-final case object ChirrtlForm extends CircuitForm(3)
+final case object ChirrtlForm extends CircuitForm(value = 3)
```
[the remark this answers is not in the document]
azidar: Scala style guide doesn't allow magic numbers greater than 2

src/main/scala/firrtl/annotations/Target.scala
azidar: I changed tokens from a Seq to a Vector, so that should do it

src/main/scala/firrtl/annotations/Target.scala
azidar: Ok, I'll add a test.

jackkoenig:
Can confirm the [de]serialization problem is fixed, but there's another issue that I'm debugging. A top-level port in our design is getting dropped, trying to figure out why

src/main/scala/firrtl/annotations/Target.scala
jackkoenig: Good point, yeah the fold is on the array so it's fine.

And good point, I think the regex is fine, if it becomes an issue we can find a different solution but you're right, it's probably fast enough.

jackkoenig:
For a moderate-sized design, this PR increases the time for Dedup from 8 seconds to 24 seconds.

azidar:
Yeah, its certainly going to increase compile time of Dedup. I think its important to get clonemodule working so we don't need to rely on dedup as much

jackkoenig:
Okay so I’ve figured out what’s causing our build some issues. This kind of relates to the notion of “forking” the firrtl build, where you have multiple tops that you want in a single compilation. My current solution involves marking the ports of each “top” as “don’t touch”, and on `master`, it works fine. I grant this isn’t a particularly elegant solution so I am open to better techniques (but I need to have one). I could completely split each compilation and I will try that out again, but last time I did that it had two undesirable consequences:
1. It was *much* slower (like 2-3x total FIRRTL runtime)
2. If you use log-level info, the screen just gets blasted because each “fork” (and there can be 100s) has to print each transform
 To see what I’m talking about, see the following example:

Given a simple firrtl circuit:
```
circuit test :
  module test :
    input x : UInt<1>
    output y : UInt<1>

    y <= x 
```                                                                                                                                                                      

And a transform that injects a module not instantiated by the top:

```scala
import firrtl._
import firrtl.ir._
import firrtl.annotations._
import firrtl.transforms._

class MyTransform extends Transform {
  def inputForm = MidForm
  def outputForm = MidForm

    val mod = state.circuit.modules.head.asInstanceOf[Module]
    val mod2 = mod.copy(name = "magic")
    val dts = mod2.ports.map { case port =>
      val t = ComponentName(port.name, ModuleName(mod2.name, CircuitName(state.circuit.main)))
      DontTouchAnnotation(t)
    }                                                                                                                                                                             
    state.copy(annotations = dts ++ state.annotations,
               circuit = state.circuit.copy(modules = mod2 +: state.circuit.modules))

  }
}
```

`firrtl.Driver -i test.fir -o test.v -X verilog -fct MyTransform`

If you run this on `master`, `magic` will show up in the Verilog. Currently with this branch, it will delete `magic` in `DeadCodeElimination`

Maybe it's a simple fix to preserve the old behavior, but I am also open to a better way of doing this/support for multi-tops / "forking" the compiler without killing performance and the log

azidar:
Hmm, well this should preserve the previous behavior, so I believe there is a bug. I'll check it out.

jackkoenig:
I think `EliminateTargetPaths` is deleting the `DontTouchAnnotation`s on modules that are not instantiated in the module hierarchy underneath the "true top" module.

src/main/scala/firrtl/annotations/analysis/DuplicationHelper.scala
```diff
@@ -0,0 +1,146 @@
...
+      path.foldLeft((top, newTop)) { case ((originalModule, newModule), (instance, ofModule)) =>
+        val newOfModule = getNewOfModule(originalModule, newModule, instance, ofModule)
+        newPath ++= Seq(instance, newOfModule)
+        (ofModule.value, newOfModule.value)
+      }
+      val module = if(newPath.nonEmpty) newPath.last.value.toString else newTop
+      t.notPath match {
+        case Nil => ModuleTarget(t.circuit, module)
+        case Instance(i) :: OfModule(m) :: Nil => ModuleTarget(t.circuit, module)
+        case Ref(r) :: components => ReferenceTarget(t.circuit, module, Nil, r, components)
```
jackkoenig: The change to Vector makes this match always fail since these all are List extractions.

Can be made generic Seq like so:
```scala
        case Seq() => ModuleTarget(t.circuit, module)
        case Instance(i) +: OfModule(m) +: Seq() => ModuleTarget(t.circuit, module)
        case Ref(r) +: components => ReferenceTarget(t.circuit, module, Nil, r, components)
```
We should probably also exercise these paths in tests

azidar: Actually all implementations of `notPath` return lists, so the tests do in fact cover this case. However, I've still changed the `::` to `+:`.

src/main/scala/firrtl/Compiler.scala
```diff
@@ -246,6 +182,12 @@ abstract class Transform extends LazyLogging {
     state.annotations.collect { case a: LegacyAnnotation if a.transform == this.getClass => a }
   }
 
+  /** Executes before any transform's execute method
+    * @param state
+    * @return
+    */
+  def prepare(state: CircuitState): CircuitState = state
```
jackkoenig: I brought this up in a meeting ages ago but I want to touch base on it again.

I like the idea of `ResolvedAnnotationPaths` and being able to mix it in and get nice behavior like automatically running `resolvePathsOf` before execute. But it's not composable, you can't create something just like `ResolvedAnnotationPaths` that also overrides `prepare`, otherwise you will get a compiler error that you have to resolve by overriding `prepare` yourself.

I am, therefore, of the opinion that we should not yet expose `prepare` as a public API until we can figure out how to make it composable. I think if we make this a public API we're in for a load of hurt if/when we need to change it in the future.

src/main/scala/firrtl/annotations/analysis/DuplicationHelper.scala
jackkoenig: If all implementations of `notPath` return lists then how did I get a match error with a Vector on this? Prior to https://github.com/freechipsproject/firrtl/pull/865/commits/792472fbc733ebf441e12bfed52c04a015f458c9 at least.

azidar: Hmm.... Well then. Not sure, let me look closer.

azidar: Ahh so in `ReferenceTarget` I append `component`, which could be a `Vector`. Good catch. Anyways, this is fixed now.

src/main/scala/firrtl/Compiler.scala
azidar: Do you have a suggestion for an alternative proposal? I agree with what you've said, but IMO the solution is schedulable transformations, which is a huge feature to add (and on my todo list) but way outside the scope of this PR. I could preemptively add a deprecation flag?

seldridge: Would making this an actual class that extends `Transform` and not a trait do it? At least then it's explicit that the user will have to override.

jackkoenig: @seldridge's suggestion would work, but in the composable analyses/prep like this world of my dreams, you could mixin multiple traits. Of course, traits have a potential ordering problem so it may just not be the right API.

In any case, I think prepare should just be `private[firrtl]` for now, then we can solve that problem later.

azidar: Ok, will do.

jackkoenig:
Thanks for all of your hard work! LGTM!

PR #867, opened 2018-08-06
Stop reporting exceptions in custom transformations as internal errors

jackkoenig:
Instead, just forward the exception

### DIFF
--- a/src/main/scala/firrtl/Compiler.scala
+++ b/src/main/scala/firrtl/Compiler.scala
@@ -295,6 +295,9 @@ trait Emitter extends Transform {
   def emit(state: CircuitState, writer: Writer): Unit
 }
 
+/** Wraps exceptions from CustomTransforms so they can be reported appropriately */
+case class CustomTransformException(cause: Throwable) extends Exception("", cause)
+
 object CompilerUtils extends LazyLogging {
   /** Generates a sequence of [[Transform]]s to lower a Firrtl circuit
     *
@@ -427,6 +430,15 @@ trait Compiler extends LazyLogging {
     compile(state.copy(annotations = emitAnno +: state.annotations), customTransforms)
   }
 
+  private def isCustomTransform(xform: Transform): Boolean = {
+    def getTopPackage(pack: java.lang.Package): java.lang.Package =
+      Package.getPackage(pack.getName.split('.').head)
+    // We use the top package of the Driver to get the top firrtl package
+    Option(xform.getClass.getPackage).map { p =>
+      getTopPackage(p) != firrtl.Driver.getClass.getPackage
+    }.getOrElse(true)
+  }
+
   /** Perform compilation
     *
     * Emission will only be performed if [[EmitAnnotation]]s are present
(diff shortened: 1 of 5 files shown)


src/main/scala/firrtl/Compiler.scala
```diff
@@ -465,10 +468,20 @@ trait Compiler extends LazyLogging {
...
-      allTransforms.foldLeft(state) { (in, xform) => xform.runTransform(in) }
+      allTransforms.foldLeft(state) { (in, xform) =>
+        try {
+          xform.runTransform(in)
+        } catch {
+          // Rethrow the exceptions which are expected or due to the runtime environment (out of memory, stack overflow)
+          case p: scala.util.control.ControlThrowable => throw p
+          // Wrap exceptions from custom transforms so they are reported as such
+          case e: Exception if isCustomTransform(xform) => throw CustomTransformException(e)
+        }
```
chick: Do you need lines 480-481?
Seems like the only thing you are actually catching is the `isCustomTransform`, all other exceptions will bubble up either way.

jackkoenig: I was worried about what would happen if we wrapped a ControlThrowable in a CustomTransformException, but since the Driver just forwards it, it'll do the right thing. I'll fix that.

I'm also going to make `isCustomTransform` a `def` since it only needs to be calculated once and in the rare case of an exception.

jackkoenig:
One potential problem with this is that it treats optional built-in transforms as Custom Transforms, so things like ReplSeqMems will be treated in the same way.

We implement this differently in a couple of ways:
1. Check if the transform is defined in the `firrtl` package
1. Create a whitelist of transforms (brittle and annoying to maintain)

I'm leaning toward 1. now. I had thought about that originally but I thought this approach would be more robust. While it is more robust, in my opinion it still has too many false negatives.

seldridge:
Is there anything that precludes a user building a library and putting it in the FIRRTL package? Does that matter?

chick:
AFAIK there isn't but what we should probably be doing is naming our packages properly as 
<umbrella>.<organizations>.firrtl
Probably too late for 3.2.x

jackkoenig:
I don't think there is anyway to prevent that. Of course a whitelist could prevent errors in such transforms from being reported as internal errors. I think it doesn't really matter though, if you inject your code into other people's packages you're always asking for trouble.

jackkoenig:
Per chisel things, I actually don't want this to treat things in `com.mycoolcompany.firrtl` as internal. Rather, only things defined in `firrtl.whatever`. I do grant that it's a problem for transforms defined in Chisel; however, there currently isn't a single one.

jackkoenig:
@chick @seldridge thoughts on the updated implementation?  Unfortunately I don't think there's a more robust to get the "top package" in which a class is defined other than splitting the package String by `.` and then loading the package named by the head of the resulting array. It looks like `package java` and `package java.lang` are two completely unrelated packages at runtime so I did some stringly-typed magic to determine the top-level package.

jackkoenig:
Okay this fails with a NullPointerException if the transform is defined in the root package, need to fix that

jackkoenig:
I guess I think it's likely that more and more build-in transforms will be run with that same mechanism and kind of the whole idea of the "Internal Error" Exception is to let the user know that this error is unhandled and is our fault. Relatedly, we should probably stop reporting user error as exceptions with useless stack traces and instead report the errors and exit

azidar:
So, should we merge this?

jackkoenig:
The implementation *could* be cleaner, but assuming it still works, I think this is very desirable. Thoughts @seldridge? The main alternative (a twist on Schuyler's proposal above), is to make all built-in transforms extend some special trait (eg. `BuiltInTransform`) so that we know what is internal and what is custom. Instead, this PR is using the package name, looking for "firrtl" which is a bit less clean but 🤷‍♂️

jackkoenig:
Fixed test and rebased (because while it's not a huge deal, the Mergify squash and merge results in intermediate commit messages being kept and I didn't want them from here). This is still fine to rebase and merge, it just won't have intermediate messages

seldridge:
FWIW, one of the gates on mergify is that the commit has to have the `Please Merge` tag. Without that, you're in complete control including to do a GitHub-based squash and merge.

jackkoenig:
Ah true, well in any case I cleaned this one up for Mergify but might just do that in the future :-P

PR #868, opened 2018-08-07
Make RemoveWires properly include registers in dependency graph

jackkoenig:
Fixes a bug where registers could be instantiated after nodes that
referred to them

Also add WRef.apply utility for nodes

### DIFF
--- a/src/main/scala/firrtl/transforms/RemoveWires.scala
+++ b/src/main/scala/firrtl/transforms/RemoveWires.scala
@@ -22,13 +22,13 @@ class RemoveWires extends Transform {
   def inputForm = LowForm
   def outputForm = LowForm
 
-  // Extract all expressions that are references to a Wire or Node
+  // Extract all expressions that are references to a Node, Wire, or Reg
   // Since we are operating on LowForm, they can only be WRefs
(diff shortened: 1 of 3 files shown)


src/main/scala/firrtl/transforms/RemoveWires.scala
```diff
@@ -80,8 +80,8 @@ class RemoveWires extends Transform {
 
     def onStmt(stmt: Statement): Statement = {
       stmt match {
-        case DefNode(info, name, expr) =>
-          netlist(we(WRef(name))) = (expr, info)
+        case node: DefNode =>
+          netlist(we(WRef(node))) = (node.value, node.info)
```
seldridge: Good cleanup.

src/main/scala/firrtl/transforms/RemoveWires.scala
```diff
@@ -60,7 +60,7 @@ class RemoveWires extends Transform {
         val (rhs, info) = netlist(key)
         kind match {
           case RegKind => regInfo(key)
-          case _ => DefNode(info, name, rhs)
+          case WireKind | NodeKind => DefNode(info, name, rhs)
```
seldridge: The explicit types is a good idea.

PR #872, opened 2018-08-21
Allow the #delay before random initialization to be overridden

aswaterman:
The default of 0.002 can be less than the Verilog time precision, which
effectively causes it to be rounded down to 0.  So, allow the user to
`define RANDOMIZE_DELAY to some other value.

If the macro is not defined, the old behavior is preserved.


aswaterman:
no I will!

PR #875, opened 2018-08-22
[F764.1] Bump scopt from 3.6.0 -> 3.7.0

seldridge:
This provides support for increased introspection of options inside of
scopt, e.g., getting an options short option (shortOpt).

Patch 1 for #764

Signed-off-by: Schuyler Eldridge <schuyler.eldridge@ibm.com>

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -62,7 +62,7 @@ libraryDependencies += "org.scalatest" %% "scalatest" % "3.0.1" % "test"
 
 libraryDependencies += "org.scalacheck" %% "scalacheck" % "1.13.4" % "test"
 
-libraryDependencies += "com.github.scopt" %% "scopt" % "3.6.0"
+libraryDependencies += "com.github.scopt" %% "scopt" % "3.7.0"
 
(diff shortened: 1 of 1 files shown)


jackkoenig:
What happens if you use FIRRTL with scopt 3.7.0 in Chisel with scopt 3.6.0? Is this a "bump everyone at the same time" or can we be more decoupled?

seldridge:
This causes 3.7.0 to be used by Chisel. See:
```
[warn] Found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[warn] 	* com.github.scopt:scopt_2.11:3.7.0 is selected over 3.6.0
[warn] 	    +- edu.berkeley.cs:firrtl_2.11:1.2-SNAPSHOT           (depends on 3.7.0)
[warn] 	    +- edu.berkeley.cs:chisel3_2.11:3.2-SNAPSHOT          (depends on 3.6.0)
```
Shouldn't be any functional problems, but I'll throw a PR at Chisel.

PR #876, opened 2018-08-23
Fix NoDedupMem to be cognizant of Module scope

jackkoenig:
Previously, mems marked no dedup would prevent mems with the same
instance name in other modules from deduping

This is *technically* changing an API because I changed the implementation of public utility method `eqMems`. I did so because the noDeDupeMems argument doesn't really work as intended. Honestly I doubt anyone is using it.

### DIFF
--- a/src/main/scala/firrtl/passes/memlib/ResolveMemoryReference.scala
+++ b/src/main/scala/firrtl/passes/memlib/ResolveMemoryReference.scala
@@ -30,9 +30,7 @@ class ResolveMemoryReference extends Transform {
       case _ => false
     }
   }
-  private implicit def wrapMem(x: DefAnnotatedMemory) = new WrappedDefAnnoMemory(x)
-
-  import scala.language.implicitConversions
+  private def wrap(mem: DefAnnotatedMemory) = new WrappedDefAnnoMemory(mem)
 
   // Values are Tuple of Module Name and Memory Instance Name
   private type AnnotatedMemories = collection.mutable.HashMap[WrappedDefAnnoMemory, (String, String)]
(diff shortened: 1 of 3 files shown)


colinschmidt:
Seems like a good bugfix to me. One question:

In your test, did you verify that the 2 extmodules that exist are the correct ones. I believe this change is being made to make sure the two memories in the parent module are deduped, but it seems the test would also pass if it just ignored the `NoDedupMems` annotation.
Perhaps there is another test that makes sure that doesn't happen, if so ignore me.

jackkoenig:
There are other tests that check that `NoDedupMemsAnnotations` are acknowledged, but this test actually does as well. There are 3 identical SeqMems in use. If the annotation were ignored, there would be 1 ExtModule created, if you run the test without the bug fix, there are 3 ExtModules created, thus by checking for 2 we check both of those issues.

jackkoenig:
I'll add that explanation as a comment to the test to make the logic clear to us in the future.

colinschmidt:
Ok thanks for the clarification.

jackkoenig:
Clarification added to the test, rebased, and pushed

src/main/scala/firrtl/passes/memlib/ResolveMemoryReference.scala
```diff
@@ -21,28 +21,40 @@ class ResolveMemoryReference extends Transform {
...
-  def updateMemStmts(mname: String, uniqueMems: AnnotatedMemories, noDeDupeMems: Seq[String])(s: Statement): Statement = s match {
+  def updateMemStmts(mname: String,
+                     uniqueMems: AnnotatedMemories,
+                     noDedupMap: Map[String, Set[String]])
+                    (s: Statement): Statement = s match {
     case m: DefAnnotatedMemory =>
-      uniqueMems find (x => eqMems(x._2, m, noDeDupeMems)) match {
+      uniqueMems.find { case (mname2, m2) =>
+        !found(noDedupMap, mname2, m2.name) &&
+        !found(noDedupMap, mname, m.name) &&
```
colinschmidt: The check on `m` could be hoisted out of the find right?
Not sure if that would be more or less clear, it took me a second to read it this way, so I think it might help readability slightly although it might be uglier scala.

jackkoenig: My previous response was actually just to Colin, you responded between my loading the page and submitting my comment 😛. In any case, this is definitely inefficient and originally I was just fixing the issue at hand but I'm wondering if just cleaning up the implementation wouldn't make more sense...

jackkoenig:
@seldridge @colinschmidt 

I've pushed a cleanup that rather than n^2 find, uses a wrapping class so that the Mems can just be looked up in a HashMap.

PR #879, opened 2018-08-23
[F764.5] Add firrtl.options package, but do not use

seldridge:
This adds a new package, `firrtl.options`, that provides a framework for working with options inside FIRRTL and in other frontends, backends, or toolflows. This treats all options, metadata, etc. as an `Annotation`. An `AnnotationSeq` then forms the ground truth for all options being passed to a toolflow. 

This is the necessary framework for the "Options are Annotations" refactor, #764. __However, this does not use any of the `firrtl.options` package!__ The existing `ExecutionOptionsManager` approach is left unchanged and continued to be used.

For this, the new `Stage` will replace the old `Driver`s. In parallel to this, I'm working on a compatibility layer that keeps the old `Driver`s intact.

Patch 5 for #764.

Merge strategy: @seldridge will merge via merge commit after a rebase

### DIFF
--- a/src/main/scala/firrtl/options/OptionParser.scala
+++ b/src/main/scala/firrtl/options/OptionParser.scala
@@ -1,0 +1,39 @@
+// See LICENSE for license details.
+
+package firrtl.options
+
+import firrtl.{FIRRTLException, AnnotationSeq}
+
+import scopt.OptionParser
+
+/** Causes an OptionParser to not call exit (call `sys.exit`) if the `--help` option is passed
+  */
+trait DoNotTerminateOnExit { this: OptionParser[_] =>
+  override def terminate(exitState: Either[String, Unit]): Unit = Unit
+}
+
+/** A modified OptionParser with mutable termination and additional checks
+  */
+trait DuplicateHandling extends OptionParser[AnnotationSeq] {
+
+  override def parse(args: Seq[String], init: AnnotationSeq): Option[AnnotationSeq] = {
+
+    /** Message for found duplicate options */
+    def msg(x: String, y: String) = s"""Duplicate $x "$y" (did your custom Transform or OptionsManager add this?)"""
+
+    val longDups = options.map(_.name).groupBy(identity).collect{ case (k, v) if v.size > 1 && k != "" => k }
+    val shortDups = options.map(_.shortOpt).flatten.groupBy(identity).collect{ case (k, v) if v.size > 1 => k }
+
+
+    if (longDups.nonEmpty)  {
+      throw new OptionsException(msg("long option", longDups.map("--" + _).mkString(",")), new IllegalArgumentException)
+    }
+
+    if (shortDups.nonEmpty) {
+      throw new OptionsException(msg("short option", shortDups.map("-" + _).mkString(",")), new IllegalArgumentException)
+    }
+
+    super.parse(args, init)
+  }
+
+}
--- a/src/main/scala/firrtl/options/OptionsView.scala
+++ b/src/main/scala/firrtl/options/OptionsView.scala
@@ -1,0 +1,27 @@
+// See LICENSE for license details.
+
+package firrtl.options
+
+import firrtl.AnnotationSeq
+
+/** Type class defining a "view" of an [[AnnotationSeq]]
+  * @tparam T the type to which this viewer converts an [[AnnotationSeq]] to
+  */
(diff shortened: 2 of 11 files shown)


seldridge:
Ping @azidar. I need to write the compatibility layer, but this is good for a review (more from a software architecture perspective if this is the model that we want to move towards).

src/main/scala/firrtl/options/OptionParser.scala
```diff
@@ -0,0 +1,43 @@
...
+}
+
+/** Causes an [[OptionParser]] to not call exit (call `sys.exit`) if the `--help` option is passed
+  */
+trait DoNotTerminateOnExit { this: OptionParser[_] =>
+  override def terminate(exitState: Either[String, Unit]): Unit = Unit
+}
+
+/** A modified [[scopt.OptionParser]] with mutable termination and additional checks
+  * @param name the name of the application
```
chick: fossil @param statement

src/main/scala/firrtl/options/OptionsView.scala
```diff
@@ -0,0 +1,27 @@
+// See LICENSE for license details
```
chick: The license line is missing a period

src/main/scala/firrtl/options/OptionParser.scala
```diff
@@ -0,0 +1,43 @@
+// See LICENSE for license details
```
chick: The license line is missing a period at end

seldridge:
Thanks for the review, @chick. 

Nah, is okay. I can take care of the style stuff. 

For the deprecated things, I'll need to take a look. I think what was going on is that I intended to write a compatibility layer in `firrtl.options` and then realized that this could be entirely hoisted out. I'll check on that now.

src/main/scala/firrtl/options/OptionParser.scala
```diff
@@ -0,0 +1,43 @@
...
+
+package firrtl.options
+
+import firrtl.{FIRRTLException, AnnotationSeq}
+import scopt.OptionParser
+
+/** Provides a mutable approach to enable/disable exiting when the `--help` option is passed
+  */
+@deprecated("Use immutable DoNotTerminateOnExit", "3.2.0")
```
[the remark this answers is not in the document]
seldridge: Yeah, this is silly now that I look at it. I think I was trying to preserve functionality in the new options manager for a time. However, this isn't necessary. We're free to define the new `firrtl.options.ExecutionOptionsManager` and it's parser however we want. I'll remove this.

src/main/scala/firrtl/options/Registration.scala
```diff
@@ -0,0 +1,36 @@
+// See LICENSE for license details.
+
+package firrtl.options
+
+import firrtl.{AnnotationSeq, Transform}
+
+import scopt.OptionParser
+
+/** Indicates that this class/object includes options (but does not add these as a registered class
```
azidar: Missing end ")"

src/main/scala/firrtl/options/Registration.scala
```diff
@@ -0,0 +1,36 @@
...
+  /** This method will be called to add options to an OptionParser
+    * @param p an option parser
+    */
+  def addOptions(p: OptionParser[AnnotationSeq]): Unit
+}
+
+/** A [[Transform]] that includes options that should be exposed at the top level.
+  *
+  * @note To complete registration, include an entry in
+  * src/main/resources/META-INF/services/firrtl.options.RegisteredTransform */
```
azidar: What does this mean? Do you need to update this file for each custom transform with options?

src/main/scala/firrtl/options/Registration.scala
seldridge: Good catch, fixed in the rebase.

seldridge:
@azidar: Thanks for the review!
@chick: Let me know if you're okay with this. I changed `ExecutionOptionsManager` to `Shell`. Short and sweet and also mostly correctly as this is what provides a user interface to a stage. No substantive changes otherwise, except to address feedback.

seldridge:
Ping @chick. Trying to see if all the FIRRTL stuff re: annotations/options can be merged before CCC. I can dismiss your review on this, but I'd prefer to get signoff on the `Shell` name.

chick:
Looks good to me.

PR #880, opened 2018-08-24
Enable "Unserializable" JSON Annotations

seldridge:
__This no longer is needed by the F764 patch series. I'm hesitant on whether or not this is the right approach. I'm leaving this here for the time being.__

This adds a trait `Unserializable` that indicates two things when mixed into an `Annotation`:

1. This indicates that this annotation cannot be serialized to JSON by json4s natively because it contains something "weird" like an anonymous function, e.g., `() => RawModule`.
2. By implementing an abstract method, `toJsonSerializable`, this `Annotation` can be converted to an `AnnotationSeq` _that can be serialized_.
3. When the `JsonProtocol` tries to serialize it, it will first convert it to something that is "good" for JSON, e.g., serializing `() => RawModule` would elaborate the module first.

JSON serialization is modified to ignore any `DeletedAnnotation`s that contain an unserializable annotation.

This includes a test that unserializable things that the user doesn't implement properly except _at run time_ and that an unserializable annotation is converted to what you expect.

Note: I used a type-unsafe approach here for the abstract method `toJsonSerializable`. I need this to return an `Annotation` that does not mixin `Unserializable`. I wasn't aware of an approach that could do that in a method signature. I fell back to having a second, final method `_toJsonSerializable` call `toJsonSerializable` and do run-time type checks. This is what is used for automatic conversion during JSON serialization.

### DIFF
--- a/src/main/scala/firrtl/annotations/Annotation.scala
+++ b/src/main/scala/firrtl/annotations/Annotation.scala
@@ -1,10 +1,8 @@
 // See LICENSE for license details.
 
-package firrtl
-package annotations
+package firrtl.annotations
 
-import net.jcazevedo.moultingyaml._
-import firrtl.annotations.AnnotationYamlProtocol._
+import firrtl._
 
 case class AnnotationException(message: String) extends Exception(message)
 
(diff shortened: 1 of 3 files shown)


seldridge:
Closing this as it was not needed for Stage/Phase.

Issue #882, opened 2018-08-28
Inhibit multiple inclusions of the same additional Verilog source file


ucbjrl:
**Type of issue**: bug report
With the addition of the (recently) new BlackBoxSourceHelper code, additional Verilog resource files that needed to be explicitly included are now automatically added to `firrtl_black_box_resource_files.f`.
Old code will now get errors when those files are included twice.

The fix is to filter the `vSource` list of additional Verilog sources in `BackendCompilationUtilities.verilogToCpp()` so we don't include files that are already listed in `firrtl_black_box_resource_files.f`.

  * **What is the current behavior?**
The following from dsptools BlackBoxFloat.scala generates an error with current SNAPSHOT code but compiles with released code:
```scala
class BlackBoxFloatSpec extends ChiselFlatSpec {
  "A BlackBoxed FP block" should "work" in {
    assertTesterPasses({ new BlackBoxFloatTester }
      , Seq(
      "/BBFAdd.v",
      "/BBFMultiply.v",
      "/BBFEquals.v"
      )
    )
  }
```
If the additional Verilog sources are commented out or omitted, the code compiles with current SNAPSHOTs, but fails with current releases.

  * **What is the expected behavior?**
BackendCompilationUtilities should not generate potentially broken verilator invocations.

**What is the use case for changing the behavior?**
Clients should not have to modify their code (if at all possible) as a result of new features ostensibly for their benefit.

**Impact**: API modification
verilator will not be asked to include the same file multiple times and cases where these duplicate definitions caused verilator compilation failures, will now succeed since only a single copy of the same file will be supplied as a source.

**Development Phase**: proposal
To be submitted.

ucbjrl:
closed by #883

PR #883, opened 2018-08-28
Filter resource file names to avoid including the same file multiple times.

ucbjrl:
Addresses issue #882.

### DIFF
--- a/src/main/scala/firrtl/util/BackendCompilationUtilities.scala
+++ b/src/main/scala/firrtl/util/BackendCompilationUtilities.scala
@@ -9,6 +9,7 @@ import java.util.Calendar
 
 import firrtl.FirrtlExecutionOptions
 
+import scala.collection.immutable.HashMap
 import scala.sys.process.{ProcessBuilder, ProcessLogger, _}
  
 trait BackendCompilationUtilities {
@@ -79,6 +80,12 @@ trait BackendCompilationUtilities {
     * all the files which are not included elsewhere. If multiple ones exist,
     * the compilation will fail.
     *
+    * If the file BlackBoxSourceHelper.fileListName exists in the output directory,
+    * it contains a list of source files to be included. Filter out any files in the vSources
+    * sequence that are in this file so we don't include the same file multiple times.
+    * This complication is an attempt to work-around the fact that clients used to have to
+    * explicitly include additional Verilog sources. Now, more of that is automatic.
+    *
     * @param dutFile name of the DUT .v without the .v extension
     * @param dir output directory
     * @param vSources list of additional Verilog sources to compile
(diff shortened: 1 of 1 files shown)


src/main/scala/firrtl/util/BackendCompilationUtilities.scala
```diff
@@ -104,12 +111,26 @@ trait BackendCompilationUtilities {
       }
     }
 
+    // Don't include the same file multiple times.
+    // If it's in BlackBoxSourceHelper.fileListName, don't explicitly include it on the command line.
+    // Build a map with absolute file paths as keys to use as a filter to exclude already included additional Verilog sources.
+    val blackBoxHelperFiles: Map[String, Boolean] = {
+      if(list_file.exists()) {
+        io.Source.fromFile(list_file).getLines.map {
+          line => line -> true
```
edwardcwang: Is this just being used as a Set?

src/main/scala/firrtl/util/BackendCompilationUtilities.scala
```diff
@@ -104,12 +111,26 @@ trait BackendCompilationUtilities {
...
+        }.toMap
+      }
+      else {
+        Map.empty
+      }
+    }
+    val vSourcesFiltered = vSources.filterNot(f => blackBoxHelperFiles.contains(f.getAbsolutePath))
```
jackkoenig: The BlackBoxSourceHelper does write absolute paths so this will *probably* work in most cases, but I think it's brittle and possibly not portable to use Strings comparing absolute paths (eg. Windows). I think we should use proper java APIs to compare the files, eg.
https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#isSameFile-java.nio.file.Path-java.nio.file.Path-

```
scala> import java.io.File
import java.io.File

scala> import java.nio.file.Files
import java.nio.file.Files

scala> val f = new File("build.sbt")
f: java.io.File = build.sbt

scala> val f2 = new File("project/../build.sbt")
f2: java.io.File = project/../build.sbt

scala> f == f2
res0: Boolean = false

scala> f.getAbsolutePath
res1: String = /Users/jack/console/build.sbt

scala> f2.getAbsolutePath
res2: String = /Users/jack/console/project/../build.sbt

scala> f.getCanonicalPath
res3: String = /Users/jack/console/build.sbt

scala> f2.getCanonicalPath
res4: String = /Users/jack/console/build.sbt

scala> Files.isSameFile(f.toPath, f2.toPath)
res6: Boolean = true
```

jackkoenig: Unfortunately, I'm not sure of a good way doing a constant time lookup for files that are the same using something like `Files.isSameFile`, perhaps there is a way

jackkoenig:
It marked my comments as outdated, but they're not

ucbjrl:
Good point @jackkoenig. I think string comparisons are fine as long as the same strings are generated for the same resource file in all cases. We should note this in the PR and at the sites where the paths are generated.

One could argue that it's really not sufficient to just verify that the paths match (or don't) - we should be checking file contents.

PR #885, opened 2018-09-07
Bug Fixes in TopWiring

alonamid:
I found a couple of bugs in the TopWiring transform while using it in an actual scenario (rather than only unit tests). The two bugs were a results of running the transform without any relevant annotations, and the transform renaming leaf components in a bad way (separating letters with underscores. This wasn't discovered in the unit tests because all the leaf signal were single letters).
I fixed them, and added an additional unit tests for one of the bugs.
I also changed to package name from 'transform' to 'transforms' in order to match the rests of the firrtl transforms.

### DIFF
--- a/src/main/scala/firrtl/transforms/TopWiring.scala
+++ b/src/main/scala/firrtl/transforms/TopWiring.scala
@@ -1,5 +1,5 @@
 // See LICENSE for license details.
-package firrtl.transform
+package firrtl.transforms
 package TopWiring
 
 import firrtl._
@@ -198,7 +198,7 @@ class TopWiringTransform extends Transform {
                 }
                 path.size match {
                    case 1 => {
-                       val leafRef = WRef(path.head.mkString("_"))
+                       val leafRef = WRef(path.head.mkString(""))
                        Connect(NoInfo, modRef, leafRef)
                    }
                    case _ =>  {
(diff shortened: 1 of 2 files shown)


src/main/scala/firrtl/transforms/TopWiring.scala
```diff
@@ -251,19 +251,21 @@ class TopWiringTransform extends Transform {
...
-    val namespacemap = state.circuit.modules.map{ case m => (m.name -> Namespace(m)) }.toMap
-    val modulesx = state.circuit.modules map onModule(sources, portnamesmap, instgraph, namespacemap)
-    val newCircuit = state.circuit.copy(modules = modulesx)
-    val fixedCircuit = fixupCircuit(newCircuit)
-    val mappings = sources(state.circuit.main).zipWithIndex
-    //Generate output files based on the mapping.
-    outputTuples.map { case (dir, outputfunction) => outputfunction(dir, mappings, state) }
-    // fin.
-    state.copy(circuit = fixedCircuit)
+    if (!(sources.isEmpty)) {
```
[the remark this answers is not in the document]
seldridge: This condition check is needed! Good catch. Can you add a test that exercises this? Something like:
```scala
  "TopWiringTransform" should "do nothing if run without TopWiring* annotations" in {
    val input = """|circuit Top :
                   |  module Top :
                   |    input foo : UInt<1>""".stripMargin
    val inputFile = {
      val fileName = s"${testDir.getAbsolutePath}/input-no-sources.fir"
      val w = new PrintWriter(fileName)
      w.write(input)
      w.close()
      fileName
    }
    val args = Array(
      "--custom-transforms", "firrtl.transform.TopWiring.TopWiringTransform",
      "--input-file", inputFile,
      "--top-name", "Top",
      "--compiler", "low",
      "--info-mode", "ignore"
    )
    firrtl.Driver.execute(args) match {
      case FirrtlExecutionSuccess(_, emitted) => parse(emitted) should be (parse(input))
      case _ => fail
    }
  }
```

Tangentially, this likely motivates an overarching meta-test of all `Transform`s. Any custom transform given a trivial circuit should be able to run over that circuit and not barf. Entirely out of scope of this PR, though.

src/main/scala/firrtl/transforms/TopWiring.scala
```diff
@@ -251,19 +251,21 @@ class TopWiringTransform extends Transform {
...
+      val portnamesmap : mutable.Map[String,String] = mutable.Map()
```
seldridge: [nit] `portnamesmap:`

src/main/scala/firrtl/transforms/TopWiring.scala
```diff
@@ -251,19 +251,21 @@ class TopWiringTransform extends Transform {
...
+      val instgraph = new firrtl.analyses.InstanceGraph(state.circuit)
+      val namespacemap = state.circuit.modules.map{ case m => (m.name -> Namespace(m)) }.toMap
+      val modulesx = state.circuit.modules map onModule(sources, portnamesmap, instgraph, namespacemap)
+      val newCircuit = state.circuit.copy(modules = modulesx)
+      val fixedCircuit = fixupCircuit(newCircuit)
+      val mappings = sources(state.circuit.main).zipWithIndex
+      //Generate output files based on the mapping.
```
seldridge: [nit] remove

src/main/scala/firrtl/transforms/TopWiring.scala
```diff
@@ -251,19 +251,21 @@ class TopWiringTransform extends Transform {
...
+      outputTuples.map { case (dir, outputfunction) => outputfunction(dir, mappings, state) }
+      // fin.
```
seldridge: [nit] remove

alonamid:
Added the test and addressed some of the cosmetic and ordering comments

PR #892, opened 2018-09-18
Firesim - Another TopWiring Bug Fix (Multi-Level Annotations)

alonamid:
Backport to the firesim branch, same as https://github.com/freechipsproject/firrtl/pull/889
Another bug discovered in TopWiring as part Midas testing.
When different levels of the circuit were annotated, the TopWiring signals of the lower levels would "run-over" the TopWiring signals of the higher levels.
One line bug fix, and I added a test that exposes the bug.


davidbiancolin:
Merged because this is a feature branch and blocking development.

Issue #890, opened 2018-09-18
Specification syntax examples rejected by current parser implementation.


ben-marshall:
**Type of issue**: bug report

  * **If possible, please directly provide the FIRRTL file or a relevant fragment if the file is large**
```
circuit myCircuit:                        
    module myCircuit:                     
        input  clk: Clock                 
        input  k  : UInt<1>               
        output c  : UInt<1>               
        when k : (c <= k) else : (c <= k) 
```

  * **What is the current behavior?**
```
ben@benm-lt:~/tools/firrtl/syntax-tests$ ../firrtl/utils/bin/firrtl  -i circuit2.fir -o output.v
line 6:17 mismatched input '(' expecting {':', '[', '.'}
Exception in thread "main" firrtl.SyntaxErrorsException: 1 syntax error(s) detected
	at firrtl.Parser$.$anonfun$parseCharStream$1(Parser.scala:45)
	at firrtl.Utils$.time(Utils.scala:182)
	at firrtl.Parser$.parseCharStream(Parser.scala:33)
	at firrtl.Parser$.parseFile(Parser.scala:25)
	at firrtl.Driver$.$anonfun$getCircuit$5(Driver.scala:199)
	at scala.Option.getOrElse(Option.scala:121)
	at firrtl.Driver$.$anonfun$getCircuit$3(Driver.scala:182)
	at scala.Option.getOrElse(Option.scala:121)
	at firrtl.Driver$.$anonfun$getCircuit$1(Driver.scala:182)
	at scala.util.Try$.apply(Try.scala:209)
	at firrtl.Driver$.getCircuit(Driver.scala:170)
	at firrtl.Driver$.$anonfun$execute$1(Driver.scala:226)
	at logger.Logger$.$anonfun$makeScope$2(Logger.scala:138)
	at scala.util.DynamicVariable.withValue(DynamicVariable.scala:58)
	at logger.Logger$.makeScope(Logger.scala:136)
	at firrtl.Driver$.execute(Driver.scala:223)
	at firrtl.Driver$.execute(Driver.scala:301)
	at firrtl.Driver$.main(Driver.scala:317)
	at firrtl.Driver.main(Driver.scala)

```

  * **What is the expected behavior?**

According to the current specification document in the repository, line 6 of the example is valid syntax. See section 12, page 51 - "Details about syntax" where the spec says that
```
when c:
    a <= b
else:
    c <= d
    e <= f
```
> can be equivalently expressed on a single line as follows.
```
[...]

PR #894, opened 2018-09-20
Number all code examples & add specification build to Makefile

ben-marshall:
This PR is the first step of work identified in issue #890 - namely that the specification is out-of-date with respect to what the current FIRRTL implementation accepts as valid input.

- Added `spec/Makefile` which uses `pdflatex` to build the specification document in place.
- Added a top-level `Makefile` target: `specification`
- Added `spec/Makefile` `clean` target to the top level `Makefile` `clean` target.
- Changed all code example `verbatim` environments into `lstlisting` environments to allow for automatic numbering. This is present in the compiled `spec.pdf` file to be merged.
- Added basic keyword hilighting for all code examples.

This commit allows all code examples to be referred too numerically and unambiguously. This is essential in any subsequent work relating to updating the specification of FIRRTL syntax to match the current implementation.

Side note 1: Going forward is it sensible to keep the `spec/spec.pdf` file under version control? Would it be better as some sort of artifact in a CI pipeline like the other Travis stuff?

Side note 2: I hope this PR is useful as a standalone. If I (or anyone else for that matter) try and make any further updates to the spec, do the maintainers prefer more smaller pull requests or fewer larger ones? What's the best / easiest way to contribute?

Thanks,
Ben


jackkoenig:
Hello and thank you for your contribution! In general smaller contributions are better and I think you've done a great job of that here.

In general I think this is good. Being able to refer to code examples is very important. I think this represents a great starting point for future improvements!

> Going forward is it sensible to keep the spec/spec.pdf file under version control? Would it be better as some sort of artifact in a CI pipeline like the other Travis stuff?

I think it probably would be better not no longer keep the `.pdf` file under version control. I also wonder if we maybe shouldn't just move the spec to another repo entirely and then host the pdf version on a website. (perhaps https://chisel.eecs.berkeley.edu/documentation.html, soon to be renamed to chisel-lang.org)

I have a few other thoughts/questions as well:

1. I'm happy to merge this as is, but should we perhaps make a `spec-0.3` branch where we can collaborate on a new spec that we can then merge more changes to master in one go?
1. As mentioned above, should we make a new repo for the spec?

ben-marshall:
>Hello and thank you for your contribution! In general smaller contributions are better

I'll continue this way in the future then! :)

> I think it probably would be better not no longer keep the .pdf file under version control. I also wonder if we maybe shouldn't just move the spec to another repo entirely and then host the pdf version on a website

That sounds like a good idea. Splitting specification from implementation is important because you don't want the implementation *becoming* the spec, especially for something like this. I think it really helps people separate them conceptually as well, even it one has to be a bit more careful synchronizing changes across the two.

> I'm happy to merge this as is, but should we perhaps make a spec-0.3 branch 

Okay. Given this is essentially a non-invasive change that doesn't affect any syntax or semantics, this PR can be merged, and then all future work wrt the spec I/we do can be on a new branch you create? I think you're right to create a new branch, otherwise the master branch spec (in this or some new repo) will be changing unhelpfully frequently.

> As mentioned above, should we make a new repo for the spec?

For what it's worth, I'd say that's a good idea for maintaining the spec long term.

azidar:
Re keeping the pdf in the repo - I agree its best to build it off of master and host it on the website. I included it in the repo because it is convenient to point to, so as long as we have a serviceable replacement, I'm game. Thanks for the PR!

aswaterman:
After much grief, I managed to get the RISC-V ISA spec repo to auto-generate a new PDF on every merge, without polluting the repo with the PDF. The scheme isn’t perfect, because it automatically creates a new unique tag on every merge, so it’s best to separate the doc into its own repo. But it’s working well for us, so you might consider it here, too.

jackkoenig:
Alright I'm going to merge this as it's a nice first step and then look into spinning the spec off to a different repo, thanks!

jackkoenig:
@ben-marshall I see these are both clean commits, do you want to rebase your fork to remove the master merge commit or should I just squash and merge this?

ben-marshall:
Squash and merge is fine with me.

PR #898, opened 2018-09-25
Enforce port uniqueness in Chirrtl/High Checks

seldridge:
This adds checks inside `CheckHighForm` and `CheckChirrtl` that all ports are uniquely named. This fixes a very old TODO.

This came up when trying to use `CheckHighForm` and `CheckChirrtl` to verify that inlining and Verilog renaming was not introducing port conflicts.

### DIFF
--- a/src/test/scala/firrtlTests/CheckSpec.scala
+++ b/src/test/scala/firrtlTests/CheckSpec.scala
@@ -282,4 +282,32 @@ class CheckSpec extends FlatSpec with Matchers {
       }
     }
   }
+
+  behavior of "Uniqueness"
+  for ((description, input) <- CheckSpec.nonUniqueExamples) {
+    it should s"be asserted for $description" in {
+      assertThrows[CheckHighForm.NotUniqueException] {
+        Seq(ToWorkingIR, CheckHighForm).foldLeft(Parser.parse(input)){ case (c, tx) => tx.run(c) }
+      }
+    }
+  }
 }
+
+object CheckSpec {
+  val nonUniqueExamples = List(
+    ("two ports with the same name",
+     """|circuit Top:
+        |  module Top:
+        |    input a: UInt<1>
+        |    input a: UInt<1>""".stripMargin),
+    ("two nodes with the same name",
+     """|circuit Top:
+        |  module Top:
+        |    node a = UInt<1>("h0")
+        |    node a = UInt<1>("h0")""".stripMargin),
+    ("a port and a node with the same name",
+     """|circuit Top:
+        |  module Top:
+        |    input a: UInt<1>
+        |    node a = UInt<1>("h0") """.stripMargin) )
+  }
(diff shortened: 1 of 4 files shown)


src/test/scala/firrtlTests/ChirrtlSpec.scala
```diff
@@ -71,6 +71,15 @@ class ChirrtlSpec extends FirrtlFlatSpec {
       }
     }
   }
+
+  behavior of "Uniqueness"
+  CheckSpec.nonUniqueExamples.map { case (description, input) =>
```
jackkoenig: Super nitpicky but generally should use `foreach` over `map` when you don't care about return value.

I also stylistically prefer using `for` loops for stuff like this:
```scala
for ((description, input) <- CheckSpec.nonUniqueExamples)  { ...
```
Neither of these matter very much though

src/test/scala/firrtlTests/CheckSpec.scala
```diff
@@ -282,4 +282,29 @@ class CheckSpec extends FlatSpec with Matchers {
...
+      assertThrows[CheckHighForm.NotUniqueException] {
+        Seq(ToWorkingIR, CheckHighForm).foldLeft(Parser.parse(input)){ case (c, tx) => tx.run(c) }
+      }
+    }
+  }
+}
+
+object CheckSpec {
+  val nonUniqueExamples =
+    Map("two ports with the same name"         -> """|circuit Top:
```
jackkoenig: Nit, I would just make this a `List`, with 3 entries the iteration order will always be the same but if there were more than 4 it would be indeterminate.

src/test/scala/firrtlTests/ChirrtlSpec.scala
seldridge: Usually I'm pretty good with `foreach` vs. `map`. Thanks for catching that.

Stylistically, I think this makes sense for a `for` loop. There's no real structure to unpack. Do you have any rough guidelines on what is "stuff like this" vs. what isn't "stuff like this" (my own curiosity). My default is always use `map` with the only exception being when multiple generators would clearly make things easier (outer products). 

Nits (and super-nits) are totally fine. The code should be in as good a shape as we can make it. Tests have a tendency to wind up as a dumping ground of bad practices due to their write once nature. This conflicts with their utility as examples for users. Making new additions there especially good code seems critical to me.

src/test/scala/firrtlTests/CheckSpec.scala
seldridge: Makes sense to me. I'm not using any `Map` features. `List` seems right.

src/test/scala/firrtlTests/ChirrtlSpec.scala
jackkoenig: Yeah the "stuff like this" is super fuzzy. I think generally things that call for a multi-line `foreach`, I generally prefer `for`.  It's kind of nice to have the keyword first, but I don't really have a hard and fast rule.

Yeah, I think the "tests as examples" is something we need to keep in mind more.

PR #900, opened 2018-09-27
Add Utils.expandPrefixes as Prefix Unique helper

seldridge:
This adds a utility, expandPrefixes, that expands a string into all
possible prefixes based on a delimiter. Any repeated occurrence of the
delimiter is viewed as a contributing to a prefix. E.g., "foo_bar" expands
to Seq("foo_", "foo_bar"). This is useful for inlining and keyword
mangling on LowForm. You would like to be able to generate a new name that
is _prefix unique_ with respect to a namespace.

The definition of _prefix uniqueness_ is included in the spec in Section 10.2.

The intended use of this is to expand all prefixes in a `Namespace` and then use something like `Uniquify.findValidPrefix` to generate a new prefix-unique name.

### DIFF
--- a/src/main/scala/firrtl/Utils.scala
+++ b/src/main/scala/firrtl/Utils.scala
@@ -9,6 +9,7 @@ import firrtl.WrappedExpression._
 import firrtl.WrappedType._
 import scala.collection.mutable
 import scala.collection.mutable.{StringBuilder, ArrayBuffer, LinkedHashMap, HashMap, HashSet}
+import scala.util.matching.Regex
 import java.io.PrintWriter
 import logger.LazyLogging
 
@@ -699,11 +700,11 @@ object Utils extends LazyLogging {
     * @return the signal name and any prefixes
     */
   def expandPrefixes(name: String, prefixDelim: String = "_"): Seq[String] = {
-    val regex = s"$prefixDelim[A-Za-z\\$$]".r
+    val regex = ("(" + Regex.quote(prefixDelim) + ")+[A-Za-z0-9$]").r
 
     name +: regex
       .findAllMatchIn(name)
-      .map(_.start + 1)
+      .map(_.end - 1)
       .toSeq
       .foldLeft(Seq[String]()){ case (seq, id) => seq :+ name.splitAt(id)._1 }
   }
--- a/src/main/scala/firrtl/Utils.scala
+++ b/src/main/scala/firrtl/Utils.scala
@@ -691,6 +691,22 @@ object Utils extends LazyLogging {
     "SYNTHESIS",
     "PRINTF_COND",
     "VCS")
+
+  /** Expand a name into its prefixes, e.g., 'foo_bar__baz' becomes 'Seq[foo_, foo_bar__, foo_bar__baz]'. This can be used
+    * to produce better names when generating prefix unique names.
+    * @param name a signal name
+    * @param prefixDelim a prefix delimiter (default is "_")
+    * @return the signal name and any prefixes
+    */
+  def expandPrefixes(name: String, prefixDelim: String = "_"): Seq[String] = {
+    val regex = s"$prefixDelim[A-Za-z\\$$]".r
+
+    name +: regex
+      .findAllMatchIn(name)
+      .map(_.start + 1)
+      .toSeq
+      .foldLeft(Seq[String]()){ case (seq, id) => seq :+ name.splitAt(id)._1 }
+  }
 }
 
 object MemoizedHash {
--- a/src/test/scala/firrtlTests/UtilsSpec.scala
+++ b/src/test/scala/firrtlTests/UtilsSpec.scala
@@ -10,18 +10,17 @@ class UtilsSpec extends FlatSpec {
   behavior of "Utils.expandPrefix"
 
   val expandPrefixTests = List(
-    ("expand an unprefixed name",
-     "foo", Set("foo")),
-    ("expand a name ending in a prefix",
-     "foo__", Set("foo__")),
-    ("expand a name with a prefix",
-     "foo_bar", Set("foo_bar", "foo_")),
-    ("expand a complex name",
-     "foo__$ba9_X__$$$$$_", Set("foo__$ba9_X__$$$$$_", "foo__$ba9_X__", "foo__$ba9_", "foo__")),
-    ("expand a name starting with a delimiter",
-     "__foo_bar", Set("__", "__foo_", "__foo_bar")))
+    ("return a name without prefixes", "_", "foo", Set("foo")),
+    ("expand a name ending with prefixes", "_", "foo__", Set("foo__")),
+    ("expand a name with on prefix", "_", "foo_bar", Set("foo_bar", "foo_")),
+    ("expand a name with complex prefixes", "_",
+     "foo__$ba9_9X__$$$$$_", Set("foo__$ba9_9X__$$$$$_", "foo__$ba9_9X__", "foo__$ba9_", "foo__")),
+    ("expand a name starting with a delimiter", "_", "__foo_bar", Set("__", "__foo_", "__foo_bar")),
+    ("expand a name with a $ delimiter", "$", "foo$bar$$$baz", Set("foo$", "foo$bar$$$", "foo$bar$$$baz")),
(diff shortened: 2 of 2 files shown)


src/main/scala/firrtl/Utils.scala
```diff
@@ -691,6 +691,22 @@ object Utils extends LazyLogging {
...
     "VCS")
+
+  /** Expand a name into its prefixes, e.g., 'foo_bar__baz' becomes 'Seq[foo_, foo_bar__, foo_bar__baz]'. This can be used
+    * to produce better names when generating prefix unique names.
+    * @param name a signal name
+    * @param prefixDelim a prefix delimiter (default is "_")
+    * @return the signal name and any prefixes
+    */
+  def expandPrefixes(name: String, prefixDelim: String = "_"): Seq[String] = {
+    val regex = s"$prefixDelim[A-Za-z\\$$]".r
```
edwardcwang: Is there ever a case where we need to escape prefixDelim? e.g. if I set `prefixDelim = ?`

seldridge: This is a very good question...

Valid FIRRTL names, as far as I understand, are the same as Verilog, i.e., `[A-Za-z_][$A-Za-z_0-9]*`. However, if the user is trying to be clever they can have an identity transform that only computes some metadata and wants to track information using flattening, lowering, etc., but do it with a non-standard identifier. This is, with high probability, an anti-pattern, but it's allowable.

Due to what ClockList is doing, this should at least support `$`. If it works for that, then it should work for any input, I expect. _This currently doesn't split correctly on `$`..._

Do you have any thoughts on how this could work?

The `prefixDelim` should be, reasonably, any `String` that will unconditionally be viewed as a `String`. The user shouldn't be passing in regex. It's possible that this could be limited to a `Char`.

edwardcwang: It's fine if we don't support arbitrary delimiters here, but then we should document this very clearly since it could lead to surprises otherwise.

src/main/scala/firrtl/Utils.scala
```diff
@@ -691,6 +691,22 @@ object Utils extends LazyLogging {
...
+
+    name +: regex
+      .findAllMatchIn(name)
+      .map(_.start + 1)
+      .toSeq
+      .foldLeft(Seq[String]()){ case (seq, id) => seq :+ name.splitAt(id)._1 }
```
jackkoenig: I don't want to prematurely optimize so we should measure this in its actual use, but I suspect this might be kind of a performance problem if used a lot, regex are not the fastest and it has multiple iterations (and appends to, admittedly short, Strings)

jackkoenig: Can you explain why this is necessary? Are you intending to re-infer the prefixes that were already flattened by LowerTypes for use in an application of the Uniquify algorithm in Inline?

If that's the case, I'm not sure if I think that's right approach but I could be convinced otherwise...

seldridge: This worried me, too. Worst case scenario would be flattening a full Rocket Chip. There are two options here: 
1. I understand the performance and make sure it's not an issue
2. The use of this can be turned off for inlining and Verilog renaming

Re: (2) prefix uniqueness is more of a nice feature to have than it is a mandatory requirement. It does make the resulting low FIRRTL / Verilog better for humans and preserves bundles/modules. Genus doesn't care, though.

Follow-up, un-filed PRs for inlining and Verilog keyword renaming do (currently) use this.

seldridge: Yeah, for Inline, [I'm computing a single Inline prefix that is unique with respect to both the namespace and all discovered prefixes](https://github.com/freechipsproject/firrtl/compare/master...seldridge:fix-inlining?expand=1#diff-12726516d9e2fa3ce2f1abf8291bf810R166). Clarifying with an example below.

Here, I want to flatten `A`:
```
circuit Top:
  module Top:
    inst a of A
    node a_foo
  module A:
    node bar
```

If I inline `A` only respecting `Top`'s namespace, then I get the following:
```
circuit Top:
  module Top:
    node a_bar
    node a_foo
```

If I respect `Top`'s namespace _and prefixes_ then I get the following:
```
circuit Top:
  module Top:
    node a__bar
    node a_foo
```

This isn't at all necessary, but it occurred to me that this is qualitatively "nicer". Obvious to me now, but if we really cared about actually preserving this information then it should be an emitted `Annotation`. This was more about trying to choose the correct prefix.

What do you think?

src/main/scala/firrtl/Utils.scala
seldridge: I wasn't aware of `Regex.quote`. That took care of it. Also, I updated this to treat multiple copies of the delimiter as contributing to one prefix. This would allow for shenanigans with likely unique delimiters, e.g., "FOOBAR" could be a delimiter. I'll hold off on merging pending the discussion below.

src/main/scala/firrtl/Utils.scala
jackkoenig: I see, yeah I definitely think it's worth experimenting with since if it makes the emitted Verilog nicer and more clear that's obviously a good thing.

One thing I was wondering with the renewed interest in improving and updating the spec, should algorithms like this one and the one used in Uniquify be part of the spec or is that more of an implementation kind of thing? At first I thought this algorithm was a bit complicated but in reality is isn't so if we choose to use it perhaps it should be defined in the spec?

seldridge:
Sounds good, Jack. I've reworked Inline and VerilogRename to not use this. Prefix uniqueness will not be guaranteed until the performance implications are understood (in addition to the whether or not that's actually needed).

src/main/scala/firrtl/Utils.scala
jackkoenig: For posterity, I came up with a quick alternative implementation (that I did not benchmark) that avoids the regex but relies on the delimiter being a character:

```scala
def expandPrefixes(name: String, prefixDelim: Char = '_'): List[String] = {
  val parts = name.foldLeft((List(new StringBuilder), false)) {
    case ((acc, prev), char) => 
      if (prev && char != prefixDelim) {
        (new StringBuilder(char.toString) :: acc, false)
      } else {
        (acc.head + char :: acc.tail, char == prefixDelim)
      }
  }._1
  parts.reverse.scanLeft("")(_ + _).tail
}
```

Still has at least 3 traversals (foldLeft, reverse, scanLeft) so may still be slow

seldridge: Yeah, I think all that should probably be in the spec. This only occurred to me now, but the spec should (I think?) contain all the information necessary to build a FIRRTL compiler that produces the same output that this repository does. I'm aware of at least one other FIRRTL compiler out there, but ambiguities in the spec may actually be super critical.

Section 10 pushes close to defining some of this stuff this as it discusses the name expansion algorithm and explicitly states that `$` is supposed to be used. (So, this is out of date.)

There are two statements that are a bit odd in Section 10 that we'll likely need to hash out with @azidar:

> All declarations within a module must have unique names. __Furthermore, the set of component declarations within a module must be prefix unique__.

Digging into prefix uniqueness, the spec currently says:

> After the lowering transformation, the names of the lowered circuit components are guaranteed by the name expansion algorithm and thus can be reliably referenced by users to pair meta-data or other annotations with named circuit components.

This seems to imply that aggregate type information (which extends to Inlined modules) must be preserved across transformations. I'm not sure if existing passes are respecting this currently, but this seems to imply that prefix uniqueness is supposed to be guaranteed by any lowering (and that assumedly extends to mangling).

Nonetheless, this part of the spec is pretty old and annotations kind of changes all this as they get renamed (by nice passes/transforms) along with their components. I think I'm largely in favor of the prefix uniqueness guarantee as it seems elegant.

Great approach! Using a `Char` should be fine. By posterity, you mean a PR next week? Errr... I'd accept this right now, actually. This looks good (and what I was too lazy to write). I can do it if you want.

jackkoenig: Yeah the section on prefix uniqueness is just flat out of date. It was checked long ago but hasn't been for ages so needs to be updated very badly.

For posterity I just meant we can try the merged implementation, if it's slow we can try the above. Even the above can be optimized to remove the reverse so only have 2 traversals.

PR #901, opened 2018-09-27
Use "_" for Inline Name Mangling, Respect Namespaces

seldridge:
### Summary of changes
  - Inline pass now uses `_` as a delimiter. Previously, `$` was used.
  - When determining an Inline prefix, the prefix must be unique with respect to the enclosing module's Namespace

### Overview

This does not change the underlying algorithm of Inline. The only major change is that a returned flattened instance is converted to a bundle with `Uniquify.enumerateNames`. The prefix is then generated for that bundle with respect to the enclosing Module's namespace (using `Uniquify.findValidPrefix`). Additionally, module changes are tracked via a `RenameMap` as opposed to assuming that one prefix will work.

Note: this does not inline with "prefix unique" names. This is somewhat expensive to do as it requires expanding all names in the namespace out to include their prefixes. There's a discussion of the limitations associated with this via a method that PR #900 introduces in [this thread](https://github.com/freechipsproject/firrtl/pull/900#pullrequestreview-159260352).

#### Small changes
  - The two used `Uniquify` methods are made `private [firrtl]`
  - `findValidPrefix` wants a `HashSet` as input. I expose this with a new method, `cloneUnderlying`, added to `Namespace`

#### Miscellaneous
  - Chisel tests are disabled for this PR as the existing InlineSpec will fail (due to explicit checks that `$` is used)
  - https://github.com/freechipsproject/chisel3/pull/898 gets Chisel tests passing

### DIFF
--- a/src/main/scala/firrtl/passes/Inline.scala
+++ b/src/main/scala/firrtl/passes/Inline.scala
@@ -6,6 +6,7 @@ package passes
 import firrtl.ir._
 import firrtl.Mappers._
 import firrtl.annotations._
+import firrtl.analyses.InstanceGraph
 
 // Datastructures
 import scala.collection.mutable
@@ -15,13 +16,14 @@ case class InlineAnnotation(target: Named) extends SingleTargetAnnotation[Named]
   def duplicate(n: Named) = InlineAnnotation(n)
 }
 
-// Only use on legal Firrtl. Specifically, the restriction of
-//  instance loops must have been checked, or else this pass can
-//  infinitely recurse
+/** Inline instances as indicated by existing [[InlineAnnotation]]s
+  * @note Only use on legal Firrtl. Specifically, the restriction of instance loops must have been checked, or else this
+  * pass can infinitely recurse.
+  */
 class InlineInstances extends Transform {
    def inputForm = LowForm
    def outputForm = LowForm
-   val inlineDelim = "$"
+   private [firrtl] val inlineDelim: String = "_"
 
    private def collectAnns(circuit: Circuit, anns: Iterable[Annotation]): (Set[ModuleName], Set[ComponentName]) =
      anns.foldLeft(Set.empty[ModuleName], Set.empty[ComponentName]) {
@@ -52,7 +54,7 @@ class InlineInstances extends Transform {
    // 3) All annotated instances exist, and their modules can be inline
    def check(c: Circuit, moduleNames: Set[ModuleName], instanceNames: Set[ComponentName]): Unit = {
       val errors = mutable.ArrayBuffer[PassException]()
-      val moduleMap = (for(m <- c.modules) yield m.name -> m).toMap
+      val moduleMap = new InstanceGraph(c).moduleMap
       def checkExists(name: String): Unit =
          if (!moduleMap.contains(name))
             errors += new PassException(s"Annotated module does not exist: $name")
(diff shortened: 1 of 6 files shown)


src/main/scala/firrtl/passes/Inline.scala
```diff
@@ -15,13 +16,14 @@ case class InlineAnnotation(target: Named) extends SingleTargetAnnotation[Named]
   def duplicate(n: Named) = InlineAnnotation(n)
 }
 
-// Only use on legal Firrtl. Specifically, the restriction of
-//  instance loops must have been checked, or else this pass can
-//  infinitely recurse
+/** Inline instances as indicated by existing [[InlineAnnotation]]s
+  * @note Only use on legal Firrtl. Specifically, the restriction of instance loops must have been checked, or else this
```
chick: Just curious if a requirement like this could be enforced by having the necessary check marked with an annotation

seldridge:
In terms of implications:
  - Anybody using inlining/flattening will now see `_` in their output Verilog as opposed to `$`. People relying on explicit signal names will have to update. 
  - Anybody who wrote a transform that relies on counting `$` in a flattened design will break. 

I expect that all of these are exceedingly rare cases.

@chick: Since this breaks Chisel temporarily, what's the correct strategy to merge this? I'm assuming that I merge this and then merge the companion PR once Jenkins pulls in upstream FIRRTL. Jenkins does pull in upstream FIRRTL, yes? While I'm doing this, all FIRRTL PRs will break as they'll be trying to use upstream Chisel. After everything is updated, this will then requires developers to use `publishLocal` verions of Chisel and FIRRTL upstream going forward until 3.2. I assume this is normal and okay?

@jackkoenig: Are you okay with this without prefix uniqueness? I was thinking that would be a follow-up PR once I look at the performance implications, if any, on a large design like Rocket Chip.

Pending these questions, I'll intend to get this in and then quickly merge https://github.com/freechipsproject/chisel3/pull/898 (which is already approved). I have a follow-up ready to go for FIRRTL that will fix Verilog keyword name mangling.

ucbjrl:
I believe if you include the string `[skip chisel tests]` in the commit message, travis will bypass the chisel tests.

Jenkins does build the FIRRTL master branch as part of its chisel testing, so this may temporarily cause chisel tests to fail (depending on how quickly you can commit the chisel companion PR). This does happen from time to time, so it's not unexpected.

We can specify specific branches/PRs/SHA1s for all of the BIG6 repositories when triggering a manual Jenkins build. I would like to do this (before starting the chain of commits to masters) to verify that we do have a consistent global state.

seldridge:
@ucbjrl: Great. Thanks for the pointer to `[skip chisel tests]`. I'm currently using that to get this PR to pass.

The specific PRs that should be tested together with the BIG6 (freechips is starting to sound more and more like a college football conference...) are:
  - [__F901__](https://github.com/freechipsproject/firrtl/pull/901)
  - [__C898__](https://github.com/freechipsproject/chisel3/pull/898)
  
These are both rebased onto master, approved, and ready to go. I've tested them locally for compatibility, but have not done a local BIG6 test.

A BIG6 test on Jenkins before serial master commits makes sense to me. Good suggestion.

Is the manual Jenkins build with branch pointers something that I can initiate (with similar tags that Jenkins will understand), or is that something that you will have to do?

ucbjrl:
You need to have Jenkins credentials (be able to log in) to run a manual Jenkins build. We don't have an anonymous public interface for this.

ucbjrl:
These passed the Jenkins manual build-the-world.

seldridge:
Thanks a lot @ucbjrl. I'll take care of the serial merge later today.

PR #903, opened 2018-10-01
add BlackBoxPathAnno

albertchen-sifive:
- add `BlackBoxPathAnno` to allow specifying locations of Verilog source files with relative or absolute paths
- add tests for checking that `BlackBoxSourceHelper` copies files

### DIFF
--- a/src/main/scala/firrtl/transforms/BlackBoxSourceHelper.scala
+++ b/src/main/scala/firrtl/transforms/BlackBoxSourceHelper.scala
@@ -82,8 +82,8 @@ class BlackBoxSourceHelper extends firrtl.Transform {
         val fromFile = new File(path)
         val toFile = new File(targetDir, fileName)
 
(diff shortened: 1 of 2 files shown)


jackkoenig:
@chick Is there any reason why only resources were supported originally?

chick:
@jackkoenig No, I don't think so it was probably an oversight

jackkoenig:
Is this still marked pending for anyone else? Clicking Details it seems the Travis tests have passed but why is it not reporting back?

seldridge:
@jackkoenig: It's pending on my end, too, even though it has passed. [Travis status](https://www.traviscistatus.com/) and [GitHub status](https://status.github.com/messages) aren't reporting anything anomalous. Maybe kick _just one_ of the Travis tests?

jackkoenig:
Alright I kicked one

PR #905, opened 2018-10-01
Remove type parameter to Evaluator (0.2.8 compatibility).

ucbjrl:
Don't merge this until we've converted to mill 0.2.8


ucbjrl:
Merged (locally) into mill branch since we're updating to mill 0.3.5.

PR #909, opened 2018-10-05
Verilog renaming uses "_", works on whole AST

### LINKED ISSUE #729: Firrtl generates verilog with odd variable names (trailing '$')

seldridge:
Fixes #729 

This fixes two problems with the current `passes.VerilogRename`:
  1. For #729 , some Verilog parsers (user written or otherwise) may have problems with names containing `$`. The current rename strategy is to blindly append `$` to any keyword that conflicts with a set of Verilog keywords. This changes that to use `_` for name disambiguation.
  2. The existing rename strategy would not rename with respect to namespaces. This only worked before because names with `$` are exceedingly rare.

To address these, this PR generalizes `passes.VerilogRename` as `transforms.RemoveKeywordCollisions` that is parameterized by a set of keywords. The new `transforms.VerilogRename` is just passing in the set of Verilog keywords. This will naturally extend to, e.g, a VHDL renamer.

When renaming with respect to a namespace, renaming of ports is problematic. These have to be renamed inside a module's namespace, but the new port names have to be propagated upwards to all parents instantiating a module with renamed ports. To accomplish this, this uses the usual traversal strategy for this problem: it traverses modules in reverse topological order (from leafs to root) communicating information about what is renamed via `RenameMap`s. This allows for the renames to be propagated in the standard way out of a `Transform`.

This is API breaking since it changes the behavior of output Verilog and since `VerilogRename` is now an object.

I'll merge this myself once it's ready as I'm intending to rebase first.

### Original Commit Message
Summary of changes to firrtl.passes.VerilogRename:
  - Use `_` to mangle names that conflict with Verilog
    keywords (previously `$`)
  - Bugfix to prevent namespace collisions
  - Rewrite to operate on the whole AST to propogate mangled ports and
    module names
  - Make VerilogRename a Transform (was previously a Pass)
  - Renames are now propagated
  - Adds documentation for new VerilogRename

This makes the VerilogRename Transform (previously a Pass) use an
underscore (`_`) instead of a dollar sign (`$`) to mangle names that
conflict with Verilog keywords. This prevents problems with potentially
buggy tools that are not expecting `$` in Verilog names.

This reimplements VerilogRename to be safe for name collisions that may
occur anywhere in the AST, e.g., in ports, module names, circuit names, or
in any statements/expressions. Previously, names were only mangled in
statements and in place. This resulted in problems where renames of ports
in a child's namespace would not be guaranteed to be mangled the same way
in a parent's namespace. The algorithm is reimplemented to walk all
modules in reverse topological order (from leafs to main) and relying on a
RenameMap to track name changes.


src/main/scala/firrtl/passes/Passes.scala
```diff
@@ -247,23 +249,204 @@ object Legalize extends Pass {
...
+      case Some(com: ComponentName) => ComponentName(s"${com.name}.$name", com.module)
+    }
+
+    val named = wrap(n, scope)
+
+    // If this has already been renamed use that name. If it conflicts with a keyword, determine a new, safe name and
+    // update the renames. Otherwise, leave it alone.
+    val namedx: Seq[Named] = renames.get(named) match {
+      case Some(x) => x
+      case None if (v_keywords(n)) =>
```
chick: Unnecessary parens

src/main/scala/firrtl/passes/Passes.scala
```diff
@@ -247,23 +249,204 @@ object Legalize extends Pass {
...
+        val sn = wrap(safeName(n, ns), scope)
+        renames.rename(named, sn)
+        Seq(sn)
+      case _ => Seq(wrap(n, scope))
+    }
+
+    namedx match {
+      case (ComponentName(n, _) :: Nil) => n
```
chick: Unnecessary parens, 3 times

src/main/scala/firrtl/passes/Passes.scala
```diff
@@ -247,23 +249,204 @@ object Legalize extends Pass {
...
+    * $implicitRename
+    * $implicitNamespace
+    * $implicitScope
+    * @return a [[Type]] with updated names
+    * @note This is not intended for fixing arbitrary types, only [[BundleType]] in instance [[WRef]]s
+    */
+  private def onType(t: Type)
+                    (implicit renames: RenameMap, ns: Namespace, scope: Option[ModuleName]): Type = t match {
+    case b: BundleType => b.copy(fields = b.fields.map(f => f.copy(name = onName(f.name))))
+    case t => t
```
chick: suspicious shadowing, this is a trivial case but I believe firrtl convention is to go `tt => tt` here, or `_ => t`

src/main/scala/firrtl/passes/Passes.scala
```diff
@@ -247,23 +249,204 @@ object Legalize extends Pass {
...
+  }
+
+  /** Rename an [[Expression]] to respect existing renames and avoid keyword collisions
+    * @param e the [[Expression]] to rename
+    * $implicitRename
+    * $implicitNamespace
+    * $implicitScope
+    * @return an [[Expression] without keyword conflicts
```
chick: Unclosed scala doc tag

src/main/scala/firrtl/passes/Passes.scala
```diff
@@ -247,23 +249,204 @@ object Legalize extends Pass {
...
+                          iToM: mutable.Map[ComponentName, ModuleName]): Statement = s match {
+    case wdi: WDefInstance =>
+      val subModule = ModuleName(wdi.module, scope.get.circuit)
+      val modulex = renames.get(subModule).orElse(Some(Seq(subModule))).get.head.name
+      val wdix = wdi.copy(module = modulex,
+                          name = onName(wdi.name),
+                          tpe = onType(wdi.tpe)(renames, ns, Some(ModuleName(modulex, scope.get.circuit))))
+      iToM(ComponentName(wdi.name, scope.get)) = ModuleName(wdix.module, scope.get.circuit)
+      wdix
+    case s => s
```
chick: another shadowing case

src/main/scala/firrtl/passes/Passes.scala
```diff
@@ -247,23 +249,204 @@ object Legalize extends Pass {
...
+
+    // Reorder the renamed modules into the original circuit order.
+    val modulesxx: Seq[DefModule] = c.modules.flatMap{ orig =>
+      val named = ModuleName(orig.name, scope.get)
+      modulesx(renames.get(named).orElse(Some(Seq(named))).get.head)
+    }
+
+    // Rename the circuit if the top module was renamed
+    val mainx = renames.get(ModuleName(c.main, CircuitName(c.main))) match {
+      case Some((ModuleName(m, _) :: Nil)) =>
```
chick: Unnecessary parens

src/main/scala/firrtl/passes/Passes.scala
```diff
@@ -247,23 +249,204 @@ object Legalize extends Pass {
...
+    p.copy(name = onName(p.name))
+
+  /** Rename a [[DefModule]] and it's internals (ports and statements) to fix keyword collisions and update instance
+    * references to respect previous renames
+    * @param renames a [[RenameMap]]
+    * @param circuit the enclosing [[CircuitName]]
+    * @return a [[DefModule]] without keyword conflicts
+    */
+  private def onModule(renames: RenameMap, circuit: CircuitName)(m: DefModule): DefModule = {
+    implicit val moduleNamespace = Namespace(m)
```
chick: IntelliJ feels that these lines here should have explicit types

src/main/scala/firrtl/passes/Passes.scala
```diff
@@ -247,23 +249,204 @@ object Legalize extends Pass {
...
+      .map(onName(_: String)(renames, moduleNamespace, Some(circuit)))
+  }
+
+  /** Fix any Verilog keyword collisions in a [[Circuit]]
+    * @param c a [[Circuit]] with possible name collisions
+    * @param renames a [[RenameMap]] to update. If you don't want to propagate renames, this can be ignored.
+    * @return a [[Circuit]] without keyword conflicts
+    */
+  def run(c: Circuit, renames: RenameMap = RenameMap()): Circuit = {
+    implicit val circuitNamespace = Namespace(c)
```
chick: More explicit typing recommended

chick:
I'd lean towards making it a new file in this PR but this issue hasn't come up that much. Maybe @jackkoenig or @azidar has an opinion

src/main/scala/firrtl/passes/Passes.scala
seldridge: I'm less enthusiastic about the wills of our IntelliJ overlords... I wonder why IntelliJ wants this?

Did some digging... found one thread about this. We could turn this off in IntelliJ, but this may be mandatory in Dotty. I'll change these. Type annotations are usually always good. https://blog.jetbrains.com/scala/2016/10/05/beyond-code-style/#comment-198824

chick: Thanks for the link, That's a decent discussion of the issues.

src/main/scala/firrtl/transforms/RemoveKeywordCollisions.scala
```diff
@@ -0,0 +1,217 @@
...
+    // on subfields where only the local instance name is available.
+    implicit val iToM: mutable.Map[ComponentName, ModuleName] = mutable.Map.empty
+
+    m
+      .map(onPort)
+      .map(onStatement)
+      .map(onName(_: String)(renames, moduleNamespace, Some(circuit)))
+  }
+
+  /** Fix any Verilog keyword collisions in a [[Circuit]]
```
chick: FYI sbt doc complains that [[Circuit]] get erros Could not find any member to link for "Circuit", I think this problem occurs throughout project.

jackkoenig: I think it's because `firrtl.ir` is imported rather than `firrtl.ir._`, so you can either change it to `[[ir.Circuit]]` or switch to importing all of `firrtl.ir._`

seldridge:
Thanks @chick. I'll clean up the scaladoc and make sure the links work before merging. Haven't had a chance to get back to this... It may be a little while.

chick:
Do you think we should merge this and do the cleanup's in a later PR. I hate to see approved PR's languish and get harder to merge later. Maybe @jackkoenig or @azidar can take a quick look and give final vet.

seldridge:
Let me give this a quick cleanup and merge it. It's functionally a-okay and it fixes a real bug.

src/main/scala/firrtl/transforms/RemoveKeywordCollisions.scala
```diff
@@ -0,0 +1,217 @@
...
+import firrtl.passes.{Uniquify, PassException}
+import firrtl.Utils.v_keywords
+import firrtl.Mappers._
+import scala.collection.mutable
+
+/** Transform that removes collisions with reserved keywords
+  * @param keywords a set of reserved words
+  * @define implicitRename @param renames the [[RenameMap]] to query when renaming
+  * @define implicitNamespace @param ns an encolosing [[Namespace]] with which new names must not conflict
+  * @define implicitScope @param scope the enclosing scope of this name. If [[None]], then this is a [[Circuit]] name
```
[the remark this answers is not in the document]
seldridge: As far as I'm aware, these `@define`s only are made available for later use. However, I'm only using them for `private` members. This is potentially over-documentation on my part, but it avoids later boilerplate and I assumed that I could turn on documentation of private members for a local copy if I wanted to do so (or if a developer wanted to do so).

PR #919, opened 2018-10-24
[F764.6] Add, but do not use Options-mirroring Annotations

seldridge:
This adds annotations for all FIRRTL command line options.

Depends on #879 (this includes commits from that PR)

Merge strategy: @seldridge will merge after rebase

### DIFF
--- a/src/main/scala/firrtl/Driver.scala
+++ b/src/main/scala/firrtl/Driver.scala
@@ -17,6 +17,7 @@ import firrtl.annotations.AnnotationYamlProtocol._
 import firrtl.passes.{PassException, PassExceptions}
 import firrtl.transforms._
 import firrtl.Utils.throwInternalError
+import firrtl.stage.TargetDirAnnotation
 
 
 /**
--- a/src/main/scala/firrtl/ExecutionOptionsManager.scala
+++ b/src/main/scala/firrtl/ExecutionOptionsManager.scala
@@ -68,9 +68,6 @@ case class CommonOptions(
   }
 }
 
-/** Annotation that contains the [[CommonOptions]] target directory */
-case class TargetDirAnnotation(value: String) extends NoTargetAnnotation
-
 trait HasCommonOptions {
   self: ExecutionOptionsManager =>
   var commonOptions = CommonOptions()
--- a/src/main/scala/firrtl/stage/FirrtlAnnotations.scala
+++ b/src/main/scala/firrtl/stage/FirrtlAnnotations.scala
@@ -1,0 +1,290 @@
+// See LICENSE for license details
+
+package firrtl.stage
+
+import firrtl._
+import firrtl.ir.Circuit
+import firrtl.annotations.NoTargetAnnotation
+import firrtl.transforms.BlackBoxTargetDirAnno
+import firrtl.options.HasScoptOptions
+
+import logger.LogLevel
+
+import scopt.OptionParser
+
+/** Indicates that a subclass is an [[firrtl.annotations Annotation]] that includes a command line option
+  *
+  * This must be mixed into a subclass of [[annotations.Annotation]]
+  */
+sealed trait FirrtlOption extends HasScoptOptions
+
+/** Holds the name of the top module
+  *  - set on the command line with `-tn/--top-name`
+  * @param value top module name
+  */
+case class TopNameAnnotation(topName: String) extends NoTargetAnnotation with FirrtlOption {
+  def addOptions(p: OptionParser[AnnotationSeq]): Unit = p.opt[String]("top-name")
+    .abbr("tn")
+    .valueName("<top-level-circuit-name>")
+    .action( (x, c) => c :+ TopNameAnnotation(x) )
+    .unbounded() // See [Note 1]
+    .text("This options defines the top level circuit, defaults to dut when possible")
+}
+
+object TopNameAnnotation {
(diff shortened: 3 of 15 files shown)


src/main/scala/firrtl/stage/FirrtlAnnotations.scala
```diff
@@ -0,0 +1,304 @@
...
+
+import firrtl.ir.Circuit
+import firrtl.annotations.NoTargetAnnotation
+import firrtl.transforms.BlackBoxTargetDirAnno
+import firrtl.options.HasScoptOptions
+import logger.LogLevel
+import scopt.OptionParser
+
+/** Indicates that a subclass is an [[annotations.Annotation]] with an
+  * option consummable by [[HasFirrtlExecutionOptions]]
```
chick: I think HasFirrtlExecutionOptions is no longer relevant

src/main/scala/firrtl/options/Stage.scala
```diff
@@ -0,0 +1,61 @@
+// See LICENSE for license details.
```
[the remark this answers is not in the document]
seldridge: Cool! @azidar had that comment a while back, too. "Stage" also seems to be the correct term in compiler literature, e.g., "three stage compiler".

FWIW, all the names behind `firrtl.options` and `firrtl.stage` are capable of being bikeshed if we can think of clearer names, e.g., we're free to change `firrtl.options.ExecutionOptionsManager` to something else.

src/main/scala/firrtl/Driver.scala
```diff
@@ -17,6 +17,7 @@ import firrtl.annotations.AnnotationYamlProtocol._
 import firrtl.passes.{PassException, PassExceptions}
 import firrtl.transforms._
 import firrtl.Utils.throwInternalError
+import firrtl.stage.TargetDirAnnotation
```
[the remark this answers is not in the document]
seldridge: This one is actually used further down.

src/main/scala/firrtl/stage/FirrtlAnnotations.scala
seldridge: You're right.

src/main/scala/firrtl/options/ExecutionOptionsManager.scala
```diff
@@ -0,0 +1,63 @@
...
+
+/** Indicate an error in [[firrtl.options]]
+  * @param msg a message to print
+  */
+case class OptionsException(msg: String, cause: Throwable = null) extends Exception(msg, cause)
+
+/** A utility for working with command line options
+  * @param applicationName the application associated with these command line options
+  */
+class ExecutionOptionsManager(val applicationName: String) {
```
chick: @seldridge You suggested we might change this name and the more I think about it the more I agree. I think it will be confusing to have two of them, the name has always been a mouthful and there's some question about what execution means to me too. Do you have any good names up your sleeve for this (after your triumphant selection of the word Stage). It's not a deal breaker if you'd like to keep it in order to get this stuff over, but I vote for changing it now. Let me know.

seldridge: Yeah, this name is clearly broken (and should probably be fixed in [F764.5](https://github.com/freechipsproject/firrtl/pull/879)). In the follow-on commit, [F764.7](https://github.com/freechipsproject/firrtl/pull/920) I define `FirrtlCli` and `LoggerCli` which are self-typed as `ExecutionOptionsManager`. It made sense back when this was a store of execution options, but this patch series moves the options into the annotatinos.

So, what is this doing now?
1. Defines the name of the program presented to the user via a CLI
2. Defines a parser
3. Registers options
4. Defines a `parse` method to convert `Array[String] -> AnnotationSeq` using the parser

With that in mind, how about:
- `CommandLineInterface`
- `CLI`
- `Console`
- `Shell`

`CommandLineInterface` is clean as it describes what's going on, but I kind of like the shorter `Shell`. Maybe open it up for some quality developer bikeshedding.

seldridge: Settled on `Shell` and Updated F764.5 to use that.

src/main/scala/firrtl/Emitter.scala
```diff
@@ -19,6 +19,7 @@ import firrtl.PrimOps._
 import firrtl.WrappedExpression._
 import Utils._
 import MemPortUtils.{memPortField, memType}
+import firrtl.stage.EmitOneFilePerModuleAnnotation
```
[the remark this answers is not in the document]
seldridge: Fixed in the rebase.

src/main/scala/firrtl/ExecutionOptionsManager.scala
```diff
@@ -9,6 +9,7 @@ import firrtl.passes.memlib.{InferReadWriteAnnotation, ReplSeqMemAnnotation}
 import firrtl.passes.clocklist.ClockListAnnotation
 import logger.LogLevel
 import scopt.OptionParser
+import firrtl.stage.EmitOneFilePerModuleAnnotation
```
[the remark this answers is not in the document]
seldridge: Fixed in rebase.

seldridge:
Ping @chick. I verified that the nits were cleared and rebased this onto master. Ready for re-review.

Issue #793, opened 2018-04-20
Don't Emit Emitter Annotations nor Deleted Annotations by default


jackkoenig:
Emitter annotations are just massive and redundant. I also think Deleted Annotations clutter everything up, perhaps there could be an option?

* **Type of issue**
  - [ ] Bug report
  - [x] Feature request
  - [ ] Other enhancement

* **What is the use case for changing the behavior?**

Make looking at the emitted annotations actually useful, especially if trying to chain invocations of firrtl together.

* **Impact**
  - [ ] no functional change
  - [ ] API addition (no impact on existing code)
  - [x] API modification (I doubt anyone is relying on this but 🤷‍♀️)
  - [ ] unknown

* **Development Phase**
  - [x] request
  - [ ] proposal

seldridge:
@jackkoenig, I think I'm in agreement with this. Likely, there should be some flag (annotation) that turns this on or off.

Related to this, `DeletedAnnotation`s aren't even emitted correctly. I was doing some tests and realized that they can't be deserialized after serialization (due to them not being serialized correctly). E.g., the following `DeletedAnnotation` should include the `class` of the deleted annotation, but it doesn't. Fix should be trivial, but likely a case for just preventing emission.
```json
  {
    "class":"firrtl.annotations.DeletedAnnotation",
    "xFormName":"AddCircuit$",
    "anno":{
      "file":"test_run_dir/FirrtlStageSpec/implict-annotation-file/Top.fir"
    }
  }
```

PR #920, opened 2018-10-24
[F764.7] Switch to FIRRTL Stage, Driver is compatibility layer

### LINKED ISSUE #485: Funny Behavior when topName isn't specified

### LINKED ISSUE #764: [RFC] Annotations subsume ExecutionOptions?

seldridge:
### Metadata

- @seldridge will rebase and merge

### Overview

This implements a refactor of FIRRTL to use Annotations as a universal source of knowledge. This provides a replacement for both the `Driver` and the `ExecutionOptionsManager` framework. 

This provides:
  - `FirrtlStage`, a replacement for `Driver`
  - A refactor of the `Driver` to act as a compatibility layer around `FirrtlStage`

### Details

#879 introduces `Stage` and #945 introduce `Phase`. A `Phase` is a class that performs a mathematical transformation of an `AnnotationSeq`. In effect, a generalization of FIRRTL's `Transform`. A `Stage` is a `Phase` with a user interface. 

#### `FirrtlStage`

This PR adds `FirrtlStage` which provides a concrete `Stage` that runs the FIRRTL compiler and the chosen emitter. Internally, `FirrtlStage` is broken up into a number of preprocessing phases that transform input annotations in the following steps:
1. Addition of default annotations
2. Validation checking
3. Conversion of compiler name to emitter name
4. Expansion of any input-like annotations in full circuits
5. Addition of an implicit output file if no output file was specified

Differing from previous attempts at an Annotations/Options refactor, all preprocessing work is broken out into these phases. The extraction of information that the `FirrtlStage` needs is done with a lightweight `view[FirrtlOptions]`. This in (aforementioned) previous attempts did much more work (adding defaults, etc.). This is all removed here making the `view` very basic.

Old logic (running the compiler, emitting the circuit) that used to live in `Driver` was hoisted out and put into `FirrtlStage`. This could be targeted by this PR, but is likely out of scope and will be tackled in a subsequent PR.

#### `Driver` as Compatibility Layer

The `Driver` remains largely intact as a shim around `FirrtlStage`. This uses some different `DriverCompatibility` phases for preprocessing that differ from `FirrtlStage`:
1. Try-hard top name determination
2. Implicit annotation file inclusion
3. Implicit FIRRTL file (based on top name)

The logic surrounding these is inordinately complex and likely confusing to a user, but is left in to align with the old `Driver`.

#### Summary

- Fixes #764 
- Fixes #485 
- Uses #793 strategy for annotation pruning
- Depends on #879 and #919  (this includes commits from that PR)

### Benchmarking

I think the performance and memory usage is under control now. This results in a small increase in runtime and a (sometimes) larger increase in memory footprint.
- Memory Increas: -1% up to 10%
- Performance Slowdown: -1% up to 2.5%

### API Changes for FIRRTL Stage

- This changes the FIRRTL fat jar helper script to use the new FIRRTL stage
- When using the FIRRTL stage, one or more emitters can be specified (`-E/--emit-circuit` or `-e/--emit-modules` for whole circuit or one-file-per-module). However, if no emitter is specified, no emitter runs. __This is a change that requires a PR on Rocket Chip.__
- String arguments are now case sensitive and mandate all lowercase, e.g., the compiler is `verilog` not `Verilog`

TODO:

- [x] ~`FirrtlStageResult` should extend `DriverExecutionResult`?~ (This isn't how things work. `DriverExecutionResult` was removed. There's no real need for this.)
- [x] Remove `Driver.dramatic*` stuff in favor of `FirrtlStage.dramatic*`
- [x] `FirrtlStage` Comment on `emittedRes` doesn't make sense (line ~65)
- [x] Should we massage things in the old Driver to fixup the annotations?
  - Yes, this is done via some `Phase`s in `firrtl.stage.phases.DriverCompatibility`
- [x] ~Switch to `view(AnnotationSeq): T` and `viewOption(AnnotationSeq): Option[T]`~
  - Supplanted by #945. The `view` method is `AnnotationSeq => T`.
- [x] Remove `//scalastyle:ignore` in Logger `view`
- [x] Test that both `firrtl.Driver` and `firrtl.stage.FirrtlStage` work with Rocket Chip
- [x] Rocket Chip performance test
  - ~I'm seeing a 4x slowdown~ Caching the `hashCode` of big [Annotation]s fixes this
  - ~Strip `CircuitOption` annotations in `FirrtlStage`~ This shouldn't be necessary
  - [x] `DefaultConfig` is the same
  - [x] @jackkoenig Rocket Chip performance suite
- [x] Fix compatibility problems in `FirrtlStage`
  - [x] `--info-mode` isn't specified
- [x] Remove `--split-modules` in favor of explicit user specification of an emitter to use. I'm not sure on the API here, but `-E/-e` may work for a normal emitter and a one-file-per-module emitter. This would then remove the implicit emitter that `-X` provides. This would be API breaking if we swing the FIRRTL fat jar to use FIRRTL stage.

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -19,6 +19,9 @@ import firrtl.PrimOps._
 import firrtl.WrappedExpression._
 import Utils._
 import MemPortUtils.{memPortField, memType}
+import firrtl.options.{HasScoptOptions, StageUtils, PhaseException}
+import firrtl.stage.RunFirrtlTransformAnnotation
+import scopt.OptionParser
 // Datastructures
 import scala.collection.mutable.{ArrayBuffer, LinkedHashMap, HashSet}
 
@@ -31,17 +34,70 @@ sealed trait EmitAnnotation extends NoTargetAnnotation {
 case class EmitCircuitAnnotation(emitter: Class[_ <: Emitter]) extends EmitAnnotation
 case class EmitAllModulesAnnotation(emitter: Class[_ <: Emitter]) extends EmitAnnotation
 
+object EmitCircuitAnnotation extends HasScoptOptions {
+  def addOptions(p: OptionParser[AnnotationSeq]): Unit = p
+    .opt[String]("emit-circuit")
+    .abbr("E")
+    .valueName("<chirrtl|high|middle|low|verilog|sverilog>")
+    .unbounded()
+    .action{ (x, c) =>
+      val xx = x match {
+        case "chirrtl"  => Seq(RunFirrtlTransformAnnotation(new ChirrtlEmitter),
+                               EmitCircuitAnnotation(classOf[ChirrtlEmitter]))
+        case "high"     => Seq(RunFirrtlTransformAnnotation(new HighFirrtlEmitter),
+                               EmitCircuitAnnotation(classOf[HighFirrtlEmitter]))
+        case "middle"   => Seq(RunFirrtlTransformAnnotation(new MiddleFirrtlEmitter),
+                               EmitCircuitAnnotation(classOf[MiddleFirrtlEmitter]))
+        case "low"      => Seq(RunFirrtlTransformAnnotation(new LowFirrtlEmitter),
+                               EmitCircuitAnnotation(classOf[LowFirrtlEmitter]))
+        case "verilog"  => Seq(RunFirrtlTransformAnnotation(new VerilogEmitter),
+                               EmitCircuitAnnotation(classOf[VerilogEmitter]))
+        case "sverilog" => Seq(RunFirrtlTransformAnnotation(new SystemVerilogEmitter),
+                               EmitCircuitAnnotation(classOf[SystemVerilogEmitter]))
+        case _          => throw new PhaseException(s"Unknown emitter '$x'! (Did you misspell it?)")
+      }
+      xx ++ c }
+    .text("Run the specified circuit emitter (all modules in one file)")
+}
+
+object EmitAllModulesAnnotation extends HasScoptOptions {
+  def addOptions(p: OptionParser[AnnotationSeq]): Unit = p
+    .opt[String]("emit-modules")
+    .abbr("e")
+    .valueName("<none|high|middle|low|verilog|sverilog>")
+    .unbounded()
+    .action{ (x, c) =>
+      val xx = x match {
+        case "chirrtl"  => Seq(RunFirrtlTransformAnnotation(new ChirrtlEmitter),
+                               EmitAllModulesAnnotation(classOf[ChirrtlEmitter]))
+        case "high"     => Seq(RunFirrtlTransformAnnotation(new HighFirrtlEmitter),
+                               EmitAllModulesAnnotation(classOf[HighFirrtlEmitter]))
+        case "middle"   => Seq(RunFirrtlTransformAnnotation(new MiddleFirrtlEmitter),
+                               EmitAllModulesAnnotation(classOf[MiddleFirrtlEmitter]))
+        case "low"      => Seq(RunFirrtlTransformAnnotation(new LowFirrtlEmitter),
+                               EmitAllModulesAnnotation(classOf[LowFirrtlEmitter]))
+        case "verilog"  => Seq(RunFirrtlTransformAnnotation(new VerilogEmitter),
+                               EmitAllModulesAnnotation(classOf[VerilogEmitter]))
+        case "sverilog" => Seq(RunFirrtlTransformAnnotation(new SystemVerilogEmitter),
+                               EmitAllModulesAnnotation(classOf[SystemVerilogEmitter]))
+        case _          => throw new PhaseException(s"Unknown emitter '$x'! (Did you misspell it?)")
+      }
+      xx ++ c }
+    .text("Run the specified module emitter (one file per module)")
+}
+
 // ***** Annotations for results of emission *****
 sealed abstract class EmittedComponent {
   def name: String
   def value: String
+  def outputSuffix: String
 }
 sealed abstract class EmittedCircuit extends EmittedComponent
-final case class EmittedFirrtlCircuit(name: String, value: String) extends EmittedCircuit
-final case class EmittedVerilogCircuit(name: String, value: String) extends EmittedCircuit
+final case class EmittedFirrtlCircuit(name: String, value: String, outputSuffix: String) extends EmittedCircuit
+final case class EmittedVerilogCircuit(name: String, value: String, outputSuffix: String) extends EmittedCircuit
 sealed abstract class EmittedModule extends EmittedComponent
-final case class EmittedFirrtlModule(name: String, value: String) extends EmittedModule
-final case class EmittedVerilogModule(name: String, value: String) extends EmittedModule
+final case class EmittedFirrtlModule(name: String, value: String, outputSuffix: String) extends EmittedModule
+final case class EmittedVerilogModule(name: String, value: String, outputSuffix: String) extends EmittedModule
 
 /** Traits for Annotations containing emitted components */
 sealed trait EmittedAnnotation[T <: EmittedComponent] extends NoTargetAnnotation {
@@ -63,6 +119,8 @@ case class EmittedVerilogModuleAnnotation(value: EmittedVerilogModule)
 sealed abstract class FirrtlEmitter(form: CircuitForm) extends Transform with Emitter {
   def inputForm = form
   def outputForm = form
+
+  val outputSuffix: String = form.outputSuffix
 
   private def emitAllModules(circuit: Circuit): Seq[EmittedFirrtlModule] = {
     // For a given module, returns a Seq of all modules instantited inside of it
@@ -87,15 +145,15 @@ sealed abstract class FirrtlEmitter(form: CircuitForm) extends Transform with Em
         case ext: ExtModule => ext
       }
       val newCircuit = Circuit(m.info, extModules :+ m, m.name)
-      EmittedFirrtlModule(m.name, newCircuit.serialize)
+      EmittedFirrtlModule(m.name, newCircuit.serialize, outputSuffix)
     }
   }
 
   override def execute(state: CircuitState): CircuitState = {
     val newAnnos = state.annotations.flatMap {
       case EmitCircuitAnnotation(_) =>
-        Seq(EmittedFirrtlCircuitAnnotation.apply(
-              EmittedFirrtlCircuit(state.circuit.main, state.circuit.serialize)))
+        Seq(EmittedFirrtlCircuitAnnotation(
+              EmittedFirrtlCircuit(state.circuit.main, state.circuit.serialize, outputSuffix)))
       case EmitAllModulesAnnotation(_) =>
         emitAllModules(state.circuit) map (EmittedFirrtlModuleAnnotation(_))
       case _ => Seq()
@@ -129,6 +187,7 @@ case class VRandom(width: BigInt) extends Expression {
 class VerilogEmitter extends SeqTransform with Emitter {
   def inputForm = LowForm
   def outputForm = LowForm
+  val outputSuffix = ".v"
   val tab = "  "
   def AND(e1: WrappedExpression, e2: WrappedExpression): Expression = {
     if (e1 == e2) e1.e1
@@ -859,7 +918,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
       case EmitCircuitAnnotation(_) =>
         val writer = new java.io.StringWriter
         emit(state, writer)
-        Seq(EmittedVerilogCircuitAnnotation(EmittedVerilogCircuit(state.circuit.main, writer.toString)))
+        Seq(EmittedVerilogCircuitAnnotation(EmittedVerilogCircuit(state.circuit.main, writer.toString, outputSuffix)))
 
       case EmitAllModulesAnnotation(_) =>
         val circuit = runTransforms(state).circuit
@@ -870,16 +929,22 @@ class VerilogEmitter extends SeqTransform with Emitter {
             val writer = new java.io.StringWriter
             val renderer = new VerilogRender(d, pds, module, moduleMap)(writer)
             renderer.emit_verilog()
-            Some(EmittedVerilogModuleAnnotation(EmittedVerilogModule(module.name, writer.toString)))
+            Some(EmittedVerilogModuleAnnotation(EmittedVerilogModule(module.name, writer.toString, outputSuffix)))
           case module: Module =>
             val writer = new java.io.StringWriter
             val renderer = new VerilogRender(module, moduleMap)(writer)
             renderer.emit_verilog()
-            Some(EmittedVerilogModuleAnnotation(EmittedVerilogModule(module.name, writer.toString)))
+            Some(EmittedVerilogModuleAnnotation(EmittedVerilogModule(module.name, writer.toString, outputSuffix)))
           case _ => None
         }
       case _ => Seq()
     }
     state.copy(annotations = newAnnos ++ state.annotations)
   }
+}
+
+class SystemVerilogEmitter extends VerilogEmitter {
+  StageUtils.dramaticWarning("SystemVerilog Emitter is the same as the Verilog Emitter!")
+
+  override val outputSuffix: String = ".sv"
 }
--- a/src/main/scala/firrtl/ExecutionOptionsManager.scala
+++ b/src/main/scala/firrtl/ExecutionOptionsManager.scala
@@ -8,7 +8,13 @@ import firrtl.ir.Circuit
 import firrtl.passes.memlib.{InferReadWriteAnnotation, ReplSeqMemAnnotation}
 import firrtl.passes.clocklist.ClockListAnnotation
 import logger.LogLevel
+import logger.{ClassLogLevelAnnotation, LogClassNamesAnnotation, LogFileAnnotation, LogLevelAnnotation}
 import scopt.OptionParser
+import firrtl.stage.{CompilerAnnotation, FirrtlCircuitAnnotation, FirrtlFileAnnotation, FirrtlSourceAnnotation,
+  InfoModeAnnotation, OutputFileAnnotation, RunFirrtlTransformAnnotation}
+import firrtl.stage.phases.DriverCompatibility.{TopNameAnnotation, EmitOneFilePerModuleAnnotation}
+import firrtl.options.{InputAnnotationFileAnnotation, OutputAnnotationFileAnnotation, ProgramArgsAnnotation, StageUtils}
+import firrtl.transforms.{DontCheckCombLoopsAnnotation, NoDCEAnnotation}
 
 import scala.collection.Seq
 
@@ -18,8 +24,10 @@ import scala.collection.Seq
   * '''NOTE''' In all derived trait/classes, if you intend on maintaining backwards compatibility,
   *  be sure to add new options at the end of the current ones and don't remove any existing ones.
   */
+@deprecated("Use firrtl.options.HasScoptOptions and/or library/transform registration", "1.2")
 trait ComposableOptions
 
+@deprecated("Use firrtl.options.{ExecutionOptionsManager, TerminateOnExit, DuplicateHandling}", "1.2")
 abstract class HasParser(applicationName: String) {
   final val parser = new OptionParser[Unit](applicationName) {
     var terminateOnExit = true
@@ -48,6 +56,7 @@ abstract class HasParser(applicationName: String) {
   * For example, in chisel, by deferring this it is possible for the execute there to first elaborate the
   * circuit and then set the topName from that if it has not already been set.
   */
+@deprecated("Use a FirrtlOptionsView, LoggerOptionsView, or construct your own view of an AnnotationSeq", "1.2")
 case class CommonOptions(
     topName:           String         = "",
     targetDirName:     String         = ".",
@@ -66,8 +75,17 @@ case class CommonOptions(
       optionsManager.getBuildFileName("log")
     }
   }
+
+  def toAnnotations: AnnotationSeq = (if (topName.nonEmpty) Seq(TopNameAnnotation(topName)) else Seq()) ++
+    (if (targetDirName != ".") Some(TargetDirAnnotation(targetDirName)) else None) ++
+    Some(LogLevelAnnotation(globalLogLevel)) ++
+    (if (logToFile) { Some(LogFileAnnotation(None)) } else { None }) ++
+    (if (logClassNames) { Some(LogClassNamesAnnotation) } else { None }) ++
+    classLogLevels.map{ case (c, v) => ClassLogLevelAnnotation(c, v) } ++
+    programArgs.map( a => ProgramArgsAnnotation(a) )
 }
 
+@deprecated("Specify command line arguments in an Annotation mixing in HasScoptOptions", "1.2")
 trait HasCommonOptions {
   self: ExecutionOptionsManager =>
   var commonOptions = CommonOptions()
@@ -168,6 +186,7 @@ final case class OneFilePerModule(targetDir: String) extends OutputConfig
   * @param compilerName           which compiler to use
   * @param annotations            annotations to pass to compiler
   */
+@deprecated("Use a FirrtlOptionsView or construct your own view of an AnnotationSeq", "1.2")
 case class FirrtlExecutionOptions(
     inputFileNameOverride:  String = "",
     outputFileNameOverride: String = "",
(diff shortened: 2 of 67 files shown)


src/main/scala/firrtl/Emitter.scala
```diff
@@ -96,15 +93,18 @@ sealed abstract class FirrtlEmitter(form: CircuitForm) extends Transform with Em
...
 
   override def execute(state: CircuitState): CircuitState = {
-    val newAnnos = state.annotations.flatMap {
-      case EmitCircuitAnnotation(_) =>
-        Seq(EmittedFirrtlCircuitAnnotation.apply(
-              EmittedFirrtlCircuit(state.circuit.main, state.circuit.serialize)))
-      case EmitAllModulesAnnotation(_) =>
-        emitAllModules(state.circuit) map (EmittedFirrtlModuleAnnotation(_))
-      case _ => Seq()
+    val (e, other) = state.annotations.partition{
```
chick: A comment on what is going on with annotations here would be helpful

src/main/scala/firrtl/Emitter.scala
```diff
@@ -863,13 +863,9 @@ class VerilogEmitter extends SeqTransform with Emitter {
...
 
   override def execute(state: CircuitState): CircuitState = {
-    val newAnnos = state.annotations.flatMap {
-      case EmitCircuitAnnotation(_) =>
-        val writer = new java.io.StringWriter
-        emit(state, writer)
-        Seq(EmittedVerilogCircuitAnnotation(EmittedVerilogCircuit(state.circuit.main, writer.toString)))
-
-      case EmitAllModulesAnnotation(_) =>
+    val oneFilePerModule = state.annotations.collect{ case a: EmitOneFilePerModuleAnnotation.type => a }.nonEmpty
```
chick: collectFirst{...}.isDefined might be faster because it can exit early

src/main/scala/firrtl/stage/phases/Checks.scala
```diff
@@ -0,0 +1,84 @@
...
+import firrtl.stage._
+
+object Checks extends Phase {
+
+  /** Determine if an annotations are sane
+    *
+    * @param annos a sequence of [[annotation.Annotation]]
+    * @return true if all checks pass
+    */
+  def transform(annos: AnnotationSeq): AnnotationSeq = {
```
chick: I like this being a Transform

src/main/scala/firrtl/stage/FirrtlOptions.scala
```diff
@@ -0,0 +1,93 @@
...
+
+import firrtl._
+import firrtl.options.{OptionsException, StageOptions}
+import firrtl.annotations.Annotation
+import firrtl.ir.Circuit
+import logger.LogLevel
+
+/** Internal options used to control the FIRRTL compiler stage
+  *
+  * @param topName the name of the top module
```
chick: First five tags here appear to be fossils

src/main/scala/firrtl/stage/FirrtlOptions.scala
```diff
@@ -0,0 +1,93 @@
...
+  * @param logClassNames indicates logging verbosity on a class-by-class basis
+  * @param programArgs explicit program arguments
+  * @param inputFileNameOverride input FIRRTL file, default: `targetDir/topName.fir`
+  * @param outputFileNameOverride output file, default: `targetDir/topName.SUFFIX` with `SUFFIX`
+  *                               as determined by the compiler
+  * @param outputAnnotationFileName the output annotation filename, default: `targetDir/topName.anno`
+  * @param compilerName which compiler to use (default: "verilog")
+  * @param infoModeName the policy for generating [[firrtl.ir.Info]] when processing FIRRTL (default: "append")
+  * @param customTransforms any custom [[Transform]] to run
+  * @param firrtlSource explicit input FIRRTL as a [[scala.String]]
```
chick: I think this is Predef.String, or just String

src/main/scala/firrtl/stage/FirrtlOptions.scala
```diff
@@ -0,0 +1,93 @@
+// See LICENSE for license details.
+
+package firrtl.stage
+
+import scala.util.Try
```
chick: Unused imports

src/main/scala/firrtl/stage/FirrtlOptions.scala
```diff
@@ -0,0 +1,93 @@
...
+  /** The top (main) in the FIRRTL Circuit
+    * @return the top, if a circuit exists, [[None]] otherwise
+    */
+  def topName: String = firrtlCircuit.map(_.main).getOrElse("a")
+
+  def outputFileName: String = outputFileNameOverride.getOrElse(topName)
+
+  /** Get the user-specified [[OutputConfig]]
+    *
+    * @param optionsManager this is needed to access build function and its common options
```
chick: Bad tag in scaladoc

src/main/scala/firrtl/stage/FirrtlOptions.scala
```diff
@@ -0,0 +1,93 @@
...
+    * @return the output configuration
+    */
+  def getOutputConfig(): OutputConfig =
+    if (emitOneFilePerModule) { OneFilePerModule(stageOptions.targetDir) }
+    else                      { SingleFile(stageOptions.getBuildFileName(outputFileName, Some(outputSuffix))) }
```
chick: outputSuffix is declared as requiring empty parens, this should change or the declaration should

src/main/scala/firrtl/stage/FirrtlStage.scala
```diff
@@ -0,0 +1,105 @@
+// See LICENSE for license details.
+
+package firrtl.stage
+
+import firrtl._
+
+import firrtl.options.{Stage, Phase, StageUtils, Shell, OptionsException, StageOptions}
```
chick: Some un-used imports here

src/main/scala/firrtl/stage/FirrtlStage.scala
```diff
@@ -0,0 +1,105 @@
...
+    catch {
+      // Rethrow the exceptions which are expected or due to the runtime environment (out of memory, stack overflow)
+      case p @ (_: ControlThrowable | _: PassException | _: PassExceptions | _: FIRRTLException) =>
+        throw p
+      // Treat remaining exceptions as internal errors.
+      case e: Exception => throwInternalError(exception = Some(e))
+    }
+
+    // Do emission
+    val emittedRes = firrtlOptions.getOutputConfig match {
```
chick: getOutputConfig should have parens

src/main/scala/firrtl/stage/FirrtlStageUtils.scala
```diff
@@ -0,0 +1,25 @@
+// See LICENSE for license details.
+
+package firrtl.stage
+
+import firrtl._
```
chick: Every input here is unused

src/main/scala/firrtl/stage/package.scala
```diff
@@ -0,0 +1,44 @@
+// See LICENSE for license details.
+
+package firrtl
+
+import firrtl.options.{OptionsView, StageUtils, StageOption, TargetDirAnnotation, InputAnnotationFileAnnotation,
```
chick: Lot's of unused imports here

src/main/scala/firrtl/stage/phases/AddCircuit.scala
```diff
@@ -0,0 +1,55 @@
+package firrtl.stage.phases
```
chick: missing license comment here

src/main/scala/firrtl/stage/phases/AddCircuit.scala
```diff
@@ -0,0 +1,55 @@
...
+import firrtl.options.{OptionsException, Phase}
+
+import java.io.FileNotFoundException
+
+object AddCircuit extends Phase {
+
+  /** Extract the info mode from an [[AnnotationSeq]]
+    * @param annos some annotations
+    * @return the info mode
+    * @note This will only work properly for an [[AnnotationSeq]] that passes [[checkAnnotations]]
```
chick: checkAnnotations here does not seem to be imported, for scaladoc link to work

src/main/scala/firrtl/stage/phases/AddCircuit.scala
```diff
@@ -0,0 +1,55 @@
...
+    */
+  private def infoMode(annos: AnnotationSeq): Parser.InfoMode = {
+    val infoMode = annos.collectFirst{ case InfoModeAnnotation(i) => i }
+    val inputFile = annos.collectFirst{ case InputFileAnnotation(f) => f }
+
+    infoMode.getOrElse { throw new FirrtlStageException("No info mode specified") } match {
```
chick: new is not needed here

src/main/scala/firrtl/stage/phases/AddCircuit.scala
```diff
@@ -0,0 +1,55 @@
...
+  def transform(annos: AnnotationSeq): AnnotationSeq = {
+    val info = infoMode(annos)
+    val phaseName = this.getClass.getName
+    annos.flatMap {
+      case a @ InputFileAnnotation(f) => val circuit = try {
+        FirrtlStageUtils.getFileExtension(f) match {
+          case ProtoBufFile => proto.FromProto.fromFile(f)
+          case FirrtlFile => Parser.parseFile(f, info) }
+      } catch {
+        case _: FileNotFoundException => throw new OptionsException(s"Input file '$f' not found")
```
chick: new not needed here

src/main/scala/firrtl/stage/phases/AddDefaults.scala
```diff
@@ -0,0 +1,34 @@
+package firrtl.stage.phases
```
chick: Missing license comment

src/main/scala/firrtl/stage/phases/AddDefaults.scala
```diff
@@ -0,0 +1,34 @@
...
+object AddDefaults extends Phase {
+
+  /** Append any missing default annotations to an annotation sequence
+    *
+    * @param annos annotation sequence to examine
+    * @return the annotation sequence with default annotations added
+    */
+  def transform(annos: AnnotationSeq): AnnotationSeq = {
+    var bb, c, im = true
+    annos.collect{ case a: FirrtlOption => a }.map{
```
chick: map should be foreach

src/main/scala/firrtl/stage/phases/AddImplicitOutputFile.scala
```diff
@@ -0,0 +1,25 @@
+package firrtl.stage.phases
```
chick: Missing license comment

src/main/scala/firrtl/stage/phases/Checks.scala
```diff
@@ -0,0 +1,84 @@
+package firrtl.stage.phases
```
chick: Missing license comment

src/main/scala/firrtl/stage/phases/DriverCompatibility.scala
```diff
@@ -0,0 +1,102 @@
+package firrtl.stage.phases
```
chick: missing license comment

src/main/scala/firrtl/stage/phases/DriverCompatibility.scala
```diff
@@ -0,0 +1,102 @@
...
+
+import java.io.File
+
+/** Provides compatibility methods to create old Driver behavior where command line options, if missing, are inferred
+  * based on other options.
+  */
+object DriverCompatibility {
+  /** Determine the top name using the following precedence (highest to lowest):
+    *  - Explicitly from the user-specified `--top-name`
+    *  - Implicitly from the top module ([[ir.Circuit.main]]) of [[FirrtlCircuitAnnotation]]
```
chick: should be firrtl.ir.Circuit for all these lines.
HashFirrtlOptions not found at lines 25 and 44

src/main/scala/firrtl/Emitter.scala
seldridge: I've added comments for the three major steps going on here.

src/main/scala/firrtl/Emitter.scala
seldridge: Makes sense to me. Fix pending.

src/main/scala/firrtl/stage/phases/Checks.scala
seldridge: Great! I'm glad this makes sense. And I assume you mean a `Phase` with a `transform` method?

FWIW, #945 grew out of the fact that I was constantly writing methods with the type signature of `AnnotationSeq => AnnotationSeq`. These weren't a `Stage`, but something else. Additionally, the `view` really has no business running checks. The common pattern seems to be that the type you're viewing an `AnnotationSeq` as is a `case class` with default values. Hence, the checks are something transform-like that live outside the `view`.

src/main/scala/firrtl/stage/FirrtlOptions.scala
seldridge: Thanks. I removed the fossil tags and added the missing tags.

src/main/scala/firrtl/stage/FirrtlOptions.scala
seldridge: Neither seems to get unidoc to do the right thing. I'm leaving this off. I don't see this being used in the Scala documentation.

src/main/scala/firrtl/stage/FirrtlOptions.scala
seldridge: Fixed. Thanks for checking on this.

src/main/scala/firrtl/stage/FirrtlOptions.scala
seldridge: Fixed.

src/main/scala/firrtl/stage/FirrtlOptions.scala
seldridge: Apparently, side effect-free methods are supposed to omit parentheses: https://docs.scala-lang.org/style/method-invocation.html. I'll update everything in `FirrtlOptions` to follow that (as nothing in here has side effects).

src/main/scala/firrtl/stage/FirrtlStage.scala
seldridge: Unused imports _below_ this line are cleaned up.

src/main/scala/firrtl/stage/FirrtlStage.scala
seldridge: Leaving this as is since I removed the parens from `getOutputConfig` in `FirrtlOptions`.

src/main/scala/firrtl/stage/FirrtlStageUtils.scala
seldridge: Fixed.

src/main/scala/firrtl/stage/package.scala
seldridge: Cleaned up.

src/main/scala/firrtl/stage/phases/AddCircuit.scala
seldridge: Done.

src/main/scala/firrtl/stage/phases/AddCircuit.scala
seldridge: Should be `Checks`. Fixed.

src/main/scala/firrtl/stage/phases/AddCircuit.scala
seldridge: Fixed.

src/main/scala/firrtl/stage/phases/AddCircuit.scala
seldridge: Fixed.

src/main/scala/firrtl/stage/phases/AddDefaults.scala
seldridge: Fixed.

src/main/scala/firrtl/stage/phases/AddDefaults.scala
seldridge: Fixed.

src/main/scala/firrtl/stage/phases/AddImplicitOutputFile.scala
seldridge: Fixed.

src/main/scala/firrtl/stage/phases/Checks.scala
seldridge: Fixed.

src/main/scala/firrtl/stage/phases/DriverCompatibility.scala
seldridge: Fixed.

src/main/scala/firrtl/stage/phases/DriverCompatibility.scala
seldridge: Fixed up. I made all of `DriverCompatibility` `private [firrtl]` as this really shouldn't be exposed externally.

seldridge:
@chick: Thanks! This is updated based on your review.

@jackkoenig: Would you be able to make sure that this doesn't break Rocket Chip performance-wise? The problem with #950 is cleaned up [here](https://github.com/freechipsproject/firrtl/pull/920/files#diff-0b0f62077e2b0d287c468a62087cae9fR196), but there may be other lurking performance problems. While this adds `firrtl.stage.FirrtlStage`, you should be able to continue using `firrtl.Driver` [like Rocket Chip does](https://github.com/freechipsproject/rocket-chip/blob/master/Makefrag#L24) as a compatibility layer around `firrtl.stage.FirrtlStage` without interruption.

jackkoenig:
With this I see a 15-40% slowdown, you can run my benchmark yourself with the following:

```
git clone git@github.com:jackkoenig/firrtlbench.git
git submodule update --init
cd firrtl
git checkout -b seldridge-f764.7 master
git pull https://github.com/seldridge/firrtl.git f764.7
cd ..
./benchmark.py --designs firrtl/regress/*.fir -N 10 --versions master seldridge-f764.7
```

seldridge:
Thanks for the analysis @jackkoenig. I'll dig into it.

seldridge:
@jackkoenig: Thanks for the pointer to `firrtlbench`. I did some tests with that and things seem to not be blowing up anymore. I don't have big sample sizes (`-N2`), but I'm seeing on `DefaultConfig` Rocket Chip the following normalized data:
  - `Driver` on `HEAD`: 1 memory, 1 time
  - `Driver` (compatibility) on #920: 1.08 memory, 0.97 time
  - `FirrtlStage` on #920: 0.95 memory, 0.88 time

This seems far better that I thought (I was expecting no possibility for performance gain...). I can kick off some longer tests tomorrow and make sure that this aligns with what I'm seeing here.

This includes your idea from #793. It made sense to prune annotations once they're no longer needed, e.g., the `CircuitAnnotation` has no business being passed around (even though it has a cached hash code in case you want to do that). This is only use when using the `FirrtlStage` without a new `--dont-strip` option.

Did you test this with large numbers of annotations or only with `firrtlbench`? I'm trying to make sure that I'm covering all angles...

@chick: Line count explosion due to added tests. Nothing huge has changed relative to when you looked at it last. Do you still want to proceed with a group review? I can work on scheduling a time.

src/main/scala/firrtl/options/Stage.scala
```diff
@@ -45,8 +46,14 @@ abstract class Stage extends Phase {
...
     */
-  final def execute(args: Array[String], annotations: AnnotationSeq): AnnotationSeq =
+  final def execute(args: Array[String], annotations: AnnotationSeq): AnnotationSeq = try {
     transform(shell.parse(args, annotations))
+  } catch {
+    case OptionsHelpException => annotations
+    case OptionsException(msg, _) =>
+      StageUtils.dramaticUsageError(msg)
+      annotations
+  }
```
seldridge: Why is this continuing and not fully erroring?

src/main/scala/firrtl/options/StageAnnotations.scala
```diff
@@ -40,3 +40,36 @@ case class TargetDirAnnotation(dir: String = ".") extends NoTargetAnnotation wit
...
+  def addOptions(p: OptionParser[AnnotationSeq]): Unit = p.arg[String]("<arg>...")
+    .unbounded()
+    .optional()
+    .action( (x, c) => c :+ ProgramArgsAnnotation(x) )
+    .text("optional unbounded args")
+}
+
+object ProgramArgsAnnotation {
+  private [options] def apply(): ProgramArgsAnnotation = ProgramArgsAnnotation("")
+}
```
seldridge: Better way to do this would be to make `HasScoptOptions` mixed into the companion object?

src/main/scala/firrtl/options/Stage.scala
seldridge: This is now updated to call `System.exit(1)` here. This should clean up all of the issues with the FIRRTL fat jar not returning normal exit codes. Tests are added to make sure that the exit codes behave as expected.

src/main/scala/firrtl/options/StageAnnotations.scala
seldridge: @jackkoenig, you're 100% right about this approach. In the updated PR, `HasScoptOptions`, for `Annotation`s, is only mixed into the companion object.

src/main/scala/firrtl/ExecutionOptionsManager.scala
```diff
@@ -284,8 +306,27 @@ extends ComposableOptions {
   def getAnnotationFileName(optionsManager: ExecutionOptionsManager): String = {
     optionsManager.getBuildFileName("anno", annotationFileNameOverride)
   }
+
+  def toAnnotations: AnnotationSeq =
+    (if (inputFileNameOverride.nonEmpty) Seq(FirrtlFileAnnotation(inputFileNameOverride)) else Seq()) ++
+      (if (outputFileNameOverride.nonEmpty) { Some(OutputFileAnnotation(outputFileNameOverride)) } else { None }) ++
+      Some(CompilerNameAnnotation(compilerName)) ++
+      Some(InfoModeAnnotation(infoModeName)) ++
+      // WTF is inferRW doing?
```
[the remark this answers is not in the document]
seldridge: This now shows a `dramaticWarning` if the user tries to run this after setting `inferRW` explicitly in a `FirrtlExecutionOptions`.

seldridge:
@jackkoenig, @chick: This is approaching the point where it's ready to review (again). The line count is reaching comical levels, however... Fortunately, the refactoring of this PR over the past couple of weeks has resulted in this being more separable than it was previously. Is there motivation in me breaking this up? If so, is it preferable to have separate commits or separate PRs?

chick:
That's good news. Personally I would rather that it remain one commit. In terms of approving it, I think a good independent test is how well (and transparently) I can port the concepts, to the other repos in the big6. To do that it was really helpful when you did that yourself with chisel3 and I could apply that model to the remaining repos. What do you think.

seldridge:
I have a working Chisel Stage PR (https://github.com/freechipsproject/chisel3/pull/981) that adds an implementation of Chisel as a Stage, the old Driver as a compatibility layer, and uses the FIRRTL Stage that this PR provide internally. The porting is tedious, but doable. https://github.com/freechipsproject/chisel3/pull/981 includes a very rough guide of how I did this and how I think it should generally be done for other repos.

While I could continue to go back and forth on this, I think this should be ready to review. It's probably best to start with the changes to the `firrtl.options` package first and go from there.

seldridge:
@jackkoenig: I think I have the situation with transform ordering sorted out. If you have a public test I can try that. 

I was reversing the transform order for some premature list prepending optimization. This should now be corrected in [firrtl.stage.phases.Compiler.scala](https://github.com/freechipsproject/firrtl/pull/920/files#diff-ca9001803ac2df62bf647097e44229e6R97). 

This annotation ordering led me to another issue...

The Compiler phase is written such that it will do parallel compilation of separate `FirrtlCircuitAnnotation`s. These annotations can be supplied directly or generated from `FirrtlFileAnnotation`s or `FirrtlSourceAnnotation`s. The simple grammar of how annotations are then grouped into compiler runs is as follows:
1. Any annotations that come before a circuit are "global" annotations
2. Any annotations that come after a circuit are left-associative to the previous circuit

Consequently, options must predictably copy the `AnnotationSeq` to either always prepend or always append. I've currently updated all options to prepend for performance reasons. The `Stage` will then automatically reverse them. (Any initial `AnnotationSeq` is preserved before any parsed options.)

This is all well and good, except that users, defining their own transforms with options can decide to append and then suddenly things are misbehaving... None of this should matter as parallel compilation is disabled via the `firrtl.stage.phases.Checks` phase which disallows more than one `FirrtlCircuitAnnotation`. However, before parallel compilation is enabled for mass consumption, this needs to be sorted out. Likely with an API change to `HasScoptOptions` to be polymorphic in `A` and abstract in `def f(a: A): AnnotationSeq`. Furthermore, I don't like the fact that when users add their own options they are able to muck with the existing `AnnotationSeq`. A good solution here would avoid all of these problems.

Issue #923, opened 2018-10-26
Chisel tests not running


seldridge:
**Type of issue**: bug report

**If the current behavior is a bug, please provide the steps to reproduce the problem:**

Look at the Travis output for a PR. See that Chisel tests don't run because it's incorrectly parsing the full commit log. See: https://travis-ci.org/freechipsproject/firrtl/jobs/445954084#L1538

@jackkoenig and I know about this, but I'm adding an issue to track it.

 **What is the use case for changing the behavior?**

**Impact**: no functional change

**Development Phase**: request

**Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. Stack Overflow, gitter, etc)

seldridge:
Alright, so there's a further problem here. Travis is doing a shallow clone. This consequently breaks any type of `$MY_BRANCH..master` or `$MY_BRANCH...master` type `git log` parsing for a PR with a sufficiently large number of commits. The only way to guarantee that we get around this is to do a full clone (I think).

Alternatively, we can do a `git log --format=%B --no-merges -n1`. This _should_ examine the most recent non-merge commit to see if Chisel tests have to be run. I'm testing this now...

Does this make sense?

seldridge:
More concretely, the problem is in checking out the PR. Once you do that, you wind up with a mostly full history. You have a full history going from your HEAD all the way to the initial commit, but only a partial history for the PR base. You can see the explosion of commits below:

```bash
$ git clone --depth=10 https://github.com/freechipsproject/firrtl.git
$ cd firrtl
$ git fetch origin +refs/pull/926/merge:
$ git log --pretty=oneline | wc -l

16

$ git checkout -qf FETCH_HEAD
$ git log --pretty=oneline | wc -l

1414
```

jackkoenig:
The `git log` command to search for `[skip chisel tests]` is *supposed* to only search the commits in the PR, ie. `TRAVIS_COMMIT_RANGE` should be just the range of commits in the PR. Clearly something is wrong with that but ideally the command should work no matter how much history is included in the repo.

seldridge:
[...]

Issue #922, opened 2018-10-25
Instance Annotations exposed bugs in Chisel StrongEnum and LoadMemory


seldridge:
**Type of issue**: bug report

**If the current behavior is a bug, please provide the steps to reproduce the problem:**

I'm seeing strong enum and load memory failures in FIRRTL with FIRRTL and Chisel master.

To reproduce: run Chisel's tests with FIRRTL 7e2f787 published locally.

Apparently Chisel tests may have not run on this?

  * **If possible, please directly provide the FIRRTL file or a relevant fragment if the file is large**

I'm not sure where the problem is, but this seems to be Target related from Instance Annotations.

  * **What is the current behavior?**

StrongEnumSpec and LoadMemoryFromFileSpec are barfing.

  * **What is the expected behavior?**

These should pass...

  * **Please tell us about your environment:**

FIRRTL: 7e2f787
Chisel: 6004052

**What is the use case for changing the behavior?**

It's broken.

**Impact**: unknown

**Development Phase**: request

(I don't know what's going on here, so it's a request...)

**Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. Stack Overflow, gitter, etc)

StrongEnum failure:
```
[info] StrongEnum FSM
[info] - should work *** FAILED ***
[info]   firrtl.FIRRTLException: Internal Error! Please file an issue at https://github.com/ucb-bar/firrtl/issues
[info]   at firrtl.Utils$.error(Utils.scala:396)
[info]   at firrtl.Utils$.throwInternalError(Utils.scala:178)
[info]   at firrtl.Driver$.$anonfun$execute$1(Driver.scala:251)
[info]   at logger.Logger$.$anonfun$makeScope$2(Logger.scala:138)
[info]   at scala.util.DynamicVariable.withValue(DynamicVariable.scala:58)
[info]   at logger.Logger$.makeScope(Logger.scala:136)
[info]   at firrtl.Driver$.execute(Driver.scala:223)
[info]   at chisel3.testers.TesterDriver$.execute(TesterDriver.scala:54)
[info]   at chiselTests.ChiselRunners.runTester(ChiselSpec.scala:24)
[info]   at chiselTests.ChiselRunners.runTester$(ChiselSpec.scala:23)
[info]   ...
[info]   Cause: java.lang.NumberFormatException: For input string: "value"
[info]   at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
[...]

PR #924, opened 2018-10-26
Fix TRAVIS_COMMIT_RANGE in .run_chisel_tests, replace ... with ..

jackkoenig:
This makes "skip chisel tests" work properly

Fixes https://github.com/freechipsproject/firrtl/issues/923 (I think)


jackkoenig:
This failing is expected due to https://github.com/freechipsproject/firrtl/issues/922, once https://github.com/freechipsproject/firrtl/issues/922 is resolved, we can rerun and it should pass

PR #926, opened 2018-10-29
Instance Annotations, Try No. 2

azidar:
Followup from #925 which was a revert of #865 .

Additionally includes significant performance and memory improvements of the Dedup pass.


seldridge:
giphy

PR #928, opened 2018-10-31
Dont append to lists

jackkoenig:
It is generally ill advised.

This is a scattershot of performance improvements by removing appending to Lists at places I could find them. You too can see these performance problems by trying to compile the following:
```
circuit test :
  module test :
    output out : UInt<1>[10000]
    out is invalid
```

Each commit is fairly small and describes what it does.

On examples like the above with very large Vecs, this helps a ton. On some large designs, this reduces total runtime by ~10%

### DIFF
--- a/src/main/scala/firrtl/Utils.scala
+++ b/src/main/scala/firrtl/Utils.scala
@@ -300,6 +300,7 @@ object Utils extends LazyLogging {
     onExp(expression)
     ReferenceTarget(main, module, Nil, ref, tokens)
   }
+   @deprecated("get_flip is fundamentally slow, use to_flip(gender(expr))", "1.2")
    def get_flip(t: Type, i: Int, f: Orientation): Orientation = {
      if (i >= get_size(t)) throwInternalError(s"get_flip: shouldn't be here - $i >= get_size($t)")
      t match {
--- a/src/main/scala/firrtl/Utils.scala
+++ b/src/main/scala/firrtl/Utils.scala
@@ -256,10 +256,9 @@ object Utils extends LazyLogging {
     case ex: ValidIf => create_exps(ex.value) map (e1 => ValidIf(ex.cond, e1, e1.tpe))
     case ex => ex.tpe match {
       case (_: GroundType) => Seq(ex)
-      case (t: BundleType) => (t.fields foldLeft Seq[Expression]())((exps, f) =>
-        exps ++ create_exps(WSubField(ex, f.name, f.tpe,times(gender(ex), f.flip))))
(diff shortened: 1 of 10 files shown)


src/main/scala/firrtl/Utils.scala
```diff
@@ -301,6 +300,7 @@ object Utils extends LazyLogging {
     onExp(expression)
     ReferenceTarget(main, module, Nil, ref, tokens)
   }
+   @deprecated("get_flip is fundamentally slow, use to_flip(gender(expr))", "1.2")
```
edwardcwang: Is the idea that using `gender(expr)` is faster than `get_flip`?

src/test/scala/firrtlTests/WidthSpec.scala
```diff
@@ -53,6 +53,7 @@ class WidthSpec extends FirrtlFlatSpec {
       ResolveKinds,
       InferTypes,
       CheckTypes,
+      ResolveGenders,
```
azidar: Why do you need to add this now?

src/main/scala/firrtl/passes/Passes.scala
```diff
@@ -125,13 +125,13 @@ object ExpandConnects extends Pass {
           case sx: DefMemory => genders(sx.name) = MALE; sx
           case sx: DefNode => genders(sx.name) = MALE; sx
           case sx: IsInvalid =>
-            val invalids = (create_exps(sx.expr) foldLeft Seq[Statement]())(
-               (invalids,  expx) => gender(set_gender(expx)) match {
-                  case BIGENDER => invalids :+ IsInvalid(sx.info, expx)
-                  case FEMALE => invalids :+ IsInvalid(sx.info, expx)
-                  case _ => invalids
+            val invalids = create_exps(sx.expr).flatMap { case expx =>
```
azidar: Is flatMap fast?

jackkoenig: The answer is yes but now I'm doing some microbenchmarking and foldLeft with List prepend followed by a reverse might actually be faster...

jackkoenig: I think more digging is necessary, but foldLeft with list prepend then a reverse might be a bit faster for simple iteration, but I think flatMap is faster when you're combining Seqs (eg. in create_exps)

seldridge:
Nice use of separate commits. I manually verified that this is passing for each commit with:
```bash
git rebase -i --exec "sbt test" $ORIGIN/master
```

PR #929, opened 2018-11-01
Add Bundle Literals.

grebe:
This is an attempt to add bundle literals to firrtl. I'll make an associated chisel PR.


chick (commented):
This looks pretty good to me but I will defer to @azidar and @jackkoenig 
The Firrtl Spec should be changed with this PR I think.

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -8,6 +8,7 @@ import Utils.indent
 /** Intermediate Representation */
 abstract class FirrtlNode {
   def serialize: String
+  override def toString = serialize
```
azidar: Mmmm I don't think we should do this... ToString is useful to see the in-memory representation, as opposed to the concrete serialization.

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -214,6 +215,57 @@ case class FixedLiteral(value: BigInt, width: Width, point: Width) extends Liter
...
+      case lit: Literal => (s, lit)
+      case _ => throw new Exception("Oh no!")
+    }
+  })
+}
+case class VectorExpression(exprs: Seq[Expression], tpe: Type) extends Expression {
+  // val value = lits.map(x => (x.value, x.width)).foldLeft(BigInt(0)) { case (prev, (v, IntWidth(w))) =>
+  //   (prev << w.toInt) + v
+  // }
+  // val width = lits.map(_.width).reduce(_ + _)
```
azidar: Remove comments?

src/test/scala/firrtlTests/MemSpec.scala
```diff
@@ -11,5 +20,248 @@ class MemSpec extends FirrtlPropSpec {
...
+    val iWriter = new StringWriter()
+    val cWriter = new StringWriter()
+    (new HighFirrtlEmitter).emit(iResult, iWriter)
+    (new HighFirrtlEmitter).emit(cResult, cWriter)
+    val iCircuit = parse(removeJunk(iWriter.toString()))
+    val cCircuit = parse(removeJunk(cWriter.toString()))
+    iCircuit should be (cCircuit)
+  }
+
+  ignore("Writing to mems with a bundle of vector literals should work") {
```
azidar: Should you ignore this test?

azidar (changes requested):
Overall looks fine as a first pass. A few high level requests/comments/questions:

- Why are you naming one BundleLiteral, but the other VectorExpression? Shouldn't it be VectorLiteral?
- Although we discussed only adding literals, I'm leaning towards adding aggregate expressions. While on the one hand, adding literals is a strict subset (and thus backwards compatible to Chisel users) and could theoretically be added first, it isn't backwards compatible with FIRRTL transform writers who may match on VectorLiteral, and then we would later remove that and add a VectorExpression. I think we need to make a decision about aggregate expressions, and if we are adding them, do it in this PR.
- If we do add aggregate expressions, they must be able to express flipped fields. I'm personally ok this this, but others may have worries.
- Any change to the IR must also have an update to the spec.tex (and update spec.pdf)

Thoughts?

grebe:
I have switched to doing bundle and vector expressions. So far it seems to work fine for literals and nesting these expressions.

I'm currently not testing mixed direction bundle expressions. I'm not really sure what the desired behavior is, nor what the use-cases are (the main one that jumps to mine is assigning a decoupled bundle directly). Do mixed direction vectors make sense? I don't think so... but maybe it does. I *think* things should "just work™" in the sense that these are pretty thin wrappers and should be mostly equivalent to the expansion that eventually happens.

PR #931, opened 2018-11-02
Update commandline `sbt publishLocal`

Jeremygu:
With the newest version of SBT, the commandline `sbt publish-local` should be changed to `sbt publishLocal`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ sbt assembly
 ```
 sbt publish-local
 ```
+>Note that if the version of SBT installed is the [1.x](https://github.com/freechipsproject/rocket-chip/issues/1385#issuecomment-385280345), this command `sbt publish-local` should be changed to `sbt publishLocal`.
 
 ##### Useful sbt Tips
  1. Run a single test suite:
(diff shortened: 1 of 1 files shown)


jackkoenig:
+1 for just `sbt publishLocal`, SBT added the camelCase version and had both for a long time, I wonder if FIRRTL even builds with whatever version of SBT it was before camelCase was added.

azidar:
Thanks @Jeremygu ! I agree with @edwardcwang and @jackkoenig , its fine to just change 'publish-local' to 'publicLocal'. If you can fix up the PR, we will merge it.

README.md
```diff
@@ -54,8 +54,9 @@ sbt assembly
     * **Note:** You can add `utils/bin` to your path to call firrtl from other processes
  1. Publish this version locally in order to satisfy other tool chain library dependencies:
 ```
-sbt publish-local
+sbt publishLocal
 ```
+>Note that if the version of SBT installed is the [1.x](https://github.com/freechipsproject/rocket-chip/issues/1385#issuecomment-385280345), this command `sbt publish-local` should be changed to `sbt publishLocal`.
```
edwardcwang: I think we can just remove this line

Jeremygu: Yes, I have removed that line.

PR #932, opened 2018-11-03
Better error message for UninferredWidth exception

### LINKED ISSUE #269: Better error messages for uninferred widths
If you make a reg with no width, and fail to assign to it, CheckWidths throws an exception. Since this is not an internal compiler error, but the result of incorrect user input, we should print a more helpful message. (Sorry, I'm on my phone so can't provide a failing test case.)

seldridge:
This improves `CheckWidths` pass by passing along the `Target` that is currently being examined. This enables a better` UninferredWidth` exception message indicating what `Target` can't be inferred. This improves over the current exception message which only tells you what module couldn't be inferred.

Fixes #269 

Consider the following test case included in this PR:
```
circuit Foo :
  module Foo :
    input clock : Clock
    inst bar of Bar
  module Bar :
    wire a: { b : UInt<1>, c : { d : UInt<1>, e : UInt } }
```

`a.c.e` in `Bar` is uninferrable because it was never assigned to. This now produces the following error message:
```
[info]   firrtl.passes.CheckWidths$UninferredWidth: : Uninferred width for target below. (Did you forget to assign to it?)
[info]     circuit Foo:
[info]     └── module Bar:
[info]         └── a.c.e
```

This is painfully specific. However, it does require the user to understand `Target` serialization syntax. No way around this, though.

Likely, similar exception messages could be improved throughout. I'm attempting to keep this small and only take care of #269 narrowly. I'm open to suggestions on if a bigger PR would be better here.

### Metadata
  - @seldridge will rebase and merge

### DIFF
--- a/src/main/scala/firrtl/annotations/Target.scala
+++ b/src/main/scala/firrtl/annotations/Target.scala
@@ -55,6 +55,32 @@ sealed trait Target extends Named {
     }
   }
 
+  /** Pretty serialization, ideal for error messages. Cannot be deserialized.
+    * @return Human-readable serialization
+    */
+  def prettyPrint(tab: String = ""): String = {
+    val circuitString = s"""${tab}circuit ${circuitOpt.getOrElse("???")}:"""
+    val moduleString = s"""\n$tab└── module ${moduleOpt.getOrElse("???")}:"""
+    var depth = 4
+    val tokenString = tokens.map {
+      case Ref(r) => val rx = s"""\n$tab${" "*depth}└── $r"""; depth += 4; rx
+      case Instance(i) => val ix = s"""\n$tab${" "*depth}└── inst $i """; ix
+      case OfModule(o) => val ox = s"of $o:"; depth += 4; ox
+      case Field(f) => s".$f"
+      case Index(v) => s"[$v]"
+      case Clock => s"@clock"
+      case Reset => s"@reset"
+      case Init => s"@init"
+    }.mkString("")
+
+    (moduleOpt.isEmpty, tokens.isEmpty) match {
+      case (true, true) => circuitString
+      case (_, true) => circuitString + moduleString
+      case (_, _) => circuitString + moduleString + tokenString
+    }
+  }
+
+
   /** @return Converts this [[Target]] into a [[GenericTarget]] */
   def toGenericTarget: GenericTarget = GenericTarget(circuitOpt, moduleOpt, tokens.toVector)
 
--- a/src/main/scala/firrtl/passes/CheckWidths.scala
+++ b/src/main/scala/firrtl/passes/CheckWidths.scala
@@ -7,42 +7,44 @@ import firrtl.ir._
 import firrtl.PrimOps._
 import firrtl.Mappers._
 import firrtl.Utils._
+import firrtl.annotations.{Target, TargetToken, CircuitTarget, ModuleTarget}
 
 object CheckWidths extends Pass {
   /** The maximum allowed width for any circuit element */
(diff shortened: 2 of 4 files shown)


azidar:
I really like the implementation. I agree forcing Chisel users to know the Target serialization is a bit much though. Maybe we can add another, more "understandable" serialization to Target (that doesn't need to be unserializable)?

seldridge:
Good idea @azidar. I've rebased and added a `prettyPrint` method to `Target`. This provides output similar to `tree`. This method may produce badly formatted output for illegal `Target`s, but I think that's justifiable (and the existing `serialize` method will produce equivalently nonsensical output).

azidar:
Very cool. Can you make it the following with the circuit/module keywords?
```
circuit Foo:
  └── module Bar:
    └── a.b.c
```

seldridge:
Makes sense to be explicit. Thoughts on how to pretty print `Instance`/`OfModule` tokens? I pre-emptively chose the following option:
```
circuit Foo :
  └── module Bar :
    └── inst baz of Baz :
        └── a.c.e
```

azidar:
Yeah, I think that's great. Can you remove the space before the `:`?

seldridge:
Sure. Related to that space... I tend to write FIRRTL without it (assumedly because of writing Scala that way). However, IR serialization uses a space everywhere. I got the impression that the space was optional, but suggested. Is there some standard approach here (no space?)?

jackkoenig:
@seldridge this is great. How should this pretty printing relate to source locators? I don't think Chisel currently generates them for Modules nor for ports but assuming it did, how could/should this look?

seldridge:
Since `prettyPrint` is a method of `Target`, it has no affiliation with source locators (`Info`). If we had line annotations or if the `Target` included `Info` then this would be possible. It's possible for a future refactor, but it would likely be a lot of work and it may not make sense. I had a related discussion with @azidar last week about this and the conclusion was that annotations are fundamentally different from info. The former are affiliated with something in a circuit and are guaranteed to be renamed, propagated, etc. based on what transformations are applied to the circuit. These are independent of how a user may have written a specific circuit. The latter are tied to specific lines, but have no guarantees that they will be propagated or not deleted arbitrarily. These may change based on how the circuit was written, what version of a frontend generated the FIRRTL, or what order things are instantiated, etc.

Does that make sense?

So, I think these are fundamentally different. However, if `Target.prettyPrint` was provided with a `Circuit` it would have enough information to traverse the circuit and scrape out source locators to give you:
```
circuit Foo: @[line 10]
└── module Bar: @[line 20]
    └── a.c.e @[line 30]
```

However, all that has to do with `Target.prettyPrint`. The source locators for the `CheckWidths.UninferredWidth` exception _does include source locators_, but the test case doesn't have them. If I had info, you'd get something like the following.

For the following circuit:
```
circuit Foo :
  module Foo :
    input clock : Clock
    inst bar of Bar
  module Bar :
    wire a: { b : UInt<1>, c : { d : UInt<1>, e : UInt } } @[line 42]
```

You'd get the following exception:
```
[info]   firrtl.passes.CheckWidths$UninferredWidth: @[line 42] : Uninferred width for target below. (Did you forget to assign to it?)
[info]     circuit Foo:
[info]     └── module Bar:
[info]         └── a.c.e
```

azidar:
TBH the old FIRRTL in stanza had a space before the colon, but scala style is for no space. I've just gotten used to no space, and I think it looks better, but if we change the normal FIRRTL serialization, we'd need to fix up a bunch of tests...

PR #936, opened 2018-11-09
Fix bug in TargetDirAnnotation compatibility

seldridge:
TargetDirAnnotation was moved from firrtl to firrtl.stage. However, this
is only aliased as a val in the firrtl package object. This also needs to
be type aliased for unapply to work (apparently). This fixes a bug I ran
across in the visualizer.

### DIFF
--- a/src/main/scala/firrtl/package.scala
+++ b/src/main/scala/firrtl/package.scala
@@ -1,3 +1,4 @@
+// See LICENSE for license details.
 
 import firrtl.AnnotationSeq
 import firrtl.annotations.Annotation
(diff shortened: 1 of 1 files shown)


src/main/scala/firrtl/package.scala
```diff
@@ -8,5 +8,6 @@ package object firrtl {
 
   /* Options as annotations compatibility items */
   @deprecated("Use firrtl.stage.TargetDirAnnotation", "3.2")
+  type TargetDirAnnotation = firrtl.stage.TargetDirAnnotation
   val TargetDirAnnotation = firrtl.stage.TargetDirAnnotation
```
jackkoenig: I think you need to deprecate both? I'm not sure though

seldridge: Good catch. One odd interaction that I did not expect is that the deprecation warnings will only trigger when doing an assignment. I expected them to trigger when matching on type or unapplying, but it doesn't.
```scala
/* Both of these trigger compile time deprecation warnings */
val x = TargetDirAnnotation()
val y = TargetDirAnnotation

/* Neither of these match statements will result in deprecation warnings */
x match {
  case TargetDirAnnotation(xx) => println(xx)
  case xx: TargetDirAnnotation => println("hello")
}
```

Updated in the amended commit.

PR #940, opened 2018-11-16
 Bump sbt to 1.2.6; update dependencies

ucbjrl:


ucbjrl:
Why the explicit reference to junit?

I was running into problems with transitive dependencies, specifically:
Symbol 'term org.junit' is missing from the classpath.
I suspect this is due to interaction between the latest version of sbt and org.scala-sbt.ivy. It may be because junit is special - it's pure Java and independent of the Scala compiler. Forcing a more recent version of junit seemed to resolve those problems.

PR #945, opened 2018-11-21
Change firrtl.options API, add Phase

seldridge:
### tl;dr

- Adds `Phase`, a parent class of `Stage`
- Defines common preprocessing for all `Stage`s
- Defines `StageOptions`, annotations for `Stage`, and `StageOptionsView`

### Overview

In the course of cleaning up the API for the new `FirrtlStage` API (#920), it occurred to me that there exists an improved API and small type hierarchy that modifies the existing `Stage` API that was introduced in #879.

Basically, there exist two types of "things" in a Hardware Compiler Framework:

  - A `Phase` which performs a mathematical transformation on an `AnnotationSeq`
  - A `Stage` subclasses `Phase` and adds a user interface (command line or otherwise)

Certain command line options/annotations are then common across all `Stage`s (this is basically just me reviving/concretizing what the "common options" are, with the caveat that more common options like program args may be added). All `Stage`s should be able to read in annotations from a file and should be able to specify a build directory, i.e., every `Stage` supports:

  - `--annotation-file`/`-faf`/`InputAnnotationFileAnnotation`
  - `--target-dir`/`-td`/`TargetDirAnnotation`

This then precipitates some common, necessary preprocessing on the input `AnnotationSeq` that all `Stage`s must do:

  - Specified annotation files should be recursively read (but not read more than once)
  - A target directory should be added if not specified for downstream tools
  - Any legacy annotations should be converted

These latter three are realized as concrete `Phase`s that run before the body of any `Stage`:

  - `GetIncludes`
  - `ConvertLegacyAnnotations`
  - `AddDefaults`

What's going on in each of these phases (and in the updated #920) has the effect of moving work out of what I was normally doing when constructing a `view[T]` of an `AnnotationSeq`. Normally, I'd add default inclusion, validation checking, etc. inside the `view`. This resulted in views needing to encode more information than they could return (good view, unable to construct view, and bad view). This created the apparent need, to me, for `view` to return `Option[T]`. However, now that work gets moved entirely out of the `view` and into separate `Phase`s that run before you construct a view, it now makes complete sense to me that `view` should directly return a `T`. (@chick: I think your intuition was onto something here... `view`s should be lightweight with work moved to separate `Phase`s. Otherwise, the `view` is getting an incorrect snapshot of the annotations.) 

#### `firrtl.hcf` package

Thinking about this slightly more... `firrtl.options` is really the underpinnings of a hardware compiler framework. This should likely (now or later) be renamed as `firrtl.hcf` or some better name (`hardwarecompilerframework` is as bad as `contextdependentenvironments`). It may additionally make sense to move some of `firrtl` and `firrtl.annotations` into this package eventually.

### Metadata
  - Schuyler will rebase and merge

### Rebase Items
  - [ ] Commit message needs to be reworded to remove references to `ViewException`

### DIFF
--- a/src/main/scala/firrtl/options/Stage.scala
+++ b/src/main/scala/firrtl/options/Stage.scala
@@ -4,38 +4,17 @@ package firrtl.options
 
 import firrtl.AnnotationSeq
 
-/** Utilities mixed into something that looks like a [[Stage]] */
-object StageUtils {
-  /** Print a warning message (in yellow)
-    * @param message error message
-    */
-  //scalastyle:off regex
-  def dramaticWarning(message: String): Unit = {
-    println(Console.YELLOW + "-"*78)
-    println(s"Warning: $message")
-    println("-"*78 + Console.RESET)
-  }
+case class StageException(val str: String, cause: Throwable = null) extends RuntimeException(str, cause)
 
-  /** Print an error message (in red)
-    * @param message error message
-    * @note This does not stop the Driver.
-    */
-  //scalastyle:off regex
-  def dramaticError(message: String): Unit = {
-    println(Console.RED + "-"*78)
-    println(s"Error: $message")
-    println("-"*78 + Console.RESET)
-  }
-}
-
-/** A [[Stage]] represents one stage in the FIRRTL hardware compiler framework
+/** A [[Stage]] represents one stage in the FIRRTL hardware compiler framework. A [[Stage]] is, conceptually, a
+  * [[Phase]] that includes a command line interface.
   *
   * The FIRRTL compiler is a stage as well as any frontend or backend that runs before/after FIRRTL. Concretely, Chisel
   * is a [[Stage]] as is FIRRTL's Verilog emitter. Each stage performs a mathematical transformation on an
   * [[AnnotationSeq]] where some input annotations are processed to produce different annotations. Command line options
   * may be pulled in if available.
   */
-abstract class Stage {
+abstract class Stage extends Phase {
 
   /** A utility that helps convert command line options to annotations */
   val shell: Shell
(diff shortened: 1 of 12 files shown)


src/main/scala/firrtl/options/Stage.scala
```diff
@@ -44,18 +23,33 @@ abstract class Stage {
...
+
+  /** Execute this [[Stage]] on some input annotations. Annotations will be read from any input annotation files.
+    * @param annotations input annotations
+    * @return output annotations
+    */
+  final def transform(annotations: AnnotationSeq): AnnotationSeq = {
+    val preprocessing: Seq[Phase] = Seq(
+      phases.GetIncludes,
+      phases.ConvertLegacyAnnotations,
+      phases.AddDefaults )
```
azidar: Just like we are registering transforms, have you thought about registering phases? It also seems like we really need a dependency API for both phases and transforms.

seldridge: I had not thought about that. Thinking about it now... The aim of registering transforms is to allow transforms to expose to a `Stage` that they contain options that the stage should expose to the user. This comes about because transforms consume annotations and command line options and annotations are fungible. Further, transforms can be injected into the stage from a library and may not live it the stage's package. Hence, you need the stage to discover what options it should present to the user via registration.

By that logic, if a phase wants to consume annotations that then means that it may want to expose options via it's encapsualting stage. However, I'm not sure if (or it hadn't occured to me that) phases are something that the user should get to inject into a stage.

All that said, passes and transforms may subclass phase eventually...

So... Phases should likely be registerable.

Dependency API? Maybe you could clarify on what you're thinking. I've long thought that we need for passes/transforms/phases/stages to express how they interoperate so that some scheduler can figure out a directed graph that satisfies dependencies and generates the requested outputs at each leaf. Some generalization of input/ouptput form?

seldridge: One follow-up: the registree (the thing wants to be registered) will need to provide information about what stage it wants to be registered by. Options are then exposed backwards along any stages that form a path to the shell which the user is interacting with.

E.g., a FIRRTL registered transform should register options into FIRRTL and those are exposed in Chisel and a (future) diplomacy stage. Registration of external Chisel phases would then _not_ have their options exposed to downstream tools (FIRRTL, Yosys, HAMMER).

Maybe we should discuss this next week... I can work on some mockups.

PR #946, opened 2018-11-21
Remove firrtl.altIR package

seldridge:
There hasn't been any activity in a year on this and I'm not aware of anybody using it. I expect that the existing abstractions provided by `DiGraph`, `InstanceGraph`, and the PR'd `CircuitGraph` are sufficient. I also don't want people to start using something that we don't really have support or tests for.


edwardcwang:
What are the thoughts of @dt27182 on this?

seldridge:
And to clarify, this reverts #671. Informal Google discussions at the CCC indicated that they don't have transforms depending on this. Somebody might be using it, however...

seldridge:
Thanks for the confirmation @dt27182. If this is in error and we throw off the Google code base, we can revert (the revert).

PR #947, opened 2018-11-22
Add nesting threshold to register update flattening

jackkoenig:
Previously, dynamically indexing a register directly into a register
would result in very deep if-else nesting in the emitted Verilog


azidar:
Seems like yosys takes forever (runs out of memory) doing equivalence on the ICache, @jackkoenig is that symptomatic of an underlying difference between the verilog designs?

jackkoenig:
Lol I totally forgot about this, but we ought to get this in

aswaterman:
There's no doubt this is better than the status quo, but it still doesn't seem quite right.  It might even be better to revert to referencing the original combinational logic as soon as there's any repetition in the if/else structure at all.

jackkoenig:
Yeah I agree, since we're looking into improving Verilog generation anyway, perhaps we should just fix the right way this as part of that effort.

jackkoenig:
Replaced by https://github.com/freechipsproject/firrtl/pull/1727

PR #950, opened 2018-11-26
Improve propagate performance for big annotations

seldridge:
When passing around big annotations (e.g., the `FirrtlCircuitAnnotation` that #920 adds), the overhead of copying annotations in `Compiler.propagateAnnotations` explodes. This does three things to improve the performance if you have big annotations:

1. The `mutable.LinkedHashSet` introduced by #855 is changed to an `immutable.ListSet`. I'm not sure exactly what's going on here, but construction of `LinkedHashSet` seems to be approximately 6x slower than `ListSet`. :woman_shrugging: No mutable operations are used here, so this is a straightforward change. The ordered requirement of #855 is still preserved.

2. Any construction of `ListSet[Annotation]` is limited as much as possible. This removes the unnecessary construction of `created` and `unchangedSet`.

3. Some `LinkedHashSet` that were only used for debugging (i.e., used inside of `logger.debug`) are moved to live inside of a `logger.debug` scope.

#920, without this PR, results in a dramatic runtime increase for large designs. E.g., Rocket Chip `DefaultConfig` takes 16s to get through `ChirrtlToHighFirrtl`:

```
Time: 15187.3 ms
Form: HighForm
======== Finished Transform ChirrtlToHighFirrtl ========
```

#920, with this PR, gets this down to 2s:

```
Time: 1840.0 ms
Form: HighForm
======== Finished Transform ChirrtlToHighFirrtl ========
```

### Alternatives 

There are two main alternatives:

1. A `FirrtlCircuitAnnotation` (and other big annotations) shouldn't be passed around during lowering. This is a valid point and I plan to update #920 in this way. However, the problem outlined here will come up for any massive annotation that people want to move around.

2. The debug-only code could be removed outright. I leave that decision up to @azidar. If left in, then debug will be substantially slower than non-debug. This seems fine generally, but this is really crawling... There may be a further refactor here that would remove the use of `LinkedHashSet`, use something faster, or rely on (1) to get rid of the big offender here.

### DIFF
--- a/src/main/scala/firrtl/Compiler.scala
+++ b/src/main/scala/firrtl/Compiler.scala
@@ -7,7 +7,7 @@ import java.io.Writer
 
 import firrtl.RenameMap.{CircularRenameException, IllegalRenameException}
 
-import scala.collection.mutable
+import scala.collection.{mutable, immutable}
 import firrtl.annotations._
 import firrtl.ir.{Circuit, Expression}
 import firrtl.Utils.{error, throwInternalError}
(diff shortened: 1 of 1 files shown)


jackkoenig:
I don't think it's the `copying` of annotations like `FirrtlCircuitAnnotation`, I think it's the hashing process. Thus every time you have to rehash it's going to be pretty expensive.

I suspect that the performance gains come entirely from never having to hash the `FirrtlCircuitAnnotation` in a `ListSet` (since insertion and lookup is a linear search, no hashing. But for larger numbers of annotations I think we'll run into some pretty big problems using `ListSet`. In particular, I think the `(inSet -- resAnno)` is be `O(n^2)` for `ListSet` while merely `O(n)` for `LinkedHashMap`.

All of this being said, benchmarking is king. Have you benchmarked just using the default stuff in rocket-chip or have you run it with a larger number of annotations? I can run experiments with the large number of annotations we use internally tomorrow. We should probably create a more formal benchmarking suite including lots of annotations ([here's the script I'm currently using, but it doesn't do any annotations at the moment](https://github.com/jackkoenig/firrtlbench/blob/master/benchmark.py)).

seldridge:
Great insight, @jackkoenig. That makes much more sense. It should be the repeated hashing of a ~10MB circuit that's doing it. And yes `ListSet` won't work for sets with large numbers of elements... 

I've only benchmarked with `DefaultConfig` rocket (~300 annotations), so this like will not scale to large numbers of annotations. 

What would you consider a "large" number of annotations?

This obviously has to be a set backed by a hash to support large numbers of annotations. Reverting #855 to move away from the linked nature shouldn't change anything either. 

I can dig into this more... It would seem like this must use a set backed by a hash. However, that hash should be a (`lazy`?) `val` for annotations storing large things and should be copied if no input parameters change.

seldridge:
Not a bug and not a complicated solution either. Just `lazy val` caching the `hashCode` in big annotations is sufficient. :man_facepalming:

PR #953, opened 2018-11-27
Add "none" compiler

jackkoenig:
Where the high form compiler removes Chirrtl (and runs some checks),
this compiler does nothing but read in the circuit and then emit it

Also looking for a better name if anyone has suggestions. I thought about making it "Chirrtl" but since the plan is to delete Chirrtl I'm not sure it's worth introducing more things to ultimately change

### DIFF
--- a/src/main/scala/firrtl/ExecutionOptionsManager.scala
+++ b/src/main/scala/firrtl/ExecutionOptionsManager.scala
@@ -202,6 +202,7 @@ extends ComposableOptions {
 
   def compiler: Compiler = {
     compilerName match {
+      case "none"      => new NoneCompiler()
       case "high"      => new HighFirrtlCompiler()
       case "low"       => new LowFirrtlCompiler()
       case "middle"    => new MiddleFirrtlCompiler()
@@ -215,8 +216,9 @@ extends ComposableOptions {
       case "verilog"   => "v"
       case "sverilog"  => "sv"
       case "low"       => "lo.fir"
-      case "high"      => "hi.fir"
       case "middle"    => "mid.fir"
+      case "high"      => "hi.fir"
+      case "none"      => "fir"
       case _ =>
         throw new Exception(s"Illegal compiler name $compilerName")
     }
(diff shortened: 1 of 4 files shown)


src/main/scala/firrtl/ExecutionOptionsManager.scala
```diff
@@ -260,6 +262,7 @@ extends ComposableOptions {
   def getEmitterAnnos(optionsManager: ExecutionOptionsManager): Seq[Annotation] = {
     // TODO should this be a public function?
     val emitter = compilerName match {
+      case "none" => classOf[HighFirrtlEmitter]
```
edwardcwang: Is `HighFirrtlEmitter` the same as a "Chirrtl emitter"?

jackkoenig: Other than the returned CircuitState type, yes. I guess I should make a ChirrtlEmitter and have the none compiler use that

edwardcwang: Is an emitter of a higher form "compatible" with lower forms? So if I use the none compiler and pass in a low FIRRTL input, will the output still be low FIRRTL even if I use the ChirrtlEmitter?

seldridge: Theoretically, that's the idea (lower forms are subsets of higher forms). Practically, all the FIRRTL emitters are the exact same.

jackkoenig: It's "compatible" in that all FIRRTL emitters do the same thing, but the output CircuitState will always have the form of the emitter itself. In fact, the output form of any transform (or emitter) is always statically known and independent of the form of the input CircuitState. This could possibly be changed but making it statically known makes it possible to statically schedule transforms which has some utility. 

You can see the actual implementations of the various Firrtl Emitters here: https://github.com/freechipsproject/firrtl/blob/b90589f5cd9d4048ada2a05d5225874791546170/src/main/scala/firrtl/Emitter.scala#L115

jackkoenig:
@seldridge yeah fair point, we can just deprecate one when that happens

PR #954, opened 2018-11-28
Replace Mappers with Foreachers in several passes

albert-magyar:
Drops ~40 LoC and helps with clarity in some cases by eliminating spurious return value expressions.

### DIFF
--- a/src/main/scala/firrtl/passes/CheckInitialization.scala
+++ b/src/main/scala/firrtl/passes/CheckInitialization.scala
@@ -5,7 +5,7 @@ package firrtl.passes
 import firrtl._
 import firrtl.ir._
 import firrtl.Utils._
-import firrtl.Mappers._
+import firrtl.traversals.Foreachers._
 
 import annotation.tailrec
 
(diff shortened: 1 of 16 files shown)


src/main/scala/firrtl/passes/CheckInitialization.scala
```diff
@@ -85,10 +81,10 @@ object CheckInitialization extends Pass {
       }
     }
 
-    c.modules foreach {
+    c.modules.foreach({
```
chick: Nit: Parens seem extraneous here

jackkoenig:
As an added bonus, this modestly improves performance, about 3-5%. This improvement most likely comes from avoiding unnecessary object creations as done by the mappers.

albert-magyar:
> As an added bonus, this modestly improves performance, about 3-5%. This improvement most likely comes from avoiding unnecessary object creations as done by the mappers.

So what you're saying is that I can write an ISCA paper about this.

src/main/scala/firrtl/passes/CheckInitialization.scala
albert-magyar: True! I removed them.

Issue #933, opened 2018-11-06
Make CheckWidths aware of WiringTransform annotations


albertchen-sifive:
**Type of issue**: feature request

**What is the use case for changing the behavior?**
I would like to make a Chisel API that connects wires across modules using `WiringTransform`. I'm doing this by declaring a wire of the same type using `chiselTypeOf` and creating sink and source wiring annotations. However, this leads to `UninferredWidth` errors, because sink widths are not set until after `WiringTransform` which runs after `CheckWidths`.

**Impact**: unknown

**Development Phase**: request

jackkoenig:
Possible approaches:
1. Make WiringTransform run before InferWidths
1. Make InferWidths aware of Wiring annotations (could also be spun out into a special case pre-width inference transform)
1. Add WidthInference ignore annotation (for specific signals) and then rerun InferWidths after WiringTransform

I think (3) has high potential for brittleness/unintended consequences, (1) or (2) seem more reasonable to me

seldridge:
The real problem here is not `CheckWidths`, but `InferWidths` doesn't know anything about `WiringTransform` annotations. The `UninferredWidth` error is a pretty dumb check -- this just looks to see if everything has a width. If it doesn't, then it asserts.

I agree with your summary, @jackkoenig. (1) I don't think is possible, because `CheckWidths` runs as soon as you read in Chirrtl. Doing a `WiringTransform` on High FIRRTL is totally doable, but it would have to run almost before anything else (which may actually make sense. (2) is likely tractable because `WiringTransform` soft connects are just ports. When building constraints these could just be treated as ports and everything should work as it does now. (3) would be a fallback if (1) or (2) proved intractable for whatever reason. The problem with (3) is that it means that use of `WiringTransform` annotations would result in breaking guarantees on High/Mid/Low FIRRTL. Custom Transforms that are relying on this would barf.

PR #956, opened 2018-12-01
Add Width Constraints with Annotations

albertchen-sifive:
This changes the `InferWidths` transform. `InferWidths` now collects extra constraints from `WidthGeqConstraintAnnotation` and runs with those extra constraints. Also makes `InferWidths` a `Transform` class instead of a `Pass` object.

This transform can be used to add width constraints that are introduced by later transforms like `WiringTransform`.

Related issue: https://github.com/freechipsproject/firrtl/issues/933


src/main/scala/firrtl/transforms/InferWidthsWithAnnos.scala
```diff
@@ -0,0 +1,95 @@
...
+        case w: DefWire => Some(w.name -> w.tpe)
+        case r: DefRegister => Some(r.name -> r.tpe)
+        case n: DefNode => Some(n.name -> n.value.tpe)
+        case i: WDefInstance => Some(i.name -> i.tpe)
+        case m: DefMemory => Some(m.name -> MemPortUtils.memType(m))
+        case other => None
+      }
+      pairOpt.foreach { case (ref, tpe) =>
+        typeMap += (ReferenceTarget(circuitName, modName, Nil, ref, Nil) -> tpe)
+      }
```
azidar: Why don't you update the typeMap directly in the stmt match?

albertchen-sifive: no real reason, `Some(ref -> tpe)` was just much less typing than `typeMap += (ReferenceTarget(circuitName, modName, Nil, ref, Nil) -> tpe)`

jackkoenig: It is a little bit DRYer this way

src/main/scala/firrtl/passes/InferWidths.scala
```diff
@@ -360,10 +386,56 @@ object InferWidths extends Pass {
...
+      mod.ports.foreach { port =>
+        typeMap += (ReferenceTarget(circuitName, mod.name, Nil, port.name, Nil) -> port.tpe)
+      }
+      mod.foreachStmt(getDeclTypes(mod.name))
+    }
+
+    val extraConstraints = state.annotations.flatMap {
+      case anno: WidthGeqConstraintAnnotation if anno.loc.isLocal && anno.exp.isLocal  =>
+        val locType :: expType :: Nil = Seq(anno.loc, anno.exp) map { target =>
+          val baseType = typeMap(target.copy(component = Seq.empty))
```
azidar: Need to check whether typeMap contains the reference, its possible that something else went wrong and they don't match up.

src/main/scala/firrtl/passes/InferWidths.scala
```diff
@@ -360,10 +386,56 @@ object InferWidths extends Pass {
...
+        case m: DefMemory => Some(m.name -> MemPortUtils.memType(m))
+        case other => None
+      }
+      pairOpt.foreach { case (ref, tpe) =>
+        typeMap += (ReferenceTarget(circuitName, modName, Nil, ref, Nil) -> tpe)
+      }
+      stmt.foreachStmt(getDeclTypes(modName))
+    }
+
+    state.circuit.modules.foreach { mod =>
```
azidar: Maybe only walk AST after confirming that you do have annotations? That way you save an AST walk in the common case.

azidar:
Also, please update the PR description to say that it affects the InferWidths transform, not that it adds a new transform.

azidar:
Will commenting trigger mergify?

jackkoenig:
lol apparently

PR #958, opened 2018-12-04
Fix bug in dedup where lots of annotations could prevent dedup

jackkoenig:
Iterating on a HashSet could cause identical modules (including
annotations) to not dedup

Changing builder (which is an `ArrayBuffer[Any]`) to just include the HashSet as a member (thus letting it call it's own `.hashCode`) rather than iterating on it.

This is a hard thing to test for, so I wrote a test that just has lots of annotations and failed (at least for me) without this change.

@ucbjrl I think this should be included in 1.1.4 if possible


jackkoenig:
A related issue that I noticed but haven't yet fixed is that dedup just uses the hashcode as a key to determine if things can dedup. @azidar does it actually check if they are equal? While the odds aren't super high, it is possible that modules could have colliding hashcodes and thus do an incorrect dedup if we don't actually check equality

Issue #219, opened 2016-08-04
Support for asynchronously-resettable registers


ben-k:
FIRRTL is currently only capable of emitting registers with synchronous reset.  This is often OK, but sometimes an asynchronous reset is desirable (e.g., see our recent discussion at ucb-bar/rocket-chip#190, as well as [previous discussion](https://groups.google.com/forum/#!topic/chisel-users/4cc4SyB5mk8) from a variety of users).  Ultimately, the designer should have the flexibility to choose which type of reset to instantiate.  (See [this classic discussion](http://www.sunburst-design.com/papers/CummingsSNUG2002SJ_Resets.pdf) of the issue for more detail.)

Asynchronously-resettable flip-flops are a common verilog design idiom and are well-supported by EDA tools.  ucb-bar/chisel#549 has a good writeup of the desired feature; I've copied and cleaned up the code example here for convenience.  The option to select the reset polarity would also be nice.

``` verilog
always @(posedge clk or posedge reset) begin
  if (reset) begin //async-resettable flops
    flop1 <= FLOP1_RESET_VALUE;
    flop2 <= FLOP2_RESET_VALUE;
  end else // D -> Q state updates for async-resettable flops
    flop1 <= flop1_next;
    flop2 <= flop2_next;
  end
end

always @(posedge clk) begin
  if (reset) begin //sync-resettable flops
    flopA <= FLOPA_RESET_VALUE;
    flopB <= FLOPB_RESET_VALUE;
  end else begin // D -> Q state updates for sync-resettable flops and flops with no reset
    flopA <= flopA_next;
    flopB <= flopB_next;
    flopC <= flopC_next; // flopC has no sync or async reset
  end
end
```

ben-k:
@azidar, per our discussion earlier.

aswaterman:
Async resets are just clocks, so if this feature is implemented, it should be as a generalization of reg that allows multiple clocks.

azidar:
To implement this feature, it seems one would only need to modify the Verilog emitter to pattern match for a Clock-type reset and emit the corresponding Verilog.

janboeye:
@azidar 
Why this feature is low priority? If we want to use Chisel in ASIC chip, this is a must-have feature. In this stage, is there any workaround for this?
Thanks a lot!

aswaterman:
[...]

PR #960, opened 2018-12-07
Add asynchronous reset support

dalance:
This PR is an implementation of #219.

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -414,6 +414,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
     val attachSynAssigns = ArrayBuffer.empty[Seq[Any]]
     val attachAliases = ArrayBuffer.empty[Seq[Any]]
     val at_clock = mutable.LinkedHashMap[Expression, ArrayBuffer[Seq[Any]]]()
+    val at_clock_reg = ArrayBuffer.empty[(Expression,Seq[Seq[Any]])]
     val initials = ArrayBuffer[Seq[Any]]()
     val simulates = ArrayBuffer[Seq[Any]]()
 
@@ -473,7 +474,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
         }
       }
 
-      at_clock.getOrElseUpdate(clk, ArrayBuffer[Seq[Any]]()) ++= addUpdate(netlist(r), "")
+      at_clock_reg += ((clk, addUpdate(netlist(r), "")))
     }
 
     def update(e: Expression, value: Expression, clk: Expression, en: Expression, info: Info) = {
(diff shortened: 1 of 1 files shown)


dalance:
Hi @azidar , Thanks.
I agree that IR level solution is better than this PR.
I'll wait for the meeting result.

aswaterman:
I think this is problematic for a few reasons:

- It's not correct in general to simply replace synchronous resets with asynchronous resets.  What if the reset is computed from combinational logic, so can glitch?

- It's hard-coded to a signal named reset.  This is just a convention, not a requirement.  I might have a signal called "reset" whose purpose is not that of a reset.

- The ifdef approach is going to cause a backwards-compatibility issue once async reset is represented properly in the IR.  At that point, the polarity will also be part of the IR, not something specified by a macro.

schoeberl:
Using synchronous or asynchronous reset is a property of the receiving flip-flop (a technology question). Sometimes asynchronous (e.g., in older FPGAs), sometimes synchronous may be cheaper.

Independent from which reset input type one uses, the register signal has to be synchronous to the clock. Reset (also synchronous version) shall not violate setup and hold time. If reset comes from a pin as a synchronous signal it has to go through an input synchronizer (two flip-flops).

That also means that a reset coming from a register (output) is fine for a synchronous and an asynchronous reset. If one wants to add logic to that signal (which is probably not good design style in my opinion) it is ok as long as it fits into one clock cycle. Same for both cases.

I agree that this is a useful short-time solution and would give it a try.

schoeberl:
> On 7 Dec, 2018, at 14:43, Andrew Waterman <notifications@github.com> wrote:
> 
> I think this is problematic for a few reasons:
> 
> It's not correct in general to simply replace synchronous resets with asynchronous resets. What if the reset is computed from combinational logic, so can glitch?
> 
Yes, this is a gotcha. (Mistake in my former comment) Agree when you have logic in your reset path you should not switch to an asynchronous reset and stick to the synchronous reset. You should first remove the logic from the reset path ;-)

A synchronous reset is basically only logic in front of the flip-flop (mux with one input the reset value). As it is so common, FPGAs start to add this signal extra to the filp-flops to not wast LUTs.

Martin

dalance:
I think the issue of reset logic and "reset" name can't be resolved by current IR because IR don't have the information about reset.
The issue of "ifdef" compatibility may be resolved by command-line option like below?

```
firrtl --reg-reset <sync|async_pos|async_neg>
```

But I can't find the way to pass an option to VerilogEmitter...

azidar:
So the dev meeting was mixed in its feelings towards this - we are seeking additional feedback from @aswaterman and @schoeberl and anyone else interested in commenting.

jackkoenig:
Superseded by https://github.com/freechipsproject/firrtl/pull/1011

PR #962, opened 2018-12-11
Enhance CheckCombLoops to support annotated ExtModule paths

albert-magyar:
This adds an `ExtModulePathAnnotation` that specifies the existence of a combinational path from one port of an `ExtModule` to another.

Feedback would be appreciated on the public API for the annotation:
- Should it be one annotation per path (as it is now), or should one annotation per blackbox cover all paths?
- Should this have a public API in Chisel like `dontTouch`?
- It could in theory be merged with `CombinationalPath`, but I currently like the separation, since it is clear that this will only work with `ExtModules` and that there are semantic differences.

### DIFF
--- a/src/main/scala/firrtl/transforms/CheckCombLoops.scala
+++ b/src/main/scala/firrtl/transforms/CheckCombLoops.scala
@@ -7,6 +7,8 @@ import scala.collection.immutable.HashSet
 import scala.collection.immutable.HashMap
 import annotation.tailrec
 
+import Function.tupled
+
 import firrtl._
 import firrtl.ir._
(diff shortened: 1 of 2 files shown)


davidbiancolin:
From the dev meeting: 
- Keep anno separate from existing CombinationalPath annotation for now; 
- One anno per path (consistent with existing schemes)
- Chisel land annotator can just be added as a method on BlackBox; one invocation per annotation (output).

albert-magyar:
Great, it seems like that's all in line with how this PR currently implements things. @jackkoenig any thoughts / reviews?

PR #964, opened 2018-12-12
Try getting travis to use multiple OSs.

grebe:
Windows does not support Scala, but OS X does. This requires some reworking of how deps are installed.

Ideally, travis will support JVM on Windows and we can also test there. Related: [this](https://github.com/freechipsproject/chisel3/issues/874) issue and [this](https://github.com/freechipsproject/chisel3/issues/874) one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,10 @@
+os:
+  - linux
+  - osx
+#  - windows
 language: scala
+scala:
+    - 2.12.4
 sudo: false
 
 cache:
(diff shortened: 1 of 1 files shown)


XVilka:
@grebe you can consider to use AppVeyor for Windows jobs - in fact, their machines are faster and you have more choice.

XVilka:
For example this https://github.com/ensime/scala-debugger/blob/master/appveyor.yml configuration.

grebe:
@XVilka thanks for the heads up, looks like windows + jvm is supported a little better there

grebe:
@colinschmidt @ucbjrl this isn't messing up the cache, right?

colinschmidt:
Sad. Do we need explicit versioning?

jackkoenig:
To expand a bit on why it breaks is that Travis includes all matrix configuration stuff in the key for the cache, thus multiple Scala versions (and probably multiple OSes) will trample on each other to create the cache.

The best solution I can think of is to put all of the dependencies in Docker, then we can use the Travis build matrix stuff and just not have to do any caching. This would also make it trivial to run regressions on different OSes, although it would not test setup instructions or anything (which I think should be included in some set of regressions, somewhere, but probably not on regular PR CI).

grebe:
So should I kill this PR?

jackkoenig:
As is? Probably. If we switch to using docker (which we should, someone just has to do it), then we should revive it

Issue #965, opened 2018-12-13
ReferenceTargets with paths and TargetTokens not renamed correctly


albertchen-sifive:
**Type of issue**: bug report

**If the current behavior is a bug, please provide the steps to reproduce the problem:**
  * **If possible, please directly provide the FIRRTL file or a relevant fragment if the file is large**
test case:
```scala
  it should "rename reference targets with paths and target tokens" in {
    val cir = CircuitTarget("Top")
    val modTop = cir.module("Top")
    val modA = cir.module("A")

    val aggregate = modA.ref("agg")
    val subField1 = aggregate.field("field1")
    val subField2 = aggregate.field("field2")

    val lowered1 = aggregate.copy(ref = "agg_field1")
    val lowered2 = aggregate.copy(ref = "agg_field2")

    // simulating LowerTypes transform
    val renames = RenameMap()
    renames.record(subField1, lowered1)
    renames.record(subField2, lowered2)
    renames.record(aggregate, Seq(lowered1, lowered2))

    val path = modTop.instOf("b", "B").instOf("a", "A")
    val testRef = subField1.setPathTarget(path)

    renames.get(testRef) should be {
      Some(Seq(testRef.copy(ref = "agg_field1", component = Nil)))
    }
  }
```
  * **What is the current behavior?**
`testRef` gets renamed from:
```
circuit Top:
└── module Top:
    └── inst b of B:
        └── inst a of A:
            └── agg.field1
```
to the following two targets:
```
circuit Top:
└── module Top:
    └── inst b of B:
        └── inst a of A:
            └── agg_field1.field1
circuit Top:
└── module Top:
    └── inst b of B:
        └── inst a of A:
            └── agg_field2.field1
```

running the test case results in:
```
[info] - should rename reference targets with paths and target tokens *** FAILED ***
[info]   Some(List(ReferenceTarget(Top,Top,List((Instance(b),OfModule(B)), (Instance(a),OfModule(A))),agg_field1,List(Field(field1))), ReferenceTarget(Top,Top,List((Instance(b),OfModule(B)), (Instance(a),OfModule(A))),agg_field2,List(Field(field1))))) was not equal to Some(List(ReferenceTarget(Top,Top,List((Instance(b),OfModule(B)), (Instance(a),OfModule(A))),agg_field1,List())))
```
  * **What is the expected behavior?**
I expected `testRef` to be renamed from:
```
circuit Top:
└── module Top:
    └── inst b of B:
        └── inst a of A:
[...]

PR #967, opened 2018-12-14
Fix renaming of annotations with paths

albertchen-sifive:
Added test case from bug report. Added `isLocal` check in `RenameMap` when renaming `ReferenceTarget`

related issue: https://github.com/freechipsproject/firrtl/issues/965


jackkoenig:
A glance from @azidar would be good. Fair warning to anyone who feels like looking at this; it made my head hurt 🤕

azidar:
What about renaming case of `~Top|Module/i1:I1>foo.field`, where rename map contains `~Top|Module/i1:I1>foo` -> `~Top|Module/i1:I1>bar`? Can you write a test case for this?

jackkoenig:
I was wrong, the case @azidar pointed out fails

albertchen-sifive:
What is the correct behavior when there are multiple possible renames in the rename map for a `ReferenceTarget`?
For example, suppose the RenameMap contains `~Top|Top/a:A/b:B>ref.f1.f2` -> `~Top|Top/a:A/b:B>ref.f1.bar` and `~Top|B>ref.f1.f2` -> `~Top|B>ref.foo.f2`. Should `~Top|Top/a:A/b:B>ref.f1.f2` be renamed to `~Top|Top/a:A/b:B>ref.foo.bar` or `~Top|Top/a:A/b:B>ref.f1.bar`?

What about when the rename map contains `~Top|Top/a:A/b:B>ref.f1.f2` -> `~Top|Top/a:A/b:B>ref.foo.f1` and `~Top|B>ref.f1.f2` -> `~Top|B>ref.f1.bar`?
I also have the same question for instance paths.

azidar:
Given my intention, when renaming x -> y, its possible for y to still require renaming. Thus:

Example 1:

`~Top|Top/a:A/b:B>ref.f1.f2` -> `~Top|Top/a:A/b:B>ref.f1.bar`
`~Top|B>ref.f1.f2 -> ~Top|B>ref.foo.f2`

`~Top|Top/a:A/b:B>ref.f1.f2` should be renamed to `~Top|Top/a:A/b:B>ref.foo.bar`


Example 2:

`~Top|Top/a:A/b:B>ref.f1.f2` -> `~Top|Top/a:A/b:B>ref.foo.f2`
`~Top|B>ref.f1.f2` -> `~Top|B>ref.f1.bar`

`~Top|Top/a:A/b:B>ref.f1.f2` should also be renamed to `~Top|Top/a:A/b:B>ref.foo.bar`

It should be the same with instance paths. Whether this is possible, however, is a separate question. As I'm sure you are realizing, its a pretty complicated problem and I do appreciate you looking at it as well!

azidar:
I can also see how its not possible to get that, given the current code. I'm not sure I see a possible way to do it, either. Thoughts?

albertchen-sifive:
I think allowing multiple renames might lead to some unexpected renames. For example, if ref `foo.bar` gets lowered to `foo_bar` while at the same time a different ref `foo_bar` is renamed to `foo_foo`. Only allowing one round of renaming also matches the behavior of the old `RenameMap`. I think either option is fine though. There just needs to be more detailed documentation on the exact behavior of `RenameMap`.

azidar:
Regarding instances, I believe this should work as advertised though. I believe the problem with the component renaming is that there is no way to express the partial rename of `f2`, that is independent of `f1`. Thus, whichever rename you apply first, it won't trigger the other one.

azidar:
> For example, if ref foo.bar gets lowered to foo_bar while at the same time a different ref foo_bar is renamed to foo_foo.

This shouldn't be a problem, as `Uniquify` guarantees that the lowering of a signal will not conflict with an existing signal name.

> Only allowing one round of renaming also matches the behavior of the old RenameMap.

Actually that isn't the case, in the previous RenameMap, you could rename a signal name and a signal's module name separately, and it would do the double rename.

> I don't really know of a good way of doing the multiple renaming thing without blowing up the runtime

Actually I think the runtime shouldn't blow up because of the caching of the results of renaming. Its basically dynamic programming, and so we shouldn't worry about spawning multiple recursions per recursive call. I believe the final complexity is O(N^2), where N is the path+component length. Also, results are cached between multiple calls to get, so it should still be quite fast. We should do a more extensive performance analysis, however.

albertchen-sifive:
> > Only allowing one round of renaming also matches the behavior of the old RenameMap.
> 
> Actually that isn't the case, in the previous RenameMap, you could rename a signal name and a signal's module name separately, and it would do the double rename.
> 
I meant just the component name portion. I think the old rename map would not rename `ref.f1.f2` to `ref.foo.bar`, it would only rename on the first exact match from `ref.f1.f2` to `ref.f1.bar`, then rename the module name portion.

albertchen-sifive:
> Regarding instances, I believe this should work as advertised though. I believe the problem with the component renaming is that there is no way to express the partial rename of `f2`, that is independent of `f1`. Thus, whichever rename you apply first, it won't trigger the other one.

what about finding all renames of the original annotation then merging them at the end instead of chaining renames together? so instead of doing
1. `ref.f1.f2` -> `ref.foo.f2`
2. `ref.foo.f2` -> `ref.foo.bar`

do
1. `ref.f1.f2` -> `ref.foo.f2`
2. `ref.f1.f2` -> `ref.f1.bar`
3. merge `ref.foo.f2`, `ref.f1.bar` into `ref.foo.bar`

maybe do some sort of edit-distance thing for the merging part?

azidar:
> I meant just the component name portion. 

Ah yes. You are right.

> what about finding all renames of the original annotation then merging them at the end instead of chaining renames together?

Haha this feels almost like a DNA analysis-type program... I think it could certainly work, and it may un-complicate the actual code. IMO its definitely an idea worth exploring.

azidar:
So for instance paths:

`~Top|Top/a:A/b:B` -> `~Top|Top/foo:Foo/b:B`
`~Top|Top/a:A/b:B` -> `~Top|Top/a:A/bar:Bar`

I don't think should `~Top|Top/a:A/b:B` should rename to `~Top|Top/foo:Foo/bar:Bar`. It should probably rename to both `~Top|Top/foo:Foo/b:B` and `~Top|Top/a:A/bar:Bar`.

albertchen-sifive:
> Hmm, so I think the rename semantics should actually be that longer paths/components get priority over naming, and then iteratively you check smaller bits.

ok so for the previous examples
Example 1:

~Top|Top/a:A/b:B>ref.f1.f2 -> ~Top|Top/a:A/b:B>ref.f1.bar
~Top|B>ref.f1.f2 -> ~Top|B>ref.foo.f2

~Top|Top/a:A/b:B>ref.f1.f2 should be renamed to ~Top|Top/a:A/b:B>ref.foo.bar

Example 2:

~Top|Top/a:A/b:B>ref.f1.f2 -> ~Top|Top/a:A/b:B>ref.foo.f2
~Top|B>ref.f1.f2 -> ~Top|B>ref.f1.bar

~Top|Top/a:A/b:B>ref.f1.f2 should be renamed to ~Top|Top/a:A/b:B>ref.foo.f2?

also, do longer components or longer paths have priority?

azidar:
No, I think Example 1 is:

~Top|Top/a:A/b:B>ref.f1.f2 -> ~Top|Top/a:A/b:B>ref.f1.bar
~Top|B>ref.f1.f2 -> ~Top|B>ref.foo.f2

~Top|Top/a:A/b:B>ref.f1.f2 should be renamed to ~Top|Top/a:A/b:B>ref.*f1*.bar

Example 2 is correct.

albertchen-sifive:
ok, I see where I went wrong. I like your proposal, but I'm still not sure what the correct order of paths is. The way I interpreted your proposal for example 1 would be to check paths in this order:
```
1. ~Top|Top/a:A/b:B>ref.f1.f2
2. ~Top|Top/a:A/b:B>ref.f1
3. ~Top|Top/a:A/b:B>ref
4. ~Top|A/b:B>ref.f1.bar
5. ~Top|A/b:B>ref.f1
6. ~Top|A/b:B>ref
7. ~Top|B>ref.f1.bar
8. ~Top|B>ref.f1
9. ~Top|B>ref
```

but doing something like this:
```
1. ~Top|Top/a:A/b:B>ref.f1.f2
2. ~Top|A/b:B>ref.f1.bar
3. ~Top|B>ref.f1.bar
4. ~Top|Top/a:A/b:B>ref.f1
5. ~Top|A/b:B>ref.f1
6. ~Top|B>ref.f1
7. ~Top|Top/a:A/b:B>ref
8. ~Top|A/b:B>ref
9. ~Top|B>ref
```

also gives to the same result

jackkoenig:
We talked about this a bit during the developer's meeting and then @albertchen-sifive and I continued talking about it some more.

> Hmm, so I think the rename semantics should actually be that longer paths/components get priority over naming, and then iteratively you check smaller bits.

I too agree with this, because it supports the main two classes of renaming that I can think of in a fairly intuitive way:

### Uniquify-style 
I want to rename whole groups of things
In module `B`, given `wire w : { a : UInt<8>, b : UInt<8> }`
* RenameMap
    * `~Top|B>w -> ~Top|B>w_`
* Examples renames
    * `~Top|B>w.a -> ~Top|B>w_.a`
    * `~Top|Top/a:A/b:B>w -> ~Top|Top/a:A/b:B>w_`

### Lower Types
I want to rename long components precisely and split aggregates:
Given same wire as above
* Rename Map
    * `~Top|B>w -> ~Top|B>w_a, ~Top|B>w_b`
    * `~Top|B>w.a -> ~Top|B>w_a`
    * `~Top|B>w.b -> ~Top|B>w_b`
* Examples renames
    * `~Top|B>w.a -> ~Top|B>w_.a`
    * `~Top|Top/a:A/b:B>w -> ~Top|Top/a:A/b:B>w_a, ~Top|Top/a:A/b:B>w_b`

______

## Now for the potentially controversial bit

I don't think it makes sense to allow renaming Modules or instances at the same time as components. For example, Adam's corrected Example 1:

`~Top|Top/a:A/b:B>ref.f1.f2 -> ~Top|Top/a:A/b:B>ref.f1.bar`
`~Top|B>ref.f1.f2 -> ~Top|B>ref.foo.f2`

`~Top|Top/a:A/b:B>ref.f1.f2` should be renamed to `~Top|Top/a:A/b:B>ref.*f1*.bar`

The two pieces of the `RenameMap` are contradictory; how can you rename a field in just one instance of a module with multiple instances? I don't know what a circuit corresponding to that could possibly look like.

In order to do stuff like this you should be using `EliminateTargetPaths` first, and then just doing simple renames:
`~Top|B_1>ref.f1.f2 -> ~Top|B_1>ref.f1.bar`
`~Top|B>ref.f1.f1 -> ~Top|B>ref.foo.f2`

If my intuition here is right, then this can greatly simplify the implementation of renaming, we could split it into two much simpler implementations: `ModuleRenameMap` and `ComponentRenameMap`. This would have the added benefit of making it impossible to express illegal renames like the (Example 1) above.

azidar:
@albertchen-sifive so where do we stand with this? Is this something we need to talk about more?

albertchen-sifive:
I agree with all the things @jackkoenig said. I also think there should be something like a `PathlessReferenceTarget`. especially if we are going to disallow renaming of `ReferenceTargets` with paths.

azidar:
Ah, I think over another dev meeting we discussed how Jack's "controversial" proposal doesn't allow for expressing things like inlining, where an instance can be renamed to a module, e.g. `~Top|Top/a:A>ref -> ~Top|Top/a.ref`. The other thing to think about is that if someone wants to do something complicated, they are going to have to figure out the renaming logic themselves anyways. Why not spare them the burden and implement it so they don't have to think it through?

Regarding a `PathlessReferenceTarget`, I had tried implementing it when I first was playing with adding `Target` stuff, but it just felt unnecessary because `EliminateTargetPaths` easily enables any `PathlessReferenceTarget` to be referenced via a `ReferenceTarget`, and thus it didn't really make sense for annotations to enforce this type safety requirement. I'm open to playing with it more, however.

I agree we should pick an ordering. My opinion is to check (1) subcomponents, (2) paths, (3) change root module by descending into next instance. Given we are renaming `~Top|Top/a:A>ref.f1`, we should check the subparts in this order:

1. `~Top|Top/a:A>ref.f1`
2. `~Top|Top/a:A>ref`
3. `~Top|Top/a:A`
4. `~Top|Top>a`
5. `~Top|Top`
6. `~Top|A>ref.f1`
7. `~Top|A>ref`
8. `~Top|A`

azidar:
Three rules:
1) From -> To, To must be final (no subpiece of To will be renamed)
2) Check all relative paths to component (and subcomponents), before checking path to leaf module
3) Enable multiple rename maps to be created from one transform (e.g. Dedup has one renamemap containing module to instance (in terms of original modules), and second mapping module to deduped module).

Below is the checking order for renaming the top target, with no renames.

```scala
~Top|Top/a:A/b:B/c:C>f.g
~Top|Top/a:A/b:B/c:C>f
~Top|A/b:B/c:C>f.g
~Top|A/b:B/c:C>f
~Top|B/c:C>f.g
~Top|B/c:C>f
~Top|C>f.g
~Top|C>f
~Top|Top/a:A/b:B/c:C
~Top|Top/a:A/b:B>c
~Top|A/b:B/c:C
~Top|A/b:B>c
~Top|B/c:C
~Top|B>c
~Top|C
~Top|Top/a:A/b:B
~Top|Top/a:A>b
~Top|A/b:B
~Top|A>b
~Top|B
~Top|Top/a:A
~Top|Top>a
~Top|A
~Top|Top
```

albertchen-sifive:
There is a test case that I'm not sure about.  Suppose the rename map contains:
```
~Top|C -> ~Top|D/e:E/f:F
```
what should `~Top|A/b:B/c:C` be renamed to?

I think above rename might occur if C was undeduped by `EliminateTargetPaths` then split off by `GroupComponents`? in this case is we might want 
```
~Top|A/b:B/c:D/e:E/f:F
```

But I think for dedup the rhs is probably an absolute path? then we might want
```
~Top|D/e:E/f:F
```

Also should the behavior be the same if a module gets renamed to a path as part of a `ReferenceTarget` rename?

albertchen-sifive:
also, is renaming of reference targets with components that point to instances allowed?
for example, should `~Top|A>foo:Foo/bar:D` be renamed to `~Top|A/foo:Foo/bar:Bar` if the rename map contains `~Top|A/foo:Foo/bar:D -> ~Top|A/foo:Foo/bar:Bar`?

azidar:
RE:

```
~Top|C -> ~Top|D/e:E/f:F
```

If C is undeduped by `EliminateTargetPaths` and then split off by `GroupComponents`, it seems that we should split that into two rename maps, no?

If that is the case, then I don't think we would have to really care about the semantics, because its possible (I think, though we should check) to express either one with different rename maps. Thus, I think

```
~Top|A/b:B/c:C -> ~Top|A/b:B/c:D/e:E/f:F
```

is acceptable.

> Also should the behavior be the same if a module gets renamed to a path as part of a ReferenceTarget rename?

Yes, I think so. If someone does this though, they better rename all of the signals in the module so that other references to signals in the module get renamed before our algorithm gets to the module name.

> also, is renaming of reference targets with components that point to instances allowed?

I'm not sure exactly what you mean, but `~Top|A>foo:Foo/bar:D` is not a legal target. ReferenceTarget components cannot have `Instance` or `OfModule` `TargetToken`s in them, they can only have `Field`, `Index`, `Clock`, `Init`, and `Reset`.

albertchen-sifive:
ok thanks for the clarification. wasn't sure about the instance/ofModule component thing because there doesn't seem like there is anything preventing people from putting those tokens into a ReferenceTarget.

azidar:
> there doesn't seem like there is anything preventing people from putting those tokens into a ReferenceTarget

That's true. Maybe we should think about how to limit that.

src/main/scala/firrtl/RenameMap.scala
```diff
@@ -178,124 +178,257 @@ final class RenameMap private () {
...
+  }
+
+  /** Recursively renames a target so the returned targets are complete renamed
+    * @param errors Used to record illegal renames
+    * @param key Target to rename
+    * @return Renamed targets
+    */
+  private def recursiveGet(errors: mutable.ArrayBuffer[String])(key: CompleteTarget): Seq[CompleteTarget] = {
+    if(getCache.contains(key)) {
+      getCache(key)
```
azidar: Can we cache the other recursive calls like traverse right and traverse left? I think this could be important if lots of signals of a deep instance get renamed (or something like that). I'm open to not caching those functions as well, if you feel it has limited use.

src/main/scala/firrtl/passes/RemoveCHIRRTL.scala
```diff
@@ -120,8 +120,6 @@ object RemoveCHIRRTL extends Transform {
           clks += "clk"
           ens += "en"
           masks ++= create_exps(SubField(portRef, "wmask", createMask(sx.tpe)))
-          renames.rename(sx.name, s"${sx.mem}.${sx.name}.rdata")
-          renames.rename(sx.name, s"${sx.mem}.${sx.name}.wdata")
```
azidar: Why did you remove these lines?

src/main/scala/firrtl/passes/RemoveCHIRRTL.scala
```diff
@@ -134,7 +132,6 @@ object RemoveCHIRRTL extends Transform {
           clks += "clk"
           ens += "en"
           masks ++= create_exps(SubField(portRef, "mask", createMask(sx.tpe)))
-          renames.rename(sx.name, s"${sx.mem}.${sx.name}.data")
```
azidar: Ditto - Why did you remove these lines?

src/main/scala/firrtl/passes/RemoveCHIRRTL.scala
```diff
@@ -149,7 +146,6 @@ object RemoveCHIRRTL extends Transform {
               raddrs(e.name) = SubField(portRef, "en", ut)
             case _ => ens += "en"
           }
-          renames.rename(sx.name, s"${sx.mem}.${sx.name}.data")
```
azidar: Ditto - Why did you remove these lines?

src/main/scala/firrtl/RenameMap.scala
```diff
@@ -178,13 +178,151 @@ final class RenameMap private () {
...
+        Some(underlying(key))
+      } else {
+        key match {
+          case t: ReferenceTarget if t.component.nonEmpty =>
+            val last = t.component.last
+            val parent = t.copy(component = t.component.dropRight(1))
+            traverseTokens(parent).map(_.flatMap { x =>
+              (x, last) match {
+                case (t2: ReferenceTarget, Field(f)) => Some(t2.field(f))
+                case (t2: ReferenceTarget, Index(i)) => Some(t2.index(i))
```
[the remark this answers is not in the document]
albertchen-sifive: I think this is not allowed according to this test case? https://github.com/freechipsproject/firrtl/pull/967/files#diff-d4b2eb4d442e9245d392a91f972f5a5aR278
It would be easy to allow this case though.

src/main/scala/firrtl/passes/RemoveCHIRRTL.scala
albertchen-sifive: um don't remember why. I'll remove these. They're probably from some other thing I was working on.

src/main/scala/firrtl/RenameMap.scala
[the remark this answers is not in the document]
albertchen-sifive: there is also another case that I think I'm missing. for rename map `~Top|Top/i:Inline>b -> ~Top|Top>i_b` should we be able to rename `~Top|Top>i.b` to `~Top|Top>i_b`? It seems like something we should support since we allow renaming of references to instances.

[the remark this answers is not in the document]
albertchen-sifive: i think we need `case (t2: InstanceTarget, Field(f)) => Some(t2.ref(f))` here too? for the case where the base reference gets renamed to an instance.

and I don't think we need the `~Top|Top>foo.bar -> ~Top|Foo>bar` case. I think it makes things more clear to only allow renaming from `IsComponent` to `IsComponent` and `IsModule` to `IsModule`.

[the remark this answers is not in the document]
azidar: > and I don't think we need the `~Top|Top>foo.bar -> ~Top|Foo>bar` case. I think it makes things more clear to only allow renaming from `IsComponent` to `IsComponent` and `IsModule` to `IsModule`.
Yeah, I agree with you that this tradeoff is worth it. Let's not support that case.

> rename map: `~Top|Top/i:Inline>b -> ~Top|Top>i_b`, rename `~Top|Top>i.b` to `~Top|Top>i_b`?
I agree it is difficult because we don't know the `OfModule`. I think we have to punt on it and force the user to add `~Top|Top>i.b -> ~Top|Top>i_b` to the rename map.

src/main/scala/firrtl/RenameMap.scala
```diff
@@ -178,13 +178,151 @@ final class RenameMap private () {
...
+      val getOpt = key match {
+        case t: InstanceTarget if underlying.contains(t) => underlying.get(t)
+        case t: InstanceTarget => underlying.get(t.asReference)
+        case t: ModuleTarget => underlying.get(t)
+      }
+
+      if (getOpt.nonEmpty) {
+        getOpt.map(_.flatMap {
+          case isMod: IsModule => Some(isMod)
+          case ref @ ReferenceTarget(c, m, p, r, Nil) =>
```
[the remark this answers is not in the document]
azidar: Also, can you add a test case demonstrating your fix works?

azidar:
Another thought - I think it would be straightforward to enable multiple rename maps in this PR.

We just need to
(1) add a new argument to `RenameMap`, `earlier: Option[RenameMap]`. Default construction of a `RenameMap` sets this to `None` (e.g. in `RenameMap.apply`)
(2) add `def nextRenameMap(): RenameMap` which creates a new `RenameMap` whose `earlier` argument is `Some(this)`.
(3) update `def completeGet(...)` so, if `earlier.nonEmpty == true`, first call `earlier.get.get(key)`, and then do normal rename on result.

This should allow chaining of RenameMaps, as each one will try to rename, then call the previous (recursively) until all renames are complete. Thoughts? Do you think you could add this to this PR?

Also I don't know why travis is being weird....

jackkoenig:
I would propose a slightly different approach, rather than continuing to pile things into the existing `RenameMap`, I would open the API up and have multiple types, `DefaultRenameMap` which is default version, `MultiRenameMap` which chains lookups to multiple, and some way to DIY since I'm realizing that certain transforms do a ton of work recording all possible renames into a map structure, when in reality they could just provide a function to do the renaming.

azidar:
> some way to DIY since I'm realizing that certain transforms do a ton of work recording all possible renames into a map structure, when in reality they could just provide a function to do the renaming.

Ooof, this would need to be done with extreeeeeme care. It has been hard enough to figure out a sane rename function, we should be very very careful about this. One option I'm more ok with is for the underlying map for tracking the rename changes to be `Target -> (Target => Seq[Target])` instead of `Target -> Seq[Target]`, thus allowing more flexibility for recording changes, but still letting us check the return values of the renaming function to preserve existing rename semantics. However, I'm very very hesitant about letting people override something like `get`.

> I would open the API up and have multiple types, DefaultRenameMap which is default version, MultiRenameMap which chains lookups to multiple

That might work, I'd need more details. What I like about what I proposed is I can see a deprecation path towards a lazy renaming scheme where annotations are not updated until the transform that consumes them (sort of like how `ResolvedAnnotationPaths` turns per-instance targets into local targets). By restricting people to this mechanism, then we can check between transforms that they always use `def nextRenameMap(): RenameMap` instead of creating a new RenameMap.

jackkoenig:
@azidar I agree with your trepidation toward opening up the API.

> One option I'm more ok with is for the underlying map for tracking the rename changes to be Target -> (Target => Seq[Target]) instead of Target -> Seq[Target]

Thats an interesting proposal. I agree that the rename function has been such a nightmare that I also worry about allowing people to potentially screw it up. The one thing that makes me think it still may be worth considering opening up is that I suspect much of the complexity of this implementation comes from the fact that it implements something of a catch all. I suspect that using different data structures might result in simpler renaming code. In any case, I think we should table my idea until I or someone else actually writes what a `LowerTypesRenameMap` would look like so that we can actually compare.

The one thing I don't like about the `Target -> (Target => Seq[Target])` idea is that you still have to populate a data structure with all of the targets you might rename. In the case of `LowerTypes` or `Uniquify`, you still have to "unfold" any aggregates out to all of their leaves and populate the Map with those Targets even if they all point to the same function at the end.

> That might work, I'd need more details. What I like about what I proposed is I can see a deprecation path towards a lazy renaming scheme where annotations are not updated until the transform that consumes them

I agree with this vision, but I don't think any changes to the API are needed for this. We currently have Transforms return a `RenameMap`, there's nothing that says those renames need to be applied (yet). Since we also already pass transforms an `AnnotationSeq` (not `Seq[Annotation]`), the rename maps could be stored in that object such that when given annotations are requested we only rename as necessary. I'm unsure exactly how to implement this, but perhaps some sort of `collectLike` function could do it (some tricky details come to mind, you can't do a general collect because the PartialFunction can peer into the Annotations and inspect their Targets so you would need a `collectByClass` or something that returns Annotations of the right type).

Forgetting about my proposal above to "open" the API, we could still have `MultiRenameMap` which allows Transforms to specify that you need to apply multiple renamings.

jackkoenig:
That's a lot so maybe I should get back to the more direct issue at hand in this PR.

Rather than adding any notion of multiple `RenameMaps` to `RenameMap`, I would just make a new type `MultiRenameMap`. It seems to me that they should implement a common interface, where looking stuff up in the `MultiRenameMap` delegates to calling each one of its constituent `RenaneMaps` in order, thoughts @azidar?

jackkoenig:
Discussed in dev meeting, decided not to have `MultiRenameMap`, but to have this notion of chained rename maps, perhaps with an `andThen` method or something similar.

@azidar something that just occurred to me is that there is awkwardness with mutability here, if I have `renameMap1.andThen(renameMap2)`, does this mutate `renameMap1` or return a new `renameMap`? If it returns a new one, what happens if you mutate it? I would like the ability to have immutable rename maps but that's probably future work.

azidar:
I think you have to do `val renameMap2 = renameMap1.andThen()`?

albertchen-sifive:
and the `++` operator should recursively call `++` on the `earlier` rename maps if they both exist?

jackkoenig:
If we're mutating, I don't think we should call it `++` nor `andThen` because both of those imply new objects in the Scala collections.

We also shouldn't use `++` anyway because of how this analogizes to Maps:
```scala
scala> val m1 = Map("a" -> "b")
m1: scala.collection.immutable.Map[String,String] = Map(a -> b)

scala> val m2 = Map("b" -> "c")
m2: scala.collection.immutable.Map[String,String] = Map(b -> c)

scala> val added = m1 ++ m2
added: scala.collection.immutable.Map[String,String] = Map(a -> b, b -> c)

scala> val chained = m1.andThen(m2)
chained: PartialFunction[String,String] = <function1>

scala> added("a")
res0: String = b

scala> chained("a")
res1: String = c
```

albertchen-sifive:
> If we're mutating, I don't think we should call it `++` nor `andThen` because both of those imply new objects in the Scala collections, we should probably just pick something else.

I wasn't proposing to call it `++`. There is already a `++` operator defined in the current `RenameMap` https://github.com/freechipsproject/firrtl/blob/master/src/main/scala/firrtl/RenameMap.scala#L116

I just wanted to clarify what the behavior of `++` should be if we call `++` on two chained `RenameMaps`.

jackkoenig:
> I just wanted to clarify what the behavior of ++ should be if we call ++ on two chained RenameMaps.

@albertchen-sifive right, sorry. I *think* `m1 ++ m2` is analogous to `m2.orElse(m1)` in that later ones overrule earlier ones. Thus I think the functionality is actually checking from later to earlier until the key is found.

albertchen-sifive:
> RE:
> 
> ```
> ~Top|C -> ~Top|D/e:E/f:F
> ```
> If C is undeduped by `EliminateTargetPaths` and then split off by `GroupComponents`, it seems that we should split that into two rename maps, no?
> 
> If that is the case, then I don't think we would have to really care about the semantics, because its possible (I think, though we should check) to express either one with different rename maps. Thus, I think
> 
> ```
> ~Top|A/b:B/c:C -> ~Top|A/b:B/c:D/e:E/f:F
> ```
> is acceptable.
> 
> > Also should the behavior be the same if a module gets renamed to a path as part of a ReferenceTarget rename?
> 
> Yes, I think so. If someone does this though, they better rename all of the signals in the module so that other references to signals in the module get renamed before our algorithm gets to the module name.
> 
> > also, is renaming of reference targets with components that point to instances allowed?
> 
> I'm not sure exactly what you mean, but `~Top|A>foo:Foo/bar:D` is not a legal target. ReferenceTarget components cannot have `Instance` or `OfModule` `TargetToken`s in them, they can only have `Field`, `Index`, `Clock`, `Init`, and `Reset`.

now that I think about it some more I don't think we should inline the paths of renamed `OfModule`s. `~Top|A/b:B/c:C` should just remain the same in this case. If we rename `OfModule`s to instance paths then that will mess up other instance paths; for example renaming `~Top|A/b:B/c:C/z:Z` to `~Top|A/b:B/c:D/e:E/f:F/z:Z` doesn't make much sense, but it is currently unavoidable if we also want to do `~Top|C -> ~Top|D/e:E/f:F`

albertchen-sifive:
I think we should still be able to express all the renames we need without allowing `OfModule` to `InstanceTarget` renames. For dedup we would first rename all `ModuleTarget`s and relative `InstanceTarget`s paths to absolute paths. Then chain that rename map with a map that only renames `ModuleTargets` to deduped `ModuleTargets`. So for the following example:
```
    top: Top
       |
      b: B
     /     \
    /       \
 a1: A1   a2: A2
```

the rename map would contain
```
~Top|A1 -> ~Top|Top/b:B/a1:A1
~Top|B/a1:A1 -> ~Top|Top/b:B/a1:A1
~Top|A2 -> ~Top|Top/b:B/a2:A2
~Top|B/a2:A2 -> ~Top|Top/b:B/a2:A2
```

chained with
```
~Top|A1 -> ~Top|A
~Top|A2 -> ~Top|A
```
we wouldn't want to rename `OfModule`s to instances, otherwise `~Top|B/a1:A1` would be renamed to `~Top|Top/b:B/a1:Top/b:B/a:A`

src/main/scala/firrtl/RenameMap.scala
```diff
@@ -35,7 +35,17 @@ object RenameMap {
   * These are mutable datastructures for convenience
   */
 // TODO This should probably be refactored into immutable and mutable versions
-final class RenameMap private () {
+final class RenameMap private (earlier: Option[RenameMap] = None, previous: Option[RenameMap] = None) {
```
jackkoenig: earlier and previous both sound the same to me

albertchen-sifive:
sorry, just realized an issue the above renaming strategy; it messes up existing absolute paths. one solution is to allow renaming annotations to themselves to short-circuit our instance path renaming algorithm. so in order to prevent  `~Top|Top/b:B/a2:A2` from being renamed to `~Top|Top/Top/b:B/a2:A2` we would have to have `~Top|Top/b:B/a2:A2 -> ~Top|Top/b:B/a2:A2` in the rename map

azidar:
Sorry this took a while to get to. I do think we want to rename `OfModule`s in the instance path. 

Given 

```
~Top|C -> ~Top|D/e:E/f:F
```

I think it actually makes sense to rename

```
~Top|A/b:B/c:C/z:Z -> ~Top|A/b:B/c:D/e:E/f:F/z:Z
```

To me, the rename map is basically saying "for all modules of `C`, we've added two wrapper modules `D` and `E`, and also renamed `C` to `F`". Thus, `z` should still be a member of `F`. If it isn't, then we should delete `~Top|C>z` with another renaming rule.

chick:
@jackkoenig suggests we push this back to 1.3. Would like to hear @azidar weigh in on this

albertchen-sifive:
ready for re-review. summary of changes
- remove support for reference targets that point to instances
- short-circuit renaming algorithm if the rename map returns an absolute target
  - absolute targets are targets where the `module` is the same as the `circuit`
- make `andThen` take a `RenameMap` as an argument and return a new, chained `RenameMap` without copying the `underlying` maps of the chained `RenameMap`s
- rename `earlier` to `chained`
- deprecate `++`
- get rid of `orElse`

src/main/scala/firrtl/RenameMap.scala
```diff
@@ -176,140 +197,289 @@ final class RenameMap private () {
...
+
+  /** Checks for renames of only the path portion of an [[InstanceTarget]]
+    * Recursively checks parent [[IsModule]]s until a match is found
+    * First checks all parent paths from longest to shortest. Then
+    * recursively checks paths leading to the encapsulating module.
+    * Stops on the first match found.
+    *
+    * For example, the order that targets are checked for ~Top|Top/a:A/b:B/c:C is:
+    * ~Top|Top/a:A/b:B/c:C
+    * ~Top|Top/a:A/b:B>c
```
azidar: Update comment to remove the reference targets

src/main/scala/firrtl/RenameMap.scala
```diff
@@ -332,8 +506,7 @@ final class RenameMap private () {
   @deprecated("Use record with IsMember instead, this will be removed in 1.3", "1.2")
   def rename(from: ComponentName, to: ComponentName): Unit = record(from, to)
 
-  @deprecated("Use record with IsMember instead, this will be removed in 1.3", "1.2")
-  def rename(from: ComponentName, tos: Seq[ComponentName]): Unit = record(from, tos.map(_.toTarget))
+  @deprecated("Use record with IsMember instead, this will be removed in 1.3", "1.2")def rename(from: ComponentName, tos: Seq[ComponentName]): Unit = record(from, tos.map(_.toTarget))
```
azidar: Missing a newline after the deprecation?

albertchen-sifive:
@azidar comments addressed

jackkoenig:
I'm so excited we finally got this merged! Great job @albertchen-sifive and @azidar!!!

PR #969, opened 2018-12-14
[Top Wiring] Expand top wiring to work on aggregates

davidbiancolin:
There was fundamentally no reason top wiring could not act on aggregates. I changed the form to reflect this, and added a couple tests to check cases involving aggregates.

### DIFF
--- a/src/main/scala/firrtl/transforms/TopWiring.scala
+++ b/src/main/scala/firrtl/transforms/TopWiring.scala
@@ -36,8 +36,8 @@ case class TopWiringAnnotation(target: ComponentName, prefix: String) extends
   * @note This *does* work for deduped modules
   */
 class TopWiringTransform extends Transform {
-  def inputForm: CircuitForm = LowForm
-  def outputForm: CircuitForm = LowForm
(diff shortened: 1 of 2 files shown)


albert-magyar:
I see there was some discussion in #798 about the pros and cons of this. @alonamid, want to chime in?

davidbiancolin:
Derp, i should have looked at that.

alonamid:
I agree that midForm might be more useful from a general wiring perspective. However, from the perspective of my original use-case (ILA wiring in FireSim), and probably any other use-case that performs this top wiring for the purposes of verilog-hacks, I still need details about the individual wires.

albert-magyar:
There’s nothing stopping you from using it on a no-aggregate circuit,
though.

davidbiancolin:
Exactly -- for the ILA wiring we can just schedule this once we have low form. For print synthesis my life is made much easier if i can wire out aggregates.

alonamid:
OK. As long as I can get a "verilog" wire level mapping using low form (by scheduling it once we have low form), then I'm fine with it.

PR #971, opened 2018-12-14
Refactor "xTests" into "x"

### LINKED ISSUE #970: Refactor firrtlTests under firrtl package
**Type of issue**: other enhancement

The separation of `firrtlTests` from `firrtl` results in some annoyances (or lots of monkeying around) to test private behavior. E.g., consider:

```scala
package firrtl

class X private [firrtl] 
```

If I can access this constructor from `firrtl`, it seems logical that I'd like to test it's behavior in `firrtlTests`. However, by placing all the tests inside `firrtlTests`, this is private to testing. There's some hackery you can do to get at private methods, but that seems inelegant.

As far as I understand it, there's no need to place tests in a separate package. Those jars are built separately and not included in any fat jar assembly or publishing (as far as I'm aware).

I'd go one step further with this and suggest that the testing packages should match the source packages exactly in terms of names and directory structure.

Make it so that you can test private things in package `firrtl` or sub-packages.

This problem exists for some of the other BIG6 packages (Chisel notably).

seldridge:
This canonicalizes the test package names. Summarily:
  - `firrtlTests` -> `firrtl`
  - `firrtlTests.annotationTests` -> `firrtl.annotations`
  - `loggertests` -> `logger`

This does not touch the structure of the old `firrtlTests`.

Fixes #970.

Merging this will likely introduce merge conflicts for all filed PRs that touch tests...


azidar:
I think since its just cosmetic changes, it should be fine for existing PR's

jackkoenig:
Do you think it should be `firrtlTests` -> `firrtl.tests`? The main reason I ask is if we did some day publish the firrtl testing stuff for whatever reason, we probably don't want it to pollute the `firrtl` package?

seldridge:
Good thought, @jackkoenig. That hadn't occurred to me. That would then mean that there would be a `tests` package below every package node, e.g., the tests for `x` would be in `x.tests` while `x.y.z` would be in `x.y.z.tests`. That makes sense to me.

seldridge:
Closing this in favor of fine-grained packaging as needed.

PR #976, opened 2018-12-19
Fix NoneCompiler outputForm

seldridge:
Every `Compiler` has an implicit assumption that it's transform sequence is non-empty as `inputForm` and `outputForm` are defined as follows:

```scala
val transforms: Seq[Transform]
val inputForm: CircuitForm = transforms.head.inputForm
val outputForm: CircuitForm = transforms.last.outputForm
```

This causes problems for `NoneCompiler` as this is an empty transform sequence. This PR fixes this in three ways:

1. An `IdentityTransform` is added that applies an identity function to some `CircuitState`
2. The `IdentityTransform` is used as the lone transform in the `NoneCompiler`
3. A run time requirement is added that checks to see if a `Compiler` was constructed with an empty sequence of transforms.

I'd like to be able to do (3)'s check at compile time, but I think that would require dependent types...

### DIFF
--- a/src/main/scala/firrtl/Compiler.scala
+++ b/src/main/scala/firrtl/Compiler.scala
@@ -373,6 +373,10 @@ trait Compiler extends LazyLogging {
     */
   def transforms: Seq[Transform]
 
+  require(transforms.size >= 1,
+          s"Compiler transforms for '${this.getClass.getName}' must have at least ONE Transform! " +
+            "Use IdentityTransform if you need an identity/no-op transform.")
+
   // Similar to (input|output)Form on [[Transform]] but derived from this Compiler's transforms
   def inputForm: CircuitForm = transforms.head.inputForm
   def outputForm: CircuitForm = transforms.last.outputForm
--- a/src/main/scala/firrtl/LoweringCompilers.scala
+++ b/src/main/scala/firrtl/LoweringCompilers.scala
@@ -2,6 +2,8 @@
 
 package firrtl
 
+import firrtl.transforms.IdentityTransform
+
 sealed abstract class CoreTransform extends SeqTransform
 
 /** This transforms "CHIRRTL", the chisel3 IR, to "Firrtl". Note the resulting
@@ -132,7 +134,7 @@ import firrtl.transforms.BlackBoxSourceHelper
   */
 class NoneCompiler extends Compiler {
   def emitter = new ChirrtlEmitter
-  def transforms: Seq[Transform] = Seq.empty
+  def transforms: Seq[Transform] = Seq(new IdentityTransform(ChirrtlForm))
 }
 
 /** Emits input circuit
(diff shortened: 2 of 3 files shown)


seldridge:
This is updated to use a better solution that fixes the general problem of `Compiler` requiring a non-empty transform sequence. The PR text is updated accordingly.

src/main/scala/firrtl/Compiler.scala
```diff
@@ -373,6 +373,9 @@ trait Compiler extends LazyLogging {
     */
   def transforms: Seq[Transform]
 
+  require(transforms.size >= 1,
+          s"Compiler transforms for '${this.getClass.getName}' was empty. A compiler must have at least ONE Transform!")
```
jackkoenig: It would be backwards incompatible, but we could consider changing the type of transforms to `::[Transform]` and then that would enforce non-empty transforms in the type system.

src/main/scala/firrtl/LoweringCompilers.scala
```diff
@@ -132,7 +134,7 @@ import firrtl.transforms.BlackBoxSourceHelper
   */
 class NoneCompiler extends Compiler {
   def emitter = new ChirrtlEmitter
-  def transforms: Seq[Transform] = Seq.empty
+  lazy val transforms: Seq[Transform] = Seq(new IdentityTransform(ChirrtlForm))
```
jackkoenig: Why `lazy val`? It doesn't really matter either way I guess

src/main/scala/firrtl/Compiler.scala
seldridge: This is way cleaner (and I had no idea you could do that!). I would like to break it, but I defer to you re: backwards incompatible changes.

src/main/scala/firrtl/LoweringCompilers.scala
seldridge: The require I added is eager, so this is either a def or a lazy val. It seemed better to be a val to prevent reconstructing the transform sequence everytime with a def.

Removing the require and doing `::` could make this a val again.

src/main/scala/firrtl/Compiler.scala
edwardcwang: Is it worth explicitly adding something like `IdentityTransform can be used as an identity/no-op transform.`?

seldridge: I propose that we kick this down the road, potentially for more discussion. I think that the Dependency API may deprecate the Compiler. It's likely not worth it to introduce multiple backwards incompatible changes for the purposes of type safety.

src/main/scala/firrtl/LoweringCompilers.scala
seldridge: This is the odd man out here, changing back to def.

src/main/scala/firrtl/Compiler.scala
seldridge: Good idea, I'll include this.

PR #979, opened 2018-12-21
Remove GhpagesPlugin.

maloneymr:
I don't know if this GhpagesPlugin is important to anyone, but I'm proposing removing it from firrtl and Chisel3 to eliminate some friction it is causing in another project.

If this plugin is in fact useful to anyway, I would like to make it configurable so that it may be disabled where it is not needed.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,10 +4,6 @@
 
 enablePlugins(SiteScaladocPlugin)
 
-enablePlugins(GhpagesPlugin)
-
-git.remoteRepo := "git@github.com:freechipsproject/firrtl.git"
-
 // Firrtl code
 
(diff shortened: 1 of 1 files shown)


maloneymr:
The related PR for Chisel3: https://github.com/freechipsproject/chisel3/pull/966

azidar:
Thanks @maloneymr !

Git blame: https://github.com/freechipsproject/firrtl/commit/c5627f355767d4c4f5232fee81a61b7e8510672f

I'll let @ucbjrl respond?

ucbjrl:
I'm merging this into a new branch first so I can revive the old source repository reference.

PR #983, opened 2018-12-27
Allow loadMemoryFromFile(MemBase[T <: Bundle])

nhynes:
@see https://github.com/freechipsproject/chisel3/issues/968

Linked with https://github.com/freechipsproject/chisel3/pull/970 and https://github.com/freechipsproject/treadle/pull/77. 

This PR is _almost_ ready to go. During testing, it seems as though there's a second pass in which the original annotation exists but the memories have already been split.

Update: the second pass happens [here](https://github.com/freechipsproject/treadle/blob/master/src/main/scala/treadle/executable/ExecutionEngine.scala#L430). I'm still not sure why the original `LoadMemoryAnnotation`s are still there, though.


chick:
@nhynes we appreciate your digging into this. I thought I had already done some testing of this case and I thought the annotation renaming pass in firrtl should have taken care of this problem. I want to check that out first, then I will go over your PR here. Thanks for the help

nhynes:
Hrm. If it makes any difference, I'm using `chisel-testers2` because it actually handles `DecoupledIO` in a not-[unreasonable way](https://github.com/freechipsproject/chisel-testers/issues/220).

Thanks a lot for your help!

nhynes:
Actually, I think that this might be a bug elsewhere. Upon closer inspection, the renames are working initially, but aren't preserved on the curious second pass.

PR #984, opened 2018-12-27
Correctly handle dots in loaded memory paths

nhynes:
**Test code:**

```scala
loadMemoryFromFile(theBundleMem, getClass.getResource("/init_mem").getPath)
```

**Current behavior:**&nbsp;&nbsp;&nbsp; `.../target/scala-2_SUBDATA.12/test-classes/init_mem`

**Proposed behavior:** `.../target/scala-2.12/test-classes/init_mem_SUBDATA`


nhynes:
comments? like in the code or the issue? the example in "Current behavior" is indeed an actual example if you replace SUBDATA with the name of a bundle element.

seldridge:
Possibly even better would be a testcase.

chick:
I didn't notice that, got stuck looking at the diff. I get it now. Let me look at it again.  And yes a test case would be nice, not a full blown test, just create the annotation and call getFileName. It's hard to complain too much about the lack of a test when the original code did not have a test, but if you can do a quick one that would be great.

PR #987, opened 2019-01-04
Fix grouping

jackkoenig:
This fixes 2 bugs in GroupComponents, from the commit messages:

1. Make GroupComponents run ResolveKinds

    This fixes an issue where expressions created by GroupComponents would
be improperly lowered because they were not marked as references to
instance ports.

1. Fix GroupComponents to work with unused components

    Previously, components that did not affect the output would cause
exceptions because they were missing from the label2group Map. This
commit treats them as "reachable" by the ports so they are included in
the default "ungrouped" group.

### DIFF
--- a/src/main/scala/firrtl/transforms/GroupComponents.scala
+++ b/src/main/scala/firrtl/transforms/GroupComponents.scala
@@ -4,7 +4,7 @@ import firrtl._
 import firrtl.Mappers._
 import firrtl.ir._
 import firrtl.annotations.{Annotation, ComponentName}
-import firrtl.passes.{InferTypes, LowerTypes, MemPortUtils}
+import firrtl.passes.{InferTypes, LowerTypes, MemPortUtils, ResolveKinds}
 import firrtl.Utils.kind
 import firrtl.graph.{DiGraph, MutableDiGraph}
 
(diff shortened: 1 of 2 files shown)


jackkoenig:
To elaborate on the resolve kinds issue, the Expressions referring to ports of a groups submodule were not `InstanceKind`s, so they were being lowered, turning the `.` into an `_`

src/main/scala/firrtl/transforms/GroupComponents.scala
```diff
@@ -119,6 +119,11 @@ class GroupComponents extends firrtl.Transform {
       }
     }
 
+    // Unused nodes are reachable from any group nor the root, add them to root group
```
chick: Minor quibble, but I don't understand this comment

jackkoenig: The comment is missing an `not` in front of `reachable`, should be:
`Unused nodes are not reachable from any group nor the root, add them to root group`

`reachableNodes` is a Map from labels to the nodes that label can reach. The "default" or "root" label (in the code as `""`) should retain any nodes that are not part of one of the groups.

Since the code is using DiGraph reachability, nodes that are not connected to anything don't get put in any of the groups; thus giving an error when they're looked up later.

Have any ideas on how to better communicate this?

chick: I think adding the missing *not* is enough, I thought that was what it was supposed to be but wasn't sure

jackkoenig: I updated the comment

PR #991, opened 2019-01-11
Towards FPGA's programming-reset support

johnsbrew:
### Commit message
Current state of firrtl unchanged
Tests passed flawlessly except for the serialized protobuf GCDTester.pb that needed an upgrade
Note: made a python script to upgrade such *.pb files automatically, feel free to ask it

## Warning
This first commit does not break anything (except chisel) but work is still in progress for an actual integration of the added 'prog' => value construct.

### Context
- My first FPGA-oriented contribution as advertised on the [chisel-user group](https://groups.google.com/forum/#!msg/chisel-users/T_2jYc97Nos/vncZHkTBFAAJ)
- Call for comments and base of work for today's dev meeting 

### The issue (and usefulness) of FPGA's programming reset
1. Routing FPGA is hard and long (especially Xilinx's ones but that is not the point)
2. Reset is primordial on ASIC since at t=0 (power-up) register's states are unknown
3. FPGA offers the ability to set any register at a precise state at t=0 (programming)
and a simple way of expressing this in verilog is for example reg my_valid = '0;
4. Using this "programming-reset" allows us to kick out the reset tree routing constraint which highly improves the routing capabilities of the design

5. Chisel offers automatic clock & reset inference, which is nice but comes with a drawback for us :
val my_valid  = RegInit(false.B) will infer a synchronous reset process for my_valid which is not what we want.

I would hence like to somehow add an option to FIRRTL in order to would translate Chisel's val x = RegInit(value) into (downto verilog) reg x = value;

### DIFF
--- a/src/main/scala/firrtl/util/ClassUtils.scala
+++ b/src/main/scala/firrtl/util/ClassUtils.scala
@@ -1,0 +1,19 @@
+package firrtl.util
+
+object ClassUtils {
+  /** Determine if a named class is loaded.
+    *
+    * @param name - name of the class: "foo.bar" or "org.foo.bar"
+    * @return true if the class has been loaded (is accessible), false otherwise.
+    */
+  def isClassLoaded(name: String): Boolean = {
+    val found = try {
+      Class.forName(name, false, getClass.getClassLoader) != null
+    } catch {
+      case e: ClassNotFoundException => false
+      case x: Throwable => throw x
+    }
+//    println(s"isClassLoaded: %s $name".format(if (found) "found" else "didn't find"))
+    found
+  }
+}
--- a/src/main/scala/firrtl/util/TestOptions.scala
+++ b/src/main/scala/firrtl/util/TestOptions.scala
@@ -1,0 +1,13 @@
+package firrtl.util
+
+import ClassUtils.isClassLoaded
+
+object TestOptions {
+  // Our timing is inaccurate if we're running tests under coverage.
+  // If any of the classes known to be associated with evaluating coverage are loaded,
+  //  assume we're running tests under coverage.
+  // NOTE: We assume we need only ask the class loader that loaded us.
+  // If it was loaded by another class loader (outside of our hierarchy), it wouldn't be available to us.
+  val coverageClasses = List("scoverage.Platform", "com.intellij.rt.coverage.instrumentation.TouchCounter")
+  val accurateTiming = !coverageClasses.exists(isClassLoaded(_))
+}
--- a/src/test/scala/firrtlTests/UniquifySpec.scala
+++ b/src/test/scala/firrtlTests/UniquifySpec.scala
@@ -12,6 +12,7 @@ import firrtl._
 import firrtl.annotations._
 import firrtl.annotations.TargetToken._
 import firrtl.transforms.DontTouchAnnotation
+import firrtl.util.TestOptions
 
 class UniquifySpec extends FirrtlFlatSpec {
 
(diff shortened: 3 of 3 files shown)


johnsbrew:
Note: just seen the failed check : it fails on chisel's side so I guess this test will always fail on every single API modification => might need to review the flow somehow ?

seldridge:
First time contributor, woot. :confetti_ball: Thanks for the PR.

On a purely logistical side, for API breakages that involve Chisel, you can include the `[skip chisel tests]` in any commit message in that PR range. That will cause Travis on the FIRRTL side to skip running Chisel tests. Obviously, merging one of these commits will immediately cause all Chisel CI to break... For that reason, PRs on the FIRRTL side that break Chisel are generally expected to have a companion Chisel PR ready to merge.

As this is an IR change, this makes sense to have a discussion about first before investing time in aligning Chisel, IMO.

Towards actual feedback: I think what you're trying to do can be done without modifications to the IR using the third option you present in the chisel-users discussion. What I think you're saying is that if you're targeting an FPGA, you want to have different Verilog emission behavior, but I'm not convinced that you need any additional information encoded in the IR to do this. The information (that you want to use different reset emission) would seem to be better encoded as an annotation (e.g., `InitializeRegistersForFpgaAnnotation`). Register randomization is already blocked via Verilog `ifdefs. One current annoyance with this approach is that there's not a good way to inject transforms that change Verilog emission behavior unless you modify the emitter. I should have infrastructure in place that will make this easier soon... 

Complete drive-by analysis, though.

johnsbrew:
Hi @seldridge and thank you for the feedback

I discussed the 3 options with my advisor and our conclusion was the following:
- standard reset are also very useful for many (if not all) FPGA's design since using "programming-reset" comes with some drawbacks
- even within a given module one may need to handle both flavors (already seen in our designs)
- the case of a register with both flavors at the same time does not make sense to us however
- hence the need for a choice within chisel for every single register and not a global setting, even module by module
- same for randomization, it should be desactivated only on those registers (we actually foresee it as a great way of finding bugs when man forgets to set the programming reset) 

I do understand your concern about adding an option to the IR that concerns only FPGAs, but IMO it is just like with verilog : some statements (like the ones we are talking about generating) only make sense for a given target.

Concerning the "time investement" part, I see it more like an onboarding into Chisel and Firrtl code base which is quite nice :)

Let's discuss all that at Today's meeting

seldridge:
Note: it doesn't have to be an all-or-nothing thing due to the ability of FIRRTL's Annotations to be attached to one or more specific targets within a circuit.

It would be entirely reasonable to use annotations to control the emission on a register-by-register basis. This can then be easily exposed via a Chisel-level API that doesn't impact FIRRTL IR or require a new Chisel core hardware type.

Good attitude about the time investment. I share your thoughts.

jackkoenig:
Let's definitely discuss at the developer's meeting today (well in a few minutes), but I had a couple of questions and comments.

Reusability is perhaps the most important design ethos of Chisel/FIRRTL. Ideally any Chisel design can be mapped efficiently to any implementation (using backend-specific FIRRTL optimizations and emission).

> * standard reset are also very useful for many (if not all) FPGA's design since using "programming-reset" comes with some drawbacks
> * even within a given module one may need to handle both flavors (already seen in our designs)

Can you provide more context or examples as to what these cases are? Are they mechanically differentiable so that it could be automated?

My intuition for how this could be done with a transform is that all registers reset to a literal and connected to a single reset source from the top are treated as "programming-reset" registers, while registers who are reset based on logic are not (since that would be functionally incorrect). I confess I'm not super familiar with FPGA "programming-reset" and design methodologies that use it, so I may be entirely off base.

johnsbrew:
Thank you for your appreciated feedbacks on Friday, I will now change the direction and start to investigate on:
- annotations 
- FIRRTL verilog emitter
Cheers

PR #994, opened 2019-01-17
Improve Shl codegen; eliminate Shlw WIR node

aswaterman:
If we emit `shl(x, k)` as `{x, k'h0}` instead of `x << k`, then there's no need for Verilog-specific padding in the PadWidths pass.  Avoiding the redundant padding improves compiler/simulator performance and renders the `Shlw` working-IR node unnecessary.

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -303,8 +303,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
          case (_: SIntType) => Seq(cast(a0)," >>> ", a1)
          case (_) => Seq(cast(a0), " >> ", a1)
        }
-       case Shlw => Seq(cast(a0), " << ", c0)
-       case Shl => Seq(cast(a0), " << ", c0)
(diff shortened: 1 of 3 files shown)


aswaterman:
I assume your response was meant for my other Chisel PR? :)

seldridge:
Browser tabs ruining my life.

Unrelated: now that this is the first PR under the CLA, should we add a license and start updating license text of files touched as also Apache v2?

aswaterman:
Yeah I think so.

ucbjrl:
I'd like to get clarity on this from Krste/Jonathan before we claim this is licensed under Apache v2.
(We should probably move the src/LICENSE.txt file up to the top of the repository.)

jackkoenig:
I can confirm that the failing test (ICache LEC) is a false negative, it either ran out of time or memory. I ran it manually on my laptop and it passed.

schoeberl:
Historically BSD 2-clause license would fit best :-) Right?

Cheers,
Martin

aswaterman:
Ping.

PR #937, opened 2018-11-09
Enabling Connectivity Analyses in FIRRTL

azidar:
Previously, it was difficult to do analyses of the FIRRTL AST that required connectivity information. A good example of this is dead-code elimination, which required many custom datastructures in order to work properly. Due to this start-up complexity, a novice FIRRTL-transform writer could not easily write transforms which needed this feature.

This PR aims to lower the threshold for transform-writers to write powerful transformations that use connectivity information. As an example, this PR also includes a clock-domain analyses transform, which given annotated signals, determines which clock signal (or clock signals) is driving it.

LFC:

- `CircuitGraph`/`ConnectionGraph`/`IRLookup` software architecture
- More use cases for netlist-type analyses

This PR adds both infrastructure and a new FIRRTL feature.

- `ConnectionGraph`: A connectivity-graph of FIRRTL local targets, but virtualizes nodes as an instance-viewed node as the graph is walked. This enables connectivity analyses with a smaller memory footprint. Also, previously explored modules will cache input-output connections and shortcut future trips through the module
- `IRLookup`: A class that enables correlating AST information (declarations, types, gender, kind) with targets, like the ones used as nodes in the `ConnectionGraph`
- `ClockSourceFinder`: A subclass of `ConnectionGraph` that lets you determine, given a signal, what clock source is it synchronized with. Intermediate results are cached to speed up queries based on past queries. It also can be done hierarchically, which means that results from leaf modules can also speed up queries from parent modules.
- `FindClockSources`: A transform that, given `GetClockSources` annotations, produces `ClockSources` annotations. It also consumes earlier `ClockSources` annotations to speed up new analyses, as well as enabling users to provide their own mapping (e.g. associating top-level inputs with the top-level clock port)

TODO:

- [x] ScalaDoc CircuitGraph functions
- [ ] Test CircuitGraph functions
- [x] Write RFC for expression tags (add new Tag type, could be super class of reference strings, which is optional for DoPrim, Mux, and ValidIf (and other WExpressions like WInvalid)
- [x] Move UnnamedToken etc. to new expression tags PR
- [x] Try to abstract more of `ClockSourceFinder` into `ConnectionGraph`, specifically the `recordClock` functions.
- [ ] ScalaDoc ConnectionGraph functions
- [ ] Add tagNode functions to ConnectionGraph

PR #995, opened 2019-01-18
Towards RegPreset support: Add locally available annotations and verilog emission customization

johnsbrew:
### For context see #991

As stated last week during developers' meeting, I rewrote my proposal with
- no impact on the IR
- no chisel compatibility issue

### What I've done
- (chisel, to be commited) created a new Object in chisel called RegPreset which is basically a RegInit annotated
- (chisel, to be commited) created a Chisel annotation 
- created a new kind of firrtl annotation` LocalAnnotations `derived from `SingleTargetAnnotation[Target]`
- added a transform that applies all `LocalAnnotations` to their respective Target
- modified some transforms and passes such as applied `LocalAnnotations` do not get deleted along the way
- modified the verilog emitter

I know I've made a bold move when applying the annotation which does not seem to be the way they are supposed  to work, however I thought it would be a shame to check multiple times each register to get their annotations instead of storing them along once and for all.

With all of this changes (and the chisel below) I've the whole chain working as I would expect: `val my_valid = RegPreset(false.B)` infers a `reg my_valid = 1'h0;`

Please give me your feedback on it :)

### Note
I've not achieved to integrate my chisel lib as a part of the chiselFrontend so here it is:
```scala
package my_working_package

import chisel3._
import chisel3.internal.InstanceId
import chisel3.experimental.{annotate, ChiselAnnotation}

import firrtl.annotations.PresetRegAnnotation

// for implicit 
import chisel3.internal.sourceinfo.SourceInfo
import chisel3.core.CompileOptions


case class PresetRegChiselAnnotation(target: InstanceId)
    extends ChiselAnnotation {
  def toFirrtl = PresetRegAnnotation(target.toNamed)
}

object RegPreset {
  /** @usecase def apply[T <: Data](t: T, preset: T): T
    *   Construct a [[Reg]] from a type template preset to the specified value on FPGA startup (note: does not make sense for ASIC design)
    *   @param t The type template used to construct this [[Reg]]
    *   @param preset The value the [[Reg]] is preset to on FPGA startup
    */
  def apply[T <: Data](t: T, preset: T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T = {
  // def apply[T <: Data](t: T, preset: T): T = {
    val reg = RegInit(t, preset)
    val anno = PresetRegChiselAnnotation(reg)
    annotate(anno)
    reg
  }

  /** @usecase def apply[T <: Data](preset: T): T
    *   Construct a [[Reg]] preset to the specified value.
    *   @param preset The value that serves as a type template and preset value
    */
  def apply[T <: Data](preset: T)(implicit sourceInfo: SourceInfo, compileOptions: CompileOptions): T = {
  // def apply[T <: Data](preset: T): T = {
    val model = chiselTypeOf(preset).asInstanceOf[T]
    RegPreset(model, preset)
  }
}
```

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -417,6 +417,13 @@ class VerilogEmitter extends SeqTransform with Emitter {
     val initials = ArrayBuffer[Seq[Any]]()
     val simulates = ArrayBuffer[Seq[Any]]()
 
+    def declare(b: String, n: String, t: Type, info: Info, preset: Expression) = t match {
+      case tx: VectorType =>
+        declares += Seq(b, " ", tx.tpe, " ", n, " [0:", tx.size - 1, "] = ", preset, ";", info)
+      case tx =>
+        declares += Seq(b, " ", tx, " ", n, " = ", preset, ";", info)
+    }
+
     def declare(b: String, n: String, t: Type, info: Info) = t match {
       case tx: VectorType =>
         declares += Seq(b, " ", tx.tpe, " ", n, " [0:", tx.size - 1, "];", info)
@@ -613,10 +620,23 @@ class VerilogEmitter extends SeqTransform with Emitter {
         case sx: DefWire =>
           declare("wire", sx.name, sx.tpe, sx.info)
         case sx: DefRegister =>
-          declare("reg", sx.name, sx.tpe, sx.info)
+          var useInitAsPreset = false
+          var disableRandomization = false
+          for (anno <- sx.annos) {
+            anno match {
+              case reo : RegisterEmissionOptions =>
+                useInitAsPreset = reo.useInitAsPreset
+                disableRandomization = reo.disableRandomization
+            }
+          }
+          if (useInitAsPreset)
+            declare("reg", sx.name, sx.tpe, sx.info, sx.init)
+          else
+            declare("reg", sx.name, sx.tpe, sx.info)
           val e = wref(sx.name, sx.tpe)
           regUpdate(e, sx.clock)
-          initialize(e)
+          if (!disableRandomization)
+            initialize(e)
         case sx: DefNode =>
           declare("wire", sx.name, sx.value.tpe, sx.info)
           assign(WRef(sx.name, sx.value.tpe, NodeKind, MALE), sx.value, sx.info)
--- a/src/main/scala/firrtl/LoweringCompilers.scala
+++ b/src/main/scala/firrtl/LoweringCompilers.scala
@@ -72,7 +72,8 @@ class HighFirrtlToMiddleFirrtl extends CoreTransform {
     passes.CheckWidths,
     passes.ConvertFixedToSInt,
     passes.ZeroWidth,
-    passes.InferTypes)
+    passes.InferTypes,
+    new firrtl.transforms.ApplyLocalAnnotations)
 }
 
 /** Expands all aggregate types into many ground-typed components. Must
--- a/src/main/scala/firrtl/annotations/EmissionAnnotations.scala
+++ b/src/main/scala/firrtl/annotations/EmissionAnnotations.scala
@@ -1,0 +1,21 @@
+// See LICENSE for license details.
+
+package firrtl
+package annotations
+
+trait EmissionOptions
+
+trait RegisterEmissionOptions extends EmissionOptions {
+  def useInitAsPreset : Boolean
+  def disableRandomization : Boolean
+}
+
+abstract class LocalAnnotations extends SingleTargetAnnotation[Target]
+
+case class PresetRegAnnotation(
+    target: Target
+) extends LocalAnnotations with RegisterEmissionOptions {
+  def duplicate(n: Target) = this.copy(target = n)
+  def useInitAsPreset = true
+  def disableRandomization = true
+}
--- a/src/main/scala/firrtl/ir/IR.scala
+++ b/src/main/scala/firrtl/ir/IR.scala
@@ -4,9 +4,10 @@ package firrtl
 package ir
 
 import Utils.indent
+import firrtl.annotations.Annotation
 
 /** Intermediate Representation */
-abstract class FirrtlNode {
+abstract class FirrtlNode(val annos : List[Annotation] = List()) {
   def serialize: String
 }
 
@@ -227,7 +228,7 @@ case class DoPrim(op: PrimOp, args: Seq[Expression], consts: Seq[BigInt], tpe: T
   def foreachWidth(f: Width => Unit): Unit = Unit
 }
 
-abstract class Statement extends FirrtlNode {
+abstract class Statement(annos : List[Annotation] = List()) extends FirrtlNode(annos) {
   def mapStmt(f: Statement => Statement): Statement
   def mapExpr(f: Expression => Expression): Statement
   def mapType(f: Type => Type): Statement
@@ -258,13 +259,14 @@ case class DefRegister(
     tpe: Type,
     clock: Expression,
     reset: Expression,
-    init: Expression) extends Statement with IsDeclaration {
+    init: Expression,
+    override val annos : List[Annotation] = List()) extends Statement(annos) with IsDeclaration {
   def serialize: String =
     s"reg $name : ${tpe.serialize}, ${clock.serialize} with :" +
     indent("\n" + s"reset => (${reset.serialize}, ${init.serialize})" + info.serialize)
   def mapStmt(f: Statement => Statement): Statement = this
   def mapExpr(f: Expression => Expression): Statement =
-    DefRegister(info, name, tpe, f(clock), f(reset), f(init))
+    this.copy(clock = f(clock), reset = f(reset), init = f(init))
   def mapType(f: Type => Type): Statement = this.copy(tpe = f(tpe))
   def mapString(f: String => String): Statement = this.copy(name = f(name))
   def mapInfo(f: Info => Info): Statement = this.copy(info = f(info))
--- a/src/main/scala/firrtl/passes/ConvertFixedToSInt.scala
+++ b/src/main/scala/firrtl/passes/ConvertFixedToSInt.scala
@@ -64,10 +64,10 @@ object ConvertFixedToSInt extends Pass {
         }
       }
       def updateStmtType(s: Statement): Statement = s match {
-        case DefRegister(info, name, tpe, clock, reset, init) =>
-          val newType = toSIntType(tpe)
-          types(name) = newType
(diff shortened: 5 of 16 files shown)


johnsbrew:
@azidar @jackkoenig @seldridge (and anyone interested) could we plan a meeting to talk about this proposal ?
I have a lots of questions about the philosophy behind the use of the IR and the annotations.
I wonder if this work on preset registers shouldn't be part of a bit more global upgrade of resets options to provide an even support for the 3 of them: sync reset, async reset and preset.

Give me your thoughts :)

johnsbrew:
> A meeting would be fine.

Let's settle when : anytime before 11am PST would be fine on any working day, can be later if needed.


> Are the annotations copied to each of the flattened `n` things? Are you allowed to copy the annotations? (...)

That is precisely an issue I have with the LowerTypes transform since I need the annotation to be copied to children in my case

> By separating out the metadata from the IR, you make a lot of this more straightforward for transform writers. Whatever you want to do, just do it. The transform's only job is to return the modified circuit and a `RenameMap` describing what it changed. The FIRRTL compiler then updates all annotations _using annotation-defined semantics_. (...)

I understand your point: it is actually great to have this done outside the scope of the transform! 
It is currently supported when inheriting from a particular class of `Annotation`or is it something to do ? I would be interested into contributing if needed.

> This is a shotgun spread of words and not anything concrete... but, it may provide some ideas into the split on IR and annotations.

Actually quite precise and detailed explanation, thank you very much!

> Related to your PR, I'd be very, very hesitant to introduce annotations into the IR.

The way I've done it: agreed given your previous points
However in a more generic way I think this need further investigation: IMHO a transform writer need a simple and clean way to know if any particular `FirrtlNode`is a `Target` of one `Annotation`.
By clean I mean:
   - 1 line (1 function call)
   - O(1) complexity per call: no research along the potentially huge list, `Target` kind `Annotation` should be indexed in Firrtl cicruit logic way: `Circuit`>`Module`>`Statement`

Please correct me if it is already possible, I might have missed something.

> The asynchronous reset is likely way more involved as that does need to introduce new IR constructs. All resets are `UInt<1>` in FIRRTL and asynchronous reset needs to either introduce an actual reset super type and then place reset and asyncreset under that. 

Why not adding `Preset` as another type under the `Reset` super type ? Even if it is a quite simpler construct in the emitted verilog, it would uniformize the treatment of this 3 kinds of resets that are (almost) evenly popular in FPGA-specific designs .

> a full async reset solution isn't an emitter change. 

100% agreed

> What you're trying to do, I think, is only a modification of the emitter. What you're proposing here seems more like a customization of the Verilog emitter with the following things:
> 
>     1. Build up a `val resetMap: Map[Target, Literal]` based on annotations
> 
>     2. Use the custom emission strategy you added to `Emitter`
> 

True, that was actually my first attempt using annotations, `RegPreset` defined as a simple annotated version (with value in annotations) of chisel's base `Reg`object . This attempt aborted before any PR because of the following issues : 
 - target modification during transforms (simple example stated above: flattening in LowerTypes)
 - preset value modification must follow the one of its type 
 - how to pass a chisel object through annotations and have it undergoing all the transformation it should have ? => for me it means that the type/value couple belongs to the IR 

For those reasons I chose to annotate a RegInit and pass the preset value through the IR, where it undergoes all the necessary transforms. As far as I understood I had no other choices but I might be mistaken on this one. 

Note: Now that I am writing this, I am thinking that in addition to annotate the `RegInit` `RegPreset`object should also set the reset to `0` instead of `val reset = Builder.forcedReset.ref` to avoid the modification of the RemoveReset transform.

> I'd be interested in thinking about how the emission could be encoded inside of the annotation.

Defining a restricted subset of `EmissionOptions` in an enable/disable style is an obvious simple approach that I've started in this PR but I would be quite interested in something much more modular.

> You could likely do this with a mixin to an IR node or a subclassed IR node and then modify the emitter to be composed of partial functions collected from any "custom node" annotations. 

This could be very nice indeed: potentially trying/customizing  different kind of verilog emission to fit best the next tools verilog front-end.
I am currenly thinking to
 - Vivado way of interpreting very specific constructs (e.g. for async resets or inferring BRAMs)
 - reducing the number of intermediate simple nodes to speed-up event-driven simulations (Although I don't know if it would actually have any significant impact on simulation speed after their own optimization steps) 

### Conclusion
- Meeting : let me know when ?
- I forget about embedding the target annotation into the IR
- I try to figure out how to have a specific kind of targets annotations behave as I would like to (e.g. survive flattening)
- I try to figure out how to provide simple local access to annotation (without embedding it into the nodes) 
- I create RegPrest such as it does not push a  `val reset = Builder.forcedReset.ref` => hence no need for modification of RemoveReset transform

azidar:
> IMHO a transform writer need a simple and clean way to know if any particular FirrtlNode is a Target of one Annotation

Why is that the case? From what I've found, a transform doesn't need to know all the possible annotations referencing a given target, but just the specific annotations it cares about. To be clear, an O(1) and single-line call to get all annotations referencing a given target might still be useful, I just suspect its probably not the correct pattern to be using.

My recommendation is to put preset information in an annotation in AnnotationSeq (not in the AST). These annotations will get automatically updated when the registers they are referring to get lowered. For example, you could have:

```scala
case class PResetAnnotation(map: Map[ReferenceTarget, Int]) extends Annotation {
   def update(renameMap: RenameMap): Seq[PResetAnnotation] = {
      ...
   }
}
```

You would need to implement the update method (given changes to the AST, how should your annotation update itself). However, then any reference to a register that got lowered would enable you to indicate exactly how to update its preset behavior. Note that ReferenceTarget's can refer to subcomponents (e.g. reg.a, or reg.b). Check out (https://github.com/freechipsproject/firrtl/pull/937/files#diff-89ebb61f6ebbe761036b389a4cdaf83eR23) for an example of how to do this.

Then, in your transform, you can find all annotations of type PResetAnnotation, combine their maps, and generate your emitter-specific annotations. Finally, the emitter runs and emits the correct Verilog.

I do think this is best done over a meeting, let's try to schedule that soon.

johnsbrew:
Thank you @seldridge & @azidar for your detailed explanations.
As I guessed at some point, I missed some of the features you've already had implemented!

I am available 0800--0900 PT on Friday: let's discuss it in details tomorrow

seldridge:
The 0800 is likely too early for the west coast folks, but it's fine on my end. Discussing tomorrow should work. I'll set something up.

seldridge:
Talked to @johnsbrew. Some of the ideas discussed are below. Here, `RegPreset` is a hypothetical Chisel API similar to `RegInit`, but will cause registers with preset-style resets to pop out of the Verilog emitter.

`RemoveResets` causes some problems for a potential `RegPreset`. This transform will, unless a reset line for a register is connected to zero, will generate a self connection in place of the initial value and convert the reset logic to a mux. The proposed solution here is an annotation that can disable this conversion for a specific register. I believe that this won't cause any problems for later transforms, but I'm not sure. I.e., does this violate LowForm or some assumption made by an interpreter like Treadle?

A `RegPreset` can be fully realized then in the `Emitter`. However, there then needs to be some way of disabling the preset-style register emission if we're not targeting an FPGA backend. E.g., @johnsbrew publishes a library that's generally useful and uses `RegPreset` in it. This will generate annotations that disable `RemoveReset` and cause custom register emission to happen. This needs to be disabled in some clean way. One idea here is to have a transform that can remove annotations before FIRRTL starts. In effect a `StripAnnotationAnnotation(a: Class[_ <: Annotation])`.

chick:
I don't see any obvious behavioral differences that would be a problem for Treadle.

seldridge:
@johnsbrew: I just talked to @jackkoenig... an even cleaner approach is to, in spite of my initial hesitation, tie this to `AsyncReset`. An asynchronous reset can't be removed by `RemoveReset`. Then you can get away with doing this with only one annotation that is consumed by the emitter. This would be cleaner, but it puts a dependency on @jackkoenig to finish the `AsyncReset` idea. (He's working on it now, though!)

I'm in favor of this approach over what I suggested.

seldridge:
After some additional discussion in the developer meeting, @jackkoenig and I think that there should likely *NOT* be a Chisel API for this. We should strive, to the utmost, to not expose FPGA-specific constructs in Chisel. 

What we're thinking here is that there is a global flag (an Annotation) that would convert asynchronous resets to use preset initialization. So, at this time, I think the minimal approach would be to prototype a PR to change the emission (which this PR should already do). We can then hold off while the asynchronous reset PR comes together.

@jackkoenig: does this agree with what you're thinking?

jackkoenig:
I agree but would just like to add a little

> I think that there should likely NOT be a Chisel API for this. We should strive, to the utmost, to not expose FPGA-specific constructs in Chisel.

I **think** we can accomplish everything you want without having to expose FPGA-specific APIs. **If** we can pull that off, the overall result will be better because the same Chisel-source code will be agnostic to its implementation technology (such compatibility is part of the ethos of Chisel).

johnsbrew:
Thank you all for spending some time on this discussion.
This new direction looks promising to us.
It makes sense to link `Preset` and `AsyncReset` as `Preset` can be seen as an hidden `AsyncReset` already in place for the whole FPGA.

@jackkoenig could we schedule a meeting to talk about the implementation of async resets as this will shape the API for `Preset` at the same time? 
- One of our concern is being able to annotate at least on a module basis such as it is inferring either `AsyncReset` (required for busses interconnection, eg AXI streams) or `Preset`
- Another is the ability to use both `AsyncReset`(or `Preset`) and `Reset` in the same module (also required for IPs interconnections)

Let's share about that, is there already a Chisel/firrtl discussion about it ? That could be a good first topic for the chisel-dev ML ;)

jackkoenig:
@johnsbrew yes, let's chat. I've set up a when2meet: https://www.when2meet.com/?7482171-eoBuE
Anyone else is welcome to join.

With my current understanding, I think that annotating at a module-level as well as on an AsyncReset signal level (ie. transform all registers connected to a given AsyncReset into PreReset registers) could provide the necessary functionality. Let's chat, prototype, and see!

I'll send out the when2meet on chisel-dev as well in case others are interested!

jackkoenig:
@johnsbrew AsyncReset has been merged

johnsbrew:
@jackkoenig thank you very much for having AsyncReset working that quickly :)
I didn't manage to get involved in the discussion but I tried to follow closely what you were doing and I will resume the work on preset as we discussed it as soon as possible.

azidar:
I'm going to add a DO NOT MERGE label on this PR for now, until you continue working :)

johnsbrew:
Hi there,

I'm (at last) getting back to work on this `Preset` feature.
I will close this PR to keep it for the records and I will come back shortly with a clean one :)

PR #1000, opened 2019-01-23
Attempt to deal with timing vagaries in UniquifySpec.quicklyrenamedeepbundles

ucbjrl:
Switching to Scala 2.12.8 causes this test to start failing on OSX. Try earlier scheme to compare shallow vs deep to reduce brittleness.
The `threshold` parameter calculation is quite _ad hoc_.


jackkoenig:
WE HAVE A WINNER!!! #1000!!!!

src/test/scala/firrtlTests/UniquifySpec.scala
```diff
@@ -286,29 +286,34 @@ class UniquifySpec extends FirrtlFlatSpec {
   }
 
   it should "quickly rename deep bundles" in {
+    val depth = 500
     // We use a fixed time to determine if this test passed or failed.
```
azidar: This comment seems wrong..?

ucbjrl: Updated comment.

PR #1004, opened 2019-01-25
Add "mverilog" Compiler Option, MinimumVerilogEmitter

### LINKED ISSUE #423: Have a way to disable optimization pass for debugging purposes
So that it doesn't optimize away some debug signals

seldridge:
This adds "mverilog" to the "--compiler" command line option. This
will run the MinimumVerilogCompiler.

This additionally fixes the `MinimumVerilogCompiler` such that
`DeadCodeElimination` will not be run (it's not supposed to be).
Previously, the `VerilogEmitter` would run `DeadCodeElimination`. This
resulted in the `MinimumVerilogCompiler` only effectively disabling
`ConstantPropagation`.

Additionally, `BlackBoxSourceHelper` is removed from the
`MinimumVerilogCompiler` since this will be run by the `VerilogEmitter`
already.

@jackkoenig, @azidar: Question about `DeadCodeElimination`. Is there a reason why this was being included in the `VerilogEmitter` directly?

Fixes #423.

TODO:

- [ ] Add `RemoveValidIf` to `MinimumLowFirrtlOptimization`
- [ ] Handle literals with bit extract

### DIFF
--- a/src/main/scala/firrtl/ExecutionOptionsManager.scala
+++ b/src/main/scala/firrtl/ExecutionOptionsManager.scala
@@ -207,18 +207,19 @@ extends ComposableOptions {
       case "low"       => new LowFirrtlCompiler()
       case "middle"    => new MiddleFirrtlCompiler()
       case "verilog"   => new VerilogCompiler()
+      case "mverilog"  => new MinimumVerilogCompiler()
       case "sverilog"  => new SystemVerilogCompiler()
     }
   }
 
   def outputSuffix: String = {
     compilerName match {
-      case "verilog"   => "v"
-      case "sverilog"  => "sv"
-      case "low"       => "lo.fir"
-      case "middle"    => "mid.fir"
-      case "high"      => "hi.fir"
-      case "none"      => "fir"
+      case "verilog" | "mverilog" => "v"
+      case "sverilog"             => "sv"
+      case "low"                  => "lo.fir"
+      case "middle"               => "mid.fir"
+      case "high"                 => "hi.fir"
+      case "none"                 => "fir"
       case _ =>
         throw new Exception(s"Illegal compiler name $compilerName")
     }
@@ -267,6 +268,7 @@ extends ComposableOptions {
       case "middle" => classOf[MiddleFirrtlEmitter]
       case "low" => classOf[LowFirrtlEmitter]
       case "verilog" => classOf[VerilogEmitter]
+      case "mverilog" => classOf[MinimumVerilogEmitter]
       case "sverilog" => classOf[VerilogEmitter]
     }
     getOutputConfig(optionsManager) match {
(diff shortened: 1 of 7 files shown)


jackkoenig:
DCE in the Emitter is just to clean up after [`FlattenRegUpdate`](https://github.com/freechipsproject/firrtl/blob/1738c1ef0ac95fae25d52586b3b0348de80de2ff/src/main/scala/firrtl/Emitter.scala#L835) which is simply there to clean up no longer used nodes (see https://github.com/freechipsproject/firrtl/pull/692 for more motivation).

I think there are a few ways to make `mverilog` work better with this in mind:
1. Don't include DCE in a `MinimumVerilogEmitter`
2. Remove DCE from the emitter in favor of a more specialized "mini-DCE" that only removes nodes made dead by `FlattenRegUpdate`
3. Use an improved dependency API that runs DCE allows DCE to only run once and have `FlattenRegUpdate` depend on it so that the regular emitter will add that dependency and delay running DCE while just compiling to optimized low FIRRTL will still run DCE itself.

seldridge:
I think (1) is the most reasonable approach now. (3) is the way of the future, but won't make the 1.2 window. (2) is tractable, but I'd rather keep DCE unified, even though it's pretty simple as is.

Rebased to implement approach (1).

Additionally, I updated `DriverSpec` to run `--compiler mverilog` and `--compiler sverilog` to improve coverage somewhat.

seldridge:
@jackkoenig: for the latter case, this is actually `shr` that's doing it. Bit extract gets properly handled in `Legalize`, but the `shr` on literals isn't. Does it make sense to modify `Legalize` to handle this? 

This is effectively the "minimal constant propagation" idea, but it's not an explicit additional `Transform`.

aswaterman:
That approach makes sense to me.

src/main/scala/firrtl/ExecutionOptionsManager.scala
```diff
@@ -344,13 +346,16 @@ trait HasFirrtlOptions {
 
   parser.opt[String]("compiler")
     .abbr("X")
-    .valueName ("<high|middle|low|verilog|sverilog|none>")
+    .valueName ("<high|middle|low|verilog|mverilog|sverilog|none>")
```
jackkoenig: I'm okay with this being a "compiler" alternative but I think we should consider making it an orthogonal argument. This is more akin to `-O0` while the default is `-O2`

seldridge: I'd alsl be in favor of a `-O` style optimization level flags and explicit `-fno-constant-propagation`, etc. in the future. 

I'd propose to kick that can down the road to the dependency API discussion. Or: if we have optimization levels, I'd prefer for those to not alias to concrete `Compiler`s. I'm open to other ideas, though.

jackkoenig: Yeah I think that makes sense

PR #1005, opened 2019-01-25
Stage/Phase

### LINKED ISSUE #485: Funny Behavior when topName isn't specified

### LINKED ISSUE #764: [RFC] Annotations subsume ExecutionOptions?

### LINKED ISSUE #793: Don't Emit Emitter Annotations nor Deleted Annotations by default

seldridge:
# Overview

1. Updates to `firrtl.options` package
2. This adds `firrtl.stage.FirrtlStage`, a refactor of `firrtl.Driver` as a `firrtl.options.Stage` composed of a sequence of internal `firrtl.options.Phase`s. 
3. The old `firrtl.Driver` is modified to be a compatibility wrapper around `firrtl.stage.FirrtlStage`.

## Todo

- [x] Deprecate `FirrtlExecutionResultView`
- [x] Copy automatic `Annotation` deletion behavior from `Transform`
- [x] Verify that `FirrtlStage` runs on SiFive workloads (@jackkoenig)
- [x] Add custom error text if a user enters a `Driver` option into `FirrtlStage`, e.g., `--top-name`
- [x] Simplify option entry for users
- [x] Consider switching to a better paradigm than `runTransform`
- [x] Switch to using a default emitter if no emitter specified

## Modifications to `firrtl.options` Package

#879 and #945 introduced the `firrtl.options` package. This centers around the concept of a `Phase` which transforms an `AnnotationSeq`. A `Stage` extends `Phase` with a user interface, a `Shell`, that converts command line options (`Array[String]`) to an `AnnotationSeq`.

This PR modifies the type hierarchy to add `TransformLike`, `Translator`, and `FirrtlStage`. The figure below shows the new type hiearchy:

transformlike-hierarchy

**`TransformLike[A]`** arose because of a discussion with @azidar on the need for `Phase` registration (the ability of external libraries to add `Phase`s) and for a Dependency API. See: https://github.com/freechipsproject/firrtl/pull/945#discussion_r235552905. `Transform`s currently support registration and have a pseudo-Dependency API via `getLoweringCompilers`. However, this leads to the observation that a `Phase` and a `Transform` are really the same thing. A `Transform` is a mathematical transformation on a `CircuitState` while a `Phase` is a mathematical transformation on an `AnnotationSeq`. `TransformLike` seeks to unify these two under a common type hierarchy. *This PR does not make `Transform` extend `TransformLike[CircuitState]` nor does it add a Dependency API nor does it introduce `Phase` registration. All that is intended to happen in follow-up PRs.*

**`Translator[A, B]`** arises from the need to insert a `TransformLike[B]` inside a `TransformLike[A]`. A `Translator` provides an interface such that externally it looks like a `TransformLike[A]`, but internally performs a transform over `B`. Concretely, this is needed to implement the interface between the `FirrtlStage` and the FIRRTL compiler (which is a `Transform`).

**`Stage`** is modified to always run two additional phases: `firrtl.options.phases.Checks` and `firrtl.options.phases.WriteOutputAnnotations`. The former will check that `StageOptions` make sense. The latter will write the `AnnotationSeq` to a file excluding always excluding `firrtl.options.Unserializable` and excluding `DeletedAnnotation`s if a `WriteDeletedAnnotation` is absent (see: https://github.com/freechipsproject/firrtl/issues/793). Additionally, a `Stage` will now automatically setup a `Logger` scope based on explicit input `Annotation`s *and those read from a a file*.

**`Shell`** is changed to include `Logger` options to align with the inclusion of a `Logger` scope in `Stage`.

After this PR, every `Stage` now provides the following options/annotations:

- `TargetDirAnnotation` / `-td` / `--target-dir`
- `ProgramArgsAnnotation` / "args" style options on the command line
- `InputAnnotationFileAnnotation` / `-faf` / `--annotation-file`
- `OutputAnnotationFileAnnotation` / `-foaf` / `--output-annotation-file`
- `WriteDeletedAnnotation` / `--write-deleted`
- `LogLevelAnnotation` / `-ll` / `--log-level`
- `ClassLogLevelAnnotation` / `-cll` / `--class-log-level`
- `LogFileAnnotation` / `--log-file`
- `LogClassNamesAnnotation` / `-lcn` / `--log-class-names`

Other minor modifications include:

- New exceptions added:
  - `PhaseException`
  - `OptionsException`
  - `PhasePrerequisiteException`
  - `OptionsHelpException`
- Addition of the `--show-registrations` option to dump discovered registered libraries/transforms
- A `Stage`, on an `OptionsException`, will now return an exit code of `1`

## Modifications/Additions to `firrtl.stage` Package

Previously, the `firrtl.stage` package only contained unused `Annotation`s related to FIRRTL. This is now fleshed out to implement `FirrtlStage`, a reimplementation of the original `firrtl.Driver` as a Stage.

This pares down the existing FIRRTL `Annotations` in the following ways:

- `TopNameAnnotation` is removed into a Driver compatibility layer
- `TargetDirAnnotation`, `InputAnnotationFileAnnotation`, `OutputAnnotationFileAnnotation`, `ProgramArgsAnnotation`, `LogLevelAnnotation`, and `ClassLogLevelAnnotation` are moved out of `firrtl.stage` as these are common to all `Stage`s

**`FirrtlOptions`** encodes information about the `FirrtlStage`. This is noticeably slimmer than its forebearer, `FirrtlExecutionOptions` (it has 4 members as opposed to 15). The reason for this is because some of these options were moved into `Logger` and `Stage`. However, the pipeline like nature of `Stage`/`Phase` means that most `Phase`s consume an `Annotation` and then delete it. Very rarely do you have an `Annotation` that is intended to hang around long term.

**`FirrtlStage`** implements the FIRRTL `Stage`. This is incredibly terse and hides the heavy lifting behind a sequence of `Phase`s. This is what runs the FIRRTL compiler and emitter. Most of the internal `Phase`s are self-explanatory based on their name. They have scaladoc internally to prime you on what they're doing.

- **`stage.phases.AddCircuit`** expands `FirrtlFileAnnotation`s and `FirrtlSourceAnnotation`s into `FirrtlCircuitAnnotation`s. The intended aim of this is that later `Phase`s only have to operate on `FirrtlCircuitAnnotation`s.
- **`stage.phases.AddDefaults`** adds default `Annotation`s
- **`stage.phases.AddImplicitOutputFile`** adds an `OutputFileAnnotation` derived from a `FirrtlCircuitAnnotation` if one does not already exist
- **`stage.phases.Checks`** does sanity checking on the `AnnotationSeq`. Due to the fungibility of `Annotation`s and command line options, no checking is possible in `scopt`. All old `scopt`-level checks are moved in here.
- **`stage.phases.Compiler`** is the most interesting of these and warrants more discussion below. This is a `Translator[AnnotationSeq, CompilerRun]` that looks like a `Phase` but is internally running the FIRRTL compiler in parallel on discovered `FirrtlCircuitAnnotation`s. This acts as the interface between the `Phase` world and the `Transform` world.
- **`stage.phases.DriverCompatibility`** is discussed in the Driver compatibility section. This provides `Phase`s necessary to replicate the behavior of the old Driver.
- **`stage.phases.WriteEmitted`** writes an `EmittedAnnotation`s to files.

**`stage.phases.Compiler`** is the interface to the actual FIRRTL compiler. This is implemented as a `Translator[AnnotationSeq, CompilerRun]` where a `CompilerRun` is a private case class used to encapsulate all the information necessary to run the FIRRTL compiler. This will take every `FirrtlCircuitAnnotation` it finds and run the FIRRTL compiler on them (in parallel if more than one is found). This then necessitates some grammar to group `Annotation`s to circuits. What I settled on is in the scaladoc, but seems reasonable: *Any annotations that come before the first circuit are "global" and apply to all circuits. Any annotations that come after a circuit are left-associative to only that circuit.* Consider:

```
The following AnnotationSeq:
  A(a), FirrtlCircuitAnnotation(x), B, FirrtlCircuitAnnotation(y), A(b), C, FirrtlCircuitAnnotation(z)

Will result in three compiler runs:
  - FirrtlCircuitAnnotation(x): A(a), B
  - FirrtlCircuitAnnotation(y): A(a), A(b), C
  - FirrtlCircuitAnnotation(z): A(a)
```

This has implications for `Annotation` ordering. Consequently, there needs to be a unified convention that either `Annotation`s are always prepended or appended to an `AnnotationSeq`. Mixing these will cause weird behavior. The current convention is to always prepend (for performance) and then [reverse after annotations are parsed](https://github.com/freechipsproject/firrtl/pull/1005/files#diff-e7891d14b8c28a115ee467abd3b02187L58). Critically, this convention needs to extend to libraries registering options.

For these reasons, parallel compilation is not available unless a user directly uses `firrtl.stage.phases.Compiler`. Presently, `firrtl.stage.phase.Checks` will fail if a user has multiple `Annotation`s that can be converted to a FIRRTL circuit.

**`FirrtlOptionsView`** and **`FirrtlExecutonResultsView`** provide "views" of an `AnnotationSeq` as their eponymous type. These are included in the `firrtl.stage` package object.

Summarily, the `FirrtlStage` now provides the following options/annotations:

- `FirrtlFileAnnotation` / `-i` / `--input-file`
- `OutputFileAnnotation` / `-o` / `--output-file`
- `InfoModeAnnotation` / `-info-mode`
- `FirrtlSourceAnnotation` / `--firrtl-source`
- `CompilerAnnotation` / `-X` / `--compiler`
- `RunFirrtlTransformAnnotation` / `-fct` / `--custom-transforms`
- `EmitCircuitAnnotation` / `-E` / `--emit-circuit`
- `EmitAllModulesAnnotation` / `-e` / `--emit-modules`

## `firrtl.Driver` Compatibility

The old `firrtl.Driver` has its API unmodified. The intention, and validated aim, here is that if a user wants to use the old `firrtl.Driver` and all its associated classes (`ExecutionOptionsManager` with mutable state, etc.) they can. This will continue to behave as before, but internally it is using the new `Stage`. This is achieved through a four-step process:

1. All "options" (`FirrtlExecutionOptions`, `CommonOptions`) get a `toAnnotations` option that converts them to an `AnnotationSeq`.
2. The `AnnotationSeq` is pre-processed through a number of compatibility `Phase`s. These are discussed below.
3. `FirrtlStage` runs
4. The output `AnnotationSeq` is post-processed by "viewing" it as a `FirrtlExecutionResult`.

More information on each of the compatibility `Phase`s follows:

- **`firrtl.stage.phases.DriverCompatibility.AddImplicitAnnotationFile`** adds an `InputAnnotationFileAnnotation` derived from the "top name" of the circuit if no `InputAnnotationFileAnnotation` already exists. [Top name determination is a minefield of weird behavior and unintended interactions](https://github.com/freechipsproject/firrtl/pull/1005/files#diff-a8449cd2e78f422a34c4ccf4d2f24ce0R49). Nonetheless, this is the old behavior. It's preserved here, but this is the motivation for removing the whole "top name" concept in `FirrtlStage`.
- **`firrtl.stage.phases.DriverCompatibility.AddImplicitFirrtlFile`** adds an `InputFileAnnotation` derived from the "top name".
- **`firrtl.stage.phases.DriverCompatibility.AddImplicitOutputFile`** adds an `OutputFileAnnotation` derived from the "top name".
- **`firrtl.stage.phases.DriverCompatibility.AddImplicitEmitterFile`** stems from a change in behavior of `FirrtlStage`. With `FirrtlStage`, a `--compiler` no longer implies an emitter. This is because we should support multiple emitters since they're really just custom transforms. However, including an implicit *circuit* emission has all sorts of odd behavior. This then requires the `--split-modules` option to change the implicit emitter to a one-file-per module emitter. How do we then handle explicit emitters specified as `--custom-transforms`? `FirrtlStage` makes the emitter explicit and this adds back the old behavior.

The figure below shows how all of this fits together. This shows all `Phase`s run from one call to `Driver.execute`. As this encompasses `FirrtlStage`, you can use it to see the path for just `FirrtlStage`.

firrtlstage

#### Differences between `FirrtlStage` and `Driver`

- `FirrtlStage` has no implicit emitter. Users need to specify `-E` or `-e`. Users can specify this multiple times.
- A number of options have been removed from `FirrtlStage`. Notably `--top-name`, `--split-modules`. 
- String arguments that were case insensitive in `Driver` are now case sensitive in `FirrtlStage`, e.g., the compiler is `verilog` not `Verilog`.

## Updates to Testing Utilities

This PR adds a boatload of tests that are all intended to be additive. During this some additional testing utilities were developed to cover situations with things that are generally difficult to test:

- What was written to standard out / standard error
- What the exception code was of a `Stage`
- Whether or not a file was written

To facilitates testing these, this PR adds the `FirrtlSpec.Utils` object which defines thunk-wrapping methods that can be used to capture standard out and standard error. Additionally, through the use of two custom [Java `SecurityManager`s](https://docs.oracle.com/javase/8/docs/api/java/lang/SecurityManager.html), this can run a well-behaved thunk to check its exit code (without causing the sbt session to shut down) and to check if it tries to write to any files. These three utilities are provided as follows with examples in the new tests:

- `grabStdOutStdErr`
- `catchStatus`
- `catchWrites`

## Other Modifications

- Make `Emitter` abstract in an `outputSuffix`
- Add `outputSuffix` field to `EmittedComponent` 

## Questions

- `TransformLike[A]` is abstract in a method `def transform(a: AnnotationSeq): AnnotationSeq`. If `Transform` extends this, it's equivalent method is called `execute`. It's not a problem to have `final def transform(a: AnnotationSeq): AnnotationSeq = execute(a)`, but it would be nice to have one name. For simplicity and backwards compatibility, we could rename `transform` to `execute`. Thoughts?
- This PR currently changes the `./utils/bin/firrtl` helper script to point at `FirrtlStage`. This needs a companion rocket-chip PR to add the explicit emitter option. Do we want to change this now?

## Benchmarking

I'm still not 100% sure where this is coming from, but if you move to `FirrtlStage` there are some gains. If using the `DriverCompatibility` layer performance is constant and heap usage may go up. `FirrtlStage` is faster and uses less memory.

Tested on 6-core Ryzen 5 with `Linux xxx 4.9.0-8-amd64 #1 SMP Debian 4.9.130-2 (2018-10-27) x86_64 GNU/Linux` using https://github.com/jackkoenig/firrtlbench.

2019-01-29-214733_573x344_scrot
2019-01-29-214745_574x342_scrot

Raw data:

```csv
revision,design,max heap,SD,runtime,SD
df3a34f0,firrtl/regress/FPU.fir,405.72,8.73,4145.50,152.58
,firrtl/regress/ICache.fir,235.97,3.97,2162.00,35.37
,firrtl/regress/Rob.fir,1524.04,63.12,18156.50,735.31
,firrtl/regress/RocketCore.fir,419.18,15.08,4351.00,140.12
105efac3,firrtl/regress/FPU.fir,502.88,9.16,3980.50,95.18
,firrtl/regress/ICache.fir,239.54,4.19,2117.00,143.96
,firrtl/regress/Rob.fir,1517.08,43.20,17889.00,1194.04
,firrtl/regress/RocketCore.fir,504.01,11.69,4354.00,77.91
105efac3-FirrtlStage,firrtl/regress/FPU.fir,325.55,4.95,2585.00,63.49
,firrtl/regress/ICache.fir,200.51,1.70,1173.50,22.42
,firrtl/regress/Rob.fir,1409.79,55.51,15471.00,214.57
,firrtl/regress/RocketCore.fir,378.67,9.75,2878.00,68.30
```

## Metadata

- Fixes #764 
- Fixes #485 
- Fixes #793
- Depends on #879 and #919  (this includes commits from that PR)

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -19,7 +19,7 @@ import firrtl.PrimOps._
 import firrtl.WrappedExpression._
 import Utils._
 import MemPortUtils.{memPortField, memType}
-import firrtl.options.{HasScoptOptions, StageUtils, PhaseException}
+import firrtl.options.{HasShellOptions, ShellOption, StageUtils, PhaseException}
 import firrtl.stage.RunFirrtlTransformAnnotation
 import scopt.OptionParser
 // Datastructures
@@ -34,14 +34,12 @@ sealed trait EmitAnnotation extends NoTargetAnnotation {
 case class EmitCircuitAnnotation(emitter: Class[_ <: Emitter]) extends EmitAnnotation
 case class EmitAllModulesAnnotation(emitter: Class[_ <: Emitter]) extends EmitAnnotation
 
-object EmitCircuitAnnotation extends HasScoptOptions {
-  def addOptions(p: OptionParser[AnnotationSeq]): Unit = p
-    .opt[String]("emit-circuit")
-    .abbr("E")
-    .valueName("<chirrtl|high|middle|low|verilog|mverilog|sverilog>")
-    .unbounded()
-    .action{ (x, c) =>
-      val xx = x match {
+object EmitCircuitAnnotation extends HasShellOptions {
+
+  val options = Seq(
+    new ShellOption[String](
+      longOption = "emit-circuit",
+      toAnnotationSeq = (a: String) => a match {
         case "chirrtl"              => Seq(RunFirrtlTransformAnnotation(new ChirrtlEmitter),
                                            EmitCircuitAnnotation(classOf[ChirrtlEmitter]))
         case "high"                 => Seq(RunFirrtlTransformAnnotation(new HighFirrtlEmitter),
@@ -54,20 +52,19 @@ object EmitCircuitAnnotation extends HasScoptOptions {
                                            EmitCircuitAnnotation(classOf[VerilogEmitter]))
         case "sverilog"             => Seq(RunFirrtlTransformAnnotation(new SystemVerilogEmitter),
                                            EmitCircuitAnnotation(classOf[SystemVerilogEmitter]))
-        case _                      => throw new PhaseException(s"Unknown emitter '$x'! (Did you misspell it?)")
-      }
-      xx ++ c }
-    .text("Run the specified circuit emitter (all modules in one file)")
+        case _                      => throw new PhaseException(s"Unknown emitter '$a'! (Did you misspell it?)") },
+      helpText = "Run the specified circuit emitter (all modules in one file)",
+      shortOption = Some("E"),
+      helpValueName = Some("<chirrtl|high|middle|low|verilog|mverilog|sverilog>") ) )
+
 }
 
-object EmitAllModulesAnnotation extends HasScoptOptions {
-  def addOptions(p: OptionParser[AnnotationSeq]): Unit = p
-    .opt[String]("emit-modules")
-    .abbr("e")
-    .valueName("<none|high|middle|low|verilog|mverilog|sverilog>")
-    .unbounded()
-    .action{ (x, c) =>
-      val xx = x match {
+object EmitAllModulesAnnotation extends HasShellOptions {
+
+  val options = Seq(
+    new ShellOption[String](
+      longOption = "emit-modules",
+      toAnnotationSeq = (a: String) => a match {
         case "chirrtl"              => Seq(RunFirrtlTransformAnnotation(new ChirrtlEmitter),
                                            EmitAllModulesAnnotation(classOf[ChirrtlEmitter]))
         case "high"                 => Seq(RunFirrtlTransformAnnotation(new HighFirrtlEmitter),
@@ -80,10 +77,11 @@ object EmitAllModulesAnnotation extends HasScoptOptions {
                                            EmitAllModulesAnnotation(classOf[VerilogEmitter]))
         case "sverilog"             => Seq(RunFirrtlTransformAnnotation(new SystemVerilogEmitter),
                                            EmitAllModulesAnnotation(classOf[SystemVerilogEmitter]))
-        case _                      => throw new PhaseException(s"Unknown emitter '$x'! (Did you misspell it?)")
-      }
-      xx ++ c }
-    .text("Run the specified module emitter (one file per module)")
+        case _                      => throw new PhaseException(s"Unknown emitter '$a'! (Did you misspell it?)") },
+      helpText = "Run the specified module emitter (one file per module)",
+      shortOption = Some("e"),
+      helpValueName = Some("<chirrtl|high|middle|low|verilog|mverilog|sverilog>") ) )
+
 }
 
 // ***** Annotations for results of emission *****
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -19,6 +19,9 @@ import firrtl.PrimOps._
 import firrtl.WrappedExpression._
 import Utils._
 import MemPortUtils.{memPortField, memType}
+import firrtl.options.{HasScoptOptions, StageUtils, PhaseException}
+import firrtl.stage.RunFirrtlTransformAnnotation
+import scopt.OptionParser
 // Datastructures
 import scala.collection.mutable.{ArrayBuffer, LinkedHashMap, HashSet}
 
@@ -31,17 +34,70 @@ sealed trait EmitAnnotation extends NoTargetAnnotation {
 case class EmitCircuitAnnotation(emitter: Class[_ <: Emitter]) extends EmitAnnotation
 case class EmitAllModulesAnnotation(emitter: Class[_ <: Emitter]) extends EmitAnnotation
 
+object EmitCircuitAnnotation extends HasScoptOptions {
+  def addOptions(p: OptionParser[AnnotationSeq]): Unit = p
+    .opt[String]("emit-circuit")
+    .abbr("E")
+    .valueName("<chirrtl|high|middle|low|verilog|mverilog|sverilog>")
+    .unbounded()
+    .action{ (x, c) =>
+      val xx = x match {
+        case "chirrtl"              => Seq(RunFirrtlTransformAnnotation(new ChirrtlEmitter),
+                                           EmitCircuitAnnotation(classOf[ChirrtlEmitter]))
+        case "high"                 => Seq(RunFirrtlTransformAnnotation(new HighFirrtlEmitter),
+                                           EmitCircuitAnnotation(classOf[HighFirrtlEmitter]))
+        case "middle"               => Seq(RunFirrtlTransformAnnotation(new MiddleFirrtlEmitter),
+                                           EmitCircuitAnnotation(classOf[MiddleFirrtlEmitter]))
+        case "low"                  => Seq(RunFirrtlTransformAnnotation(new LowFirrtlEmitter),
+                                           EmitCircuitAnnotation(classOf[LowFirrtlEmitter]))
+        case "verilog" | "mverilog" => Seq(RunFirrtlTransformAnnotation(new VerilogEmitter),
+                                           EmitCircuitAnnotation(classOf[VerilogEmitter]))
+        case "sverilog"             => Seq(RunFirrtlTransformAnnotation(new SystemVerilogEmitter),
+                                           EmitCircuitAnnotation(classOf[SystemVerilogEmitter]))
+        case _                      => throw new PhaseException(s"Unknown emitter '$x'! (Did you misspell it?)")
+      }
+      xx ++ c }
+    .text("Run the specified circuit emitter (all modules in one file)")
+}
+
+object EmitAllModulesAnnotation extends HasScoptOptions {
+  def addOptions(p: OptionParser[AnnotationSeq]): Unit = p
+    .opt[String]("emit-modules")
+    .abbr("e")
+    .valueName("<none|high|middle|low|verilog|mverilog|sverilog>")
+    .unbounded()
+    .action{ (x, c) =>
+      val xx = x match {
+        case "chirrtl"              => Seq(RunFirrtlTransformAnnotation(new ChirrtlEmitter),
+                                           EmitAllModulesAnnotation(classOf[ChirrtlEmitter]))
+        case "high"                 => Seq(RunFirrtlTransformAnnotation(new HighFirrtlEmitter),
+                                           EmitAllModulesAnnotation(classOf[HighFirrtlEmitter]))
+        case "middle"               => Seq(RunFirrtlTransformAnnotation(new MiddleFirrtlEmitter),
+                                           EmitAllModulesAnnotation(classOf[MiddleFirrtlEmitter]))
+        case "low"                  => Seq(RunFirrtlTransformAnnotation(new LowFirrtlEmitter),
+                                           EmitAllModulesAnnotation(classOf[LowFirrtlEmitter]))
+        case "verilog" | "mverilog" => Seq(RunFirrtlTransformAnnotation(new VerilogEmitter),
+                                           EmitAllModulesAnnotation(classOf[VerilogEmitter]))
+        case "sverilog"             => Seq(RunFirrtlTransformAnnotation(new SystemVerilogEmitter),
+                                           EmitAllModulesAnnotation(classOf[SystemVerilogEmitter]))
+        case _                      => throw new PhaseException(s"Unknown emitter '$x'! (Did you misspell it?)")
+      }
+      xx ++ c }
+    .text("Run the specified module emitter (one file per module)")
+}
+
 // ***** Annotations for results of emission *****
 sealed abstract class EmittedComponent {
   def name: String
   def value: String
+  def outputSuffix: String
 }
 sealed abstract class EmittedCircuit extends EmittedComponent
-final case class EmittedFirrtlCircuit(name: String, value: String) extends EmittedCircuit
-final case class EmittedVerilogCircuit(name: String, value: String) extends EmittedCircuit
+final case class EmittedFirrtlCircuit(name: String, value: String, outputSuffix: String) extends EmittedCircuit
+final case class EmittedVerilogCircuit(name: String, value: String, outputSuffix: String) extends EmittedCircuit
 sealed abstract class EmittedModule extends EmittedComponent
-final case class EmittedFirrtlModule(name: String, value: String) extends EmittedModule
-final case class EmittedVerilogModule(name: String, value: String) extends EmittedModule
+final case class EmittedFirrtlModule(name: String, value: String, outputSuffix: String) extends EmittedModule
+final case class EmittedVerilogModule(name: String, value: String, outputSuffix: String) extends EmittedModule
 
 /** Traits for Annotations containing emitted components */
 sealed trait EmittedAnnotation[T <: EmittedComponent] extends NoTargetAnnotation {
@@ -51,18 +107,20 @@ sealed trait EmittedCircuitAnnotation[T <: EmittedCircuit] extends EmittedAnnota
 sealed trait EmittedModuleAnnotation[T <: EmittedModule] extends EmittedAnnotation[T]
 
 case class EmittedFirrtlCircuitAnnotation(value: EmittedFirrtlCircuit)
-  extends EmittedCircuitAnnotation[EmittedFirrtlCircuit]
+    extends EmittedCircuitAnnotation[EmittedFirrtlCircuit]
 case class EmittedVerilogCircuitAnnotation(value: EmittedVerilogCircuit)
-  extends EmittedCircuitAnnotation[EmittedVerilogCircuit]
+    extends EmittedCircuitAnnotation[EmittedVerilogCircuit]
 
 case class EmittedFirrtlModuleAnnotation(value: EmittedFirrtlModule)
-  extends EmittedModuleAnnotation[EmittedFirrtlModule]
+    extends EmittedModuleAnnotation[EmittedFirrtlModule]
 case class EmittedVerilogModuleAnnotation(value: EmittedVerilogModule)
-  extends EmittedModuleAnnotation[EmittedVerilogModule]
+    extends EmittedModuleAnnotation[EmittedVerilogModule]
 
 sealed abstract class FirrtlEmitter(form: CircuitForm) extends Transform with Emitter {
   def inputForm = form
   def outputForm = form
+
+  val outputSuffix: String = form.outputSuffix
 
   private def emitAllModules(circuit: Circuit): Seq[EmittedFirrtlModule] = {
     // For a given module, returns a Seq of all modules instantited inside of it
@@ -87,15 +145,15 @@ sealed abstract class FirrtlEmitter(form: CircuitForm) extends Transform with Em
         case ext: ExtModule => ext
       }
       val newCircuit = Circuit(m.info, extModules :+ m, m.name)
-      EmittedFirrtlModule(m.name, newCircuit.serialize)
+      EmittedFirrtlModule(m.name, newCircuit.serialize, outputSuffix)
     }
   }
 
   override def execute(state: CircuitState): CircuitState = {
     val newAnnos = state.annotations.flatMap {
       case EmitCircuitAnnotation(_) =>
-        Seq(EmittedFirrtlCircuitAnnotation.apply(
-              EmittedFirrtlCircuit(state.circuit.main, state.circuit.serialize)))
+        Seq(EmittedFirrtlCircuitAnnotation(
+              EmittedFirrtlCircuit(state.circuit.main, state.circuit.serialize, outputSuffix)))
       case EmitAllModulesAnnotation(_) =>
         emitAllModules(state.circuit) map (EmittedFirrtlModuleAnnotation(_))
       case _ => Seq()
@@ -129,6 +187,7 @@ case class VRandom(width: BigInt) extends Expression {
 class VerilogEmitter extends SeqTransform with Emitter {
   def inputForm = LowForm
   def outputForm = LowForm
+  val outputSuffix = ".v"
   val tab = "  "
   def AND(e1: WrappedExpression, e2: WrappedExpression): Expression = {
     if (e1 == e2) e1.e1
@@ -363,8 +422,8 @@ class VerilogEmitter extends SeqTransform with Emitter {
     * @return             the render reference
     */
   def getRenderer(descriptions: Seq[DescriptionAnnotation],
-    m: Module,
-    moduleMap: Map[String, DefModule])(implicit writer: Writer): VerilogRender = {
+                  m: Module,
+                  moduleMap: Map[String, DefModule])(implicit writer: Writer): VerilogRender = {
     val newMod = new AddDescriptionNodes().executeModule(m, descriptions)
 
     newMod match {
@@ -384,9 +443,9 @@ class VerilogEmitter extends SeqTransform with Emitter {
     * @param writer           where rendered information is placed.
     */
   class VerilogRender(description: Description,
-    portDescriptions: Map[String, Description],
-    m: Module,
-    moduleMap: Map[String, DefModule])(implicit writer: Writer) {
+                      portDescriptions: Map[String, Description],
+                      m: Module,
+                      moduleMap: Map[String, DefModule])(implicit writer: Writer) {
 
     def this(m: Module, moduleMap: Map[String, DefModule])(implicit writer: Writer) {
       this(EmptyDescription, Map.empty, m, moduleMap)(writer)
@@ -458,7 +517,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
       assigns += Seq("assign ", e, " = ", syn, ";", info)
       assigns += Seq("`else")
       assigns += Seq("assign ", e, " = ", garbageCond, " ? ", rand_string(syn.tpe), " : ", syn,
-        ";", info)
+                     ";", info)
       assigns += Seq("`endif // RANDOMIZE_GARBAGE_ASSIGN")
     }
 
@@ -550,7 +609,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
       initials += Seq("`ifdef RANDOMIZE_MEM_INIT")
       initials += Seq("for (initvar = 0; initvar < ", bigIntToVLit(s.depth), "; initvar = initvar+1)")
       initials += Seq(tab, WSubAccess(wref(s.name, s.dataType), index, s.dataType, FEMALE),
-        " = ", rstring, ";")
+                      " = ", rstring, ";")
       initials += Seq("`endif // RANDOMIZE_MEM_INIT")
     }
 
@@ -696,14 +755,14 @@ class VerilogEmitter extends SeqTransform with Emitter {
           instdeclares += Seq(");")
         case sx: DefMemory =>
           val fullSize = sx.depth * (sx.dataType match {
-            case GroundType(IntWidth(width)) => width
-          })
+                                       case GroundType(IntWidth(width)) => width
+                                     })
           val decl = if (fullSize > (1 << 29)) "reg /* sparse */" else "reg"
           declareVectorType(decl, sx.name, sx.dataType, sx.depth, sx.info)
           initialize_mem(sx)
           if (sx.readLatency != 0 || sx.writeLatency != 1)
             throw EmitterException("All memories should be transformed into " +
-              "blackboxes or combinational by previous passses")
+                                     "blackboxes or combinational by previous passses")
           for (r <- sx.readers) {
             val data = memPortField(sx, r, "data")
             val addr = memPortField(sx, r, "addr")
@@ -717,7 +776,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
 
             //; Read port
             assign(addr, netlist(addr), NoInfo) // Info should come from addr connection
-            // assign(en, netlist(en))     //;Connects value to m.r.en
+                                                // assign(en, netlist(en))     //;Connects value to m.r.en
             val mem = WRef(sx.name, memType(sx), MemKind, UNKNOWNGENDER)
             val memPort = WSubAccess(mem, addr, sx.dataType, UNKNOWNGENDER)
             val depthValue = UIntLiteral(sx.depth, IntWidth(sx.depth.bitLength))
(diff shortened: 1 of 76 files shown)


chick:
This is a spectacularly tiny nit, but what do you think about adding an AnnotationSeq.empty as a convenience method. Higher up the toolchain there's sometimes a need to provide a default list of annotations.

seldridge:
Clarifying, @chick: Are you proposing a convenience method for `AnnotationSeq.empty` or a convenience method for generating default FIRRTL annotations, i.e., `firrtl.stage.phases.AddDefaults.transform(AnnotationSeq.empty)`?

chick:
I was specifically thinking about creating an empty AnnotationSeq, more than once I've created methods like `def f(a: A, annotations: AnnotationSeq = Seq.empty)` but perhaps that's less likely now since there should in almost all uses cases be some annotations, and, as long as the type is explicit, the Seq.empty works.

seldridge:
I'd be fine with it, but can see what others think. You're right, though. The only time that you would typically be creating an empty `AnnotationSeq` is in a `Stage`. This should be relatively infrequent as typically you're chaining `Phase`s together.

src/main/scala/firrtl/options/Phase.scala
```diff
@@ -3,18 +3,90 @@
...
+
+  /** The name of this [[Phase]]. This will be used to generate debug/error messages or when deleting annotations. This
+    * will default to the `simpleName` of the class.
+    * @return this phase's name
+    * @note Override this with your own implementation for different naming behavior.
+    */
+  lazy val name: String = this.getClass.getName
+
+  /** Perform the transform of [[transform]] on an [[firrtl.AnnotationSeq AnnotationSeq]] and add
+    * [[firrtl.annotations.DeletedAnnotation DeletedAnnotation]]s for any deleted [[firrtl.annotations.Annotation
```
azidar: Scaladoc looks weird here? Did you mean to repeat AnnotationSeq and DeletedAnnotation?

src/main/scala/firrtl/options/Phase.scala
```diff
@@ -3,18 +3,90 @@
...
+    */
+  final def runTransform(annotations: AnnotationSeq): AnnotationSeq = {
+    val ax = transform(annotations)
+
+    val (in, out) = (mutable.LinkedHashSet() ++ annotations, mutable.LinkedHashSet() ++ ax)
+
+    (in -- out).map {
+      case DeletedAnnotation(n, a) => DeletedAnnotation(s"$n+$name", a)
+      case a                       => DeletedAnnotation(name, a)
+    }.toSeq ++ ax
```
azidar: Does it matter that you are prepending, rather than appending, for the parallel compilation grammar?

src/main/scala/firrtl/options/Phase.scala
```diff
@@ -3,18 +3,90 @@
...
+    * @note Override this with your own implementation for different naming behavior.
+    */
+  lazy val name: String = this.getClass.getName
+
+  /** Perform the transform of [[transform]] on an [[firrtl.AnnotationSeq AnnotationSeq]] and add
+    * [[firrtl.annotations.DeletedAnnotation DeletedAnnotation]]s for any deleted [[firrtl.annotations.Annotation
+    * Annotation]]s.
+    * @param a
+    */
+  final def runTransform(annotations: AnnotationSeq): AnnotationSeq = {
```
azidar: So I'm not convinced that this internal transform + external runTransform paradigm is really the best way to do this. Perhaps its worth brainstorming for a bit to see if we could improve it?

src/main/scala/firrtl/options/StageAnnotations.scala
```diff
@@ -3,40 +3,89 @@
...
 
 import scopt.OptionParser
 
-sealed trait StageOption extends HasScoptOptions
+sealed trait StageOption { this: Annotation => }
+
+/** An annotation that should not be serialized automatically [[phases.WriteOutputAnnotations WriteOutputAnnotations]].
+  * This usually means that this is an annotation that is used only internally to a [[Stage]].
+  */
+trait Unserializable { this: Annotation => }
```
azidar: Should there be a class of Unserializables that error if you try to serialize them, instead of silently not serializing?

src/main/scala/firrtl/options/StageAnnotations.scala
```diff
@@ -3,40 +3,89 @@
...
-object InputAnnotationFileAnnotation {
-  private [firrtl] def apply(): InputAnnotationFileAnnotation = InputAnnotationFileAnnotation("")
+/** An explicit output _annotation_ file to write to
+  *  - set with `-foaf/--output-annotation-file`
+  * @param value output annotation filename
+  */
+case class OutputAnnotationFileAnnotation(file: String) extends NoTargetAnnotation with StageOption
+
+object OutputAnnotationFileAnnotation extends HasScoptOptions {
+  def addOptions(p: OptionParser[AnnotationSeq]): Unit = p.opt[String]("output-annotation-file")
```
azidar: What do you think about making HasScoptOptions have unimplemented members corresponding to opt, abbr, valueName, action, and text, and then implementing addOptions in the parent class? This would hide scopt from annotations writers who want commandline options but don't want to learn about scopt.

It would also enable you to hide the prepend/append business with another unimplemented member called isGlobal or something like that.

src/main/scala/firrtl/options/phases/Checks.scala
```diff
@@ -0,0 +1,43 @@
...
+  * as [[StageOptions]] are guaranteed to succeed.
+  */
+object Checks extends Phase {
+
+  /** Validate an [[AnnotationSeq]] for [[StageOptions]]
+    * @throws OptionsException if annotations are invalid
+    */
+  def transform(annotations: AnnotationSeq): AnnotationSeq = {
+
+    val td, outA = collection.mutable.ListBuffer[Annotation]()
```
azidar: I had no idea you could do this. Wow, learned something new!

src/main/scala/firrtl/stage/FirrtlCli.scala
```diff
@@ -0,0 +1,23 @@
+// See LICENSE for license details.
+
+package firrtl.stage
+
+import firrtl.options.Shell
+
+/** [[firrtl.options.Shell Shell]] mixin that provides command line options for FIRRTL. This does not include any
+  * [[firrtl.options.RegisteredLibrary RegisteredLibrary]] or [[firrtl.options.RegisteredTransform RegisteredTransform]]
+  * as those are automatically loaded by the [[firrtl.options.Stage Stage]] using this [[firrtl.options.Shell Shell]].
```
azidar: Duplicated names here again, is this just a rendering thing?

src/main/scala/firrtl/stage/phases/Checks.scala
```diff
@@ -0,0 +1,104 @@
...
+  * extremeley strict. Nothing is inferred or assumed to take a default value (for default value resolution see
+  * [[AddDefaults]]).
+  *
+  * The intent of this approach is that after running this [[firrtl.options.Phase Phase]], a user can be absolutely
+  * certain that other [[firrtl.options.Phase Phase]]s or views will succeed. See [[FirrtlStage]] for a list of
+  * [[firrtl.options.Phase Phase]] that commonly run before this.
+  */
+object Checks extends Phase {
+
+  /** Determine if an annotations are sane
```
azidar: grammar

src/test/scala/firrtlTests/options/phases/GetIncludesSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+class GetIncludesSpec extends FlatSpec with Matchers with BackendCompilationUtilities with firrtlTests.Utils {
+
+
+  val dir = new File("test_run_dir/GetIncludesSpec")
+  dir.mkdirs()
+
+  def ref(filename: String): InputAnnotationFileAnnotation = InputAnnotationFileAnnotation(s"$dir/$filename.anno.json")
+
+  def checkAnnos(a: AnnotationSeq, b: AnnotationSeq): Unit = {
+    info("read the exepcted number of annotations")
```
azidar: spelling

src/test/scala/firrtlTests/stage/FirrtlStageSpec.scala
```diff
@@ -0,0 +1,415 @@
...
+      out should not include ("Warning:")
+
+      And("no error should be printed")
+      out should not include ("Error:")
+
+      And("the exit code should be 0")
+      result shouldBe a [Right[_,_]]
+    }
+
+    scenario("User provides unsupported legacy annotations") {
```
azidar: This scenario thing is really cool, I would be good to hear about your experience with it.

src/main/scala/firrtl/options/Phase.scala
seldridge: I pushed a commit that handles this more manually with a `DeletedWrapper` `Phase`. See what you think. I like this as it's scalable... We can keep wrapping `Phase`s in additional processing steps as needed.

src/main/scala/firrtl/options/Phase.scala
seldridge: I think this is safe as a `DeletedAnnotation` won't be used in parallel compilation. There's a question of preserving ordering as the `Set` operations will mangle that. I do think that's okay, though. This logic is now moved into the `DeletedWrapper`.

src/main/scala/firrtl/options/Phase.scala
seldridge: Removed due to removal of `runTransform` method.

src/main/scala/firrtl/options/StageAnnotations.scala
seldridge: I actually think that all `Unserializable`s are the same. However, you want to be able to change the behavior of this in certain scenarios. Two scenarios come to mind:

1. User writes a `Stage` and the automatic `WriteOutputAnnotations` runs. 
2. User wants to run `WriteOutputAnnotations` after every `Phase` for debugging reasons.

In the former case, an `Unserializable` is a "stage local" annotation that was supposed to be removed, deleted, or just not supposed to exist at the end of the `Stage`. In the latter case, an `Unserializable` may exist, but this is okay and should be ignored. 

It initially seemed like this means `WriteOutputAnnotations` should be parameterized. However, I now think that `WriteOutputAnnotations` should ignore `Unserializable`. I need to add additional `Checks` that run at the end of a `Stage` to verify that all `Unserializable` were removed.

The assumption here is that `Unserializable`s are not supposed to exit a `Stage`.

What do you think? Thoughts @chick?

chick:
The problem for testers is that they need the instantiated chisel module in order to find the top level ports. It's the ChiselStage that produces an unserializable ChiselCircuitAnnotation from a unserializable ChiselGeneratorAnnotation. I don't see an easy way to get around need this to make it back. Testers calls the Treadle stage which uses the emitted firrtl is used to generate a TreadleTestersAnnotation, also unserializable. These are all unserializable things that are needed after the phases that generate them

src/main/scala/firrtl/options/StageAnnotations.scala
seldridge: Thanks for the thoughts below @chick. Certain `Unserializable` annotations do currently need to exit `ChiselStage` and these therefore can't be entirely purged.

I think things should likely then be left the way they are for now.

I do think that this hints at a larger problem: there needs to be some way of passing information from earlier in the pipeline to later in the pipeline. This can likely be done either by splitting the `AnnotationSeq` as it goes along or with a modification to the existing `DeletedAnnotation`. 

`DeletedAnnotation` doesn't really communicate what you need here. You want to have some history about what was used to create a specific `Annotation`. There could be an optional `evidence` member added to an `Annotation` that can include information about what was used to create that `Annotation`. 

E.g., we currently have the following:

  - `ChiselGeneratorAnnotation(gen: () => RawModule)`
  - `ChiselCircuitAnnotation(circuit: chisel3.internal.firrtl.Circuit)`
  - `FirrtlCircuitAnnotation(circuit: firrtl.ir.Circuit)`
  - `EmittedAnnotation[T <: EmittedComponent]{ def value: T }`

The end result of a run of `ChiselStage` -> `FirrtlStage` would then be one or more `EmittedAnnotation` that includes evidence about how it got to that point. The evidence would be excluded from the serialization or could be converted to `DeletedAnnotation`s.

Alternatively, the notion of splitting an `AnnotationSeq` could be resolved with the `DependencyAPI`. 

As a further alternative, testers and treadle could use the constituent `Phase`s to construct a different `Stage` if that made sense.

src/main/scala/firrtl/stage/FirrtlCli.scala
seldridge: I think the names are okay here. Both `RegisteredLibrary` and `RegisteredTransform` should likely be mentioned.

src/main/scala/firrtl/options/phases/Checks.scala
seldridge: Yep... it makes things easy when you have 12 of these things...

src/main/scala/firrtl/stage/phases/Checks.scala
seldridge: Fixed in rebase.

src/test/scala/firrtlTests/options/phases/GetIncludesSpec.scala
seldridge: Thanks! Fixed in rebase.

src/test/scala/firrtlTests/stage/FirrtlStageSpec.scala
seldridge: All the scaladoc testing styles are interesting. The use of `FeatureSpec with `GivenWhenThen` is that it enables more "acceptance" style testing. You can get somewhat creative with trying to think of how a user would use a `Stage` from the command line versus testing all the individual components. I generally think this makes sense for testing user-exposed `Stage`s, but not so much for the internal stuff. I usually like `FlatSpec` with a lot of `info()` methods there.

Generally, it may be worth a discussion at some point about aligning to some testing style.

src/main/scala/firrtl/Emitter.scala
```diff
@@ -31,17 +34,68 @@ sealed trait EmitAnnotation extends NoTargetAnnotation {
...
+      shortOption = Some("E"),
+      helpValueName = Some("<chirrtl|high|middle|low|verilog|mverilog|sverilog>") ) )
+
+}
+
+object EmitAllModulesAnnotation extends HasShellOptions {
+
+  val options = Seq(
+    new ShellOption[String](
+      longOption = "emit-modules",
```
jackkoenig: I can't seem to find where `--emit-modules ` was deleted but can you add a better message for when people try to use it, similarly to what you did with `--top-name`?

seldridge: Just to clarify, you're talking about `--split-modules` yes?

jackkoenig: Yeah sorry, updated

src/main/scala/firrtl/stage/phases/DriverCompatibility.scala
```diff
@@ -0,0 +1,223 @@
...
+    * as opposed to an [[EmitCircuitAnnotation]].
+    */
+  private [firrtl] case object EmitOneFilePerModuleAnnotation extends NoTargetAnnotation {
+
+    def addOptions(p: OptionParser[AnnotationSeq]): Unit = p
+      .opt[Unit]("split-modules")
+      .abbr("fsm")
+      .hidden
+      .unbounded
+      .action( (_, _) => throw new OptionsException(optionRemoved("--split-modules/-fsm")) )
```
seldridge: @jackkoenig: This will now provide the same behavior as `--top-name/-tn`.

src/main/scala/firrtl/Emitter.scala
seldridge: Fixed in rebase. This includes a test, like with `--top-name/-tn`.

jackkoenig:
I've been able to narrow down that when using `firrtl.stage.FirrtlStage`, it runs the emitter always (even if you don't give a `-E/-e` directive), and if you *do* give `-E/-e`, it runs the emitter twice

jackkoenig:
I'm looking into a fix, but don't have one yet, and I'm out tomorrow through Sunday

seldridge:
I would prefer if we could remove the auto-emitter execution (make compilers no longer abstract in emitter). I was messing with that a while ago, but didn't get far.

Should be coming up for air next week...

Thanks for digging into this @jackkoenig.

seldridge:
@jackkoenig: I've rebased with what I think is a fix for the emitter issue. 

The problem with the way emitters worked before was that there were two types of emitters that could run:

1. The explicit emitter that a `Compiler` would run after all transforms for that compiler
2. An emitter scheduled as a `Transform`

The fix here removes the explicit execution of an emitter (the former option 1) and adds an implicit emitter that is scheduled like the latter option 2. 

Relying solely on option 2 doesn't work for custom transforms, however. E.g., if you add a custom transform that runs on HighFirrtl (`-X high --custom-transforms Foo`) then it's not clear what the user intends. This expands to `CompilerAnnotation(HighFirrtlCompiler), RunFirrtlTransformAnnotation(Foo)`. The new `AddImplicitTransform` phase then expands the compiler annotation into `CompilerAnnotation(HighFirrtlCompiler), RunFirrtlTransformAnnotation(HighFirrtlEmitter), EmitterAnnotation(class[HighFirrtlEmitter])`. However, the transform ordering is now confused. Should the emitter run before or after `Foo`?

This avoids the issue by scheduling emitters as late as possible. `Compiler.mergeTransforms` is modified to sort all emitters last before making scheduling decisions. 

Note: this causes misuses of `*TransformSpec` to break. If you are using this to test a transform that is lower than that transform spec, the emitter will run before your transform. `Compiler.mergeTransforms` has an assumption that things are ordered sanely.

**A real fix would be part of a dependency API.**

seldridge:
ping: @jackkoenig, @azidar, @chick. Corporate wheels moving... CLA signed... LFG :rocket:

If this is going in the FIRRTL 1.2 / Chisel 3.2 window, how do we want to proceed?

Backlogging: the only lingering question was about the `Unserializable` annotation. We can add something that hard errors if you try to serialize it (e.g., to indicate that an annotation was expected to be deleted, but was not).

jackkoenig:
I think we should proceed by merging. Let me rerun some tests but I'm pretty sure we're ready to pull the trigger

src/main/scala/firrtl/stage/FirrtlStage.scala
```diff
@@ -0,0 +1,38 @@
...
+import firrtl.{AnnotationSeq, CustomTransformException, FIRRTLException, Utils}
+import firrtl.options.{Stage, Phase, PhaseException, Shell, OptionsException}
+import firrtl.options.phases.DeletedWrapper
+import firrtl.passes.{PassException, PassExceptions}
+
+import scala.util.control.ControlThrowable
+
+import java.io.PrintWriter
+
+class FirrtlStage extends Stage {
```
jackkoenig: Should this be an object?
```
$ java -cp utils/bin/firrtl.jar firrtl.stage.FirrtlStage -i regress/ICache.fir -o out -X verilog
Error: Main method is not static in class firrtl.options.Stage, please define the main method as:
   public static void main(String[] args)
```

Interestly, and I don't pretend to know why this worked, but:
```
$ java -jar utils/bin/firrtl.jar firrtl.stage.FirrtlStage -i regress/ICache.fir -o out -X verilog                        
------------------------------------------------------------------------------
Warning: firrtl.Driver is deprecated since 1.2!
Please switch to firrtl.stage.FirrtlStage
------------------------------------------------------------------------------
Total FIRRTL Compile Time: 1081.4 ms
]
```

seldridge: I changed everything to classes so that the dependencies could be defined as `Class[A <: DependencyAPI[A]]` (see: https://github.com/freechipsproject/firrtl/pull/1015/files#diff-8cd99845a8f2c996dfdce4295c252b3cR26).

I think the clean approach is to add an object that constructs a `FirrtlStage`.

seldridge: Current solution: 
  - Move the `main` method inside `Stage` into a new class `StageMain`
  - Add an `object FirrtlMain extends StageMain(new FirrtlStage)`
  - Change name of `FirrtlStageSpec` to `FirrtlMainSpec` and update this to use `FirrtlMain`

ucbjrl:
We're failing in chisel-testers due to the fact that we have multiple `targetDir`s. This in turn appears to be due to the following in firrtl Driver.scala:217
```scala
    val annos = optionsManager.firrtlOptions.toAnnotations ++ optionsManager.commonOptions.toAnnotations
```
and the fact that we have a TargetDirAnnotation in both `firrtlOptions` and `commonOptions`.

I think this is a symptom of _who's on top_ which will eventually be addressed by the SiFive global restructuring, but until then, we need some way for the _top level_ driver to compose potentially conflicting annotations. For annotations that **must be unique**, we need a mechanism to choose one.

seldridge:
Ugh... Thanks @ucbjrl. If we need to revert, do it.

That should only be producing one `TargetDirAnnotaion`, I think, as only `commonOptions` can store it and only produces one annotation. Hmmm...

ucbjrl:
No need to revert. I think there should be a quick short-term fix. I'll investigate on Monday.

seldridge:
Sounds good. I did some poking around. The underlying issue is replicated across the backends. The `firrtl.Driver.getAnnotations` method is called (which includes a `TargetDirAnnotation`). These are then passed to an options manager in its `annotations` member. That then causes a duplication of the `TargetDirAnnotation`. E.g.:
  - https://github.com/freechipsproject/chisel-testers/blob/master/src/main/scala/chisel3/iotesters/FirrtlTerpBackend.scala#L134

I'm not sure of the best solution here. I think `getAnnotations` should probably stay the same for backwards compatibility, but the `TargetDirAnnotation` should be filtered out of each of the backends if it does this.

PR #1007, opened 2019-01-26
Move FileUtils code from Driver to FileUtils and use java.nio methods.

ucbjrl:
This will require `import firrtl.utils.FileUtils` in client code - currently just chisel-testers in the Big6 repos.


azidar (changes requested):
Let's wait on this PR until after #1005 is merged, as they will conflict. Otherwise, this seems fine to me. Thanks!

chick:
Should we revivify this since it's been a year since #1005 was merged

Issue #978, opened 2018-12-19
Repeated Testing Exhausts Metaspace


seldridge:
**Type of issue**: bug report

**If the current behavior is a bug, please provide the steps to reproduce the problem:**

1. Open up an `sbt` shell
2. Run `test`
3. `goto` (2)

  * **What is the current behavior?**

Eventually, your `sbt` shell will hang because you ran out of metaspace. The originating exception may be pretty far up, but looks like:

```
[error] java.util.concurrent.ExecutionException: java.lang.OutOfMemoryError: Metaspace
[error] 	at java.util.concurrent.FutureTask.report(FutureTask.java:122)
[error] 	at java.util.concurrent.FutureTask.get(FutureTask.java:192)
[error] 	at sbt.ConcurrentRestrictions$$anon$4.take(ConcurrentRestrictions.scala:213)
[error] 	at sbt.Execute.next$1(Execute.scala:110)
[error] 	at sbt.Execute.processAll(Execute.scala:113)
[error] 	at sbt.Execute.runKeep(Execute.scala:90)
[error] 	at sbt.EvaluateTask$.liftedTree1$1(EvaluateTask.scala:420)
```

This is truncated as there's really no useful information here (as far as I'm aware).

Alternatively, you sometimes get the prettier:

```
Exception in thread "Timer-3" java.lang.OutOfMemoryError: Metaspace
[info] shutting down server
java.lang.OutOfMemoryError: Metaspace
Error during sbt execution: java.lang.OutOfMemoryError: Metaspace
```

  * **What is the expected behavior?**

Repeated testing should work.

  * **Please tell us about your environment:**
    (examples)
    - version: 3655ae091249a72bd424073cfb4a382a5ab170c6
    - OS: `Linux knight 4.4.0-92-generic #115-Ubuntu SMP Thu Aug 10 09:04:33 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux`
    - Scala: 2.12.7 or 2.11.12
    - `sbt`: 1.2.7 

**What is the use case for changing the behavior?**

This only really matters for developers as far as I'm aware. I like to keep an `sbt` session open and repeatedly run `test` if I'm dealing with some patch that has far reaching implications. I can avoid this with invoking `sbt test` explicitly, but then you get that `sbt` startup time.

[...]

PR #1008, opened 2019-01-26
Use apache commons for string escaping instead of reflection

jackkoenig:
In order to avoid writing escaping/unescaping logic, I used the internal Scala String escaping. It **might** be the case that this is causing the out of metaspace issues so I'm trying this out. I don't love pulling in a whole dependency just for this, so we can write it ourselves instead, but this was lower effort.

Possible Fix to #978 

This also should speed stuff up a bit

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -69,6 +69,8 @@ libraryDependencies += "net.jcazevedo" %% "moultingyaml" % "0.4.0"
 
 libraryDependencies += "org.json4s" %% "json4s-native" % "3.6.1"
 
+libraryDependencies += "org.apache.commons" % "commons-text" % "1.6"
+
 // Java PB
 
 enablePlugins(ProtobufPlugin)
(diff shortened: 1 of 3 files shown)


edwardcwang:
What caused such a dramatic performance improvement? Is the internal Scala escaping just super slow/inefficient? Is it related to having to import it?

seldridge:
I think this makes sense, but I'm still seeing the metaspace issues after ~3 runs of `sbt test` on FIRRTL. Ugh...

jackkoenig:
With some additional analysis, this does **not** seem to fix #978, but it **does** seem to fix a related issue where repeatedly running main in an sbt shell would also run out of metaspace.

For this experiment, I repeatedly ran `runMain firrtl.Driver -i test.fir -o out.v -X verilog` on a very basic firrtl file.

-----

### Without this PR

And once it got to that flatline at the end it would not run anymore, it would just error and say `java.lang.OutOfMemoryError: Metaspace`
screen shot 2019-01-31 at 2 54 32 pm

-----

### With this PR
screen shot 2019-01-31 at 2 57 53 pm

Yay! GC works!

Relatedly, as I noted above, the runtime for this tiny design dropped from 1-1.5s to 300-400ms
Honestly that's still pretty slow but **way** better.

jackkoenig:
Minor downside of this PR, it does increase the size of the fat jar by 3.6%, from 16.8 MB to 17.4 MB.

An alternative is to implement [de]escaping ourselves.

PR #1009, opened 2019-01-30
Missed constprop opportunity

aswaterman:
Failing test case is attached in the form of a PR.  Basically, the following could be constpropped but isn't:

    reg r : UInt<8>, clock with : (reset => (reset, UInt<4>("hb")))
    when cond :
      r <= UInt<4>("hb")

Naturally, I'm only guessing about what the expected output should be, since currently the output is unoptimized.


albert-magyar:
I updated the test case, since the `cond` port was missing from the expected output (no DCE).

aswaterman:
@albert-magyar 👍

albert-magyar:
I wrote a new version of register constant prop that is much more general. Basically, it walks back along nodes to ensure that all drivers of a register are a mix of a) references to the register itself and b) instances of a common constant.

It exits as soon as it finds anything other than a literal, mux, or node, which means it should do the minimal amount of circuit walking necessary to cover these cases. However, it should catch more complicated patterns (wires/logic) in concert with the pattern of running ~~`pdflatex`~~ `ConstantPropagation` three times.

I added a fun example that works now:

```
  reg r : UInt<8>, clock with : (reset => (reset, UInt<4>("h7")))
  r <= r
  when eq(cmd, UInt<3>("h0")) :
    r <= UInt<4>("h7")
  else :
    when eq(cmd, UInt<3>("h1")) :
      r <= r
    else :
      when eq(cmd, UInt<3>("h2")) :
        r <= UInt<4>("h7")
      else :
        r <= r
  z <= r
```

The register gets successfully const propped.

albert-magyar:
retest this please

azidar:
Update after chatting with @albert-magyar - looks like constprop will already handle simplifying the padding of literals, so just add a test case demonstrating it works and feel free to either add ScalaDoc or line comments.

albert-magyar:
@azidar, I added scaladoc for `regConstant` and inline comments on pattern-matched cases. I merged `compatibleConstants` into `regConstant`, since it actually cleaned up the code and it would a bit confusing to split the descriptions of the two. It also cleans up the excessive `Option`-based signature.

Issue #695, opened 2017-12-13
Negedge-Triggered Things (memories, registers, etc.)


seldridge:
Tracking issue related to a discussion of adding support for negative edge triggered things in FIRRTL. Motivated by @ccelio's Chisel3 issue: https://github.com/freechipsproject/chisel3/issues/736.

Roughly, I see this as:
  - [ ] Extending clocked IR statements (`DefRegister`, `DefMemory`) to specify "_edgedness_"
  - [ ] Define edgedness, assumedly as `Set(Posedge, Negedge, PosNegEdge)`
  - [ ] Extending the Verilog emitter to use the appropriate edgedness when generating statements
  - [ ] Update the specification document to include edgedness

Couple of questions:
  - Thoughts on the approach?
  - Is this excluded from the FIRRTL specification intentionally and is there some motivation for not addressing negative edge clocking?

Possible problems/alternative solutions:
  - The original intent is to only provide the capability to emit negedge triggered SRAMs. This may be better handled as just some tweaks to the `ReplSeqMem` transform or a similar transform that corrects clocking. However, by my rough understanding a full solution would likely require extending the IR.
  - This could be viewed as entirely out of scope and should be handled via blackboxing (a view that would likely hinder adoption).

azidar:
A few more thoughts to consider. Adding more to the IR requires all transformations to need to consider it. For example, all MIDAS transformations would have to address Negedge, or explicitly disallow it. Before we add more burden to transformation writers, I think its important to understand (1) how common this feature really is in practice, (2) whether the current practice is *best* practice and not an anti-pattern, and (3) how difficult it is to support in the toolchain (testers, interpreters). We shouldn't half-support a feature.

azidar:
Any updates on examples? We discuss this during our research meeting, and came up with the following that would need negedge support, as well as explicit latches.

- JTAG block
- Physical design of register file
- DDR signals
- SRAM (BOOM)

seldridge:
Maybe @ccelio has some thoughts on additional examples.

This seems to cover it. I'd be more inclined to limit the motivation to two groups: "negedge latches" and "negedge memories". I think everything else falls into one of these two categories (which aligns with the IR).

No updates, nor have I had a chance to work out a sketch of a PR. This isn't something that's affecting me in any way currently.

Towards your points above:
[...]

PR #1011, opened 2019-02-02
Asynchronous Reset

### LINKED ISSUE #219: Support for asynchronously-resettable registers

jackkoenig:
I think this is more-or-less good to review!

Summary of changes:
* Adds AsyncResetType (similar to ClockType)
* Registers who's reset signal is of type AsyncResetType are async reset registers
* Registers with async reset can only be reset to literal values

I've probably missed a weird interaction somewhere but this does have several tests including an execution driven one.

@edwardcwang, per discussion with @aswaterman (and please pipe up if I miss[characterize] something), but it seems that the restriction of only allowing literals for async reset registers reset values is the right way to go. You generally can't have an async reset register with the init value driven by an input because in hierarchical synthesis and place-and-route flows you are doing things hierarchically (obviously) so as far as the tool is concerned the "constant" input is just an arbitrary input. Put another way, making constant values inputs makes them treated as arbitrary inputs so the tools can't synthesize asynchronous reset registers from them. Due to this restriction, it's just not a thing that people generally need to express.

TODO:
- [ ] Add AsyncReset to the spec
- [ ] Audit for assumptions based on synchronous reset
- [ ] Use in rocket-chip as proof

META:
- Fixes #219

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -420,6 +420,16 @@ class VerilogEmitter extends SeqTransform with Emitter {
     // but Formality doesn't allow more than 1 statement inside async reset always blocks
     val asyncResetAlwaysBlocks = mutable.ArrayBuffer[(Expression, Expression, Seq[Any])]()
     val initials = ArrayBuffer[Seq[Any]]()
+    // In Verilog, async reset registers are expressed using always blocks of the form:
+    // always @(posedge clock or posedge reset) begin
+    //   if (reset) ...
+    // There is a fundamental mismatch between this representation which treats async reset
+    // registers as edge-triggered when in reality they are level-triggered.
+    // This can result in silicon-simulation mismatch in the case where async reset is held high
+    // upon power on with no clock, then async reset is dropped before the clock starts. In this
+    // circumstance, the async reset register will be randomized in simulation instead of being
+    // reset. To fix this, we need extra initial block logic for async reset registers
+    val asyncInitials = ArrayBuffer[Seq[Any]]()
     val simulates = ArrayBuffer[Seq[Any]]()
 
     def declare(b: String, n: String, t: Type, info: Info) = t match {
@@ -509,10 +519,17 @@ class VerilogEmitter extends SeqTransform with Emitter {
       Seq(nx, "[", bitWidth(t) - 1, ":0]")
     }
 
-    def initialize(e: Expression) = {
+    def initialize(e: Expression, reset: Expression, init: Expression) = {
       initials += Seq("`ifdef RANDOMIZE_REG_INIT")
       initials += Seq(e, " = ", rand_string(e.tpe), ";")
       initials += Seq("`endif // RANDOMIZE_REG_INIT")
+      reset.tpe match {
+        case AsyncResetType =>
+          asyncInitials += Seq("if (", reset, ") begin")
+          asyncInitials += Seq(tab, e, " = ", init, ";")
+          asyncInitials += Seq("end")
+        case _ => // do nothing
+      }
     }
 
     def initialize_mem(s: DefMemory) {
@@ -628,7 +645,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
           declare("reg", sx.name, sx.tpe, sx.info)
           val e = wref(sx.name, sx.tpe)
           regUpdate(e, sx.clock, sx.reset, sx.init)
-          initialize(e)
+          initialize(e, sx.reset, sx.init)
         case sx: DefNode =>
           declare("wire", sx.name, sx.value.tpe, sx.info)
           assign(WRef(sx.name, sx.value.tpe, NodeKind, MALE), sx.value, sx.info)
@@ -770,9 +787,11 @@ class VerilogEmitter extends SeqTransform with Emitter {
         emit(Seq("`ifndef RANDOM"))
         emit(Seq("`define RANDOM $random"))
         emit(Seq("`endif"))
-        emit(Seq("`ifdef RANDOMIZE"))
+        emit(Seq("`ifdef RANDOMIZE_MEM_INIT"))
         emit(Seq("  integer initvar;"))
-        emit(Seq("  initial begin"))
+        emit(Seq("`endif"))
+        emit(Seq("initial begin"))
+        emit(Seq("  `ifdef RANDOMIZE"))
         emit(Seq("    `ifdef INIT_RANDOM"))
         emit(Seq("      `INIT_RANDOM"))
         emit(Seq("    `endif"))
@@ -788,8 +807,9 @@ class VerilogEmitter extends SeqTransform with Emitter {
         emit(Seq("      `endif"))
         emit(Seq("    `endif"))
         for (x <- initials) emit(Seq(tab, x))
-        emit(Seq("  end"))
-        emit(Seq("`endif // RANDOMIZE"))
+        emit(Seq("  `endif // RANDOMIZE"))
+        for (x <- asyncInitials) emit(Seq(tab, x))
+        emit(Seq("end"))
       }
 
       for ((clk, content) <- noResetAlwaysBlocks if content.nonEmpty) {
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -156,7 +156,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
     case (_: UIntType | _: SIntType | _: AnalogType) =>
       val wx = bitWidth(tpe) - 1
       if (wx > 0) s"[$wx:0]" else ""
-    case ClockType => ""
+    case ClockType | AsyncResetType => ""
     case _ => throwInternalError(s"trying to write unsupported type in the Verilog Emitter: $tpe")
   }
   def emit(x: Any)(implicit w: Writer) { emit(x, 0) }
@@ -297,6 +297,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
        case AsUInt => Seq("$unsigned(", a0, ")")
        case AsSInt => Seq("$signed(", a0, ")")
        case AsClock => Seq(a0)
+       case AsAsyncReset => Seq(a0)
        case Dshlw => Seq(cast(a0), " << ", a1)
        case Dshl => Seq(cast(a0), " << ", a1)
        case Dshr => doprim.tpe match {
@@ -412,7 +413,12 @@ class VerilogEmitter extends SeqTransform with Emitter {
     val assigns = ArrayBuffer[Seq[Any]]()
     val attachSynAssigns = ArrayBuffer.empty[Seq[Any]]
     val attachAliases = ArrayBuffer.empty[Seq[Any]]
-    val at_clock = mutable.LinkedHashMap[Expression, ArrayBuffer[Seq[Any]]]()
+    // No (aka synchronous) always blocks, keyed by clock
+    val noResetAlwaysBlocks = mutable.LinkedHashMap[Expression, ArrayBuffer[Seq[Any]]]()
+    // One always block per async reset register, (Clock, Reset, Content)
+    // An alternative approach is to have one always block per combination of clock and async reset,
+    // but Formality doesn't allow more than 1 statement inside async reset always blocks
+    val asyncResetAlwaysBlocks = mutable.ArrayBuffer[(Expression, Expression, Seq[Any])]()
     val initials = ArrayBuffer[Seq[Any]]()
     val simulates = ArrayBuffer[Seq[Any]]()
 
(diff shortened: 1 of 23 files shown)


src/main/scala/firrtl/Emitter.scala
```diff
@@ -413,8 +413,10 @@ class VerilogEmitter extends SeqTransform with Emitter {
     val assigns = ArrayBuffer[Seq[Any]]()
     val attachSynAssigns = ArrayBuffer.empty[Seq[Any]]
     val attachAliases = ArrayBuffer.empty[Seq[Any]]
-    // Always blocks, keyed by (clock, asyncReset)
-    val alwaysBlocks = mutable.LinkedHashMap[(Expression, Option[Expression]), ArrayBuffer[Seq[Any]]]()
+    // No (aka synchronous) always blocks, keyed by clock
+    val noResetAlwaysBlocks = mutable.LinkedHashMap[Expression, ArrayBuffer[Seq[Any]]]()
+    // One always block per async reset register, (Clock, Reset, Content)
+    val asyncResetAlywaysBlocks = mutable.ArrayBuffer[(Expression, Expression, Seq[Any])]()
```
seldridge: Spelling, `alyways` -> `always`. Also, this probably needs some comment as to why you're doing it this way. I'd also accept a rant in place of a comment.

jackkoenig: Done

src/main/scala/firrtl/Emitter.scala
```diff
@@ -412,7 +413,12 @@ class VerilogEmitter extends SeqTransform with Emitter {
     val assigns = ArrayBuffer[Seq[Any]]()
     val attachSynAssigns = ArrayBuffer.empty[Seq[Any]]
     val attachAliases = ArrayBuffer.empty[Seq[Any]]
-    val at_clock = mutable.LinkedHashMap[Expression, ArrayBuffer[Seq[Any]]]()
+    // No (aka synchronous) always blocks, keyed by clock
+    val noResetAlwaysBlocks = mutable.LinkedHashMap[Expression, ArrayBuffer[Seq[Any]]]()
+    // One always block per async reset register, (Clock, Reset, Content)
+    // An alternative approach is to have one always block per combination of clock and async reset,
+    // but Formality doesn't allow more than 1 statement inside async reset always blocks
+    val asyncResetAlwaysBlocks = mutable.ArrayBuffer[(Expression, Expression, Seq[Any])]()
```
seldridge: a: "Chisel emits unreadable, verbose Verilog."
b: "Synopsys *demands* unreadable, verbose Verilog."

???

:hankey: :hankey: :hankey:

src/main/scala/firrtl/Emitter.scala
```diff
@@ -779,9 +792,15 @@ class VerilogEmitter extends SeqTransform with Emitter {
...
-        emit(Seq(tab, "always @(posedge ", clk_stream._1, ") begin"))
-        for (x <- clk_stream._2) emit(Seq(tab, tab, x))
+      for ((clk, content) <- noResetAlwaysBlocks if content.nonEmpty) {
+        emit(Seq(tab, "always @(posedge ", clk, ") begin"))
+        for (line <- content) emit(Seq(tab, tab, line))
+        emit(Seq(tab, "end"))
+      }
+
+      for ((clk, reset, content) <- asyncResetAlwaysBlocks if content.nonEmpty) {
+        emit(Seq(tab, "always @(posedge ", clk, " or posedge ", reset, ") begin"))
```
seldridge: I think this then leaves the door open to eventually negedge triggered things (#695).

src/main/scala/firrtl/passes/InferWidths.scala
```diff
@@ -282,10 +283,13 @@ object InferWidths extends Pass {
...
-        case (s: DefRegister) => v ++= (
-           get_constraints_t(s.reset.tpe, UIntType(IntWidth(1))) ++
-           get_constraints_t(UIntType(IntWidth(1)), s.reset.tpe) ++ 
-           get_constraints_t(s.tpe, s.init.tpe))
+        case (s: DefRegister) => 
+          if (s.reset.tpe != AsyncResetType ) {
+            v ++= (
+               get_constraints_t(s.reset.tpe, UIntType(IntWidth(1))) ++
+               get_constraints_t(UIntType(IntWidth(1)), s.reset.tpe))
+           }
```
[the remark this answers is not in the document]
jackkoenig: Note that this code is guarded by `if (s.reset.tpe != AsyncResetType ) {`, thus this constraint isn't applied for `AsyncReset`, it's for ensuring synchronous resets with uninferred width get the right width.

src/test/scala/firrtlTests/AsyncResetSpec.scala
```diff
@@ -0,0 +1,169 @@
...
+class AsyncResetSpec extends FirrtlFlatSpec {
+  def compile(input: String): CircuitState =
+    (new VerilogCompiler).compileAndEmit(CircuitState(parse(input), ChirrtlForm), List.empty)
+  def compileBody(body: String) = {
+    val str = """
+      |circuit Test :
+      |  module Test :
+      |""".stripMargin + body.split("\n").mkString("    ", "\n    ", "")
+    compile(str)
+  }
```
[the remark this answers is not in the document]
jackkoenig: I have found `compileBody` to be useful in multiple places, I agree it should be standardized.

src/main/scala/firrtl/Emitter.scala
jackkoenig: Profit?

colinschmidt: We can now sell the hard to read verilog as tool-friendly and kill two birds with one stone.

src/main/scala/firrtl/Emitter.scala
jackkoenig: Yeah, I think this would support that with minor modifications, I think the bigger issues involve how (if at all) negedge involves additionally checks and how (if at all) FIRRTL should reason about flops on the same clock but different edges should be handled.

src/main/scala/firrtl/transforms/ConstantPropagation.scala
```diff
@@ -405,6 +407,9 @@ class ConstantPropagation extends Transform with ResolvedAnnotationPaths {
       // Record things that should be propagated
       stmtx match {
         case x: DefNode if !dontTouches.contains(x.name) => propagateRef(x.name, x.value)
+        case reg: DefRegister if reg.reset.tpe == AsyncResetType =>
+          asyncResetRegs += reg.name
+          reg
```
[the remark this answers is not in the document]
jackkoenig: Good catch, removing it

src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -291,41 +300,43 @@ object CheckTypes extends Pass {
...
-            case ClockType    => (isUInt, isSInt, true, isFix)
-            case f: FixedType => (isUInt, isSInt, isClock, true)
+      def checkAllTypes(exprs: Seq[Expression], okUInt: Boolean, okSInt: Boolean, okClock: Boolean, okFix: Boolean, okAsync: Boolean): Unit = {
+        exprs.foldLeft((false, false, false, false, false)) {
+          case ((isUInt, isSInt, isClock, isFix, isAsync), expr) => expr.tpe match {
+            case u: UIntType       => (true, isSInt, isClock, isFix, isAsync)
+            case s: SIntType       => (isUInt, true, isClock, isFix, isAsync)
+            case ClockType         => (isUInt, isSInt, true, isFix, isAsync)
+            case f: FixedType      => (isUInt, isSInt, isClock, true, isAsync)
+            case AsyncResetType => (isUInt, isSInt, isClock, isFix, true)
```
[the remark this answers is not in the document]
jackkoenig: Hahaha yeah fair, I will push it with alignment

src/main/scala/firrtl/Emitter.scala
aswaterman: Put this man on the product marketing team

jackkoenig:
@seldridge I've addressed your feedback, rebased, and pushed. This also dismissed the review 😢

src/main/scala/firrtl/Emitter.scala
seldridge: But seriously this may actually be a good idea. There's a lot of effort in the Verilog world of top-down brittle coding styles because they produce better results [[1], [2], [3], [4]]. If these are encapsulated in either the FIRRTL optimizations or in the Verilog emitter then you don't wind up with brittle "code it this way because our Cadence rep told us to" anti-patterns. That also may help avoid vendor lock in.

At some point, it would likely be sensible to encapsulate these types of "optimizations" in something like an `-mtool` flag, e.g., `-mtool=synopsys` vs. `-mtool=cadence`. 

[1]: http://www.sunburst-design.com/papers/CummingsSNUG2000SJ_NBA.pdf
[2]: http://www.sunburst-design.com/papers/CummingsSNUG1999SJ_SynthMismatch.pdf
[3]: http://www.sunburst-design.com/papers/CummingsSNUG1998SJ_FSM.pdf
[4]: http://www.sunburst-design.com/papers/CummingsICU1997_VerilogCodingEfficiency.pdf

edwardcwang: Even better is to allow downstream users to customize the emitter as needed. e.g. I may want to ship a emitter-modifying pass with Hammer that makes adjustments depending on what CAD tools are selected.

jackkoenig: We obviously want to keep the Verilog as compatible as possible but it's abundantly clear that different styles are sometimes necessary for different tools. Customization of emission along with generated collateral to be confident that the emitted code is the same (ie. LEC or model checking scripts) seems like a reasonable approach to me.

src/main/scala/firrtl/transforms/ConstantPropagation.scala
```diff
@@ -414,7 +418,7 @@ class ConstantPropagation extends Transform with ResolvedAnnotationPaths {
           constOutputs(pname) = paddedLit
         // Const prop registers that are driven by a mux tree containing only instances of one constant or self-assigns
         // This requires that reset has been made explicit
-        case Connect(_, lref @ WRef(lname, ltpe, RegKind, _), rhs) if !dontTouches.contains(lname) =>
+        case Connect(_, lref @ WRef(lname, ltpe, RegKind, _), rhs) if !dontTouches(lname) && !asyncResetRegs(lname) =>
```
azidar: Is there a reason you aren't checking dontTouches.contains anymore?

jackkoenig: For sets, `set.contains(blah)` is the exact same as `set(blah)`, I just changed it to make it fit on one line better 😬

azidar: Wow. TIL.

jackkoenig:
@azidar @seldridge please see the most recent commit, I'll paste the commit message below:

```
Add initialization logic for async reset registers

In Verilog, async reset registers are expressed using always blocks of the form:
always @(posedge clock or posedge reset) begin
  if (reset) ...
There is a fundamental mismatch between this representation which treats async reset
registers as edge-triggered when in reality they are level-triggered.
This can result in silicon-simulation mismatch in the case where async reset is held high
upon power on with no clock, then async reset is dropped before the clock starts. In this
circumstance, the async reset register will be randomized in simulation instead of being
reset. To fix this, we need extra initial block logic for async reset registers.

At the end of the initial block, *not* guarded by ifdef RANDOMIZE, we
emit an if block for each async reset register to reset it if the
appropriate reset is asserted initially
```

Obnoxiously, I have been unable to create a test for this that fails without the extra initial block logic. It is a real problem for rocket-chip designs though, this most recent commit is essentially implementing something from the AsyncResetReg BlackBox in rocket-chip: https://github.com/freechipsproject/rocket-chip/blob/53b88546e32673b00b8dfa3e291effb0f75bdeab/src/main/resources/vsrc/AsyncResetReg.v#L64.

Now one might wonder, why can't we just mandate that async reset must be toggled in simulation? The issue is that does not match real use cases in silicon where async reset is asserted from power on and deasserted before the clock starts. As mentioned above, this is really just a mismatch between Verilog and the actual semantics of async reset registers.

jackkoenig:
Another slight tweak was to guard `initvar` by `RANDOMIZE_MEM_INIT` because that's the only place where it's used and I needed to make the initial block not be guarded by `ifdef RANDOMIZE` for these async reset initial blocks (they need to happen whether or not randomization is being used).

Issue #446, opened 2017-02-17
Enforce ordering on custom pass annotations


colinschmidt:
When directly calling a firrtl compiler with custom transforms the user has the ability to enforce an order between their passes. If annotations are used to add custom transforms there is no extant way to order them. This would be a very useful feature.

shunshou:
One thing I might add, and not sure if anyone's really tried -- for ResetInverter, i assume the invert(this) thing could also be run from some higher module that instantiates a Module with ResetInverter (or whatever it was called) trait. If that's the case, people might be able to order them to some degree... but when dealing with transforms done on the top level module, I'm not sure how one goes about doing that with Chisel Driver? I'm still kind of confused about how the (): => stuff works...

shunshou:
As another example: I *need* my AddPads Transform to come before BlackBoxHelper, but I don't know if there's any way to guarantee that, given that both transforms should occur on LowForm.

azidar:
Current thoughts: list of passes required before/after running - topo sort, then insert lowering passes to fix form mismatches

seldridge:
Current approach in #1123 is that you get two ways of defining dependencies:

1. The custom `Transform`s can define prerequisites, dependents, and invalidates
2. The order of the `RunFirrtlTransformAnnotation`s is used in seeding the topological sort. This has the effect of transforms, with the same dependencies, running in the same order they're specified. This should match the current behavior. 

Implementation-wise for (2) explicit dependencies are not added based on the order.

PR #1015, opened 2019-02-13
Dependency API

### LINKED ISSUE #948: [RFC] Dependency API
**By: Schuyler Eldridge**

## History

| Date          | Version       |
|---------------|---------------|
| Nov 25th 2018 | Initial Draft |
| Nov 27th 2018 | Add LLVM PassManager Info |
| Dec 5th 2018 | Clarify examples, add Steve Burns prerequisite injection example |

## Motivation

The FIRRTL compiler middle end is the serial application of a sequence of `firrtl.Transform`. Each `Transform` is abstract in an `execute` method with signature:

```scala
abstract class Transform {
  def inputForm: CircuitForm
  def outputForm: CircuitForm

  /* Each Transform is abstract in an execute method */
  def execute(state: CircuitState): CircuitState
}

/* The CircuitState describes a snapshot of a Circuit and some metadata */
case class CircuitState(circuit: Circuit, form: CircuitForm, annotations: AnnotationSeq, renames: Option[RenameMap])
```

The Annotations/Options refactor (Issue: #764, Selected PRs: #919, #945, #920) generalizes the concept of a `Transform` as a `Phase`. A `Phase` is abstract in a transformation of an `AnnotationSeq` which allows this to generalize to components of the Chisel/FIRRTL Hardware Compiler Framework _not_ inside the FIRRTL middle end. Additionally, #945 introduces `Stage`, a subclass of `Phase`, that exposes a user interface. _Going forward, this CIP (Chisel Improvement Process) refers to `Phase` as if it were the superclass of `Transform` and `Pass` even though that is not currently the case._ A high-level overview of `Phase` and `Stage` are outlined in the code examples below:

```scala
abstract class Phase {
  /** Mathematical transformation of an [[AnnotationSeq]] */
  def transform(annotations: AnnotationSeq): AnnotationSeq
}

abstract class stage {
  /** A command line interface utility */
  val shell: Shell

  /** Abstract transformation that a subclass implements */
  def run(annotations: AnnotationSeq): AnnotationSeq

  /** Does common preprocessing and then calls the run method */
  final def transform(annotations: AnnotationSeq): AnnotationSeq = {
    val x = // preprocessing of input `annotations`
    run(x)
  }

  /** Method that allows for both command line arguments and annotations as inputs */
  final def execute(args: Array[String], annotations: AnnotationSeq): AnnotationSeq =

  /* Command line interface API */
  final def main(args: Array[String]): Unit = execute(args, Seq.empty)
}
```

The discussion around `Stage` and `Phase` revived a long standing issue first brought up in #446:

> __How does the `Phase` scheduler derive the order in which to apply each `Phase`?__

Such an ordering is important for both:

1. An internal `Phase` that runs by default, e.g., `DeadCodeElimination`
2. An external, user-injected `Phase`, e.g., MIDAS' FAME1 transform

This CIP aims to develop two new APIs:

1. A __Dependency API__ for `Phase`
2. A __`Phase` Scheduler__ that derives a `Phase` order using (1)

At a high level, the __Dependency API__ needs to communicate two pieces of information for every `Phase`:

1. What are my prerequisites?
2. What do I update?

This information is mirrored by any build system or compiler. Consider [Make](https://en.wikipedia.org/wiki/Make_(software)). Make provides an API consisting of _rules_ and _recipes_. Each _rule_ defines a _target_ (something that this rule updates) and zero or more _prerequisites_. The _recipe_ is then called whenever any _prerequisite_ is newer than the _target_. Also consider [LLVM](http://llvm.org/docs/WritingAnLLVMPass.html#specifying-interactions-between-passes). Each pass in LLVM may optionally define a list of _prerequisite_ analyses that it needs and a list of analyses that it _invalidates_. Conceptually, the dependency API for both Make and LLVM are the same.

This CIP proposes that each `Phase` declares it's dependencies in the following way:

```scala
/** Defines the set of `Phase` that have been run */
case class PhaseStamps(phase: Set[Phase]) extends NoTargetAnnotation

abstract class Phase {
  /** A list of `Phase`s that must be run before this `Phase` */
  def prerequisites: Seq[Phase] = Seq.empty

  /** A function that is true if a particular `Phase` is invalidated by this `Phase` */
  def invalidate(phase: Phase): Boolean = true
}
```

First, this defines an `Annotation` that holds information about what `Phase`s have been run. Second, this defines the prerequisites that a specific `Phase` needs. Third, this defines a method `invalidate` that a manager can use to invalidate members of the set contained by the `PhaseStamps` annotation. By default, a `Phase` has no prerequisites and it invalidates everything.

Note: it may make sense to have the `PhaseStamps` be attached to something or a collection of something. E.g., one could consider a generalization of `FirrtlCircuitAnnotation` that, in addition to holding a `Circuit`, also holds the "state" of that circuit listing all `Phase`s that have been applied to it:
```scala
trait Stamped[T] { 
  val stamps: Set[Phase] 
  val data: Seq[T]
}

case class FirrtlCircuitAnnotation(data: Seq[Circuit], stamps: Set[Phase]) extends Stamped[Circuit]
```

This API defining `prerequisites` and `invalidates` is a generalization of the current `inputForm: CircuitForm` and `outputForm: CircuitForm`. The generalization being that it enables you to specify "partial" input and output forms and doesn't require the definition of a new `CircuitForm` and `LoweringCompiler`. This is critical if the `Phase` API is used outside of the FIRRTL middle end.

It then falls out naturally what a `PhaseManager` should do. A `PhaseManager` should, given an input `AnnotationSeq` and a `Set[Phase]` to run, determine a phase ordering (likely not optimal due to computational difficulty of doing so):

```scala
object PhaseManager {
  /** Run all necessary `Phase`s on an `AnnotationSeq` that executes all targets `Set[Phase]`
    * @throws PhaseManagerException if a cyclic Phase ordering was specified
    */
  def run(annotations: AnnotationSeq, targets: Set[Phase]): AnnotationSeq = {
    // TBD implementation
  }
}
```

# Motivating Examples

### Examples

See above for API.

### Comparison Examples

- How would `inputForm` and `outputForm` be migrated to this Dependency API?
- What is the equivalent operation for `firrtl.compilerUtils.getLoweringTransforms` and `firrtl.compilerUtils.mergeTransforms`?
- How would an injected dependency work, e.g., a user wants a specific `Phase` to run before another `Phase`? In effect, an injection of a new prerequisite for an existing `Phase` the user has not modified? (Is there a better way to do this?)

### Counter-Examples

I'm not sure what this _couldn't_ be used for. I'll need to think.

## Design

TBD.

### LLVM `PassManager`/`LegacyPassManager`

LLVM uses a `PassManager` to (by my initial skim) dynamically control the execution of passes and to error if something that looks like a cycle shows up. See [`PassManager.h`](https://github.com/llvm-mirror/llvm/blob/master/include/llvm/IR/PassManager.h#L443). 

The actual optimization appears to happen in `LegacyPassManager.cpp` in [`PMTopLevelManager::schedulePass`](https://github.com/llvm-mirror/llvm/blob/master/lib/IR/LegacyPassManager.cpp#L699).

## Counter-Examples

TBD

## Drawbacks

This is a big refactor...

TBD.

## Alternatives

TBD

## References

[1] [`PassManager.h`](https://github.com/llvm-mirror/llvm/blob/master/include/llvm/IR/PassManager.h#L443)

seldridge:
This implements a Dependency API for `TransformLike[A]`. This adds a `DependencyAPI` trait, an abstract class `DependencyManager`, and a concrete `PhaseManager`. The `DependencyAPI` defines three types of "dependencies" that, by implementing, allow a user to defer determination of a valid ordering to a `DependencyManager`. 

Dependencies are defined in terms of:
  - *prerequisites* are transforms that should run before me
  - *dependents* are transforms that should run after me (a means of prerequisite injection)
  - *invalidates* are transforms which I invalidate (e.g., a transform that runs at low form, but emits high form constructs and requires re-lowering)

The `DependencyManager` can then be used to, given a set of transforms you want to run and a set of all transforms that have been run, determine a linear ordering of transforms that satisfy your requirements. This uses uses the same algorithm proposed by @azidar in https://github.com/freechipsproject/firrtl/issues/446#issuecomment-300573107 with the exception of seeding the topological sort with information from a topological sort of the invalidates. This results in a more optimal solution that seems to minimize the number of repeated `Phase`s. Re-lowerings are handled by creating a new `DependencyManager` to solve the lowering sub-problem. 

This PR then adds dependencies to `firrtl.stage.phases` and modifies `FirrtlStage` to be implemented using a `PhaseManager` under the hood.

This includes a modification to `DiGraph` that allows for seeding the topological sort to get the necessary better linearization solution.

In preparation for using `DependencyManager` for FIRRTL transform scheduling, this makes `Transform` extend `TransformLike[CircuitState]`. 

META:
- Depends on #1005 
- Fixes #948 
- Related to #446 

TODO:
- [x] Switch to `Class[A]` instead of `A` for dependencies

## Examples

The examples below show examples of how different combinations of dependencies interact and are resolved. This uses built-in `DependencyManager` methods for generating graphviz output.

### Simple Dependency

dependencygraph dot
phaseorder dot

### One Invalidate

dependencygraph dot
phaseorder dot

### Two Invalidates

There are many valid solutions here. However, many of the solutions result in repeated re-lowerings. The algorithm proposed seems to minimize this and choose a good solution.

dependencygraph dot
phaseorder dot

### Repeated Analysis Phase

dependencygraph dot
phaseorder dot

### Repeated Analysis with Invalidates

dependencygraph dot
phaseorder dot

### Invalidation Creates a Cycle

dependencygraph dot
phaseorder dot

### Custom Transform Needing to Run *Before* Something Else

This is intended to model the @stevenmburns case of needing to run a custom transform *before* something else. Similarly, this is the same thing as what #446 is talking about. The major difference from the previous examples is that this defines the `dependents` for the custom transform.

```scala
class First extends IdentityPhase {
  override def invalidates(phase: Phase): Boolean = false
}

class Second extends IdentityPhase {
  override val prerequisites: Set[Class[Phase]] = Set(classOf[First])
  override def invalidates(phase: Phase): Boolean = false
}

class Custom extends IdentityPhase {
  override val prerequisites: Set[Class[Phase]] = Set(classOf[First])
  override val dependents: Set[Class[Phase]] = Set(classOf[Second])
  override def invalidates(phase: Phase): Boolean = false
}

implicit def f(a: Class[_ <: Phase]): Class[Phase] = a.asInstanceOf[Class[Phase]]
new PhaseManager(Set(classOf[Custom], classOf[Second]).map(_.asInstanceOf[Class[Phase]]))
  .flattenedPhaseOrder.map(_.asClass) should be (Set(classOf[First], classOf[Custom], classOf[Second]))
```

dependencygraph dot
phaseorder dot

### FIRRTL Stage

This PR additionally swings `FirrtlStage` to use the Dependency API. This allows you to be very terse with the expression of what `FirrtlStage` needs to do:

```scala
  private val phases: Seq[Phase] = PhaseManager(Set(firrtl.stage.phases.WriteEmitted))
    .transformOrder
    .map(DeletedWrapper(_))
```

The dependency graph then looks like (and note that this is a much simpler use case than some of the examples above):

firrtlstage-dependencies dot

And the determined phase ordering is as follows:

firrtlstage-order dot

### DIFF
--- a/src/test/scala/firrtlTests/options/PhaseManagerSpec.scala
+++ b/src/test/scala/firrtlTests/options/PhaseManagerSpec.scala
@@ -1,0 +1,338 @@
+// See LICENSE for license details.
+
+package firrtlTests.options
+
+import org.scalatest.{FlatSpec, Matchers}
+
+import firrtl.AnnotationSeq
+import firrtl.options.{DependencyManagerException, Phase, PhaseManager}
+import firrtl.annotations.{Annotation, NoTargetAnnotation}
+
+import java.io.{File, PrintWriter}
+
+import sys.process._
+
+trait IdentityPhase extends Phase {
+  def transform(annotations: AnnotationSeq): AnnotationSeq = annotations
+}
+
+/** Default [[Phase]] that has no prerequisites and invalidates nothing */
+class A extends IdentityPhase {
+
+  override def invalidates(phase: Phase): Boolean = false
+}
+
+/** [[Phase]] that requires [[A]] and invalidates nothing */
+class B extends IdentityPhase {
+  override def prerequisites: Set[Class[Phase]] = Set(classOf[A])
+  override def invalidates(phase: Phase): Boolean = false
+}
+
+/** [[Phase]] that requires [[B]] and invalidates nothing */
+class C extends IdentityPhase {
+  override def prerequisites: Set[Class[Phase]] = Set(classOf[A])
+  override def invalidates(phase: Phase): Boolean = false
+}
+
+/** [[Phase]] that requires [[A]] and invalidates [[A]] */
+class D extends IdentityPhase {
+  override def prerequisites: Set[Class[Phase]] = Set(classOf[A])
+  override def invalidates(phase: Phase): Boolean = phase match {
+    case _: A => true
+    case _ => false
+  }
+}
+
+/** [[Phase]] that requires [[B]] and invalidates nothing */
+class E extends IdentityPhase {
+  override def prerequisites: Set[Class[Phase]] = Set(classOf[B])
+  override def invalidates(phase: Phase): Boolean = false
+}
+
+/** [[Phase]] that requires [[B]] and [[C]] and invalidates [[E]] */
+class F extends IdentityPhase {
+  override def prerequisites: Set[Class[Phase]] = Set(classOf[B], classOf[C])
+  override def invalidates(phase: Phase): Boolean = phase match {
+    case _: E => true
+    case _ => false
+  }
+}
+
+
+/** [[Phase]] that requires [[C]] and invalidates [[F]] */
+class G extends IdentityPhase {
+  override def prerequisites: Set[Class[Phase]] = Set(classOf[C])
+  override def invalidates(phase: Phase): Boolean = phase match {
+    case _: F => true
+    case _ => false
+  }
+}
+
+class CyclicA extends IdentityPhase {
+  override def prerequisites: Set[Class[Phase]] = Set(classOf[CyclicB])
+}
+
+class CyclicB extends IdentityPhase {
+  override def prerequisites: Set[Class[Phase]] = Set(classOf[CyclicA])
+}
+
+object CyclicInvalidateFixture {
+
+  class A extends IdentityPhase {
+    override def invalidates(phase: Phase): Boolean = false
+  }
+  class B extends IdentityPhase {
+    override def prerequisites: Set[Class[Phase]] = Set(classOf[A])
+    override def invalidates(phase: Phase): Boolean = false
+  }
+  class C extends IdentityPhase {
+    override def prerequisites: Set[Class[Phase]] = Set(classOf[A])
+    override def invalidates(phase: Phase): Boolean = phase match {
+      case _: B => true
+      case _ => false
+    }
+  }
+  class D extends IdentityPhase {
+    override def prerequisites: Set[Class[Phase]] = Set(classOf[B])
+    override def invalidates(phase: Phase): Boolean = phase match {
+      case _: C | _: E => true
+      case _ => false
+    }
+  }
+  class E extends IdentityPhase {
+    override def prerequisites: Set[Class[Phase]] = Set(classOf[B])
+    override def invalidates(phase: Phase): Boolean = false
+  }
+
+}
+
+object RepeatedAnalysisFixture {
+
+  trait InvalidatesAnalysis extends IdentityPhase {
+    override def invalidates(phase: Phase): Boolean = phase match {
+      case _: Analysis => true
+      case _ => false
+    }
+  }
+
+  class Analysis extends IdentityPhase {
+    override def invalidates(phase: Phase): Boolean = false
+  }
+  class A extends InvalidatesAnalysis {
+    override def prerequisites: Set[Class[Phase]] = Set(classOf[Analysis])
+  }
+  class B extends InvalidatesAnalysis {
+    override def prerequisites: Set[Class[Phase]] = Set(classOf[A], classOf[Analysis])
+  }
+  class C extends InvalidatesAnalysis {
+    override def prerequisites: Set[Class[Phase]] = Set(classOf[B], classOf[Analysis])
+  }
+
+}
+
+object InvertedAnalysisFixture {
+
+  class Analysis extends IdentityPhase {
+    override def invalidates(phase: Phase): Boolean = false
+  }
+  class A extends IdentityPhase {
+    override def prerequisites: Set[Class[Phase]] = Set(classOf[Analysis])
+    override def invalidates(phase: Phase): Boolean = phase match {
+      case _: Analysis => true
+      case _ => false
+    }
+  }
+  class B extends IdentityPhase {
+    override def prerequisites: Set[Class[Phase]] = Set(classOf[Analysis])
+    override def invalidates(phase: Phase): Boolean = phase match {
+      case _: Analysis | _: A => true
+      case _ => false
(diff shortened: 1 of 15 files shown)


src/test/scala/firrtlTests/options/PhaseManagerSpec.scala
```diff
@@ -0,0 +1,223 @@
...
+case object CyclicB extends IdentityPhase {
+  override def prerequisites: Set[Phase] = Set(CyclicA)
+}
+
+class CyclicInvalidateFixture {
+
+  case object A extends IdentityPhase {
+    override def invalidates(phase: Phase): Boolean = false
+  }
+  case object B extends IdentityPhase {
```
azidar: Its tough to remember what all of these do. Can you give a clearer name, like A=A, B=RecA, C=RecAInvB, etc.?

seldridge:
@jackkoenig: Your idea of moving to `Class[A]` for dependency specification seems to work just fine. A `DependencyManager` will now construct a transform/phase on demand, and only construct each transform/phase once (multiple dependencies pointing to `classOf[Foo]` will resolve to the same `Foo` object).

jackkoenig:
While I'm thinking about it:

Using Sets makes sense for defining the set of things you depend on, but we should be careful that it doesn't introduce non-determinism when there are multiple legal orders. This can probably be made deterministic by just converting these sets to SortedSets before applying linearization (or before putting into DiGraph).

seldridge:
Apparently Travis does not update the `$TRAVIS_BRANCH` if you change the base branch. :fire: Will close and re-open...

PR #1027, opened 2019-02-21
Added mergify badge to README

azidar:


ucbjrl:
@ucbjrl is an approved reviewer according to GitHub

PR #1030, opened 2019-02-22
Change mergify to just require an approval

azidar:
Previously would require all requested reviewers must have approved a PR before merging.


.mergify.yml
```diff
@@ -10,9 +10,6 @@ pull_request_rules:
     conditions:
       - status-success=continuous-integration/travis-ci/pr
       - "#approved-reviews-by>=1"
-      - "#review-requested=0"
-      - "#changes-requested-reviews-by=0"
```
jackkoenig: Could this be a problem? What if one person approves and another requests changes? I guess Github itself will block the merge?

PR #1031, opened 2019-02-22
Detect and error on registers with flip in type

albert-magyar:
Thanks @edwardcwang for discovering that this doesn't error.

It's not in `CheckChirrtl` for now, but there are quite a few error checks that aren't duplicated in `CheckChirrtl`. I can add that too, but maybe syncing these up with an extra `CheckHighForm` after `ChirrtlToHighFirrtl` would be the right call?


src/test/scala/firrtlTests/CheckSpec.scala
```diff
@@ -10,7 +10,7 @@ import firrtl.ir.Circuit
 import firrtl.passes.{Pass,ToWorkingIR,CheckHighForm,ResolveKinds,InferTypes,CheckTypes,PassException,InferWidths,CheckWidths,ResolveGenders,CheckGenders}
 
 class CheckSpec extends FlatSpec with Matchers {
-  "Connecting bundles of different types" should "throw an exception" in {
+  "Memories with flip in the data type" should "throw an exception" in {
```
edwardcwang: Nice catch

albert-magyar:
image

PR #1034, opened 2019-02-25
Create a simple generic GraphViz renderer for DiGraph

chick:
There are three basic kinds
- A simple default renderer
- A ranked renderer that places nodes in columns based on depth from sources
- A sub-graph render for graphs that contain a loop
  - Renders just nodes that are part of first loop found
  - Plus the neighbors of the loop
  - Loop edges are shown in red.

### DIFF
--- a/src/main/scala/firrtl/graph/RenderDiGraph.scala
+++ b/src/main/scala/firrtl/graph/RenderDiGraph.scala
@@ -1,0 +1,233 @@
+// See LICENSE for license details.
+
+package firrtl.graph
+
+import scala.collection.mutable
+
+/**
+  * Implement a really simple graphviz dot renderer for a digraph
+  * There are three main renderers currently
+  * -
+  *
+  * @param diGraph   The DiGraph to be rendered
+  * @param graphName Name of the graph, when shown in viewer
+  * @param rankDir   Graph orientation, default is LR (left to right), most common alternative is TB (top to bottom)
+  * @tparam T        The type of the Node.
+  */
+class RenderDiGraph[T <: Any](diGraph: DiGraph[T], graphName: String = "", rankDir: String = "LR") {
+
+
+  /**
+    * override this to change the default way a node is displayed. Default is toString surrounded by double quotes
+    * {{{
+    * val rend = new RenderDiGraph(graph, "alice") {
+    *   override def renderNode(node: Symbol): String = s"\"${symbol.name}\""
+    *   }
+    * }}}
+    */
+  def renderNode(node: T): String = {
+    s""""${node.toString}""""
+  }
+
+  def findOneLoop: Set[T] = {
+    var path = Seq.empty[T]
+
+    try {
+      diGraph.linearize
+    }
+    catch {
+      case cyclicException: CyclicException =>
+        val node = cyclicException.node.asInstanceOf[T]
+        val pathFound = diGraph.getEdges(node).exists { vertex =>
+          try {
+            path = diGraph.path(vertex, node, blacklist = Set.empty)
+            true
+          }
+          catch {
+            case _: PathNotFoundException =>
+            case t: Throwable =>
+              throw t
+
+          }
+          false
+        }
+
+    }
+    path.toSet
+  }
+
+  /**
+    * Searches a DiGraph for a cycle. The first one found will be rendered as a graph that contains
+    * only the nodes in the cycle plus the neighbors of those nodes.
+    * @return a string that can be used as input to the dot command, string is empty if no loop
+    */
+  def showOnlyTheLoopAsDot: String = {
+    // finds the loop
+
+    val loop = findOneLoop
+
+    if(loop.nonEmpty) {
+
+      // Find all the children of the nodes in the loop
+      val childrenFound = diGraph.getEdgeMap.flatMap {
+        case (node, children) if loop.contains(node) => children
+        case _ => Seq.empty
+      }.toSet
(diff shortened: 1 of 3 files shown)


src/main/scala/firrtl/graph/RenderDiGraph.scala
```diff
@@ -0,0 +1,233 @@
...
+    * val rend = new RenderDiGraph(graph, "alice") {
+    *   override def renderNode(node: Symbol): String = s"\"${symbol.name}\""
+    *   }
+    * }}}
+    */
+  def renderNode(node: T): String = {
+    s""""${node.toString}""""
+  }
+
+  def findOneLoop: Set[T] = {
```
edwardcwang: Missing scaladoc?

src/main/scala/firrtl/graph/RenderDiGraph.scala
```diff
@@ -0,0 +1,233 @@
...
+      newRenderer.toDotWithLoops(loop, getRankedNodes)
+    }
+    else {
+      ""
+    }
+  }
+
+  /**
+    * returns a string that is a graphviz digraph
+    * @return
```
edwardcwang: Maybe put the return in the return field?

src/main/scala/firrtl/graph/RenderDiGraph.scala
```diff
@@ -0,0 +1,233 @@
...
+        s.append(s"""  ${renderNode(parent)} -> ${renderNode(child)};""" + "\n")
+      }
+    }
+    s.append("}\n")
+    s.toString
+  }
+
+  /**
+    * returns a string that is a graphviz digraph, but with loops highlighted
+    * @return
```
edwardcwang: Maybe put the return in the return field?

src/main/scala/firrtl/graph/RenderDiGraph.scala
```diff
@@ -0,0 +1,233 @@
...
+
+    paredRankedNodes.foreach { nodesAtRank =>
+      s.append(s"""  { rank=same; ${nodesAtRank.map(renderNode).mkString(" ")} };""" + "\n")
+    }
+
+    s.append("}\n")
+    s.toString
+  }
+
+  def getRankedNodes: mutable.ArrayBuffer[Seq[T]] = {
```
edwardcwang: Missing scaladoc

src/main/scala/firrtl/graph/RenderDiGraph.scala
```diff
@@ -0,0 +1,233 @@
...
+
+    walkByRank(diGraph.findSources.toSeq)
+    rankNodes
+  }
+  /**
+    * returns a string that is a graphviz digraph
+    * this method may look better than the toDot format. It tries to align nodes in columns based
+    * on their minimum distance to a source.
+    * Can also be faster and better behaved on large graphs
+    * @return
```
edwardcwang: ditto

src/main/scala/firrtl/graph/RenderDiGraph.scala
chick: Done, moved actual loop finding into DiGraph

src/main/scala/firrtl/graph/RenderDiGraph.scala
chick: Done

src/main/scala/firrtl/graph/RenderDiGraph.scala
chick: Done

src/main/scala/firrtl/graph/RenderDiGraph.scala
chick: Done

src/main/scala/firrtl/graph/RenderDiGraph.scala
chick: Done

albert-magyar:
@chick do you have an example of the "self path" behavior vs. what you want to see?

chick:
This is what I think is the compelling example, add this to DiGraphTests
```scala
  "DiGraph#path(a,a)" should "(a,b,d,a) or (a,c,d,a)" in {
    cyclicGraph.path("a", "a").length should be (4)
  }
```

albert-magyar:
Finding SCCs is definitely a good way to find loops, but it comes with the caveats that a) it gives SCCs, so it won't actually return a path from one specified node in the SCC back to itself and b) it runs on the whole graph, finding all SCCs at once.

In principle, the API of `path` is pretty loose: it gives you one path. In this case, the implementation is BFS, which is efficient and gives short paths. I'm not really inclined to add an extra restriction (paths must be nontrivial), as it has an undesirable side effect of making the case of `path(v,v)` in an _acyclic_ graph ill-defined and/or inconsistent.

chick:
This is not a big deal to implement by starting at the node in question's children, but why does a node have a path to itself if there is not an edge connecting it to itself

albert-magyar:
My understanding is that paths with zero edges are generally accepted as paths per most definitions. As far as I can tell, trails with one repeated edge (aka simple cycles) are less universally included in the definition of paths.

chick:
@edwardcwang I don't think this PR is going to wait or implement any more functionality to DiGraph, in particular, with the way path is interpreted and computed. I think I have fixed your other comments, when you get the chance can you check, confirm, and approve.

PR #1036, opened 2019-02-27
Add description to fixed point numbers

grebe:
Description says that a Named is a fixed point number and says what the binary point is. These will end up as comments in the emitted verilog.

With this example:
```
circuit Example :
  module Example :
    input clk: Clock
    input a: Fixed<4><<2>>
    input b: Fixed<2><<1>>
    input c: Fixed<3><<0>>
    output out: Fixed

    reg sum: Fixed<5>, clk
    reg prod: Fixed<7>, clk

    sum <= add(a, b)
    prod <= mul(sum, c)
    out <= sub(prod, sum)
```

I get the following `DescriptionAnnotations` in the output annotations file (`-foaf`):
```
{
    "class":"firrtl.DescriptionAnnotation",
    "named":"~Example|Example>sum",
    "description":"Fixed point with binary point 2"
  },
  {
    "class":"firrtl.DescriptionAnnotation",
    "named":"~Example|Example>prod",
    "description":"Fixed point with binary point 2"
  },
  {
    "class":"firrtl.DescriptionAnnotation",
    "named":"~Example|Example>a",
    "description":"Fixed point with binary point 2"
  },
  {
    "class":"firrtl.DescriptionAnnotation",
    "named":"~Example|Example>b",
    "description":"Fixed point with binary point 1"
  },
  {
    "class":"firrtl.DescriptionAnnotation",
    "named":"~Example|Example>c",
    "description":"Fixed point with binary point 0"
  },
```

but I don't get any comments in the output verilog.


azidar:
Can you add a test of a FIRRTL design with fixedpoints in the unit tests, so we can see it failing to work?

grebe:
Still failing, is there some stage/phase stuff I should be aware of?

src/main/scala/firrtl/AddDescriptionNodes.scala
```diff
@@ -108,6 +108,7 @@ class AddDescriptionNodes extends Transform {
     val compMap = annos.collect {
       case DescriptionAnnotation(ComponentName(comp, ModuleName(mod, CircuitName(circ))), desc) =>
         (mod, comp, desc)
+      case DescriptionAnnotation(ReferenceTarget(_, m, _, r, _), desc) => (m, r, desc)
```
grebe: It turns out the pass was being scheduled, it just wasn't finding any annotations it could consume! For whatever reason, everything that comes out of the earlier pass is a `ReferenceTarget`. Should they be collected here, or is there something I should be doing differently upstream?

src/main/scala/firrtl/transforms/DescribeFixed.scala
```diff
@@ -0,0 +1,50 @@
...
+object DescribeFixed extends Transform {
+  def inputForm = MidForm
+  def outputForm = MidForm
+
+  private def onModule(cir: CircuitName)(m: DefModule): Seq[DescriptionAnnotation] = {
+    val mod = ModuleName(m.name, cir)
+    var descs = Seq[DescriptionAnnotation]()
+    def onType(n: Target)(tpe: Type): Type = {
+      tpe match {
+        case FixedType(IntWidth(w), IntWidth(p)) =>
```
chick: Shouldn't second IntWidth be BinaryPoint?

src/main/scala/firrtl/transforms/DescribeFixed.scala
```diff
@@ -0,0 +1,50 @@
...
+          descs = descs :+ DescriptionAnnotation(n, s"Fixed point with binary point $p")
+        case f: FixedType => throw new Exception(s"Uh oh!")
```
chick: Better Error message here might be "Found FixedPoint with unresolved width or binary point"

chick (changes requested):
Two minor nits. Otherwise looks good.
You asked a question about Refs in beginning, what was the final outcome to that question.

src/main/scala/firrtl/transforms/DescribeFixed.scala
grebe: I think binary points are implemented in firrtl with intwidths b/c the constraint solver looks at binary points as being widths

src/main/scala/firrtl/transforms/DescribeFixed.scala
grebe: err, yes. that's debug stuff i should have caught. thanks!

chick (approved):
lgtm

chick:
@grebe are there potential performance regressions here, have you tried this with a circuit with a whole bunch of FixedPoints

grebe:
I haven't tried it with a huge circuit, although I don't think it does anything most other passes don't already do a bunch of times. In the first pass, it iterates through the whole circuit once and may potentially add a bunch of annotations. In the emitter phase, it goes through all the annotations and pulls out the descriptions. It adds more to the output verilog file too, but I think it's a worthwhile thing to do.

chick:
Just curious. And I was more curious about any consequences downstream having more annotations to wade through. I think @seldridge  did some analysis of this overhead so I wouldn't worry about it too much

src/main/scala/firrtl/transforms/DescribeFixed.scala
```diff
@@ -0,0 +1,50 @@
...
+    }
+    def onStmt(s: Statement): Unit = s match {
+      case b: Block => b.foreachStmt(onStmt)
+      case ns: Statement with IsDeclaration =>
+        val comp = ComponentName(ns.name, mod)
+        ns.mapType(onType(comp))
+      case _ =>
+    }
+    def onPort(p: Port): Unit = {
+      val comp = ComponentName(p.name, mod)
```
azidar: Replace `ComponentName` with `ReferenceTarget`, as `ComponentName` is deprecated.

src/main/scala/firrtl/transforms/DescribeFixed.scala
```diff
@@ -0,0 +1,50 @@
...
+          descs = descs :+ DescriptionAnnotation(n, s"Fixed point with binary point $p")
+        case f: FixedType => throw new Exception(s"Found unexpected fixed point number with unresolved binary point")
+        case _ =>
+      }
+      tpe
+    }
+    def onStmt(s: Statement): Unit = s match {
+      case b: Block => b.foreachStmt(onStmt)
+      case ns: Statement with IsDeclaration =>
+        val comp = ComponentName(ns.name, mod)
```
azidar: Replace `ComponentName` with `ReferenceTarget`, as `ComponentName` is deprecated.

src/main/scala/firrtl/transforms/DescribeFixed.scala
```diff
@@ -0,0 +1,50 @@
...
+      val comp = ComponentName(p.name, mod)
+      onType(comp)(p.tpe)
+    }
+    m.foreachStmt(onStmt)
+    m.foreachPort(onPort)
+    descs
+  }
+
+  def execute(state: CircuitState): CircuitState = {
+    val descriptions = state.circuit.modules.flatMap(onModule(CircuitName(state.circuit.main)))
```
azidar: Same, use `CircuitTarget` instead of `CircuitName`

src/main/scala/firrtl/transforms/DescribeFixed.scala
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Adds description to fixed point nodes giving its width and binary point
+ */
+object DescribeFixed extends Transform {
+  def inputForm = MidForm
+  def outputForm = MidForm
+
+  private def onModule(cir: CircuitName)(m: DefModule): Seq[DescriptionAnnotation] = {
+    val mod = ModuleName(m.name, cir)
```
azidar: Use `ModuleTarget` instead of `ModuleName`

Issue #852, opened 2018-07-22
Last connect does not prevent VOID from appearing in AST


jackkoenig:
Originally reported in https://github.com/freechipsproject/firrtl/issues/850

**Type of issue**: bug report

**If the current behavior is a bug, please provide the steps to reproduce the problem:**

  * **If possible, please directly provide the FIRRTL file or a relevant fragment if the file is large**

Consider the following FIRRTL:
```
circuit test :
  module test :
    input en : UInt<1>
    output y : UInt<8>

    when en :
      y <= UInt(1)
    y <= UInt(0)  
```
  * **What is the current behavior?**

After ExpandWhens you get the following FIRRTL:
```
circuit test :                                                                   
  module test :                                                                                 
    input en : UInt<1>                                                                                  
    output y : UInt<8>                                                                                  
                                                                                                           
    node _GEN_0 = mux(en, UInt<1>("h1"), VOID)                                                                             
    y <= UInt<1>("h0")
```
This `VOID` leads to a match error in `InferTypes`.
  * **What is the expected behavior?**

Last connect semantics mean that `y` is fully initialized. CheckInitialization does not error because it assumes that all `VOIDs` on nodes lead to a `Connect` but there is no `Connect` because `y` has a legitimate connection!

**What is the use case for changing the behavior?**

Make the behavior of last connect interact with initialization requirements as one would expect

**Impact**: no functional change

**Development Phase**: request

**Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. Stack Overflow, gitter, etc)

jackkoenig:
@seldridge @azidar per discussion in #973, any thoughts on how to fix this? It's a pretty gnarly error for technically legal (if awkward) code.

One possible solution (although not particularly elegant), is to do a specialized DCE where nodes injected by ExpandWhens that are unused are removed. This has the side effect of removing some things that may be useful for debugging, but since they are unused it should reduce unnecessary work. For example:
```
when en : 
  x <= a
else : 
  x <= b
x <= c
```
Currently after expand whens this looks like:
```
node _GEN_0 = mux(en, a, b)
x <= c
```
Such a "fixup" would remove `_GEN_0`.

A perhaps more focused version of this proposal would be to only removed unused nodes that include a `VOID`. It's perhaps over specialized but would fix the issue!

Thoughts?

seldridge:
Maybe it would be beneficial to think about it from this perspective:

[...]

PR #1037, opened 2019-02-27
FIRRTL Bug - WVoid leaks from ExpandWhens

azidar:
Interesting compiler bug found by @vighneshiyer (and the systolic array team).

When a partially unassigned wire is then assigned a value (thus theoretically overwriting the earlier partial assignment because of last connect semantics), the memoized nodes created by ExpandWhens (representing the partial assignment expression) are still generated. These nodes contain the WVoid expression, but are never read from. While this could be DCE'd away, DCE only runs on low FIRRTL. A later pass then trips up when it sees the WVoid expression.

Example of code that triggers the bug:
```
circuit Test :
    module Test :
        input p : UInt<1>
        wire x : UInt<32>
        when p :
            x <= UInt(1)
        x <= UInt(1)
```

Anyways, this PR includes the test case demonstrating the problem. @jackkoenig (and others) any thoughts on solutions?

### DIFF
--- a/src/main/scala/firrtl/passes/CheckInitialization.scala
+++ b/src/main/scala/firrtl/passes/CheckInitialization.scala
@@ -8,6 +8,7 @@ import firrtl.Utils._
 import firrtl.traversals.Foreachers._
 
 import annotation.tailrec
+import scala.collection.mutable
 
 /** Reports errors for any references that are not fully initialized
   *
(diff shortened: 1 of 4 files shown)


jackkoenig:
There's a little bit of discussion of this issue in https://github.com/freechipsproject/firrtl/issues/852, I couldn't come up with an elegant solution but we definitely need to address this.

azidar:
I had forgotten about that issue... I just pushed a change which should hopefully fix the problem, with minimal changes to the compiler. Basically ExpandWhens now pseudo calls CheckInitialization to figure out whether it passes but still contains bad nodes. If it would pass CheckInitialization but has bad nodes, then it should delete those nodes (and does so).

PR #1039, opened 2019-02-27
[ExpandWhens] Don't create nodes to hold Muxes with >0 void cases

albert-magyar:
Fixes https://github.com/freechipsproject/firrtl/issues/852

### DIFF
--- a/src/main/scala/firrtl/passes/ExpandWhens.scala
+++ b/src/main/scala/firrtl/passes/ExpandWhens.scala
@@ -171,7 +171,20 @@ object ExpandWhens extends Pass {
               conseqNetlist getOrElse (lvalue, altNetlist(lvalue))
           }
 
+          // Does an expression contain WVoid inserted in this pass?
+          def containsVoid(e: Expression): Boolean = e match {
+            case WVoid => true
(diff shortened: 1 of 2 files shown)


jackkoenig:
The `WVoids` are checked in `CheckInitialization` so won't this break initialization checking?

albert-magyar:
The `WVoids` are still there, they just appear in inline expressions instead of getting assigned to nodes that are now illegal.

azidar:
Do you want to grab the test case from #1037 ? Your solution is definitely better than mine :)

jackkoenig:
oh nice, this is great 🙂

albert-magyar:
I generalized it to cover cases where `ValidIf`s with a `WVoid` defaults and `Mux`es with `WVoid` cases are mixed. I also added a deeper test case that covers this scenario.

albert-magyar:
@azidar, I talked to Jack, and I ended up using `MemoizedHash` to memoize the expressions. This should provide a  similar performance benefit to hashing the LHS values, and it definitely makes the `WVoid` checks a lot easier to understand.

azidar:
Will commenting rerun mergify?

albert-magyar:
So I'm guessing no?

Issue #941, opened 2018-11-16
Custom verilog emission


grebe:
**Type of issue**:  other enhancement

**What is the use case for changing the behavior?**

@azidar mentioned that you were looking at adding a special node type to firrtl, @albertchen-sifive.

I'm interested in being able to support a use-case where a chisel annotation is used to emit custom verilog. For example, I'd like to be able to emit \#ifdef guards for simulation around `Analog` types so we can use Verilog `real`.

Is this something close to what you're working on? If so, I'd like to be involved.

**Impact**: unknown

**Development Phase**: proposal

edwardcwang:
One thing to note is that we've been using `Analog` types for `inout` wires so hopefully using Verilog `real` will be an optional change/transform, not mandatory.

grebe:
I think the current analog behavior is probably better called inout (at least from chisel), but I see this as being optional in the sense that the current behavior is probably what should continue to happen by default, but there should be a mechanism for an annotation/custom node type coming out of chisel to emit custom verilog (simulation ifdefs being valuable to me). Perhaps chisel should spit this annotation out whenever you use analog and have something called InOut that doesn't use the annotation.

edwardcwang:
Sounds good. The current `Analog` seems almost like a blackbox type e.g. maybe even something like `Net`.

albertchen-sifive:
yep, I would like to add some custom node type that can be extended to emit things like #ifdef and such. I haven't gotten around to implementing anything yet though. @azidar you mentioned that there was some cleanup on some of the firrtl passes that should be done before starting?

seldridge:
I think there are two things here. You'd like to be able to add transforms that run only for a specific emitter. You'd also like to be able to override how specific nodes are emitted and to add custom nodes where you define their emission behavior.

Issue #687, opened 2017-11-16
Emission of Verilog Inline Attributes 


davidbiancolin:
I had a short conversation with Jack over slack and there doesn't seem to be a general way to do this. I'd like to write firrtl passes that add specific Inline Annotations like 
```(* mark_debug = "true" *) wire my_wire;``` or ``` (* ram_style = "block" *) reg [] my_ram [];```

doofin:
I also need this feature,which lets us annotate at chisel and then get those annotations at verilog.What's the current status? Glad to help!

jackkoenig:
A bit late, but https://github.com/freechipsproject/firrtl/pull/874 add annotations to let users put comments in the emitted Verilog. During the developers meeting last Friday, there was wide agreement with the approach and I believe we can easily extend that mechanism to support Inline Attributes as well! I don't have an exactly timeline but I believe we will have support for this the next few weeks.

jwright6323:
Is this still in the works? Would be nice to have this! cc @colinschmidt

grebe:
See also #941

jackkoenig:
Implemented in https://github.com/freechipsproject/firrtl/pull/1643

PR #1040, opened 2019-02-28
[WiP]: Add attributes, ifdefs to emitter.

grebe:
This adds `(* ... *)` attributes and ```  `ifndef `else `endif` ``` to the verilog emitter by extending the existing `Description` functionality. For motivation, see #941 and #687.

For attributes, the way it works is pretty similar to comments- just uses `(* *)` instead of `//`.

Ifdef is a little trickier. You need to add stuff both before and after the code of interest. Hopefully the way I've done it is correct. I'm curious if anyone has any thoughts about how I could make it better.

Another issue with the ifdefs: right now, they actually emit `ifndef` because it is easier to emit the alternative clause first rather than after the code. I think ideally you should be able to emit either an ``` `ifdef ``` or ``` `ifndef ```.

I've augmented an existing test- if I get good feedback, I'll clean them up and add a bunch more test cases.


jwright6323:
@grebe is this something you're still working on?

grebe:
Yes, it's a TODO to split out attributes from ifdefs. I think ifdefs will take a while.

PR #1041, opened 2019-02-28
Add a data structure for memory conf reading and writing

jwright6323:
This provides a data structure wrapper around the existing memory conf format which contains both reading and writing methods, making it easier to write code that needs to read the format. This should produce identical .conf files to the existing code, so it should be fully backward-compatible. From ucb-bar/barstools#35

### DIFF
--- a/src/main/scala/firrtl/passes/memlib/MemConf.scala
+++ b/src/main/scala/firrtl/passes/memlib/MemConf.scala
@@ -1,0 +1,61 @@
+// See LICENSE for license details.
+
+package firrtl.passes
+package memlib
+
+import scala.util.matching._
+
+sealed abstract class MemPort(val name: String) { override def toString = name }
+
+case object ReadPort extends MemPort("read")
+case object WritePort extends MemPort("write")
+case object MaskedWritePort extends MemPort("mwrite")
+case object ReadWritePort extends MemPort("rw")
+case object MaskedReadWritePort extends MemPort("mrw")
+
+object MemPort {
+
+  val all = Set(ReadPort, WritePort, MaskedWritePort, ReadWritePort, MaskedReadWritePort)
+
+  def apply(s: String): Option[MemPort] = MemPort.all.find(_.name == s)
+
+  def fromString(s: String): Seq[MemPort] = {
+    s.split(",").toSeq.map(MemPort.apply).map(_ match {
+      case Some(x) => x
+      case _ => throw new Exception(s"Error parsing MemPort string : ${s}")
+    })
+  }
+}
+
+case class MemConf(
+  name: String,
+  depth: Int,
+  width: Int,
+  ports: Seq[MemPort],
+  maskGranularity: Option[Int]
+) {
+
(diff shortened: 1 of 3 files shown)


src/main/scala/firrtl/passes/memlib/MemConf.scala
```diff
@@ -0,0 +1,61 @@
+// See LICENSE for license details.
+
+package firrtl.passes
+package memlib
+
+import scala.util.matching._
+
+sealed abstract class MemPort(val name: String) { override def toString = name }
```
edwardcwang: Should this be more like `MemPortType`?

src/main/scala/firrtl/passes/memlib/MemConf.scala
```diff
@@ -0,0 +1,61 @@
...
+case object ReadWritePort extends MemPort("rw")
+case object MaskedReadWritePort extends MemPort("mrw")
+
+object MemPort {
+
+  val all = Set(ReadPort, WritePort, MaskedWritePort, ReadWritePort, MaskedReadWritePort)
+
+  def apply(s: String): Option[MemPort] = MemPort.all.find(_.name == s)
+
+  def fromString(s: String): Seq[MemPort] = {
```
edwardcwang: Some scaladoc may be useful here; it appears that `s` is meant to be a comma-delimited list of port types?

src/main/scala/firrtl/passes/memlib/MemConf.scala
```diff
@@ -0,0 +1,61 @@
...
+  depth: Int,
+  width: Int,
+  ports: Seq[MemPort],
+  maskGranularity: Option[Int]
+) {
+
+  private def portsStr = ports.map(_.name).mkString(",")
+  private def maskGranStr = maskGranularity.map((p) => s"mask_gran $p").getOrElse("")
+
+  override def toString() = s"name ${name} depth ${depth} width ${width} ports ${portsStr} ${maskGranStr} "
```
edwardcwang:
```suggestion
  override def toString = s"name ${name} depth ${depth} width ${width} ports ${portsStr} ${maskGranStr} "
```

src/main/scala/firrtl/passes/memlib/MemConf.scala
```diff
@@ -0,0 +1,61 @@
...
+}
+
+object MemConf {
+
+  val regex = raw"\s*name\s+(\w+)\s+depth\s+(\d+)\s+width\s+(\d+)\s+ports\s+([^\s]+)\s+(?:mask_gran\s+(\d+))?\s*".r
+
+  def fromString(s: String): Seq[MemConf] = {
```
edwardcwang: Scaladoc indicating that this takes in a string containing multiple lines delimited by `\n`?

azidar:
Do you think it would be difficult to add a test of this memory configuration stuff?

jwright6323:
It shouldn't be too hard. I will add some.

jwright6323:
@azidar Done. Let me know if you want anything else.

PR #1043, opened 2019-02-28
Make mergify run when no reviews request changes

azidar:


azidar:
It seems like sometimes mergify doesn't detect that it needs to rerun, and in the past what has triggered this case is when it fails to merge because of repo permissions. Before it was due to master being protected, and the previous one was because there was an existing changes-requested-review. There are various was to retrigger it to run, but none of them were working.

I'm hoping with this fix, we shouldn't come across the case where mergify needs to rerun because it failed to merge.

PR #1045, opened 2019-03-04
Upgrade AsyncReset non-literal values check 

### LINKED ISSUE #1044: AsyncReset literals check is NOT compatible with chisel's complex literals emission
**Type of issue**: bug report

**Steps to reproduce the bug:**
Compile the following fragment
`./firrtl -i bug_asyncreset.fir -tn Test`
  * **Relevant FIRRTL fragment:**
```
circuit Test : 
  module Test : 
    input clock : Clock
    input reset : AsyncReset
    @[NB: following lines were generated by chisel]
    @[val valid = RegInit(VecInit(Seq.fill(8)(false.B)))]
    wire _T_6 : UInt<1>[8]
    _T_6[0] <= UInt<1>("h00")
    _T_6[1] <= UInt<1>("h00")
    _T_6[2] <= UInt<1>("h00")
    _T_6[3] <= UInt<1>("h00")
    _T_6[4] <= UInt<1>("h00")
    _T_6[5] <= UInt<1>("h00")
    _T_6[6] <= UInt<1>("h00")
    _T_6[7] <= UInt<1>("h00")
    reg valid : UInt<1>[8], clock with : (reset => (reset, _T_6))
```
```
Exception in thread "main" firrtl.passes.CheckHighForm$NonLiteralAsyncResetValueException: 
@[matching_fifo.scala 57:37:bug_asyncreset.fir@15.4]: 
[module Test] AsyncReset Reg 'valid' reset to non-literal '_T_6'
```
No errors: _T_6 is actually a literal

  * **Environment:**
    - version: `3.2-SNAPSHOT`
    - OS: `Darwin 17.7.0 Darwin Kernel Version 17.7.0: Thu Dec 20 21:47:19 PST 2018; root:xnu-4570.71.22~1/RELEASE_X86_64 x86_64`

Ensuring chisel's firrtl emission compatibility, see the simple example given
`val valid = RegInit(VecInit(Seq.fill(8)(false.B)))`

**Other information**

johnsbrew:
Upgrade AsyncReset non-literal values check with in-depth recursive  lookup of complex types

Fix #1044

### DIFF
--- a/src/main/scala/firrtl/passes/Checks.scala
+++ b/src/main/scala/firrtl/passes/Checks.scala
@@ -12,6 +12,7 @@ import firrtl.WrappedType._
 object CheckHighForm extends Pass {
   type NameSet = collection.mutable.HashSet[String]
   type LiteralSet = collection.mutable.HashMap[String, Boolean]
+  type RegCheckSet = collection.mutable.HashMap[Expression, PassException]
 
   // Custom Exceptions
   class NotUniqueException(info: Info, mname: String, name: String) extends PassException(
@@ -192,7 +193,7 @@ object CheckHighForm extends Pass {
       literals(name) = literals.getOrElse(name, true) && resolveToLiteral(literals, value)
 
 
-    def checkHighFormS(minfo: Info, mname: String, names: NameSet, literals: LiteralSet)(s: Statement): Unit = {
+    def checkHighFormS(minfo: Info, mname: String, names: NameSet, literals: LiteralSet, regsToCheck: RegCheckSet)(s: Statement): Unit = {
       val info = get_info(s) match {case NoInfo => minfo case x => x}
       s foreach checkName(info, mname, names)
       s match {
@@ -200,8 +201,8 @@ object CheckHighForm extends Pass {
           literals(name) = false
           if (hasFlip(tpe))
             errors.append(new RegWithFlipException(info, mname, name))
-          if (reset.tpe == AsyncResetType && !(resolveToLiteral(literals,init)))
-            errors.append(new NonLiteralAsyncResetValueException(info, mname, name, init.serialize))
+          if (reset.tpe == AsyncResetType)
+            regsToCheck(init) = new NonLiteralAsyncResetValueException(info, mname, name, init.serialize)
         case sx: DefMemory =>
           literals(name) = false
           if (hasFlip(sx.dataType))
@@ -225,7 +226,7 @@ object CheckHighForm extends Pass {
       }
       s foreach checkHighFormT(info, mname)
       s foreach checkHighFormE(info, mname, names, literals)
-      s foreach checkHighFormS(minfo, mname, names, literals)
+      s foreach checkHighFormS(minfo, mname, names, literals, regsToCheck)
     }
 
     def checkHighFormP(mname: String, names: NameSet, literals: LiteralSet)(p: Port): Unit = {
@@ -240,8 +241,13 @@ object CheckHighForm extends Pass {
     def checkHighFormM(m: DefModule) {
       val names = new NameSet
       val literals = new LiteralSet
+      val regsToCheck = new RegCheckSet
       m foreach checkHighFormP(m.name, names, literals)
-      m foreach checkHighFormS(m.info, m.name, names, literals)
+      m foreach checkHighFormS(m.info, m.name, names, literals, regsToCheck)
+      regsToCheck foreach ( (c:(Expression, PassException)) => {
+        if(!resolveToLiteral(literals, c._1))
+          errors.append(c._2)
+      })
     }
 
     c.modules foreach checkHighFormM
--- a/src/main/scala/firrtl/passes/Checks.scala
+++ b/src/main/scala/firrtl/passes/Checks.scala
@@ -166,9 +166,8 @@ object CheckHighForm extends Pass {
     }
     
     def resolveToLiteral(literals: LiteralSet, e: Expression): Boolean = {
-      if(e.isInstanceOf[Literal])
-        return true
       e match {
+        case _: Literal => true
         case t: HasName => literals.contains(t.name) && literals(t.name)
         case t: WRef => literals.contains(t.name) && literals(t.name)
         case t: WSubIndex => resolveToLiteral(literals, t.expr)
@@ -189,12 +188,9 @@ object CheckHighForm extends Pass {
       }
     }
     
-    def validateLiteralS(info: Info, mname: String, literals: LiteralSet, name: String, value: Expression): Unit = {
-      if (literals.contains(name))
-        literals(name) = literals(name) && resolveToLiteral(literals, value)
-      else 
-        literals(name) = resolveToLiteral(literals, value)
-    }
+    def validateLiteralS(info: Info, mname: String, literals: LiteralSet, name: String, value: Expression): Unit =
+      literals(name) = literals.getOrElse(name, true) && resolveToLiteral(literals, value)
+
 
     def checkHighFormS(minfo: Info, mname: String, names: NameSet, literals: LiteralSet)(s: Statement): Unit = {
       val info = get_info(s) match {case NoInfo => minfo case x => x}
@@ -207,13 +203,13 @@ object CheckHighForm extends Pass {
           if (reset.tpe == AsyncResetType && !(resolveToLiteral(literals,init)))
             errors.append(new NonLiteralAsyncResetValueException(info, mname, name, init.serialize))
         case sx: DefMemory =>
-        literals(name) = false
+          literals(name) = false
           if (hasFlip(sx.dataType))
             errors.append(new MemWithFlipException(info, mname, sx.name))
           if (sx.depth <= 0)
             errors.append(new NegMemSizeException(info, mname))
         case sx: WDefInstance =>
-        literals(name) = false
+          literals(name) = false
           if (!moduleNames(sx.module))
             errors.append(new ModuleNotDefinedException(info, mname, sx.module))
           // Check to see if a recursive module instantiation has occured
--- a/src/main/scala/firrtl/passes/Checks.scala
+++ b/src/main/scala/firrtl/passes/Checks.scala
@@ -11,6 +11,7 @@ import firrtl.WrappedType._
 
 object CheckHighForm extends Pass {
   type NameSet = collection.mutable.HashSet[String]
+  type LiteralSet = collection.mutable.HashMap[String, Boolean]
 
   // Custom Exceptions
   class NotUniqueException(info: Info, mname: String, name: String) extends PassException(
@@ -142,7 +143,7 @@ object CheckHighForm extends Pass {
       }
     }
 
-    def checkHighFormE(info: Info, mname: String, names: NameSet)(e: Expression): Unit = {
+    def checkHighFormE(info: Info, mname: String, names: NameSet, literals: LiteralSet)(e: Expression): Unit = {
       e match {
         case ex: WRef if !names(ex.name) =>
           errors.append(new UndeclaredReferenceException(info, mname, ex.name))
@@ -155,7 +156,7 @@ object CheckHighForm extends Pass {
       }
       e foreach checkHighFormW(info, mname)
       e foreach checkHighFormT(info, mname)
-      e foreach checkHighFormE(info, mname, names)
+      e foreach checkHighFormE(info, mname, names, literals)
     }
 
     def checkName(info: Info, mname: String, names: NameSet)(name: String): Unit = {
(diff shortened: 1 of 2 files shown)


johnsbrew:
Just realized that this fix not really working since the foreach never loops...
Working on another solution

johnsbrew:
**Commit message**    
Upgrade AsyncReset non-literal values check

Use a local hashmap of declared references.
By default all references are considered as literal and they are invalidated as soon as they get asssigned to a non-literal (ports, registers, instances)

**IMPORTANT NOTICE:** this behaviour has 1 known flaw : if a reference is populated/connected to a non-literal after its use as an async reet init value, the error WILL NOT be triggered.
2 solutions to solve this issue :
    - run 2 disctincts passes (a first to build the complete hashmap of valid literal, the second to run the checks)
    - build a local map of dependancies (connections) between references (heavy !)

**NOTE:** the above-mentionned behaviour CANNOT happen during chisel's FIRRTL emission

johnsbrew:
@jackkoenig what do you think of this new proposal ?

src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -155,47 +156,84 @@ object CheckHighForm extends Pass {
...
+        return true
+      e match {
+        case t: HasName => literals.contains(t.name) && literals(t.name)
+        case t: WRef => literals.contains(t.name) && literals(t.name)
+        case t: WSubIndex => resolveToLiteral(literals, t.expr)
+        case t: WSubField => resolveToLiteral(literals, t.expr)
+        case t: WSubAccess => resolveToLiteral(literals, t.expr)
+        case _ => false
+      }
+    }
```
jackkoenig: I think this would be better expressed as:
```scala
     def resolveToLiteral(literals: LiteralSet, e: Expression): Boolean = {
        case _: Literal => true
        case t: HasName => literals.contains(t.name) && literals(t.name)
        case t: WRef => literals.contains(t.name) && literals(t.name)
        case t: WSubIndex => resolveToLiteral(literals, t.expr)
        case t: WSubField => resolveToLiteral(literals, t.expr)
        case t: WSubAccess => resolveToLiteral(literals, t.expr)
        case _ => false
      }
    }

File: src/main/scala/firrtl/passes/Checks.scala [hunk 2]
```diff
@@ -155,47 +156,84 @@ object CheckHighForm extends Pass {
...
+        case t: WSubAccess => validateLiteral(info, mname, literals, t.expr, value)
+        case _ => // do nothing 
+      }
+    }
+    
+    def validateLiteralS(info: Info, mname: String, literals: LiteralSet, name: String, value: Expression): Unit = {
+      if (literals.contains(name))
+        literals(name) = literals(name) && resolveToLiteral(literals, value)
+      else 
+        literals(name) = resolveToLiteral(literals, value)
```
jackkoenig: Normal Scala style says that multi-line `if-else` should use curly braces.

src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -155,47 +156,84 @@ object CheckHighForm extends Pass {
...
       s match {
         case DefRegister(info, name, tpe, _, reset, init) =>
+          literals(name) = false
           if (hasFlip(tpe))
             errors.append(new RegWithFlipException(info, mname, name))
-          if (reset.tpe == AsyncResetType && !init.isInstanceOf[Literal])
+          if (reset.tpe == AsyncResetType && !(resolveToLiteral(literals,init)))
             errors.append(new NonLiteralAsyncResetValueException(info, mname, name, init.serialize))
         case sx: DefMemory =>
+        literals(name) = false
```
jackkoenig: This and the below, should be tabbed to match the if

src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -155,47 +156,84 @@ object CheckHighForm extends Pass {
...
           if (!moduleNames(sx.module))
             errors.append(new ModuleNotDefinedException(info, mname, sx.module))
           // Check to see if a recursive module instantiation has occured
           val childToParent = moduleGraph add (mname, sx.module)
           if (childToParent.nonEmpty)
             errors.append(new InstanceLoop(info, mname, childToParent mkString "->"))
         case sx: Connect => checkValidLoc(info, mname, sx.loc)
+                            validateLiteral(info, mname, literals, sx.loc, sx.expr)
+        case sx: DefWire => literals(sx.name) = true
```
jackkoenig: If I'm correct that this case is handling unconnected wires, rather than defaulting `DefWires` to literals, I would explicitly do this on `IsInvalid`, I'd rather the compiler crash if something goes wrong than hide a bug.

src/main/scala/firrtl/passes/Checks.scala
jackkoenig: This really doesn't matter, but if you want to have fun with it, you could express this as:
```scala
literals(name) = literals.getOrElse(name, true) && resolveToLiteral(literals, value)
```

src/test/scala/firrtlTests/AsyncResetSpec.scala
```diff
@@ -90,6 +90,87 @@ class AsyncResetSpec extends FirrtlFlatSpec {
...
+        |complex_literal[1] <= bundle.b
+        |complex_literal[2] <= UInt<1>("h00")
+        |complex_literal[3] <= UInt<1>("h00")
+
+        |reg r : UInt<1>[4], clock with : (reset => (reset, complex_literal))
+        |r <= x
+        |z <= r""".stripMargin
+      )
+    result should containLine ("always @(posedge clock or posedge reset) begin")
+  }
```
jackkoenig: Can you add a test that this works for a wire with multiple connections?
Eg.
```
wire w : UInt<1>
w <= UInt(1)
when cond : 
  w <= a
reg r : UInt<1>, clock with : (reset => (asyncReset, w))
```
This should obviously fail, and I think you handle it correctly, just good to test.

jackkoenig:
I think it looks reasonable. Before merging, I would like to see what happens in synthesis if such a literal is not propagated to the resulting GroundType literal, this could happen if someone dontTouches the intermediate wire or turns off optimizations. Since this does not cross module boundaries, I suspect it will work, but we should definitely check.

Aggregate literals would be much cleaner but I think it's unreasonable to block fully featured AsyncReset on that.

Do you have any thoughts on this @azidar, @ducky64, @aswaterman ?

This involves changing the Chisel API to allow non-literal reset values such that they can only reasonably caught in FIRRTL. The lack of Aggregate literals in FIRRTL aside, I think there is an argument that it provides more generality because it's impossible for a person to know via the Scala type system if some `Data` is a literal or not.

Currently this works:
```scala
val reg = withReset (someAsyncReset) {
  RegInit(0.U(32.W))
}
```
While the following only works with the merging of @johnsbrew's proposal in this PR:
```scala
val reg = withReset (someAsyncReset) {
  RegInit(WireInit(0.U(32.W)))
  // Note the init argument can obviously be some function argument
}
```
There is a new source of confusion because this will allow such propagation of literals to work for async reset values only within a Module, not across modules, but I think this is desirable behavior.

johnsbrew:
Hi @jackkoenig, thank you for this careful review.
Please don't be afraid of highlighting all the (despite numerous) defects in my work: practice makes perfect =)

ducky64:
@jackkoenig It doesn't appear we currently check for literal-ness in chisel3 frontend, so errors get delayed to FIRRTL anyways. If I'm understanding this change correctly (from comments, at least, haven't looked through code), it would allow certain types of analyses that better determine whether something is truly a hardware literal rather than just if it was directly declared literal.

I think one standard to apply would be whether this should also work or other types of const-prop-like analyses, eg should `RegInit(2.U+2.U)` be (or is already) legal?

I'd also like to see aggregate literals in FIRRTL. Wasn't there a PR for that? But aggregate literals also probably won't fully address all cases enabled by this change.

src/main/scala/firrtl/passes/Checks.scala
johnsbrew: Nice one-liner indeed

src/test/scala/firrtlTests/AsyncResetSpec.scala
johnsbrew: it works as expected, testcase added

src/main/scala/firrtl/passes/Checks.scala
johnsbrew: Not exactly: this case is defining that any wire is a literal, whatever its connections/state/validy.
This is a potential flaw as I explained above (invalidating Connect might come after the DefRegister) however not the one your are talking about.
To confirm, I've just tried with the following test:

```
  "Unconnected Wire" should "NOT be allowed as reset values for AsyncReset" in {
    an [passes.CheckHighForm.NonLiteralAsyncResetValueException] shouldBe thrownBy {
      compileBody(s"""
        |input clock : Clock
        |input reset : AsyncReset
        |input x : UInt<1>
        |output z : UInt<1>
        |wire w : UInt<1>
        |reg r : UInt<1>, clock with : (reset => (reset, w))
        |r <= x
        |z <= r""".stripMargin
      )

```
and I got a `RefNotInitializedException` error from `firrtl.passes.CheckInitialization` as I would have expected (or at least hoped for ^^)
```
- should NOT be allowed as reset values for AsyncReset *** FAILED ***
[info]   Expected exception firrtl.passes.CheckHighForm$NonLiteralAsyncResetValueException to be thrown, but firrtl.passes.CheckInitialization$RefNotInitializedException was thrown (AsyncResetSpec.scala:133)
```

jackkoenig:
@ducky64, my original AsyncReset proposal did check in Chisel, but with the changes we adopted in the dev meeting (Reset as an abstract type in FIRRTL), it is no longer possible to catch all misuses in Chisel. That being said, without the change in this PR, we could still catch some.

The proposal in this PR only allows literals that are directly connected through wires. Things like `RegInit(2.U + 2.U)` will **not** be legal under this proposal. For implementation reasons (because we don't have aggregate literals), this PR also enables `RegInit(WireInit(2.U))`.

I agree with your standard of how this relates to other forms of const prop. I don't really like the idea of opening up to some const prop, but not all. I **do**, however, think there is a reasonable distinction to draw between purely connectivity constant propagation and things that involve removing logic, but I also understand (and originally proposed) not allowing that for AsyncReset.

The main reason that I'm open to this proposal is that, in my opinion, the restriction pointed out in #1044 is not acceptable, we obviously must support AsyncReset registers of aggregates.

### Possible alternative solution

There is a possible immediate term solution in lieu of proper aggregate literal support--we could have Chisel emit an annotation on wires that implement aggregate literals and then allow async reset registers to have those as reset values. This would remove the need for the recursive check. The cons are that it is less elegant and more prone to misuse than the alternative. It is fairly simple implementation though.

Thoughts on this, @johnsbrew?

edwardcwang:
The alternative solution sounds interesting in two aspects:

* The annotations can be gracefully removed when aggregate literals are implemented in FIRRTL
* For e.g. designs composed of multiple separate FIRRTL files, we could use this annotation to pass designer knowledge that a wire that it is constant. Of course the designer uses this at his/her own peril/risk

jackkoenig:
Possible problem with my proposal, even if the FIRRTL implementation doesn't require full aggregate literal support, the Chisel side does (otherwise it will be an API change to later restrict to that). Fortunately, most of the Chisel machinery is there in Chisel.

Random thought, for Chisel aggregate literals, I think `literalElement.asTypeOf(aggregateType)` should count as a literal, eg. `0.U.asTypeOf(Vec(8, UInt(8.W))` or whatever. That is how I tell people to create aggregates of all zero. This proposal aside, the API for aggregate literals still needs work.

ducky64:
One thing to note is that we typically don't do any const prop in the frontend, so propagating `literalElement.asTypeOf(aggregateType)` as literal would be a change. The advantage of doing this const prop in the frontend is that certain literal extractor operations would be possible, eg if you wanted to unpack a literal into Bundle fields and get a literal (and Scala-land number) out, though the disadvantage is possible subtle differences between Chisel and FIRRTL behavior depending on how code is written. If we did this, we should also be very clear on what kinds of const prop can be done in frontend.

aswaterman:
Given that we already do forward width propagation in the frontend, which is more finicky than forward constprop, using differences between chisel and firrtl as an excuse to not do forward constprop in the frontend is getting more tenuous by the day.

It’s also a sufficiently important compiler-performance optimization that 
we will eventually need to give in on that basis alone.

johnsbrew:
I had initially no strong opinion about where such a check should take place.
Afterwards I realized that our discussion was focused only on Chisel -> FIRRTL, but what about potential other front-end ? Say Yosys Verilog => FIRRTL for example.
I would find it awkward to have such a verilog front-end to have to generate annotations in addition to FIRRTL. For me it sounds like an acknowledgment of failure concerning the capacity of FIRRTL to be an *universal* hardware IR.
IMHO, at least every basic constructs that are *universally synthesizable* - not talking about some corner cases in a few advanced EDA tools - within verilog should expressible within FIRRTL only,  without having to use annotations.
Otherwise you have to produce a very precise standardization of all annotations that shall/should be produced/supported along with the FIRRTL specification.
As far as I understood the reason for using annotations during my discussions with @seldridge and @jackkoenig, their goal was to carry additional pieces of information that do NOT belong to this minimal hardware IR that FIRRTL is restricted to. [Footnote1]

Conclusion: IMHO either FIRRTL is fully able to do this check by itself (why should it even trust an external annotation? how to be certain to generate proper synthesizable verilog in the end?) either it is specified that this check is NOT done by FIRRTL and must be carried out  by the front-end.
- In the first case, FIRRTL takes the responsability for generating proper Verilog given any FIRRTL.
- In the second case, the front-end takes the responsability for generating proper FIRRTL and *de facto* takes the responability for improper

I really don't like this third annotation case where the responsability is somehow shared between the two.

If we go for the first case, and if we want this check to be exhaustive (in order to ensure that emitted verilog will be synthesizable , this PR should be reworked.
Indeed it currently makes one assumption about the order of statements.
This assumption is only valid for the current chisel's FIRRTL emission and I now realize how far this is an issue when it comes to consider the entire ecosystem around FIRRTL.

----

[Footnote1]: 
This point of view relaunches of course the debate about having Bitstream Preset considered as either annotated `AsyncReset` or as a fully featured type (inheriting from `Reset`, along with `SyncReset` and `AsyncReset`) within FIRRTL.
As decided during earlier discussion, I am currently working on the first solution and the corresponding PR should arise soon.
However I see here again a serious limitation and the risk for FIRRTL to miss the opportunity to become the *universal* hardware IR. Of course when I mean universal, I mean an even support for all hardware targets.
IMHO I would rather have a cli option --target, that would come with list of forbidden constructs for a given target (ex: --target asic would refuse the use of bitstream presets) instead of restricting FIRRTL to the minimal subset that is supported by all targets.

Then, from there we could add other cli options such as explicit casting (--preset-as-(a)sync). (for example a cast from Preset to AsyncReset or even )
I understand the point of view that says: "you could have it generated in a other way using, for example chisel construct `withReset`."
But here, once again, what about the other flows that are not Chisel -> FIRRTL?
What about having FIRRTL IPs libraries ?

Making Chisel & FIRRTL two distincts entities was a wonderful move, I would love them to remain as independent as possible from one another. In our discussion, I really see the generalized use of annotations as a potential threat to preserve their independence.

----

@jackkoenig @seldridge @azidar what are your thoughts about the entire ecosystem consistency?

johnsbrew:
@azidar I am sorry I couldn't be at the meeting last Friday.
I read the meetings notes and saw you are going to accept the PR.
I found a way to gracefully fix the issue I raised in my first comment https://github.com/freechipsproject/firrtl/pull/1045#issuecomment-469739138
So I added both corresponding test-case and fix :)

Just let me know if this is fine, so I can merge commits, rebase and push a clean version.

jackkoenig:
@johnsbrew I apologize for the delayed response, I was out of town so my working hours were limited and inconsistent.

Thank you for your detailed thoughts and ideas. I agree pretty much with everything you said. Let me address some of your points/concerns.

### Use of annotations for Async reset

I wasn't clear here so I apologize. My proposal for using annotations was intended as a **temporary** solution (I said "immediate", but I really meant "short term and temporary") until FIRRTL supports literal aggregates. Now, that doesn't mean it's a good idea, and your concerns apply even to temporary solutions, but I completely agree that the legality check here should be the responsibility of FIRRTL and should be able to work exclusively on the IR (no annotations involved).

### Should this PR be merged?

This is a question about approaches. My original proposal was that FIRRTL would require known literals as the async reset value (ie. it can't be connected through a wire, it must be known to FIRRTL to be a literal with no analysis). I don't feel that strongly about that approach and I am fine with what you've proposed here. It's less strict and therefore puts less onus on the frontend (which is nice). We have generally opposed requiring optimizations to check for correctness, but this is just connectivity analysis which isn't really an optimization. Relatedly, @azidar pointed out to me that it's similar in kind to combinational loop detection, so I'm down with this approach 👍.

I'm not sure if I agree with making it cross-module, but this PR has nothing to do with that so we can discuss that at length later.

### Register Preset as a first-class FIRRTL feature

Your point is well-taken--if we want FIRRTL to be able to be a *universal* hardware IR, it might be worth representing preset directly in the IR. As you pointed out, we separate Chisel and FIRRTL deliberately, so most of my concerns with the register preset API relate primarily to Chisel. Thus, the ability of Chisel to use AsyncReset and a transform to represent Preset does not mean that FIRRTL should not represent Preset in the IR.

I think this definitely warrants more discussion. One the one hand, I wish the FPGA tools could just be told that a certain async reset wire should be treated as preset (because then the Verilog you use for ASICs and FPGAs could be the same), but on the other, this should be relatively trivial to do.

Thanks for your thoughts and work on this @johnsbrew!

johnsbrew:
Thanks for your detailed answer @jackkoenig
I am glad we agree on these topics!

Your vision of having the ability to express a target-agnostic verilog is appealing but I thing it is also one of the ground reason for building a toolchain above EDA tools: we heavily depends on vendor EDA tools, and we know they do not move fast-enough for us.
I would rather ignore them than trying to make them add the support for new features. 
Just watch how far they all are to support even one single percent of SystemVerilog 2015 or VHDL-2008 features...

Regarding this PR, I think it worth it to merge it now to unlock the situation.
IMHO having cross-module litterals sounds like a bad coding practice back in chisel: parameters should have been used instead.

Whatever the upcoming discussion about extending FIRRTL to include native preset support, it would be great if you could have a look to #1050

jackkoenig:
Folded into https://github.com/freechipsproject/firrtl/pull/1068

jackkoenig:
Many thanks @johnsbrew for reporting this issue, an initial implementation, and the tests.

PR #1046, opened 2019-03-05
Advertise FIRRTL grammar support in Atom

johnsbrew:
I have just published the adaptation of the sublime-text language support to Atom.
Let's advertise it in the README :)


chick:
@edwardcwang just wondering if you or I should have just added the Please Merge tag so mergify would take over

PR #1056, opened 2019-03-13
Add serialization support for LoadMemoryFileType in LoadMemoryAnnotation

### LINKED ISSUE #1055: hexOrBinary field of LoadMemoryAnnotation isn't serialized/deserialized correctly
**Type of issue**: bug report
Try to create a LoadMemoryAnnotation with type `MemoryLoadFileType.Binary`, serialize it, then de-serialize it:
```scala
    val lma = new LoadMemoryAnnotation(
      ComponentName("ram", ModuleName("ModuleMem", CircuitName("CircuitMem"))),
      "CircuitMem.ModuleMem.ram.dat",
      hexOrBinary = MemoryLoadFileType.Binary,
      originalMemoryNameOpt = Some("memory")
    )

    val annoString = JsonProtocol.serializeTry(Seq(lma)).get
    val loadedAnnos = JsonProtocol.deserializeTry(annoString).get
```
```
No usable value for hexOrBinary
No usable value for $outer
Can't find field scala$Enumeration$$vmap from class scala.Enumeration
org.json4s.package$MappingException: No usable value for hexOrBinary
```
No exceptions.

`LoadMemoryAnnotation`s should be serializable.

Update `JsonProtocol.scala` to provide support for `LoadMemoryFileType`

ucbjrl:
Add custom LoadMemoryFileTypeSerializer.
Add test to verify LoadMemoryAnnotation can be correctly serialized/deserialized.

This resolves #1055.

### DIFF
--- a/src/main/scala/firrtl/annotations/JsonProtocol.scala
+++ b/src/main/scala/firrtl/annotations/JsonProtocol.scala
@@ -45,7 +45,19 @@ object JsonProtocol {
       }},
     { case x: Transform => JString(x.getClass.getName) }
   ))
-
+  class LoadMemoryFileTypeSerializer extends CustomSerializer[MemoryLoadFileType.FileType](format => ( {
+    case JString(s) => s match {
+      case "h" => MemoryLoadFileType.Hex
+      case "b" => MemoryLoadFileType.Binary
+      case _ => throw new FIRRTLException(s"Unrecognized MemoryLoadFileType: $s")
+    }
+  }, {
+    case named: MemoryLoadFileType.FileType => named match {
+        case MemoryLoadFileType.Hex => JString("h")
+        case MemoryLoadFileType.Binary => JString("b")
+        case _ => throw new FIRRTLException(s"Unrecognized MemoryLoadFileType: $named")
+    }
+  }))
 
   class TargetSerializer extends CustomSerializer[Target](format => (
     { case JString(s) => Target.deserialize(s) },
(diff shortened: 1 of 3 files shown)


jackkoenig:
What happens if you change the implementation of `MemoryLoadFileType` to something like:

```scala
sealed abstract class MemoryLoadFileType(val value: String)
object MemoryLoadFileType {
  case object Hex extends MemoryLoadFileType("h")
  case object Binary extends MemoryLoadFileType("b")
}
```

I ask because I think the JSON serialization should be using type tags, but it may be the case that Enumerations just don't work very well with it whereas the above might.

ucbjrl:
Changing to @jackkoenig's suggestion:
 - doesn't correctly serialize, produces:
```
[
  {
    "class":"firrtl.annotations.LoadMemoryAnnotation",
    "target":"CircuitMem.ModuleMem.ram",
    "fileName":"CircuitMem.ModuleMem.ram.dat",
    "hexOrBinary":{
      
    },
    "originalMemoryNameOpt":"memory"
  }
]
```
 - produces the following when de-serializing:
```
No usable value for hexOrBinary
No usable value for value
Did not find value which can be converted into java.lang.String
org.json4s.package$MappingException: No usable value for hexOrBinary
No usable value for value
Did not find value which can be converted into java.lang.String
	at org.json4s.reflect.package$.fail(package.scala:95)
	at org.json4s.Extraction$ClassInstanceBuilder.buildCtorArg(Extraction.scala:568)
```
I'm not sure what the goal is. Is it the elimination of `CustomSerializer`s? Shouldn't libraries/applications share annotation definitions (where possible), including their serializations?

PR #1058, opened 2019-03-13
Convert the RemoveAccesses object into a class.

### LINKED ISSUE #1057: Threading issue - RemoveAccesses object should not be shared
**Type of issue**: bug report

Running CI tests under Jenkins has revealed another concurrency issue. We find cases where tests are caught in a tight loop in:
```scala
/* inlined HashTable.findEntry0 to preserve its visibility */
private[this] def findEntry(key: A, h: Int): Entry = {
  var e = table(h).asInstanceOf[Entry]
  while (notFound(key, e))
    e = e.next
  e
}
```
due to the fact that `e.next` points to `e`.

This has been traced to the following code in src/main/scala/firrtl/passes/RemoveAccesses.scala:
```scala
// This improves the performance of this pass
private val createExpsCache = mutable.HashMap[Expression, Seq[Expression]]()
private def create_exps(e: Expression) =
  createExpsCache getOrElseUpdate (e, firrtl.Utils.create_exps(e))
```
This cache needs to be protected against simultaneous access from multiple threads and the problem can be remedied by:
- using a per-circuit cache,
- converting RemoveAccess from an object to a class and making it scoped per transform instance.

I should note that this happens with a frequency of around 1 in 100 so it's not readily reproducible.

Running tests simultaneously should be possible.

Related to issue #159

ucbjrl:
Prevent simultaneous access to common resources when tests are run in parallel.
This resolves #1057.

### DIFF
--- a/spec/spec.tex
+++ b/spec/spec.tex
@@ -1606,7 +1606,7 @@ Given a component with a bundle type, the suffix \verb|$|{\em f} is appended to 
 
 \subsection{Prefix Uniqueness} \label{prefix_unique}
 
-The {\em symbol sequence} of a name is the ordered list of strings that results from splitting the name at each occurence of the `\$' character.
+The {\em symbol sequence} of a name is the ordered list of strings that results from splitting the name at each occurrence of the `\$' character.
 
 A symbol sequence $a$ is a {\em prefix} of another symbol sequence $b$ if the strings in $a$ occur in the beginning of $b$.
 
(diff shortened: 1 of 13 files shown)


jackkoenig:
Having `object RemoveAccesses` with `def run` would prevent this change from breaking anyone's code who runs it directly, although I dunno if anyone actually does that.

For other transforms (`InferTypes`, `ResolveKinds`, etc.), that would be more important

ucbjrl:
Good points.

Issue #998, opened 2019-01-22
instance different verilog module when constant propagation.


sequencer:
**Type of issue**: other enhancement

**What is the use case for changing the behavior?**
1. optimize the constant propagating behavior by replace xor(a,1) to not(a) etc,
2. instance different verilog module when some ports are tie 0/1 rather than single instance with constant as input.    

**Impact**: API modification
**Development Phase**: request

Here is the example code by chisel3
```scala
import chisel3._

trait ConstantPropagationTest {
  class CSAOutBundle extends Bundle {
    val s: UInt = UInt(32.W)
    val c: UInt = UInt(32.W)
  }

  def csa(a: UInt, b: UInt, c: UInt): CSAOutBundle = {
    // the reason I use Module here is I wanna keep the circuit hierarchy. 
    val m = Module(new Module {
      val io = IO(new Bundle {
        val a = Input(UInt(32.W))
        val b = Input(UInt(32.W))
        val c = Input(UInt(32.W))
        val out = Output(new CSAOutBundle)
      })
      val a = Wire(Vec(32, Bool()))
      val b = Wire(Vec(32, Bool()))
      val c = Wire(Vec(32, Bool()))
      val cOut = Wire(Vec(32, Bool()))
      val sOut = Wire(Vec(32, Bool()))
      a := io.a.asTypeOf(Vec(32, Bool()))
      b := io.b.asTypeOf(Vec(32, Bool()))
      c := io.c.asTypeOf(Vec(32, Bool()))
      io.out.s := sOut.asTypeOf(UInt(32.W))
      io.out.c := cOut.asTypeOf(UInt(32.W))
      override def desiredName = "csa"
      for(i <- 0 to 31) {
        sOut(i) := a(i) ^ b(i) ^ c(i)
        cOut(i) := {
          i match {
            case j if j == 0 => false.B
            case default => (a(default - 1) && b(default - 1)) || (b(default - 1) && c(default - 1)) || (a(default - 1) && c(default - 1))
          }
        }
      }
    })
    m.io.a := a
    m.io.b := b
    m.io.c := c
    m.io.out
  }
}

class ConstantTester(dual: Boolean) extends Module
  with ConstantPropagationTest {
  val io = IO(new Bundle {
    val a = Input(UInt(32.W))
    val b = Input(UInt(32.W))
    val out0 = Output(new CSAOutBundle)
    val out1 = Output(new CSAOutBundle)
  })
  val c = WireInit("h1234abcd".U)
[...]

PR #1062, opened 2019-03-25
Constant propagation complement

sequencer:
add sub and mul for constant propagation.


sequencer:
Here I have another question:
If I wanna optimize the function of `def simplify(e: Expression, lhs: Literal, rhs: Expression) ` to `FoldAND`, `FoldOR`, `FoldXOR`, using a sequence of bits with `Not` and `Cat`, Will it be OK?

src/main/scala/firrtl/transforms/ConstantPropagation.scala
```diff
@@ -89,6 +89,32 @@ class ConstantPropagation extends Transform with ResolvedAnnotationPaths {
     }
   }
 
+  object FoldSUB extends FoldCommutativeOp {
+    def fold(c1: Literal, c2: Literal) = ((c1, c2): @unchecked) match {
+      case (_: UIntLiteral, _: UIntLiteral) => UIntLiteral(c1.value - c2.value, c1.width max c2.width)
```
aswaterman: The width should be `(c1.width max c2.width) + IntWidth(1)` here too

src/main/scala/firrtl/transforms/ConstantPropagation.scala
```diff
@@ -89,6 +89,32 @@ class ConstantPropagation extends Transform with ResolvedAnnotationPaths {
...
+  }
+
+  object FoldMUL extends FoldCommutativeOp {
+    def fold(c1: Literal, c2: Literal) = ((c1, c2): @unchecked) match {
+      case (_: UIntLiteral, _: UIntLiteral) => UIntLiteral(c1.value * c2.value, c1.width + c2.width)
+      case (_: SIntLiteral, _: SIntLiteral) => SIntLiteral(c1.value * c2.value, c1.width + c2.width)
+    }
+    def simplify(e: Expression, lhs: Literal, rhs: Expression) = lhs match {
+      case UIntLiteral(v, w) if v == BigInt(0) => UIntLiteral(0, w)
+      case SIntLiteral(v, w) if v == BigInt(0) => UIntLiteral(0, w)
```
aswaterman: `SIntLiteral(0, w)`

aswaterman: Actually for both of these you can simply return `lhs` to avoid creating a new object.

aswaterman:
As to your question about Not/Cat: could you write an example of what you're trying to do?  (Also, provided we agree, it should be saved for a separate PR)

sequencer:
for example:
```
val a = UInt(4.W)
```

sequencer:
for example:
```
val a = UInt(4.W)
val b = a ^ 8.U(4.W)
```
`b` should be `Cat(!a(3),a(2),a(1),a(0))`

sequencer:
In this comment: https://github.com/freechipsproject/firrtl/issues/998#issue-401750878
`sOut_7` can not be optimized `sOut_7 = ! _T_730` is the same problem.

aswaterman:
That optimization is best left for synthesis, because it will usually reduce simulation performance, and the synthesis tool will always get it right.

sequencer:
So what I wanna do is using `Cat` and `Not` to represent a `Expression` which has Constants. This can reduce the number of complex gates like `xor`.
Although synthesis can do it too, but I think it is more clear to make this happen in firrtl stage.

sequencer:
Oh I got it.

aswaterman (approved):

jackkoenig:
I suspect these failures are false positives, running locally (and refactoring the script so that running locally is easier to do).

sequencer:
`run_formal_checks` seems failed?

sequencer:
any update on this?

sequencer:
I believe the formal checks are false positives as jack said.

albert-magyar:
I bumped this, as it's still relevant. In light of what we've seen so far with recent subtle Verilog-emission bugs, it would be good to add a couple tests and/or ensure LEC passes.

PR #1067, opened 2019-03-28
Faster reg constprop

albert-magyar:
This reuses memoized information about nodes across the const prop checks for different registers. This lowers reg const prop from a worst-case quadratic runtime to linear runtime.


colinschmidt:
Works on my slightly larger use case. So LGTM!

src/main/scala/firrtl/transforms/ConstantPropagation.scala
```diff
@@ -353,6 +353,31 @@ class ConstantPropagation extends Transform with ResolvedAnnotationPaths {
...
+      def resolve[V >: T](that: ConstPropBinding[V]): ConstPropBinding[V] = (this, that) match {
+        case (x, y) if (x == y) => x
+        case (x, UnboundConstant) => x
+        case (UnboundConstant, y) => y
+        case _ => NonConstant
+      }
+    }
+    case class BoundConstant[T](value: T) extends ConstPropBinding[T]
+    case object NonConstant extends ConstPropBinding[Nothing]
+    case object UnboundConstant extends ConstPropBinding[Nothing]
```
jackkoenig: It doesn't matter that much, but maybe these should be hoisted out of this function, perhaps into the companion object (private of course)?

src/main/scala/firrtl/transforms/ConstantPropagation.scala
```diff
@@ -420,30 +444,36 @@ class ConstantPropagation extends Transform with ResolvedAnnotationPaths {
...
+            case _ => RegCPEntry(NonConstant, NonConstant)
           }
-          def cpExp(e: Expression) = constPropExpression(nodeMap, instMap, constSubOutputs)(e)
-          regConstant(rhs).foreach {
-            case wr: WRef => nodeMap(lname) = cpExp(pad(passes.RemoveValidIf.getGroundZero(ltpe), ltpe))
-            case e => nodeMap(lname) = cpExp(e)
+          val zero = passes.RemoveValidIf.getGroundZero(ltpe)
+          def padCPExp(e: Expression) = constPropExpression(nodeMap, instMap, constSubOutputs)(pad(e, ltpe))
+          regConstant(rhs) match {
+            case RegCPEntry(BoundConstant(name), litBinding) if (name == lname) => litBinding match {
```
jackkoenig:
```suggestion
            case RegCPEntry(BoundConstant(`lname`), litBinding) => litBinding match {
```
I prefer this to using following if guards but doesn't really matter.

src/main/scala/firrtl/transforms/ConstantPropagation.scala
```diff
@@ -420,30 +444,36 @@ class ConstantPropagation extends Transform with ResolvedAnnotationPaths {
...
+            case WRef(nodeName, _, NodeKind, _) if nodeMap.contains(nodeName) =>
+              nodeRegCPEntries.getOrElseUpdate(nodeName, { regConstant(nodeMap(nodeName)) })
+            case Mux(_, tval, fval, _) => regConstant(tval).resolve(regConstant(fval))
+            case _ => RegCPEntry(NonConstant, NonConstant)
           }
-          def cpExp(e: Expression) = constPropExpression(nodeMap, instMap, constSubOutputs)(e)
-          regConstant(rhs).foreach {
-            case wr: WRef => nodeMap(lname) = cpExp(pad(passes.RemoveValidIf.getGroundZero(ltpe), ltpe))
-            case e => nodeMap(lname) = cpExp(e)
+          val zero = passes.RemoveValidIf.getGroundZero(ltpe)
```
jackkoenig: This is going to construct a zero for every time we see this pattern, perhaps it should be a def instead?

src/main/scala/firrtl/transforms/ConstantPropagation.scala
albert-magyar: Haha, that's how I originally wrote it, but I changed it since there is so little backtick pattern match usage in firrtl, and I thought it might be confusing.

PR #1068, opened 2019-04-02
Infer reset

jackkoenig:
This PR implements `ResetType` (`Reset` in the serialize format) as an "abstract" reset. Things of type `Reset` will be inferred in a flow forward way to be either `UInt<1>` or `AsyncReset`. They are disallowed on top-level inputs and BlackBox outputs.

Currently, this pass runs after `ExpandConnects` because I didn't want to duplicate the work of expanding connections. I think it probably should run earlier, during `ResolveAndCheck`, but will require a little bit of work to do that. Whenever this pass is run, https://github.com/freechipsproject/firrtl/pull/1045 must run after it.

I'm starting this as a Draft Pull Request because it's ready for a look, but probably needs to be modified to run in `ResolveAndCheck`


jackkoenig:
This is pretty much done. What's missing:
- [ ] Unify with https://github.com/freechipsproject/firrtl/pull/1045 (this should be run before the check in #1045)
- [ ] Make this run before Dedup. Similarly to past issues with Width inference, Dedup means that you must infer the same result for two `Reset`-templated instances that dedup to the same Module
- [ ] Add tests that deal with the Dedup stuff

src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -388,13 +424,15 @@ object CheckTypes extends Pass {
...
 
     def bulk_equals(t1: Type, t2: Type, flip1: Orientation, flip2: Orientation): Boolean = {
-      //;println_all(["Inside with t1:" t1 ",t2:" t2 ",f1:" flip1 ",f2:" flip2])
       (t1, t2) match {
         case (ClockType, ClockType) => flip1 == flip2
         case (_: UIntType, _: UIntType) => flip1 == flip2
         case (_: SIntType, _: SIntType) => flip1 == flip2
         case (_: FixedType, _: FixedType) => flip1 == flip2
         case (_: AnalogType, _: AnalogType) => true
+        case (AsyncResetType, AsyncResetType) => flip1 == flip2
```
chick: Why is there an asymmetry here between AsyncResetType and ResetType.
Is it intended that ART ART is ok, RT can connect to RT, UInt and ART,
but ART cannot connect to UInt.
Could be fine, but I don't get it.

jackkoenig: The asymmetry comes from the fact that `ResetType` gets inferred by what's connected to it. Normally you have to cast all types to connect them (eg. must cast from AsyncResetType to UInt if you want to connect), but the types of things driving ResetTypes tells us what the type should be:
```
wire x : UInt<1>
wire w : Reset
w <= x ; This will cause us to infer w to be UInt<1>
```
You can't connect a Reset to a UInt (nor AsyncResetType) without casting because the inference only flows one way.

src/main/scala/firrtl/Utils.scala
```diff
@@ -477,6 +478,11 @@ object Utils extends LazyLogging {
             ilen + get_size(t1x.tpe), jlen + get_size(t2x.tpe))
         }._1
       case (ClockType, ClockType) => if (flip1 == flip2) Seq((0, 0)) else Nil
+      case (AsyncResetType, AsyncResetType) => if (flip1 == flip2) Seq((0, 0)) else Nil
+      case (ResetType, other) =>
+        if (legalResetType(other) && flip1 == Default && flip1 == flip2) Seq((0, 0)) else Nil
```
azidar: Could use a comment with an example. I think this is very subtle, and it would be good for clarity to explain (e.g. why in this case the flips must be default, but in the other case, they must be flip.

src/main/scala/firrtl/WIR.scala
```diff
@@ -215,32 +215,43 @@ case class ExpWidth(arg1: Width) extends Width with HasMapWidth {
 object WrappedType {
   def apply(t: Type) = new WrappedType(t)
   def wt(t: Type) = apply(t)
-}
-class WrappedType(val t: Type) {
-  def wt(tx: Type) = new WrappedType(tx)
-  override def equals(o: Any): Boolean = o match {
-    case (t2: WrappedType) => (t, t2.t) match {
+  private def compare(strict: Boolean)(t1: Type, t2: Type): Boolean =
```
azidar: Perhaps rename the arguments from t1/t2 to source/sink. This way it implies that the argument order is important. I'd also add comments explaining the arguments.

src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -57,6 +57,11 @@ object CheckHighForm extends Pass {
     s"$info: [module $mname] Primop $op lsb $lsb > $msb.")
   class NonLiteralAsyncResetValueException(info: Info, mname: String, reg: String, init: String) extends PassException(
     s"$info: [module $mname] AsyncReset Reg '$reg' reset to non-literal '$init'")
+  class ResetInputException(info: Info, mname: String, expr: Expression) extends PassException(
+    s"$info: [module $mname] Abstract Reset not allowed as top-level input: ${expr.serialize}")
+  class ResetExtModuleOutputException(info: Info, mname: String, expr: Expression) extends PassException(
+    s"$info [module $mname] Abstract Reset not allowed as ExtModule output: ${expr.serialize}")
```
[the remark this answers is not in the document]
azidar: Is it ok to have an input ExtModule reset type? I think it is, but I think its worth discussing?

src/main/scala/firrtl/passes/InferResets.scala
```diff
@@ -0,0 +1,200 @@
...
+import firrtl.analyses.InstanceGraph
+import firrtl.graph.{DiGraph, MutableDiGraph}
+import firrtl.annotations.{ReferenceTarget, TargetToken}
+import firrtl.Utils.toTarget
+import collection.mutable
+
+import scala.util.{Try, Success, Failure}
+
+/** Infers the concrete type of [[Reset]]s by their connections
+  * This is a global inference because ports can be of type [[Reset]]
```
azidar: I'd document the behavior of the case if two instances of the same module have inferred different reset ports.

src/main/scala/firrtl/passes/InferResets.scala
```diff
@@ -0,0 +1,200 @@
...
+          case Connect(_, lhs, rhs) if lhs.tpe == ResetType =>
+            val target = makeTarget(lhs)
+            val driver = rhs.tpe match {
+              case ResetType => TargetDriver(makeTarget(rhs))
+              case tpe       => TypeDriver(tpe, () => makeTarget(rhs))
+            }
+            map(target) = driver :: Nil
+          case WDefInstance(_, inst, module, _) =>
+            instMap += (inst -> module)
+          case Conditionally(_, _, con, alt) =>
```
azidar: Does this pass work on High FIRRTL now? If so, you may have SubAccesses in the lhs of the Connect, which I'm not sure Utils.toTarget will work properly.

src/main/scala/firrtl/passes/InferResets.scala
```diff
@@ -0,0 +1,200 @@
...
+    for ((target, _) <- map) {
+      rec(target)
+    }
+    Try {
+      errors.trigger()
+      res.toMap
+    }
+  }
+
+  private sealed trait TypeTree
```
azidar: Add comment describing the high level purpose of these.

src/main/scala/firrtl/passes/InferResets.scala
```diff
@@ -0,0 +1,200 @@
...
+  final class DifferingDriverTypesException private (msg: String) extends PassException(msg)
+  object DifferingDriverTypesException {
+    def apply(target: ReferenceTarget, tpes: Seq[(Type, Seq[TypeDriver])]): DifferingDriverTypesException = {
+      val xs = tpes.map { case (t, ds) => s"${ds.map(_.target().serialize).mkString(", ")} of type ${t.serialize}" }
+      val msg = s"${target.serialize} driven with multiple types!" + xs.mkString("\n  ", "\n  ", "")
+      new DifferingDriverTypesException(msg)
+    }
+  }
+
+  private sealed trait ResetDriver
```
azidar: Add high-level overview of what these classes represent.

src/main/scala/firrtl/passes/InferResets.scala
```diff
@@ -0,0 +1,200 @@
...
+
+  private sealed trait TypeTree
+  private case class BundleTree(fields: Map[String, TypeTree]) extends TypeTree
+  private case class VectorTree(subType: TypeTree) extends TypeTree
+  // TODO ensure is only AsyncResetType or BoolType
+  private case class GroundTree(tpe: Type) extends TypeTree
+
+  private object TypeTree {
+    // TODO make return Try[TypeTree]
+    def fromTokens(tokens: (Seq[TargetToken], Type)*): TypeTree = tokens match {
```
azidar: Clarify what the tokens first and second element represent in comment, as well as the return value.

azidar: From where this is call, I see this as converting a bunch of ReferenceTargets (and their type) which all target the subelements of one aggregate typed component. This then converts all of them into what the new aggregate type should be. Why did you need to create a new class, and not just use the ir.Type?

azidar: In addition, this seems to be a very useful function, irregardless of the async reset. We should consider putting it in a standard place, not hidden away here.

azidar: I'd recommend adding more comments to all of these private functions.

azidar:
Overall looks really good! It could use more documentation explaining some of the new classes and what many of the private functions do. Supposing this works prior to RemoveAccesses, we need to see a test assuming the loc of a Connect is a WSubAccess.

src/main/scala/firrtl/passes/Checks.scala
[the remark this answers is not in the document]
jackkoenig: I'm actually not sure, is it okay? @edwardcwang has requested the ability to have the reset values for `AsyncReset` registers be inputs from the top or ExtModules since it might be coming from some analog (or otherwise not expressible in FIRRTL) module, I guess this could be a related feature.

src/main/scala/firrtl/passes/InferResets.scala
jackkoenig: It actually requires `ExpandConnects` so runs during lowering from High to Mid. That being said, I do have it running before `RemoveAccesses` so your point is well taken, I need to test that.

src/main/scala/firrtl/passes/Checks.scala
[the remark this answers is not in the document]
edwardcwang: I mean, is there a particularly good reason to restrict it?

src/main/scala/firrtl/passes/InferResets.scala
jackkoenig: This now works on High FIRRTL

src/main/scala/firrtl/WIR.scala
jackkoenig: done

src/main/scala/firrtl/passes/InferResets.scala
jackkoenig: done

src/main/scala/firrtl/passes/InferResets.scala
jackkoenig: done

src/main/scala/firrtl/passes/InferResets.scala
jackkoenig: done

src/main/scala/firrtl/Utils.scala
jackkoenig: Done

src/main/scala/firrtl/passes/InferResets.scala
jackkoenig: I've added some notes. The purpose of this type hierarchy is to represent a lookup structure for setting types of leaf-level nodes in nested Types. I created the new type structure to make the code more self-documenting in that the structure of the lookup is pretty explicit.

It could be expressed very similarly using `ir.Type` and then performing some sort of "union" of the two structures. The thing I like about the private type structure is it only contains the necessary information whereas having to create Fields and stuff would require me to fill in lots of values that I don't know, nor care about, and it could be a bit misleading.

jackkoenig:
@azidar I have updated this by adapting some code from ExpandConnects, it now runs between ResolveAndCheck and DedupModules, it should be a fully compliant transform and thus useful to run on its own on any HighFirrtl!

XVilka:
Seems this Travis CI job stuck forever.

seldridge:
@XVilka: not sure what was going on here, but a merge conflict may gate Travis from starting. 

@jackkoenig: I locally merged with master and pushed. The force push was due to one false start... Should be clean now.

jackkoenig:
Done, including support for aggregates

azidar:
LGTM. Congrats @jackkoenig !!!!

PR #1073, opened 2019-04-19
Rename annotations in InlineInstances

albertchen-sifive:
depends on: https://github.com/freechipsproject/firrtl/pull/967

The current `InlineInstances` does not do any renaming of annotations. This PR updates `InlineInstances` to use chained rename maps for renaming of inlined targets.


azidar:
@albertchen-sifive I believe this PR can be updated now?

albertchen-sifive:
yes I'm working on it. and I think I ran into  a bug in my PR while trying to update this PR

`andThen` ignores the `chained` field of the `next` map so `a.andThen(b.andThen(c))` is missing the renames in `c`
https://github.com/freechipsproject/firrtl/blob/master/src/main/scala/firrtl/RenameMap.scala#L43

albertchen-sifive:
@azidar updated, now it depends on https://github.com/freechipsproject/firrtl/pull/1126 for `RenameMap` chaining fixes

PR #1079, opened 2019-04-26
Dependency API

### LINKED ISSUE #948: [RFC] Dependency API

seldridge:
(Re-opening #1015 as Travis is not updating `$TRAVIS_BRANCH` after changing PR base branch...)

This implements a Dependency API for `TransformLike[A]` (hereafter referred to as **transforms**). This adds a `DependencyAPI` trait, an abstract class `DependencyManager`, and a concrete `PhaseManager`. The `DependencyAPI` defines three types of "dependencies" that, by implementing, allow a user to defer determination of a valid ordering to a `DependencyManager`. 

Dependencies of transform _Foo_ are then described using:

  - **prerequisites**: transforms that should run before Foo
  - **dependents**: transforms that should run after Foo (a means of prerequisite injection)
  - **invalidates**: transforms whose work/effects Foo destroys

Prerequisites and dependents are defined as a `Seq[Class[A]]` and internally converted to `LinkedHashSet[Class[A]]`. While a public API that exposes an immutable set is preferable, this is a compromise since no such set exists. Alternatives not used are: `scala.collection.Set`, `scala.collection.mutable.LinkedHashSet`, and `ListSet`.

A `DependencyManager` can then be used to, given a set of transforms you want to run and a set of all transforms that have been run, determine a linear ordering of transforms that satisfy your requirements. This uses a modification of the algorithm proposed by @azidar in https://github.com/freechipsproject/firrtl/issues/446#issuecomment-300573107:

1. A DAG of transforms with *invalidation edges* is constructed (an *invalidates graph*)
2. A DAG of transforms with *prerequisite and dependent edges* is constructed (a *dependents graph*)
3. A toplogical sort of the dependents graph, seeded with the reverse topological sort of the invalidates graph, gives a good ordering of transforms that minimizes re-lowerings
4. This ordering is examined, node by node, cleaning up any mismatches between transforms by solving `DependencyManager` sub-problems

This PR adds the `seededLinearize` method to `DiGraph` that allows for a topological sort seeded with some starting nodes. This is required to implement the above algorithm.

Finally, in preparation for using `DependencyManager` for FIRRTL transform scheduling, this makes `Transform` extend `TransformLike[CircuitState]`. 

META:
- Depends on #1005 
- Fixes #948 
- Related to #446 

## Examples

The examples below show examples of how different combinations of dependencies interact and are resolved. This uses built-in `DependencyManager` methods for generating graphviz output. All examples below are taken from the included `PhaseManagerSpec`. 

The top graph shows dependencies with prerequisites as black arrows, dependents as red arrows, and invalidates with dashed lines. The bottom graph shows the resulting linear ordering (if one exists). Each `DependencyManager` sub-problem is shown as a cluster with it's "target transforms" and the "state" showing current transforms that have run.

### Simple Dependency

Here, `B` depends on `A` and we request to run `B`.

dependencyGraph dot
transformOrder dot

### Simple Invalidate

Here, you have a pretty basic graph with one invalidate: `D` invalidates `A`, but also requires `A`. To get everything to run, you have to solve a sub-problem that reruns `A`.

dependencyGraph dot
transformOrder dot

### Single Dependent (Prerequisite Injection)

This is intended to model the @stevenmburns case of needing to run a custom transform *before* something else. Here, `Second` depends on `First`, but we want to run the custom transform `Custom` which should be inserted between them. By making `Second` a dependent of `Custom` you get the ordering you want.

dependencyGraph dot
transformOrder dot

### Two Invalidates

In this graph, there are only two invalidations. However, this teases out problems with sub-optimal linearizations (corresponding to different possible topological sorts of the dependents graph). 

dependencyGraph dot
transformOrder dot

Here is the resulting linearization without seeding using the invalidation graph.

transformOrder dot

### Complicated Example

The graph sum of the dependents and invalidates graphs form a cycle. This is a test with multiple valid orderings where choosing a bad one will result in a lot of re-lowering.

dependencyGraph dot
transformOrder dot

Similar to the above "Two Invalidates", you wind up with a valid, but worse solution without the seeded topological sort.

transformOrder dot

### Chained Invalidation

In this graph, you start with a simple initial graph with some weird properties where your invalidations start invalidating other needed transforms. This graph has these odd properties:

- `A` invalidates `B`
- `B` invalidates `C`
- `C` invalidates `D`
- `E` depends on `[A, B, C, D]`
- You start in state `[B, C, D]` wanting to run `[A, E]`

This tests that "chained invalidation" is getting properly handled.

dependencyGraph dot
transformOrder dot

### Cyclic Prerequisite

A cycle in the prerequisistes will throw an exception.

dependencyGraph dot

### Cyclic Invalidates

A cycle in the invalidates will throw an exception.

dependencyGraph dot

### Repeated Analyses

This covers an intended use case where some "analysis" phase is repeatedly invalidated by some "consumer" phases. Concretely an analysis phase would be a phase that produces information that a consumer phase wants to use. However, each consumer phase makes modifications which require the analysis phase to be re-run (e.g., an analysis phase that constructs an `InstanceGraph` and consumer phases that rename modules).

dependencyGraph dot
transformOrder dot

### Inverted Repeated Analyses

This is equivalent to the repeated analyses above, but the consumer phases invalidate each other.

dependencyGraph dot
transformOrder dot

### Deterministic Order

This relatively big example asserts that a deterministic order is preserved for all graphs constructed inside a `DependencyManager`.

dependencyGraph dot
transformOrder dot

### DIFF
--- a/src/main/scala/firrtl/Compiler.scala
+++ b/src/main/scala/firrtl/Compiler.scala
@@ -11,7 +11,7 @@ import firrtl.annotations._
 import firrtl.ir.Circuit
 import firrtl.Utils.throwInternalError
 import firrtl.annotations.transforms.{EliminateTargetPaths, ResolvePaths}
-import firrtl.options.StageUtils
+import firrtl.options.{StageUtils, TransformLike}
 
 /** Container of all annotations for a Firrtl compiler */
 class AnnotationSeq private (private[firrtl] val underlying: List[Annotation]) {
@@ -169,7 +169,7 @@ final case object UnknownForm extends CircuitForm(-1) {
 // scalastyle:on magic.number
 
 /** The basic unit of operating on a Firrtl AST */
-abstract class Transform extends LazyLogging {
+abstract class Transform extends TransformLike[CircuitState] {
   /** A convenience function useful for debugging and error messages */
   def name: String = this.getClass.getSimpleName
   /** The [[firrtl.CircuitForm]] that this transform requires to operate on */
@@ -184,6 +184,8 @@ abstract class Transform extends LazyLogging {
     * @return A transformed Firrtl AST
     */
   protected def execute(state: CircuitState): CircuitState
+
+  def transform(state: CircuitState): CircuitState = execute(state)
 
   /** Convenience method to get annotations relevant to this Transform
     *
--- a/src/main/scala/firrtl/graph/DiGraph.scala
+++ b/src/main/scala/firrtl/graph/DiGraph.scala
@@ -64,13 +64,14 @@ class DiGraph[T] private[graph] (private[graph] val edges: LinkedHashMap[T, Link
     */
   def findSinks: Set[T] = reverse.findSources
 
-  /** Linearizes (topologically sorts) a DAG
-    *
+  /** Linearizes (topologically sorts) a DAG using a DFS. This can be seeded with an order to use for the DFS if the user
+    * wants to tease out a special ordering of the DAG.
+    * @param seed an optional sequence of vertices to use. This will default to the vertices ordering provided by getVertices.
     * @throws CyclicException if the graph is cyclic
     * @return a Map[T,T] from each visited node to its predecessor in the
     * traversal
     */
-  def linearize: Seq[T] = {
+  def seededLinearize(seed: Option[Seq[T]] = None): Seq[T] = {
     // permanently marked nodes are implicitly held in order
     val order = new mutable.ArrayBuffer[T]
     // invariant: no intersection between unmarked and tempMarked
@@ -80,7 +81,7 @@ class DiGraph[T] private[graph] (private[graph] val edges: LinkedHashMap[T, Link
     case class LinearizeFrame[T](v: T, expanded: Boolean)
     val callStack = mutable.Stack[LinearizeFrame[T]]()
 
-    unmarked ++= getVertices
+    unmarked ++= seed.getOrElse(getVertices)
     while (unmarked.nonEmpty) {
       callStack.push(LinearizeFrame(unmarked.head, false))
       while (callStack.nonEmpty) {
@@ -132,6 +133,14 @@ class DiGraph[T] private[graph] (private[graph] val edges: LinkedHashMap[T, Link
     }
     foundPath
   }
+
+  /** Linearizes (topologically sorts) a DAG
+    *
+    * @throws CyclicException if the graph is cyclic
+    * @return a Map[T,T] from each visited node to its predecessor in the
+    * traversal
+    */
+  def linearize: Seq[T] = seededLinearize(None)
 
   /** Performs breadth-first search on the directed graph
     *
@@ -187,7 +196,7 @@ class DiGraph[T] private[graph] (private[graph] val edges: LinkedHashMap[T, Link
     * @return a Seq[T] of nodes defining an arbitrary valid path
     */
   def path(start: T, end: T): Seq[T] = path(start, end, Set.empty[T])
-  
+
   /** Finds a path (if one exists) from one node to another, with a blacklist
     *
     * @param start the start node
--- a/src/main/scala/firrtl/options/DependencyManager.scala
+++ b/src/main/scala/firrtl/options/DependencyManager.scala
@@ -1,0 +1,363 @@
+// See LICENSE for license details.
+
+package firrtl.options
+
+import firrtl.AnnotationSeq
+import firrtl.graph.{DiGraph, CyclicException}
+
+import scala.collection.Set
+import scala.collection.immutable.{Set => ISet}
+import scala.collection.mutable.{ArrayBuffer, HashMap, LinkedHashMap, LinkedHashSet, Queue}
+
+/** An exception arising from an in a [[DependencyManager]] */
+case class DependencyManagerException(message: String, cause: Throwable = null) extends RuntimeException(message, cause)
+
+/** A [[firrtl.options.TransformLike TransformLike]] that resolves a linear ordering of dependencies based on
+  * requirements.
+  * @tparam A the type over which this transforms
+  * @tparam B the type of the [[firrtl.options.TransformLike TransformLike]]
+  */
+trait DependencyManager[A, B <: TransformLike[A] with DependencyAPI[B]] extends TransformLike[A] with DependencyAPI[B] {
+
+  /** Requested [[firrtl.options.TransformLike TransformLike]]s that should be run. Internally, this will be converted to
+    * a set based on the ordering defined here.
+    */
+  def targets: Seq[Class[B]]
+  private lazy val _targets: LinkedHashSet[Class[B]] = targets
+    .foldLeft(new LinkedHashSet[Class[B]]()){ case (a, b) => a += b }
+
+  /** A sequence of [[firrtl.Transform]]s that have been run. Internally, this will be converted to an ordered set.
+    */
+  def currentState: Seq[Class[B]]
+  private lazy val _currentState: LinkedHashSet[Class[B]] = currentState
+    .foldLeft(new LinkedHashSet[Class[B]]()){ case (a, b) => a += b }
+
+  /** Existing transform objects that have already been constructed */
+  def knownObjects: Set[B]
+
+  /** A sequence of wrappers to apply to the resulting [[firrtl.options.TransformLike TransformLike]] sequence. This can
+    * be used to, e.g., add automated pre-processing and post-processing.
+    */
+  def wrappers: Seq[(B) => B] = Seq.empty
+
+  /** Store of conversions between classes and objects. Objects that do not exist in the map will be lazily constructed.
+    */
+  protected lazy val classToObject: LinkedHashMap[Class[B], B] = {
+    val init = LinkedHashMap[Class[B], B](knownObjects.map(x => x.getClass.asInstanceOf[Class[B]] -> x).toSeq: _*)
+    (_targets ++ _currentState)
+      .filter(!init.contains(_))
+      .map(x => init(x) = safeConstruct(x))
+    init
+  }
+
+  /** A method that will create a copy of this [[firrtl.options.DependencyManager DependencyManager]], but with different
+    * requirements. This is used to solve sub-problems arising from invalidations.
+    */
+  protected def copy(
+    targets: Seq[Class[B]],
+    currentState: Seq[Class[B]],
+    knownObjects: ISet[B] = classToObject.values.toSet): B
+
+  /** Implicit conversion from Class[B] to B */
+  private implicit def cToO(c: Class[B]): B = classToObject.getOrElseUpdate(c, safeConstruct(c))
+
+  /** Implicit conversion from B to Class[B] */
+  private implicit def oToC(b: B): Class[B] = b.getClass.asInstanceOf[Class[B]]
+
+  /** Modified breadth-first search that supports multiple starting nodes and a custom extractor that can be used to
+    * generate/filter the edges to explore. Additionally, this will include edges to previously discovered nodes.
+    */
+  private def bfs( start: LinkedHashSet[Class[B]],
+                   blacklist: LinkedHashSet[Class[B]],
+                   extractor: B => Set[Class[B]] ): LinkedHashMap[B, LinkedHashSet[B]] = {
+
+    val (queue, edges) = {
+      val a: Queue[Class[B]]                    = Queue(start.toSeq:_*)
+      val b: LinkedHashMap[B, LinkedHashSet[B]] = LinkedHashMap[B, LinkedHashSet[B]](
+        start.map((cToO(_) -> LinkedHashSet[B]())).toSeq:_*)
+      (a, b)
+    }
+
+    while (queue.nonEmpty) {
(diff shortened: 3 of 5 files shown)


azidar:
It would be good to get @jackkoenig approval to ensure his workload works.

seldridge:
I'm hitting issues with determinism due to use of `Set`. @jackkoenig suggestion is to use `Seq` and then convert internally.

seldridge:
Several updates in an omnibus commit that should likely be broken up:

- A new `firrtl.stage.Forms` object is added that defines the old lowering compiler forms in terms of sets of transforms/passes that should be run. A cool thing to fall out of this is that higher forms are subsets of lower forms. You can then define an ordering that mirrors the old `CircuitForm.value` concept.
- The explicit transform/pass order in each lowering compiler is removed and replaced with a transform manager that solves a transform ordering sub-problem between two `firrtl.stage.Forms`.
- Dependencies are defined for all passes/transforms (used by the lowering compilers)
- A `LoweringCompilersSpec` is added that checks that the new transform ordering mostly matches the old. The "mostly" comes from additional checks being added.
- This fixes a few bugs in the way a dependency manager ordered things.

One interesting thing to fall out of this. There are a lot of transforms that you could deem optional: `CheckTypes`, `ConstantPropagation`, etc. If you define the dependencies of optional transforms using `dependents` as opposed to `prerequisites`, you can construct forms where they are used or not used. Consider the situation of `MinimumLowFirrtlOptimization` vs. `LowFirrtlOptimization`. Roughly, low firrtl optimization is the minimum optimizations with constant propagation and some other transforms mixed in. If you define the correct invalidates for transforms in min low and the correct dependents in constant propagation, you can just "mix in" constant propagation and everything works. I need to redefine all the checks this way, but this seems incredibly useful for composing superset forms from subset forms.

Also, the use of sets for defining forms gives you a nice ordering (using `_.subsetOf(_)`). This effectively mirrors the old `CircuitForm.value`. 

Stability problems of `Set` still not fixed...

seldridge:
I updated this to clear a bug that Chisel tests exposed and move to a `Seq` for dependency specification. @jackkoenig had suggested using a `ListSet` for the latter instead. I started to migrate to this, but I think the reverse insertion order of `ListSet` may make things somewhat confusing for a user (which way are they expecting things to run?). As a middle ground, I opted to include a note in the scaladoc on methods that take a `Seq` to indicate the reason for this and that it'll be converted to an  ordered `Set` internally.

@jackkoenig: once things calm down for you, can you see if this works on SiFive workloads?

seldridge:
I removed all code related to moving FIRRTL to use the Dependency API. This is a lot of additional code that makes the content of this PR get lost. I'm moving the port into a separate PR.

This PR is then a pure API addition that adds the `DependencyAPI` trait, defines `DependencyManager` and `PhaseManager`, and adds tests.

sequencer:
I have a question that, will `Phase` finally deprecate `Transform`?
Since all `Transform` is modifying the
```
CircuitState(
    circuit: Circuit,
    form: CircuitForm,
    annotations: AnnotationSeq,
    renames: Option[RenameMap])
```
which is a wrapper of `AnnotationSeq` for `Compiler`.
Since `Phase` is a `TransformLike[AnnotationSeq]`, in order to decrease redundancy of `Phase`, will `Compiler` and `Transform` deprecate in the future, so we can easily manage dependency between `Phase` and `Transform`.

seldridge:
@sequencer: That's a good question... At this point, I have not considered deprecating `Transform` in favor of `Phase`, nor has this come up in dev meetings. It could, however, be something worth discussing long term (FIRRTL 1.4). Part of the motivation of something like that is, as you mention, being able to define one `DependencyManager[A]` and have that deal with all dependencies as opposed to having to interface different `TransformLike`s together.

To elucidate the current approach ([FIRRTL#1105](https://github.com/freechipsproject/firrtl/pull/1105)) which uses a `DependencyManager[A]` (`firrtl.stage.FirrtlStage`) which internally calls a `Translator[A, B]` (`firrtl.stage.phases.Compiler`) which internally calls a `DependencyManager[B]` (`firrtl.stage.transforms.Compiler`). You can imply dependencies in this way, but it's not explicit. Maybe it's worth thinking about how we could do this...

Nevertheless, a full deprecation of `Transform` would cause a **ton** of pain... some of which I'm aware of:

- We'd need to have complete backwards compatibility for all existing third-party transforms. This likely wouldn't be too bad as you can always provide backwards compatibility for `Transform` in a `Translator[CircuitState, AnnotationSeq]`.
- The `AnnotationSeq` abstraction may not always be the best container, especially if there's something that better encapsulates the type over which you're doing a mathematical transformation. While I'm possibly motivated in ramming everything into this, it may not always be the best fit.

`AnnotationSeq` is really just a `Monad`-like pattern. While we can make this the de-facto standard, there's likely some benefit in letting users use the abstraction that makes sense for their compiler internals. E.g., abstractly, a FIRRTL transform is most natively expressed as `(Circuit) -> Circuit` and not `(AnnotationSeq) -> AnnotationSeq` even though it could be expressed that way. It's hard to know where to draw the line...

In summary, good thoughts, I'm super interested in this type of discussion, and am open to ideas. Stage/Phase and Dependency API are really about *trying to build **hardware compiler infrastructure*** and it's worth putting a lot of thought into this.

sequencer:
Thank you for your replying. Really appreciate your answer!
In current status, it's seems easy to manage the dependency among different `Transform`, that help me a lot!
BTW, will you accept a `CustomPhaseAnotation` PR for `firrtl`? After `Dependency API` finally merged, I think it's easy to add custom `Phase` to do some additional `os` works. For example, emiting `hammer` config or directly calling the EDA like hammer does.

seldridge:
Something like `CustomPhaseAnnotation` likely makes sense as does the motivation you state. It would need to have an API that includes specification of in which `Stage` you want it to run since, unlike with `Transform`s which are only associated with FIRRTL, a `Phase` could run in a lot of places (Chisel, FIRRTL, testers, etc.). I think this also means that a `Stage` needs to be a `DependencyManager` or be a very thin wrapper around one. Likely worth a separate issue or maybe on RFC.

Something like:
```scala
RunPhaseAnnotation(stage: Class[Stage], phase: Class[Phase])`
```

seldridge:
After some confirmation from @jackkoenig that this passes SiFive workloads, this is now rebased and merged.

PR #1080, opened 2019-04-27
add CustomInfo for custom transform emitted IR.

sequencer:
currently, if user wanna custom a info in transform, only to extends from `FileInfo` and override `toString`, maybe CustomInfo is need to get rid of this hack.


azidar:
Can you give an example of what you would use custom info for, that you couldn't use annotations for?

sequencer:
Sorry for the late reply. I'd like to use this Info to mark the Transform information if I add some additional information in some transforms and then try to locate it in the generated firrtl.
I'll try to finish this PR tonight.

sequencer:
@chick ready for review.

chick:
I think the code looks clean but we aren't clear on is how is what this PR does not supported with Annotations. We are very cautious about adding new ways of transferring information through the firrtl stack. Can you please describe a little more completely what you would use this for, particularly with respect to how Annotations do not help your use case.

sequencer:
Sorry for the late reply, I'm cleaning my PR these days.
I think this PR is mainly for firrtl Transform designer to annotate their own comments on the transformed IRs for the final low firrtl delivery.
Different from the debug function, it adds the `SourceInfo` which can be costumed by firrtl Transform designer for a better convenience.

seldridge:
My understanding is that `Info` is really only for file locator information. I had thought these were "FIRRTL comments", but the intent was not that. Any other information is intended to be stuck in an annotation.

Echoing @chick here: having transforms consume `Info` in order to do work is a potentially brittle anti-pattern. The real problem with having `Info` be consumed by transforms is that there are zero guarantees that the info is preserved by transformations. Annotations, in contrast, will be preserved through renaming. 

Put differently, stripping info before passing a circuit into the FIRRTL compiler should produce the same circuit as stripping info from the compiler output.

sequencer:
Got it! I agree that comment Annotation is much better than this!

PR #1081, opened 2019-04-29
Update NoCircuitDedupAnnotation so it's available from firrtl.stage.FirrtlMain

ucbjrl:


seldridge:
I just realized that this should be added to `firrtl.stage.FirrtlCli`]

PR #1082, opened 2019-05-09
Bugfix: GroupComponents

azidar:
Bugfix for GroupComponents transform.

When grouping instances with invalid input ports, buggy behavior leaves the invalid statements in the parent module, instead of moving them to the new grouped module.

Includes test and bugfix.


colinschmidt:
You seem to have only included the new test.

azidar:
Yeah, I wanted CI to show it failing, before I pushed the fix.

azidar:
I can not do that if desired, though. LMK

colinschmidt:
Doesn't matter to me, just thought it was unintentional.

seldridge:
The only time this would matter and/or we should hold it up is if you want to merge via merge commit. (Rebasing the fix before the test before merging would fix it.) I'm usually the only one going nuts with the merge commit flow, though...

Issue #968, opened 2018-12-14
Spec says cat(SInt, UInt) is valid, but compiler rejects it


aswaterman:
Currently, the Chisel concatenation operator `##` accepts any combination of UInt and SInt.  This gets translated directly to Firrtl `cat`.  The Firrtl spec claims this is valid, but the implementation issues an error when the types don't match. It does accept both UInt/UInt or SInt/SInt, and in both cases returns UInt, as the spec says it should.

@jackkoenig tells me he thinks the implementation is correct and the spec is outdated.  If that is indeed the case, we should change Chisel's definition of `##` to `asUInt` its arguments.

aswaterman:
BTW, my take is that cat should accept all combinations (and continue to return UInt unconditionally).  This doesn't add much complexity, because there's no special handling for UInt vs. SInt in this operator.

seldridge:
Tentative discussed solution: 
  - Fix the spec to disallow this
  - Update Chisel to emit casts

albert-magyar:
Closed by #1085

PR #1085, opened 2019-05-20
spec: mixed-input arguments for prim ops are no longer allowed

ekiwi:
This updates the spec to refelect the changes made in #587.
It also fixes issue #968.

### DIFF
--- a/spec/spec.tex
+++ b/spec/spec.tex
@@ -1178,8 +1178,6 @@ Notationally, the width of an argument \vv{e} is represented as \vv{w}\ts{e}.
 \begin{tabular}{ |c|c|c|c|c|c| }   
   \opheader 
 add & (e1,e2) & () & (UInt,UInt) & UInt & max(w\ts{e1},w\ts{e2})+1\\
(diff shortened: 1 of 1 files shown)


ekiwi:
Let me know if I should also check in the rebuilt pdf.

edwardcwang:
I wonder if we should just remove the cached copy of the spec to prevent it from becoming stale/out-of-date if we don't plan to update it regularly.

ekiwi:
> I wonder if we should just remove the cached copy of the spec to prevent it from becoming stale/out-of-date if we don't plan to update it regularly.

That sounds like a good idea. Maybe generate the pdf in the CI.

aswaterman:
That would be better, for a number of reasons.  You can use the RISC-V spec as an example of how to do it: https://github.com/riscv/riscv-isa-manual/blob/master/.travis.yml (note you need to generate a github access token with write perms to this repo, then go to the Travis setup for this repo, and define a variable named github_access_token with the value of your token).

PR #1087, opened 2019-05-24
Emit legal Verilog literals for ExtModule IntParams > 32-bit

jackkoenig:
As the title states, we were emitting illegal literals which is a bit of a drag.

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -205,6 +205,9 @@ class VerilogEmitter extends SeqTransform with Emitter {
   /** Turn Params into Verilog Strings */
   def stringify(param: Param): String = param match {
     case IntParam(name, value) => s".$name($value)"
+      val blen = value.bitLength
+      val lit = if (value > 0) s"$blen'd$value" else s"-${blen+1}'sd${value.abs}"
+      s".$name($lit)"
     case DoubleParam(name, value) => s".$name($value)"
     case StringParam(name, value) => s".${name}(${value.verilogEscape})"
     case RawStringParam(name, value) => s".$name($value)"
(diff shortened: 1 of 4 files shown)


src/main/scala/firrtl/Emitter.scala
```diff
@@ -205,6 +205,9 @@ class VerilogEmitter extends SeqTransform with Emitter {
   /** Turn Params into Verilog Strings */
   def stringify(param: Param): String = param match {
     case IntParam(name, value) => s".$name($value)"
+      val blen = value.bitLength
+      val lit = if (value > 0) s"$blen'd$value" else s"-${blen+1}'sd${value.abs}"
```
aswaterman: Is this right when value is 0? (Or is that impossible?)

jackkoenig: Good catch, now it does work (at least using Verilator below), because this emits as follows for the test below:
```verilog
LargeParam #(.WIDTH(1'd1), .DATA(-1'sd0))
```
That being said, it's a little weird. Maybe should just handle zero explicitly to avoid confusion.

jackkoenig:
It looks like this doesn't work for some circumstances, debugging...

jackkoenig:
Okay I updated it, now we only emit `IntParams` with width modifiers if it would not fit in a 32-bit integer literal. The test included in this PR emits as:
```verilog
  LargeParam #(.WIDTH(1), .DATA(0)) mod1 (
    .out(mod1_out)
  );
  LargeParam #(.WIDTH(128), .DATA(73'd9223372036854775807000)) mod2 (
    .out(mod2_out)
  );
  LargeParam #(.WIDTH(128), .DATA(-74'sd9223372036854775807000)) mod3 (
    .out(mod3_out)
  );
```

PR #1091, opened 2019-06-01
Fixed Mill SourcePath

colin4124:
its source directory is in the top level directory, not `firrtl/src/main/scala` .

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -43,9 +43,9 @@ trait CommonModule extends CrossUnRootedSbtModule with PublishModule {
 
 // Generic antlr4 configuration.
 // This could be simpler, but I'm trying to keep some compatibility with the sbt plugin.
-case class Antlr4Config(val sourcePath: Path) {
+case class Antlr4Config(val sourcePath: Path, val antlr4Path: Path) {
   val antlr4Version: String = "4.7.1"
-  val ANTLR4_JAR = (pwd / s"antlr-${antlr4Version}-complete.jar").toString
+  val ANTLR4_JAR = ( antlr4Path / s"antlr-${antlr4Version}-complete.jar").toString
   if (!new java.io.File(ANTLR4_JAR).exists) {
     println("Downloading ANTLR4 .jar")
     %%("wget", s"https://www.antlr.org/download/antlr-${antlr4Version}-complete.jar", "-O", ANTLR4_JAR)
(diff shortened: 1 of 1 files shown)


edwardcwang:
Just to double check, how are you currently using this as a mill submodule?

colin4124:
@edwardcwang Sorry, I don't understand the "as a mill submodule" what it means. I use this firrtl repo as what it used to be. In other words, I use `mill` in this way, I clone firrtl repo and run 

```shell
mill firrtl.compile
```

I don't have to change any directory structure. I copy this code from chisel3's build.sc https://github.com/freechipsproject/chisel3/blob/master/build.sc#L143

edwardcwang:
I have a project using mill that has firrtl as a submodule, and importing it in mill as a "foreign module" (mill terminology): https://github.com/lihaoyi/mill/blob/master/docs/pages/5%20-%20Modules.md#foreign-modules

colin4124:
@edwardcwang I found the error: **Error: Invalid or corrupt jarfile  ...antlr-4.7.1-complete.jar**

Here are my project hierarchic：

```
foo/
    build.sc
    firrtl/
    bar/
        build.sc
```

**foo/build.sc**
```scala
import mill._, scalalib._

import $file.firrtl.build

object foo extends ScalaModule{
  def compile = T{
    firrtl.build.firrtl.compile()
  }
}
```

**foo/bar/build.sc**
```scala
import mill._, scalalib._

import $file.^.firrtl.build

object bar extends ScalaModule{

  def compile = T{
    ^.firrtl.build.firrtl.compile()
  }
}
```

Because of **pwd** in `val ANTLR4_JAR = (pwd / s"antlr-${antlr4Version}-complete.jar").toString`:

-  if I run `mill foo.compile` in `foo` directory,  `ANTLR4_JAR` would be `foo/antlr-4.7.1-complete.jar`; 
- if I run `mill bar.compile` in `bar` directory, `ANTLR4_JAR` would be `foo/bar/antlr-4.7.1-complete.jar`; 
- If I run `mill firrtl.compile` in `firrtl` directory, `ANTLR4_JAR` would be `foo/firrtl/antlr-4.7.1-complete.jar`.

The `ANTLR4_JAR` should be `foo/firrtl/antlr-4.7.1-complete.jar` whatever I run the compile command in any directory.

edwardcwang:
Great! I will try this within the next few days. Thanks Colin!

azidar:
Any update on this, @edwardcwang ?

edwardcwang:
Sorry for falling behind on this. I will aim to try it out either this weekend or hopefully early next week...

sequencer:
This works fine, just tested with `mill mill.scalalib.GenIdea/idea` and `mill firrtl.compile`.

edwardcwang:
The issue is that we also want it to work in "subproject configuration"...

azidar:
So sounds like this still requires work, @edwardcwang ? perhaps you can clarify?

edwardcwang:
I think #1368 may obsolete this. Will test with #1368

colin4124:
Ok, so I close it?

PR #1098, opened 2019-06-18
Travis Verilator Issues

seldridge:
Testing PR to test what's going on with Travis.

Currently, all builds are failing because Travis' cache does not appear to be getting properly populated.

See the nightly CRON job:
  - https://travis-ci.org/freechipsproject/firrtl/builds/547060586

In the prepare build stage, [Verilator and Yosys are getting installed](https://travis-ci.org/freechipsproject/firrtl/jobs/547060590#L3477). However, the cache is then not found in any of the test stages. See: https://travis-ci.org/freechipsproject/firrtl/jobs/547060593#L440

PR #1074, opened 2019-04-19
Use scalafix to remove unused import and deprecated procedure syntax

colin4124:
I add scalafix and then run 

```shell
sbt "firrtl/scalafix ProcedureSyntax"
```
and
```shell
sbt "firrtl/scalafix RemoveUnused"
```

PR #1099, opened 2019-06-18
Try explicitly clearing the cache directories to re-enable cacheing.

ucbjrl:
See #858, #1098


ucbjrl:
This appears to have done the trick. #1074 is through two test stages that require a valid cache.

PR #1101, opened 2019-06-19
Fix typo

colin4124:


src/main/antlr4/FIRRTL.g4
```diff
@@ -90,7 +85,7 @@ simple_reset
 
 reset_block
 	: INDENT simple_reset info? NEWLINE DEDENT
-	| '(' +  simple_reset + ')'
+	| '(' simple_reset+ ')'
```
jackkoenig:
```suggestion
	| '(' simple_reset ')'
```
I think both `+` were actually mistakes, at the moment, this does fix the issue where 
`reg r : UInt<8>, clock with : (((((((((reset => (reset, UInt(0)))` would be accepted

But as is, this still accepts
`reg r : UInt<8>, clock with : (reset => (reset, UInt(0)) reset => (reset, UInt(0)))` which is also wrong.

colin4124:
@jackkoenig I have changed as your request, is there still any problem?

jackkoenig:
My apologies, looks great!

PR #1106, opened 2019-06-27
Add Test for AddDefaults phase

colin4124:


src/test/scala/firrtlTests/options/phases/AddDefaultsSpec.scala
```diff
@@ -0,0 +1,38 @@
+// See LICENSE for license details.
+
+package firrtlTests.options.phases
+
+import org.scalatest.{FlatSpec, Matchers}
+
+import firrtl.AnnotationSeq
```
seldridge:
```suggestion
```
This isn't used, so just pull it out.

src/test/scala/firrtlTests/options/phases/AddDefaultsSpec.scala
```diff
@@ -0,0 +1,38 @@
...
+import firrtl.options.{OptionsException, Phase, TargetDirAnnotation}
```
seldridge:
```suggestion
import firrtl.options.{Phase, TargetDirAnnotation}
```

src/test/scala/firrtlTests/options/phases/AddDefaultsSpec.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  it should "add a TargetDirAnnotation if it does not exist" in new Fixture {
+    val Occurrences = phase.transform(Seq.empty) count {
+      a => a match {
+        case _: TargetDirAnnotation => true
+        case _ => false
+      }
+    }
+    assert(Occurrences == 1)
+  }
```
seldridge: For tests, I think it's good to avoid a blind `assert`. I would prefer if you compare against the actual produced sequence, e.g.,:

```scala
phase.transform(Seq.empty).toSeq should be (Seq(targetDir))
```

src/test/scala/firrtlTests/options/phases/AddDefaultsSpec.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  it should "don't add a TargetDirAnnotation if it exists" in new Fixture {
+    val Occurrences = phase.transform(Seq(targetDir)) count {
+      a => a match {
+        case _: TargetDirAnnotation => true
+        case _ => false
+      }
+    }
+    assert(Occurrences == 1)
+  }
```
seldridge: Same modification as above:

```scala
phase.transform(Seq(targetDir)).toSeq should be (Seq(targetDir))
```

src/test/scala/firrtlTests/options/phases/AddDefaultsSpec.scala
```diff
@@ -0,0 +1,38 @@
...
+
+import firrtl.AnnotationSeq
+import firrtl.options.{OptionsException, Phase, TargetDirAnnotation}
+import firrtl.options.phases.AddDefaults
+
+class AddDefaultsSpec extends FlatSpec with Matchers {
+
+  val targetDir = TargetDirAnnotation("foo")
+
+  class Fixture { val phase: Phase = new AddDefaults }
```
seldridge: It probably makes sense to put the `targetDir` inside the fixture as well.

colin4124:
@seldridge Thanks for your suggestion ^_^

PR #1109, opened 2019-07-01
Implement MultiTargetAnnotation

sequencer:
I'm writing a physical design annotation library, which needs `MultiTargetAnnotation` for multi signal annotation. Since there are no historic issues, `MultiTargetAnnotation` defines targets as `Seq[Target]`.

### DIFF
--- a/src/main/scala/firrtl/annotations/Annotation.scala
+++ b/src/main/scala/firrtl/annotations/Annotation.scala
@@ -77,6 +77,28 @@ trait SingleTargetAnnotation[T <: Named] extends Annotation {
   }
 }
 
+trait MultiTargetAnnotation extends Annotation {
+  val targets: Seq[Target]
+
+  def duplicate(n: Seq[Target]): Annotation
+
+  private def crossJoin[T](list: Seq[Seq[T]]): Seq[Seq[T]] =
+    list match {
+      case Nil => Nil
+      case x :: Nil => x map (Seq(_))
+      case x :: xs =>
+        val xsJoin = crossJoin(xs)
+        for {
+          i <- x
+          j <- xsJoin
+        } yield {
+          Seq(i) ++ j
+        }
+    }
+
+  def update(renames: RenameMap): Seq[Annotation] = crossJoin(targets.map(renames(_))).map(ts => duplicate(ts.asInstanceOf[Seq[Target]]))
+}
+
 @deprecated("Just extend NoTargetAnnotation", "1.1")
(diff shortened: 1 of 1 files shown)


src/main/scala/firrtl/annotations/Annotation.scala
```diff
@@ -77,6 +77,28 @@ trait SingleTargetAnnotation[T <: Named] extends Annotation {
   }
 }
 
+trait MultiTargetAnnotation extends Annotation {
```
azidar: This needs ScalaDoc describing its behavior.

In addition, what should the desired behavior be if the two targets are renamed to a different number of new targets? I wonder if there is a better way to represent this, perhaps as a Seq[Seq[Target]], where each target is expanded (flatMap) within its container Seq?

E.g. Seq(Seq(a), Seq(b)) -> Seq(Seq(a1, a2), Seq(b1, b2, b3))

Food for thought. I agree though that we should certainly add something like this - I've found myself duplicating this code over and over again....

sequencer: I try to folded `Seq[Seq[Target]]` to `Seq[Target]` is aimed to give a same behavior with `SingleTargetAnnotation`. If you permit `Seq[Seq[Target]]` would be much better.
 I think `crossJoin` is also a useful function for the multi target folding in `RenameMap`, so do mind I keeping them as a private function with detail documentation?

azidar:
This LGTM now, @jackkoenig any thoughts before I merge?

azidar:
Also, yeah I think crossJoin is a good idea, feel free to add it as a private function with documentation.

sequencer:
done.

src/main/scala/firrtl/annotations/Annotation.scala
```diff
@@ -77,6 +77,48 @@ trait SingleTargetAnnotation[T <: Named] extends Annotation {
   }
 }
 
+/** Annotate Multi Target in a single Annotation,
+  * rather than [[SingleTargetAnnotation]] which duplicate the [[Annotation]] based on the [[RenameMap]],
+  * [[MultiTargetAnnotation]] keeps a [[RenameMap]] in the targets for better data structure
```
azidar: Reword to: "[[MultiTargetAnnotation]] keeps the renamed targets grouped within a single annotation."

sequencer: Sorry for my redundant documentations.

PR #1110, opened 2019-07-02
Make sure directory exist before writing

colin4124:
If you run `AddCircuitSpec` test only before other testcases created the directory it needed (as below), it would throw `java.io.FileNotFoundException`.

https://github.com/freechipsproject/firrtl/blob/bdd6bf92e0a1a256856005f7acada3555cf28e79/src/test/scala/firrtlTests/stage/FirrtlMainSpec.scala#L120-L122


jackkoenig:
Thanks!

PR #1111, opened 2019-07-02
Add SimplifyMems transform to lower memories without splitting

albert-magyar:
This makes a wire "adapter" instead of relying on the I/O of a blackboxed submodule or splitting the memories.

@jackkoenig, I think I mentioned this to you and you expressed some interest?

Alternatively, one could get vaguely similar output by running `ReplaceMemMacros` to blackbox memories and then later inlining those modules, but it would likely produce more obfuscated output.

### DIFF
--- a/src/main/scala/firrtl/transforms/SimplifyMems.scala
+++ b/src/main/scala/firrtl/transforms/SimplifyMems.scala
@@ -1,0 +1,83 @@
+// See LICENSE for license details.
+
+package firrtl
+package transforms
+
+import firrtl.ir._
+import firrtl.Mappers._
+import firrtl.annotations._
+import firrtl.passes._
+import firrtl.passes.memlib._
+import scala.collection.mutable
+
+import Utils._
+import AnalysisUtils._
+import MemPortUtils._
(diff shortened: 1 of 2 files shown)


src/main/scala/firrtl/transforms/SimplifyMems.scala
```diff
@@ -0,0 +1,83 @@
...
+      case s => s.map(onStmt).map(onExpr)
+    }
+
+    m.map(onStmt)
+  }
+
+  override def execute(state: CircuitState): CircuitState = {
+    val c = state.circuit
+    val renames = RenameMap()
+    CircuitState(c.map(onModule(c, renames)(_)), outputForm, state.annotations, Some(renames))
```
azidar: This is failing to build - I think you need to replace `onModule(c, renames)(_)` with `onModule(c, renames)`

albert-magyar: Thanks for the heads up! That anonymous function was inferring correctly on my machine, but it was also useless :)

PR #1118, opened 2019-07-05
Remove some warnings

colin4124:


colin4124:
I disable deprecated warnings by remove `-deprecation` from scalacOptions, and add `-Xlint` to scalacOptions in `build.sbt` as below:

```scala
scalacOptions := scalacOptionsVersion(scalaVersion.value) ++ Seq(
   "-Yrangepos",          // required by SemanticDB compiler plugin
   "-Xlint",
   "-Ywarn-unused-import" // required by `RemoveUnused` rule
 )
```

And then I run `sbt test` to find warnings, here is I have fixed.

## 1. possible missing interpolator: detected interpolated identifier

Here miss `s` prefix so that `$port` can not evaluate in `"... Port $port"`.

https://github.com/freechipsproject/firrtl/blob/648dddeacd9aece4a43cad09430dad25cba69457/src/main/scala/firrtl/Emitter.scala#L676

## 2. Adapting argument list by creating a N-tuple

The type of this **Warning** is so tricky, I take some examples to explain.

### a. foldLeft follow with tuple

Here the tuple `foldLeft (offset, Seq[Statement]())` want to be a parameter for initialization, but in the view of `foldLeft` , the parenthesis is not for tuple but for two arguments as `foldLeft( offset, Seq[Statement]() )` . So we have to explicitly add parenthesis for arguments `foldLeft( (offset, Seq[Statement]()) )`. 
https://github.com/freechipsproject/firrtl/blob/648dddeacd9aece4a43cad09430dad25cba69457/src/main/scala/firrtl/Utils.scala#L117

### b. Operator follow with tuple

Here the tuple `(f1.tpe, f1.flip, ilen)` is one argument for `->` operator, but in the view of `->`, the parenthesis is not for tuple but for three arguments as `->( f1.tpe, f1.flip, ilen )`. So we also have to explicitly add parenthesis for arguments `->( (f1.tpe, f1.flip, ilen) )`.

https://github.com/freechipsproject/firrtl/blob/648dddeacd9aece4a43cad09430dad25cba69457/src/main/scala/firrtl/Utils.scala#L457

The same as `:+` operator.

https://github.com/freechipsproject/firrtl/blob/648dddeacd9aece4a43cad09430dad25cba69457/src/main/scala/firrtl/annotations/Target.scala#L654

## 3. type parameter T defined in inner level shadows type T in outer level

This is different from I have found in https://github.com/freechipsproject/firrtl/pull/1108 .

`case class LinearizeFrame[T](v: T, expanded: Boolean)` defined in `class DiGraph[T]` so it shadows `T` in `class DiGraph[T]`. And then `LinearizeFrame` instantiate by `mutable.Stack[LinearizeFrame[T]]()` . That is to say, `T` in `LinearizeFrame[T]` should shadow `class DiGraph[T]` but the same name is ambiguous. So it should rename to remove warning, such as `case class LinearizeFrame[A](v: A, expanded: Boolean)`

https://github.com/freechipsproject/firrtl/blob/648dddeacd9aece4a43cad09430dad25cba69457/src/main/scala/firrtl/graph/DiGraph.scala#L81-L82

`T` in `def sortSeqSeq[T <: Int](x: Seq[T], y: Seq[T]): Boolean = {` is more narrow than `T` in class level, so it should rename to avoid ambiguity. Such as `def sortSeqSeq[A <: Int](x: Seq[A], y: Seq[A])`

https://github.com/freechipsproject/firrtl/blob/648dddeacd9aece4a43cad09430dad25cba69457/src/main/scala/firrtl/graph/EulerTour.scala#L115

## 3. Unused variable and import

I remove some unused variables and import, and these should be historical reason.

azidar:
Thanks for all the cleanup PR's!!

PR #1121, opened 2019-07-11
Change Dependency API to Class[_ <: A]

seldridge:
This changes `prerequisites` and `dependents` specification in the Dependency API to use `Class[_ <: A]` instead of `Class[A]`.

This avoids the need for a ton of janky `.asInstanceOf[Class[A]]` nonsense later. This should have been a mammoth code smell to me and I'm not sure why this didn't occur to me earlier...

### DIFF
--- a/src/main/scala/firrtl/options/DependencyManager.scala
+++ b/src/main/scala/firrtl/options/DependencyManager.scala
@@ -22,15 +22,15 @@ trait DependencyManager[A, B <: TransformLike[A] with DependencyAPI[B]] extends 
   /** Requested [[firrtl.options.TransformLike TransformLike]]s that should be run. Internally, this will be converted to
     * a set based on the ordering defined here.
     */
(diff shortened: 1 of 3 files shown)


seldridge:
This adds two type aliases. One which defines the generic [type of dependencies (`type Dependency`)](https://github.com/freechipsproject/firrtl/pull/1121/files#diff-8cd99845a8f2c996dfdce4295c252b3cR45) and [one for `PhaseManager` (`type PhaseDependency`)](https://github.com/freechipsproject/firrtl/pull/1121/files#diff-4aa6423289d5fcb0bcd5aee390090c4bR365). 

For the former, I was surprised that you can have parametric polymorphic type aliases. This makes sense, but I was still shocked...

This was a good suggestion and it makes the code a lot easier to reason about. Thanks @jackkoenig and @azidar.

Once this passes, I'll merge.

PR #1122, opened 2019-07-15
Update deprecations - WIP

ucbjrl:
Define intermediate types to enable switching between Named and Target (this may not be useful and should be replaced prior to merging).


src/main/scala/firrtl/transforms/RemoveKeywordCollisions.scala
```diff
@@ -6,6 +6,7 @@ import firrtl._
 
 import firrtl.analyses.InstanceGraph
 import firrtl.annotations.{Named, CircuitName, ModuleName, ComponentName}
+import firrtl.annotations.{Target, GenericTarget}
```
chick: I don't see that this import is necessary. Nothing else in file has changed in order to reference it. Was something missed here?

src/test/scala/firrtlTests/FlattenTests.scala
```diff
@@ -8,7 +8,7 @@ import org.scalatest.junit.JUnitRunner
 import firrtl.ir.Circuit
 import firrtl.Parser
 import firrtl.passes.PassExceptions
-import firrtl.annotations.{Annotation, CircuitName, ComponentName, ModuleName, Named}
+import firrtl.annotations.{Annotation, CircuitName, ComponentName, ModuleName}
```
chick: Seems like only imports at lines 11,12 here are actually used. All others are unused. But maybe this should
be saved for a style PR

chick (approved):
This looks fine, but I had one question about RemoveKeywordCollisions. 
If that's ok, looks good to merge

ucbjrl:
(I'm including the comment from the last commit as a comment since otherwise it's difficult to view.)

**NOTE**: This currently fails one test:
```
[info] - should wire subindex source to sink *** FAILED ***
[info]   java.lang.NumberFormatException: For input string: "a"
 ...
[info]   at firrtl.annotations.Target$.$anonfun$toTargetTokens$1(Target.scala:150)
[info]   at scala.collection.immutable.List.foreach(List.scala:392)
[info]   at firrtl.annotations.Target$.toTargetTokens(Target.scala:148)
[info]   at firrtl.annotations.ComponentName.toTarget(Target.scala:724)
[info]   ...
[error] Failed: Total 1380, Failed 1, Errors 0, Passed 1379, Ignored 9
[error] Failed tests:
[error]         firrtlTests.WiringTests
```

due to

```scala
  it should "wire subindex source to sink" in {
    val sinks = Seq(ModuleTarget("Top", "X"))
    val source = ComponentName("r[a]", ModuleName("A", CircuitName("Top"))).toTarget
```

and the fact that `r[a]` is not strictly a subindex access.

Either the test should be changed, or `toTargetTokens()` needs to accept a string as a successor to `[`, and not assume it's an `Int`.

**NOTE2**: This test will fail in current master if a `toTarget` is applied to the source, i,e:
```scala
    val source = ComponentName("r[a]", ModuleName("A", CircuitName("Top")))
    val t = source.toTarget
```

seldridge:
@azidar can correct me, but I believe that `Target` cannot be used to point to a sub*access*. (I incorrectly wrote the name of that test as a subindex, but it's actually a subaccess.)

By that logic, that test is invalid and should be changed. The current `Wiring` transform is treating `[a]` and `[1]` equivalently due to use of [`AnnotationUtils.tokenize`](https://github.com/freechipsproject/firrtl/blob/master/src/main/scala/firrtl/annotations/AnnotationUtils.scala#L42) which doesn't hold for `Target`.

ucbjrl:
Thanks for the explanation @seldridge. I've corrected the test and added a subindex test.

If `Target` cannot be used to represent a subaccess reference, then isn't it premature to deprecate `ComponentName` (which can represent a subaccess), especially when the associated deprecation warning is:
```
class ComponentName in package annotations is deprecated (since 1.2): Use Target instead, will be removed in 1.3
```

seldridge:
Nah, I think it's fine. Really a `ComponentName` pointing at a subaccess doesn't make sense either, it's just that this happens to work for wiring because you can cram an arbitrary string into a component. A user wanting the same behavior with wiring can always wire from a temporary. If the temporary doesn't exist then it can be added with an aspect.

This is potentially a major change for custom transform writers, but I think there are solutions.

@azidar would probably need to chime in. Feedback from custom transform writers at UCB and elsewhere would help, too.

seldridge:
The issue with `ComponentName` always being converted to `ReferenceTarget` can be resolved with a strategy like https://github.com/freechipsproject/firrtl/pull/1311 where `ReferenceTarget`s are converted to `InstanceTarget`s.

ucbjrl:
This currently (3/10/20) fails three tests:
```
 ...
[info] - LoadMemoryAnnotation should be correctly parsed from a string *** FAILED ***
[info]   firrtl.annotations.InvalidAnnotationJSONException: No usable value for target
[info] No constructor for type IsComponent, JString(~CircuitMem|ModuleMem>ram)
[info]   ...
 ...
[info] Renaming
[info] - should track deduplication *** FAILED ***
[info]   List(CombinationalPath(ReferenceTarget(Top,Top,List(),out,List()),ArrayBuffer(ReferenceTarget(Top,Top,List(),in_0,List()), ReferenceTarget(Top,Top,List(),in_1,List()))), SimpleAnno(ReferenceTarget(Top,Child,List(),x,List()),this is a value), SimpleAnno(ReferenceTarget(Top,Child_1,List(),y,List()),this is a value), ModuleAnno(ModuleTarget(Top,Child)), ModuleAnno(ModuleTarget(Top,Child_1))) did not contain element SimpleAnno(ReferenceTarget(Top,Child,List(),y,List()),this is a value) (AnnotationTests.scala:410)
 ...

[info] *** 3 TESTS FAILED ***
[error] Failed: Total 1438, Failed 3, Errors 0, Passed 1435, Ignored 10
[error] Failed tests:
[error] 	firrtlTests.annotationTests.LoadMemoryAnnotationSpec
[error] 	firrtlTests.JsonAnnotationTests
[error] 	firrtlTests.LegacyAnnotationTests
[error] (Test / test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 229 s (03:49), completed Mar 10, 2020, 12:34:56 PM
```

PR #1123, opened 2019-07-16
Dependency API 2 (take 3)

### LINKED ISSUE #446: Enforce ordering on custom pass annotations

seldridge:
### Dependencies

- Prerequisites: [#1357, #1423]

### Overview

(Apologies for the email traffic. [GitHub has restrictions on re-opening PRs that have been forced pushed](https://github.com/isaacs/github/issues/361), so I have to re-create this... again.) *Re-opened version of #1115 (which was a re-opened of #1105)...*

This adds a Dependency API version of the FIRRTL compiler.

- Adds `optionalPrerequisistes` member to `DependencyAPI`
- Adds `TransformManager` for managing transforms (a `DependencyManager[Transform]`)
- Adds `firrtl.stage.transforms.Compiler` which replaces `firrtl.Compiler` and is, itself, a `Transform`
- Defines dependencies for all transforms and passes
- Any transform/pass object is deprecated in favor of a transform/pass class
- Adds compatibility logic for converting `inputForm`/`outputForm` to Dependency API
- New tests to ensure that the resulting transform orderings are sane
- Mass deprecations

The new transform order deviates from the original `LoweringCompilers.scala` order (and is tested/documented in `LoweringCompilersSpec.scala`. Namely:

- `ResolveAndCheck`
  - One `CheckTypes` is appended
  - see the test case: https://github.com/freechipsproject/firrtl/pull/1123/files#diff-
- `HighFirrtlToMiddleFirrtl`
  - `ResolveKinds` and `InferTypes` run after `Uniquify`
  - `ExpandWhens` and `CheckInitialization` are collapsed to one transform: `ExpandWhensAndCheck`
  - `CheckResets` runs after `InferWidths`
  - `CheckWidths` does not run
  - see the test case: https://github.com/freechipsproject/firrtl/pull/1123/files#diff-ba7d72efa1d556670c3affbe82a41222R93
- `LowFirrtlOptimization`
  - An additional `Legalize` runs after the second `ConstantPropagation`
  - see the test case: https://github.com/freechipsproject/firrtl/pull/1123/files#diff-ba7d72efa1d556670c3affbe82a41222R130

In summary: 5 new transforms run, one transform is removed. Therefore, there is an expectation that this will be slower than the original compiler.

Additionally, this adds pretty printing for `DependencyManager`s with the ability for customization via a `customPrintHandling` method to print additional information about or recurse into `Transform`s which are not `DependencyManager`s, e.g., the `VerilogEmitter`. When running with `--log-level info` you can now get more structured information about what transforms will be run in what order. See the following output for `-X verilog`:

```
Computed transform order in: 406.9 ms
Determined Transform order that will be executed:
  firrtl.stage.transforms.Compiler
  ├── firrtl.passes.CheckChirrtl$
  ├── firrtl.passes.CInferTypes$
  ├── firrtl.passes.CInferMDir$
  ├── firrtl.passes.RemoveCHIRRTL$
  ├── firrtl.passes.ToWorkingIR$
  ├── firrtl.passes.CheckHighForm$
  ├── firrtl.passes.ResolveKinds$
  ├── firrtl.passes.InferTypes$
  ├── firrtl.passes.CheckTypes$
  ├── firrtl.passes.Uniquify$
  ├── firrtl.stage.TransformManager
  │   ├── firrtl.passes.ResolveKinds$
  │   └── firrtl.passes.InferTypes$
  ├── firrtl.passes.ResolveFlows$
  ├── firrtl.passes.CheckFlows$
  ├── firrtl.passes.InferBinaryPoints
  ├── firrtl.passes.TrimIntervals
  ├── firrtl.passes.InferWidths
  ├── firrtl.passes.CheckWidths$
  ├── firrtl.transforms.InferResets
  ├── firrtl.stage.TransformManager
  │   └── firrtl.passes.CheckTypes$
  ├── firrtl.transforms.DedupModules
  ├── firrtl.passes.PullMuxes$
  ├── firrtl.passes.ReplaceAccesses$
  ├── firrtl.passes.ExpandConnects$
  ├── firrtl.passes.RemoveAccesses$
  ├── firrtl.stage.TransformManager
  │   ├── firrtl.passes.Uniquify$
  │   └── firrtl.stage.TransformManager
  │       ├── firrtl.passes.ResolveKinds$
  │       └── firrtl.passes.InferTypes$
  ├── firrtl.passes.ExpandWhensAndCheck
  ├── firrtl.stage.TransformManager
  │   ├── firrtl.passes.ResolveKinds$
  │   ├── firrtl.passes.InferTypes$
  │   ├── firrtl.passes.ResolveFlows$
  │   └── firrtl.passes.InferWidths
  ├── firrtl.passes.RemoveIntervals
  ├── firrtl.passes.ConvertFixedToSInt$
  ├── firrtl.passes.ZeroWidth$
  ├── firrtl.stage.TransformManager
  │   └── firrtl.passes.InferTypes$
  ├── firrtl.passes.LowerTypes$
  ├── firrtl.stage.TransformManager
  │   ├── firrtl.passes.ResolveKinds$
  │   ├── firrtl.passes.InferTypes$
  │   ├── firrtl.passes.ResolveFlows$
  │   └── firrtl.passes.InferWidths
  ├── firrtl.passes.Legalize$
  ├── firrtl.transforms.RemoveReset$
  ├── firrtl.stage.TransformManager
  │   └── firrtl.passes.ResolveFlows$
  ├── firrtl.transforms.CheckCombLoops
  ├── firrtl.checks.CheckResets
  ├── firrtl.transforms.RemoveWires
  ├── firrtl.passes.RemoveValidIf$
  ├── firrtl.transforms.ConstantPropagation
  ├── firrtl.passes.PadWidths$
  ├── firrtl.stage.TransformManager
  │   ├── firrtl.transforms.ConstantPropagation
  │   └── firrtl.passes.Legalize$
  ├── firrtl.passes.memlib.VerilogMemDelays$
  ├── firrtl.stage.TransformManager
  │   ├── firrtl.transforms.ConstantPropagation
  │   └── firrtl.stage.TransformManager
  │       └── firrtl.passes.Legalize$
  ├── firrtl.passes.SplitExpressions$
  ├── firrtl.transforms.CombineCats
  ├── firrtl.passes.CommonSubexpressionElimination$
  ├── firrtl.transforms.DeadCodeElimination
  └── firrtl.VerilogEmitter
      ├── firrtl.transforms.BlackBoxSourceHelper
      ├── firrtl.transforms.FixAddingNegativeLiterals
      ├── firrtl.transforms.ReplaceTruncatingArithmetic
      ├── firrtl.transforms.InlineBitExtractionsTransform
      ├── firrtl.transforms.InlineCastsTransform
      ├── firrtl.transforms.LegalizeClocksTransform
      ├── firrtl.transforms.FlattenRegUpdate
      ├── firrtl.transforms.DeadCodeElimination
      ├── firrtl.passes.VerilogModulusCleanup$
      ├── firrtl.transforms.VerilogRename
      ├── firrtl.passes.VerilogPrep$
      └── firrtl.AddDescriptionNodes
```

API Changes:

- The old behavior of `-fct foo,bar,foo` would result in `foo` being run twice. With the Dependency API, this is equivalent to `-fct foo,bar`. If a user wants to have the behavior where `Seq(foo, bar, foo)` runs, then they have to express this using the Dependency API (`bar` depends on `foo` and invalidates `foo`). This old behavior is *not* preserved.

Todo:

- [x] Preserve hysteresis behavior of `outputForm = HighForm`
- [x] Test that ordering of form-based transforms still works. (This isn't the best test, but it adds to the `CustomTransformSpec` to test that the ordering of transforms specified is the same as what executes.)
- [x] Switch `Forms.HighForm` to include everything up through dedup
- [x] Fix `inputForm = Forms.LowForm` scheduling; a transform with `LowForm` input should occur after `MiddleFirrtlToLowFirrtl` if using the `LowFirrtlEmitter` and after `LowFirrtlOptimizations` if using the `*VerilogEmitter`s
- [x] Placement of `CheckResets`
- [x] Check `Legalize` placement
  - This uses the same order for `Legalize` with one additional `Legalize`
- [x] Placement of additional `InferTypes`/`ResolveFlows` in optimizations (see: https://github.com/freechipsproject/firrtl/issues/1171#issuecomment-544767661)
- [x] Jack/SiFive sign-off: as of 2020-03-05, this is passing SiFive Verilog regressions on 1.2.x with @albert-magyar changes in #1422 cherry-picked
- [x] Update Chisel3 to use `Dependency` (https://github.com/freechipsproject/chisel3/pull/1270)
- [x] Check performance: +-5% for Rocket-sized designs

On performance, there's a constant overhead which makes normalized runtime look bad for the benchmarks. However, once these get large enough in runtime, things don't really matter and it converges to about +-5%.

Meta:

- Fixes #446
- Oddities of `inputForm`/`outputForm` are documented in #1165

### DIFF
--- a/src/main/scala/firrtl/Compiler.scala
+++ b/src/main/scala/firrtl/Compiler.scala
@@ -7,11 +7,13 @@ import java.io.Writer
 
 
 import scala.collection.mutable
+
 import firrtl.annotations._
 import firrtl.ir.Circuit
 import firrtl.Utils.throwInternalError
 import firrtl.annotations.transforms.{EliminateTargetPaths, ResolvePaths}
-import firrtl.options.{StageUtils, TransformLike}
+import firrtl.options.{DependencyAPI, Dependency, PreservesAll, StageUtils, TransformLike}
+import firrtl.stage.transforms.CatchCustomTransformExceptions
 
 /** Container of all annotations for a Firrtl compiler */
 class AnnotationSeq private (private[firrtl] val underlying: List[Annotation]) {
@@ -98,6 +100,7 @@ object CircuitState {
   * strictly supersets of the "lower" forms. Thus, that any transform that
   * operates on [[HighForm]] can also operate on [[MidForm]] or [[LowForm]]
   */
+@deprecated("CircuitForm will be removed in 1.3. Switch to Seq[TransformDependency] to specify dependencies.", "1.2")
 sealed abstract class CircuitForm(private val value: Int) extends Ordered[CircuitForm] {
   // Note that value is used only to allow comparisons
   def compare(that: CircuitForm): Int = this.value - that.value
@@ -116,6 +119,7 @@ sealed abstract class CircuitForm(private val value: Int) extends Ordered[Circui
   *
   * See [[CDefMemory]] and [[CDefMPort]]
   */
+@deprecated("Form-based dependencies will be removed in 1.3. Please migrate to the new Dependency API.", "1.2")
 final case object ChirrtlForm extends CircuitForm(value = 3) {
   val outputSuffix: String = ".fir"
 }
@@ -127,6 +131,7 @@ final case object ChirrtlForm extends CircuitForm(value = 3) {
   *
   * Also see [[firrtl.ir]]
   */
+@deprecated("Form-based dependencies will be removed in 1.3. Please migrate to the new Dependency API.", "1.2")
 final case object HighForm extends CircuitForm(2) {
   val outputSuffix: String = ".hi.fir"
 }
@@ -138,6 +143,7 @@ final case object HighForm extends CircuitForm(2) {
   *  - All whens must be removed
   *  - There can only be a single connection to any element
   */
+@deprecated("Form-based dependencies will be removed in 1.3. Please migrate to the new Dependency API.", "1.2")
 final case object MidForm extends CircuitForm(1) {
   val outputSuffix: String = ".mid.fir"
 }
@@ -148,6 +154,7 @@ final case object MidForm extends CircuitForm(1) {
   *  - All aggregate types (vector/bundle) must have been removed
   *  - All implicit truncations must be made explicit
   */
+@deprecated("Form-based dependencies will be removed in 1.3. Please migrate to the new Dependency API.", "1.2")
 final case object LowForm extends CircuitForm(0) {
   val outputSuffix: String = ".lo.fir"
 }
@@ -163,6 +170,7 @@ final case object LowForm extends CircuitForm(0) {
   * TODO(azidar): Replace with PreviousForm, which more explicitly encodes
   * this requirement.
   */
+@deprecated("Form-based dependencies will be removed in 1.3. Please migrate to the new Dependency API.", "1.2")
 final case object UnknownForm extends CircuitForm(-1) {
   override def compare(that: CircuitForm): Int = { sys.error("Illegal to compare UnknownForm"); 0 }
 
@@ -171,13 +179,22 @@ final case object UnknownForm extends CircuitForm(-1) {
 // scalastyle:on magic.number
 
 /** The basic unit of operating on a Firrtl AST */
-abstract class Transform extends TransformLike[CircuitState] {
+trait Transform extends TransformLike[CircuitState] with DependencyAPI[Transform] {
+
   /** A convenience function useful for debugging and error messages */
   def name: String = this.getClass.getName
   /** The [[firrtl.CircuitForm]] that this transform requires to operate on */
+  @deprecated(
+    "InputForm/OutputForm will be removed in 1.3. Use DependencyAPI methods (prerequisites, dependents, invalidates)",
+    "1.2")
   def inputForm: CircuitForm
+
   /** The [[firrtl.CircuitForm]] that this transform outputs */
+  @deprecated(
+    "InputForm/OutputForm will be removed in 1.3. Use DependencyAPI methods (prerequisites, dependents, invalidates)",
+    "1.2")
   def outputForm: CircuitForm
+
   /** Perform the transform, encode renaming with RenameMap, and can
     *   delete annotations
     * Called by [[runTransform]].
@@ -185,9 +202,64 @@ abstract class Transform extends TransformLike[CircuitState] {
     * @param state Input Firrtl AST
     * @return A transformed Firrtl AST
     */
-  protected def execute(state: CircuitState): CircuitState
+  def execute(state: CircuitState): CircuitState
 
   def transform(state: CircuitState): CircuitState = execute(state)
+
+  import firrtl.{ChirrtlForm => C, HighForm => H, MidForm => M, LowForm => L, UnknownForm => U}
+  import firrtl.stage.Forms
+
+  override def prerequisites: Seq[Dependency[Transform]] = inputForm match {
+    case C => Nil
+    case H => Forms.Deduped
+    case M => Forms.MidForm
+    case L => Forms.LowForm
+    case U => Nil
+  }
+
+  override def optionalPrerequisites: Seq[Dependency[Transform]] = inputForm match {
+    case L => Forms.LowFormOptimized
+    case _ => Seq.empty
+  }
+
+  private lazy val fullCompilerSet = new mutable.LinkedHashSet[Dependency[Transform]] ++ Forms.VerilogOptimized
+
+  override def dependents: Seq[Dependency[Transform]] = {
+    val lowEmitters = Dependency[LowFirrtlEmitter] :: Dependency[VerilogEmitter] :: Dependency[MinimumVerilogEmitter] ::
+      Dependency[SystemVerilogEmitter] :: Nil
+
+    val emitters = inputForm match {
+      case C => Dependency[ChirrtlEmitter]      :: Dependency[HighFirrtlEmitter]   :: Dependency[MiddleFirrtlEmitter] :: lowEmitters
+      case H => Dependency[HighFirrtlEmitter]   :: Dependency[MiddleFirrtlEmitter] :: lowEmitters
+      case M => Dependency[MiddleFirrtlEmitter] :: lowEmitters
+      case L => lowEmitters
+      case U => Nil
+    }
+
+    val selfDep = Dependency.fromTransform(this)
+
+    inputForm match {
+      case C => (fullCompilerSet                           ++ emitters - selfDep).toSeq
+      case H => (fullCompilerSet -- Forms.Deduped          ++ emitters - selfDep).toSeq
+      case M => (fullCompilerSet -- Forms.MidForm          ++ emitters - selfDep).toSeq
+      case L => (fullCompilerSet -- Forms.LowFormOptimized ++ emitters - selfDep).toSeq
+      case U => Nil
+    }
+  }
+
+  private lazy val highOutputInvalidates = fullCompilerSet -- Forms.MinimalHighForm
+  private lazy val midOutputInvalidates = fullCompilerSet -- Forms.MidForm
+
+  override def invalidates(a: Transform): Boolean = {
+    (inputForm, outputForm) match {
+      case (U, _) | (_, U)  => true  // invalidate everything
+      case (i, o) if i >= o => false // invalidate nothing
+      case (_, C)           => true  // invalidate everything
+      case (_, H)           => highOutputInvalidates(Dependency.fromTransform(a))
+      case (_, M)           => midOutputInvalidates(Dependency.fromTransform(a))
+      case (_, L)           => false // invalidate nothing
+    }
+  }
 
   /** Convenience method to get annotations relevant to this Transform
     *
@@ -309,7 +381,7 @@ trait ResolvedAnnotationPaths {
 }
 
 /** Defines old API for Emission. Deprecated */
-trait Emitter extends Transform {
+trait Emitter extends Transform with PreservesAll[Transform] {
   @deprecated("Use emission annotations instead", "firrtl 1.0")
   def emit(state: CircuitState, writer: Writer): Unit
 
@@ -324,6 +396,7 @@ object CompilerUtils extends LazyLogging {
     * @param outputForm [[CircuitForm]] to lower to
     * @return Sequence of transforms that will lower if outputForm is lower than inputForm
     */
+  @deprecated("Use a TransformManager requesting which transforms you want to run. This will be removed in 1.3.", "1.2")
   def getLoweringTransforms(inputForm: CircuitForm, outputForm: CircuitForm): Seq[Transform] = {
     // If outputForm is equal-to or higher than inputForm, nothing to lower
     if (outputForm >= inputForm) {
@@ -333,9 +406,8 @@ object CompilerUtils extends LazyLogging {
         case ChirrtlForm =>
           Seq(new ChirrtlToHighFirrtl) ++ getLoweringTransforms(HighForm, outputForm)
         case HighForm =>
-          Seq(new IRToWorkingIR, new ResolveAndCheck,
-              new transforms.DedupModules, new HighFirrtlToMiddleFirrtl) ++
-              getLoweringTransforms(MidForm, outputForm)
+          Seq(new IRToWorkingIR, new ResolveAndCheck, new transforms.DedupModules, new HighFirrtlToMiddleFirrtl) ++
+            getLoweringTransforms(MidForm, outputForm)
         case MidForm => Seq(new MiddleFirrtlToLowFirrtl) ++ getLoweringTransforms(LowForm, outputForm)
         case LowForm => throwInternalError("getLoweringTransforms - LowForm") // should be caught by if above
         case UnknownForm => throwInternalError("getLoweringTransforms - UnknownForm") // should be caught by if above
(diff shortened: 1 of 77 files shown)


src/main/scala/firrtl/passes/RemoveValidIf.scala
```diff
@@ -2,27 +2,24 @@
 
 package firrtl
 package passes
+
+import firrtl.{SystemVerilogEmitter, VerilogEmitter}
```
chick: Unneeded import

src/main/scala/firrtl/passes/RemoveAccesses.scala
```diff
@@ -2,18 +2,30 @@
...
 package firrtl.passes
 
-import firrtl.{WRef, WSubAccess, WSubIndex, WSubField, Namespace}
+import firrtl.{Namespace, Transform, WRef, WSubAccess, WSubIndex, WSubField}
 import firrtl.PrimOps.{And, Eq}
 import firrtl.ir._
 import firrtl.Mappers._
 import firrtl.Utils._
 import firrtl.WrappedExpression._
+import firrtl.options.PreservesAll
```
chick: Another unneeded import?

src/main/scala/firrtl/passes/Uniquify.scala
```diff
@@ -8,9 +8,79 @@ import firrtl._
 import firrtl.ir._
 import firrtl.Utils._
 import firrtl.Mappers._
+import firrtl.options.PreservesAll
```
chick: Unneeded import

src/main/scala/firrtl/stage/phases/AddImplicitOutputFile.scala
```diff
@@ -3,22 +3,26 @@
...
+  *    below)
   *  - Use "a"
   *
   * The file suffix may or may not be specified, but this may be arbitrarily changed by the [[Emitter]].
   *
-  * @note This [[firrtl.options.Phase Phase]] has a dependency on [[AddCircuit]]. Only a [[FirrtlCircuitAnnotation]]
-  * will be used to implicitly set the [[OutputFileAnnotation]] (not other [[CircuitOption]] subclasses).
+  * @note This [[firrtl.options.Phase Phase]] has a dependency on [[AddCircuit]]. Only a
+  * [[firrtl.stage.FirrtlCircuitAnnotation FirrtlCircuitAnnotation]] will be used to implicitly set the
+  * [[OutputFileAnnotation]] (not other [[CircuitOption]] subclasses).
```
chick: CircuitOption may not link in scaladoc correctly here since it is not imported. Emitter at line 17 probably as well, but you didn't change that in this PR

src/main/scala/firrtl/stage/transforms/WrappedTransform.scala
```diff
@@ -0,0 +1,33 @@
...
+
+package firrtl.stage.transforms
+
+import firrtl.Transform
+
+/** A [[firrtl.Transform]] that "wraps" a second [[firrtl.Transform Transform]] to do some work before and after the
+  * second [[firrtl.Transform Transform]].
+  *
+  * This is intended to synergize with the [[firrtl.options.DependencyManager.wrappers]] method.
+  * @see [[firrtl.stage.transforms.CatchCustomTransformException]]
```
chick: IntelliJ says the CatchCustomTransformException link here is broken.
I think it should be plural CatchCustomTransformExceptions

src/main/scala/firrtl/stage/Forms.scala
```diff
@@ -0,0 +1,86 @@
+// See LICENSE for license details.
+
+package firrtl.stage
+
+import firrtl._
+import firrtl.Transform
```
chick: Unneeded dependency

src/main/scala/firrtl/stage/TransformManager.scala
```diff
@@ -0,0 +1,27 @@
+// See LICENSE for license details.
+
+package firrtl.stage
+
+import firrtl.{CircuitForm, CircuitState, Transform, UnknownForm}
+import firrtl.options.DependencyManager
+
+class TransformManager(
```
chick: Quibble, is this likely to be used by developers, should it have scaladoc

src/main/scala/firrtl/transforms/CombineCats.scala
```diff
@@ -2,11 +2,13 @@
 package firrtl
 package transforms
 
+import firrtl.{SystemVerilogEmitter, VerilogEmitter}
```
chick: Unused import

src/main/scala/firrtl/passes/PadWidths.scala
```diff
@@ -3,12 +3,28 @@
 package firrtl
 package passes
 
+import firrtl.{SystemVerilogEmitter, VerilogEmitter}
```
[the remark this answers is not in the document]
seldridge: You're right... Apparently specifying two packages isn't equivalent to `package firrtl.passes` and will make the contents of both in scope. I had no idea it worked that way...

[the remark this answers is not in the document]
seldridge: To be fixed in rebase.

src/main/scala/firrtl/passes/RemoveValidIf.scala
seldridge: To be fixed in rebase.

src/main/scala/firrtl/passes/RemoveAccesses.scala
seldridge: Removed in rebase.

src/main/scala/firrtl/passes/Uniquify.scala
seldridge: Removed in rebase.

src/main/scala/firrtl/stage/phases/AddImplicitOutputFile.scala
seldridge: Somehow it finds this. I have no idea how. Scaladoc links are magic... 

I'll make this explicit in the rebase.

src/main/scala/firrtl/stage/transforms/WrappedTransform.scala
seldridge: Good catch. Fixed in the rebase.

src/main/scala/firrtl/stage/Forms.scala
seldridge: Fixed in rebase.

src/main/scala/firrtl/stage/TransformManager.scala
seldridge: Yeah, it should. I'll include documentation in the rebase.

src/main/scala/firrtl/transforms/CombineCats.scala
seldridge: Fixed in rebase.

seldridge:
Thanks for taking a look, @chick. Unfortunately, this is load bearing and should probably have @jackkoenig approval before merging. I'll ask him to look at this on SiFive regressions.

chick:
@seldridge I agree, there should be another review. But you're halfway there now :-)

sequencer:
Can't wait this to be merged, LOL.
I have struggled a whole night for the strange problem that `VerilogCompiler` and `LowFirrtlCompiler` has different behavior until I changed the `inputForm` and `outputForm` into `MidForm` from `LowForm`.
Till now, I still have no idea what happened.

seldridge:
@sequencer: Some of the oddities with emitters before this PR that may be leading to what you're seeing:

1) Emitters are `Transform`s, but are scheduled in a special way. Before [the original `FirrtlStage` PR](https://github.com/freechipsproject/firrtl/pull/1005), emitters were not supposed to be scheduled as custom transforms. Emitters were used to fill out an abstract member of `Compiler`. When a `Compiler` runs, it runs all of it's transforms, adding custom transforms as specified, and then it runs the concrete emitter. Trying to add an emitter as a custom transform would possibly cause this to run earlier than you want. Trying to run a `VerilogEmitter` in this way wouldn't, I think, work. The `FirrtlStage` PR modified the logic for `mergeTransforms` such that emitters were viewed as "special transforms" that would be scheduled as late as possible. This fixed some issues where your additional emitter transform would run before your custom transform. This PR goes one step further and schedules emitters just like all other transforms.

2) Form-based specification of dependencies/ordering is insufficient to properly schedule certain emitters like they were transforms. As one example, the `VerilogEmitter` is specified as `inputForm = LowForm; outputForm = LowForm`. However, the `VerilogEmitter` really requires low FIRRTL *and an additional `LowFirrtlOptimization` sequence of transforms* in order to work. This PR attempts to fix this by allowing for more finer-grained specifications of arbitrary transform prerequisites while providing pre-defined sequences of transforms that replicate what the old forms meant.

3) There are further optimizations that happen *beyond* those specified in the prerequisite requirements for an emitter. This results in an emitter `Transform` necessarily being an identity transform because it is modifying the FIRRTL AST in non-compliant ways. One example is `VerilogPrep` that adds the WIR node `WDefInstanceConnector`. The emitter will then generate an emitter annotation containing the generated circuit, but returns the un-modified FIRRTL circuit. If this were to modify the FIRRTL circuit, then you would wind up with something which is non-compliant with the FIRRTL spec (i.e., any later transform would hit match errors on `WDefInstanceConnector`.

In summary, the logic for how compilers, their emitters, other emitters, and custom transforms work is complicated, but is getting simpler.

sequencer:
@seldridge Thank you for your clear explanation! I was working on the CCC 2019 presentation these days, sorry for the late reply.
I just found the difference of transform between `lowfirrtlEmitter` and `verilogEmitter` is not the verilogEmitter just adding new Transform after `LowFirrtlEmitter`, but added a sequence of 
```
LowFirrtlOptimization
RemoveValidIf
ConstantPropagation
PadWidths
ConstantPropagation
Legalize
VerilogMemDelays
ConstantPropagation
SplitExpressions
CombineCats
CommonSubexpressionElimination
DeadCodeElimination
```
into `verilogEmitter`, my own Transform only show up before the `Emitter` itself.
That might lead to this strange problem.

Will highfirrtl, midfirrtl, lowfirrtl become deprecated after dependency API merged? That might lead to a lot of misunderstanding.

seldridge:
@sequencer: One potentially useful resource to see what the orderings are is in one of the new tests: https://github.com/freechipsproject/firrtl/pull/1123/files#diff-ba7d72efa1d556670c3affbe82a41222R13. That is checking that the generated orderings by a `TransformManager` are correct. You can also use that if you have questions about "what transforms should be run in what order"?

Yes, everything related to form-based specifications are getting deprecated. That includes `inputForm`/`outputForm` as well as the older `firrtl.{ChirrtlForm, HighForm, MidForm, LowForm, UnknownForm}`. Yeah, I'm not sure how to handle the confusion here other than with a good document/web page that talks about everything related to Stage/Phase/Dependency API.

albert-magyar:
I know this is fairly late in the game to bring this up, but do you have any thoughts on wrapping dependencies? In the most basic form, this would make it possible to use a singleton `Transform` with just a little extra code behind the scenes.

```
override val prerequisites =
  Seq( Dependency(ObjectTransform),
       Dependency(classOf[ClassTransform]) )
```

I personally think that singletons are a language feature that is worth keeping around, especially given the fact that basically any sane `Transform` could be implemented as a singleton in the absence of (now-forbidden) constructor parameters.

seldridge:
Interesting thought @albert-magyar. I had assumed that it was either objects as dependencies or classes as dependencies. I started with the former (as it was easier to prototype), but moved to the latter after feedback from @jackkoenig. 

I think what you're showing is pretty unambiguous, however, I'm not sure how to handle/disallow objects from singlestons like `Dependency(new ClassTransform(foo))` and have these be properly deduplicated across multiple transforms that may list this as a dependency.

I'm also not sure that supporting singleton transforms is totally necessary as the dependency API results in users never interacting with transforms (singleton or a constructed transform class). So, supporting language features shouldn't be hugely necessary. (Though, this is basically mandating that dependencies and Transform managers are the only way to do things...)

So, I'm not sure, but can think about it. (And maybe you can think about it on your end.)

seldridge:
I took a look at the performance of this and the results seem to be within +-5% of `master`. There are non-trivial constant increases in runtime which manifest for small tests, but are lost in the noise of large Rocket Chip-sized builds. I.e., you'll probably see a 0.5-1 second increase due to the combination of the dependency API and the singleton work. 

How should we proceed with this one? This could probably use one more code review.

seldridge:
Ping @jackkoenig, @azidar. This is ready for review. The branch is passing locally on FIRRTL and Chisel3 master (but needs a fix for Travis building Verilator #1423).

src/main/scala/firrtl/Compiler.scala
```diff
@@ -171,24 +179,88 @@ final case object UnknownForm extends CircuitForm(-1) {
...
+
   /** Perform the transform, encode renaming with RenameMap, and can
     *   delete annotations
     * Called by [[runTransform]].
     *
     * @param state Input Firrtl AST
     * @return A transformed Firrtl AST
     */
-  protected def execute(state: CircuitState): CircuitState
+  def execute(state: CircuitState): CircuitState
```
ucbjrl: This change on the execute method (from protected to public) is causing compilation errors in other chisel projects that have been dependent on master. Is this change intentional?

seldridge: No, I don't think this was needed. Quick fix inbound.

PR #1124, opened 2019-07-17
Gender to Flow

### LINKED ISSUE #943: Direction versus male/female
**Type of issue**: other enhancement

On Bundles Chisel uses Input and Output for directions. Why introducing the terms male and female in FIRRTL? First, this introduces new terms without a real need (and needs some thinking to map). Second, it sounds a little bit sexism to me. I know we use those terms for physical connectors, but not to indicate a direction. Which is BTW another argument to not use them for a direction indication.

Anything around in/out/inout or input/output/bidirectional sounds more logical.

seldridge:
In light of the discussion around #943, this implements a structural rename of Working IR's `Gender` to `Flow`. I'm attempting to maintain backwards compatibility by deprecating anything gender-related that's public and providing implicit conversions where appropriate.

In summary, the following are the new names, which are assumedly more descriptive on what is going on. Nonetheless, I'm happy to bikeshed:
  - `gender        -> flow`
  - `MALE          -> SourceFlow`
  - `FEMALE        -> SinkFlow`
  - `BIGENDER      -> DuplexFlow`
  - `UNKNOWNGENDER -> UnknownFlow`

Fixes #943.


seldridge:
ping @azidar

src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -582,66 +582,83 @@ object CheckGenders extends Pass {
...
+object CheckGenders {
+
+  implicit def toStr(g: Gender): String = g match {
+    case MALE => "source"
+    case FEMALE => "sink"
+    case UNKNOWNGENDER => "unknown"
+    case BIGENDER => "sourceOrSink"
+  }
+
+  def run(c: Circuit): Circuit = CheckGenders.run(c)
```
jackkoenig: This looks like an infinite recursion to me

seldridge: I prefer to call this "Big Stick Deprecation Diplomacy."

...


(Fixed in rebase.)

jackkoenig: 😂

seldridge:
Updated deprecation messages and added deprecated `ResolveGenders` methods. A cursory glance seems that adding a deprecated copy method is complicated. I'm punting on that here, but can dig into it if you think otherwise.

jackkoenig:
I don't think there's anything we can do about `copy`, so I think we're good to go!

PR #1125, opened 2019-07-19
[WIP] add clock edge to enable negedge clock for some special design.

sequencer:
This PR is trying to give a native clock edge support for clock, enable the `negedge clock` in the special module like `clockgen`.
[some tools](https://github.com/ucb-bar/barstools/blob/82636b3ff43ecf6a0f0a7d46ebc2456b31e9703f/macros/src/main/scala/Utils.scala#L211) was used to generate a negedge sensitive clock, which add a `not(exp)` to clock. This work in digital logic. But in case of some synthesis tool regard the inverted clock as a new clock, I add `Edge` as an IR into the firrtl, trying to fix this fundamentally.

### DIFF
--- a/src/main/antlr4/FIRRTL.g4
+++ b/src/main/antlr4/FIRRTL.g4
@@ -46,6 +46,11 @@ dir
   | 'output'
   ;
 
+edge
+  : 'posedge'
+  | 'negedge'
+  ;
+
 type 
   : 'UInt' ('<' intLit '>')?
   | 'SInt' ('<' intLit '>')?
@@ -90,7 +95,8 @@ reset_block
 
 stmt
   : 'wire' id ':' type info?
-  | 'reg' id ':' type exp ('with' ':' reset_block)? info?
+  // use edge? for back compatiable
+  | 'reg' id ':' type edge? exp ('with' ':' reset_block)? info?
   | 'mem' id ':' info? INDENT memField* DEDENT
   | 'cmem' id ':' type info?
   | 'smem' id ':' type info?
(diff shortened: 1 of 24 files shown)


sequencer:
tests will be added soon, but I need some comments from viewer firstly, because this is a functional change, which also need some API changes from chisel3. https://github.com/freechipsproject/firrtl/issues/695 used to mention this problem.

I'm working on the backend annotation framework, which SiFive China guys wanna write clockgen in pure chisel3.
And I think till now firrtl only leaks 2 expression ability then systemverilog.
1. UDP(User Defined Primitives) for base simulation marco define, which can help the blackbox design more elegant.
2. explicit latch support. latch is the most essential digital part of PPA sensitive designs,  which systemverilog is supported with `always_latch`, but chisel only blackbox it.

In my backend annotation flow design, I wanna make some thing like barstools but give a more intact annotation design method to bridge the PaR between chisel design. So when I try to annotate something in verilog, I found the only thing is improving the expression ability of firrtl.

azidar:
Interesting. I generally like the approach you took. One thing to consider how to represent negedge with the Printf, Stop, and DefMemory constructs, which also consume the clock.

You are right this is a very large backwards-incompatible IR change, and its something we shouldn't do lightly. I'd like to solicit a lot of user feedback regarding this (from both Chisel users and FIRRTL-transform writers), just so we have an understanding of the impact of this change.

sequencer:
Yes. I'm trying to give a back compatibility for firrtl to use the `?` in the antlr4.
As for chisel, I think I just need alter the `DefRegister` by adding the default `Posedge` IR, and use annotations and firrtl Transform to make the annotated Target changing to `Negedge` as designed.

PR #1128, opened 2019-07-23
Ignore annotations when deduping modules

albertchen-sifive:
Differing annotations do not necessarily mean that two modules are different. For annotations that do cause changes to a module there is `ResolvedAnnotationPaths`.

### DIFF
--- a/src/main/scala/firrtl/analyses/InstanceGraph.scala
+++ b/src/main/scala/firrtl/analyses/InstanceGraph.scala
@@ -71,7 +71,7 @@ class InstanceGraph(c: Circuit) {
     */
   def findInstancesInHierarchy(module: String): Seq[Seq[WDefInstance]] = {
     val instances = graph.getVertices.filter(_.module == module).toSeq
-    instances flatMap { i => fullHierarchy(i) }
+    instances flatMap { i => fullHierarchy.getOrElse(i, Seq.empty) }
   }
 
   /** An [[firrtl.graph.EulerTour EulerTour]] representation of the [[firrtl.graph.DiGraph DiGraph]] */
--- a/src/main/scala/firrtl/transforms/Dedup.scala
+++ b/src/main/scala/firrtl/transforms/Dedup.scala
@@ -16,8 +16,8 @@ import scala.collection.mutable
 
 
 /** A component, e.g. register etc. Must be declared only once under the TopAnnotation */
-case class NoDedupAnnotation(target: ModuleName) extends SingleTargetAnnotation[ModuleName] {
-  def duplicate(n: ModuleName): NoDedupAnnotation = NoDedupAnnotation(n)
+case class NoDedupAnnotation(target: ModuleTarget) extends SingleTargetAnnotation[ModuleTarget] {
+  def duplicate(n: ModuleTarget): NoDedupAnnotation = NoDedupAnnotation(n)
 }
 
 /** If this [[firrtl.annotations.Annotation Annotation]] exists in an [[firrtl.AnnotationSeq AnnotationSeq]],
@@ -51,7 +51,9 @@ class DedupModules extends Transform {
     if (state.annotations.contains(NoCircuitDedupAnnotation)) {
       state
     } else {
-      val noDedups = state.annotations.collect { case NoDedupAnnotation(ModuleName(m, c)) => m }
+      val noDedups = state.annotations.collect {
+        case NoDedupAnnotation(m: ModuleTarget) => m.module
+      }
       val (newC, renameMap) = run(state.circuit, noDedups, state.annotations)
       state.copy(circuit = newC, renames = Some(renameMap))
     }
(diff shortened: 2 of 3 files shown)


azidar:
Shouldn't we be renaming targets of deduped modules to their absolute instance targets?

albertchen-sifive:
That sort of changes the meaning of annotations though. If we turn a `ModuleTarget` anno into an absolute `InstanceTarget` annotation and a later transform duplicates the target module then it would not automatically inherit the annotations of the target module. Same for relative `InstanceTarget`s.

It would also make renaming more expensive because of the extra absolute paths introduced and because `Dedup` runs pretty early.

jackkoenig:
I would argue that this PR is the thing changing the meaning of annotations, they currently prevent dedup because the argument is that an annotation that applies to one target does *not* apply to things that dedup with that target, thus annotations must be the same in order to dedup.

### This PR

This PR combined with https://github.com/freechipsproject/chisel3/pull/1128 makes it such that annotations will apply to any Target that dedups to the same Target. The only way to make sure this doesn't occur is to have an absolute annotation from the start (thus putting the onus on the annotator or annotation API writer to use that API over a less absolute annotation).

The main pro here is expressibility, it allows you to express things in such a way that a certain annotation should apply to all things that dedup with that target, whether or not the deduping thing has a similar annotation. This is useful for something like `DontTouchAnnotation` which needs to apply to all instances of the deduped module.

The con is potentially usability. Annotations applied in some library may apply to your Modules that dedup with theirs even though you didn't want the annotation. A good example might be applying an annotation to a `Queue` which now has pretty non-local effects. Yes, one can say they should have used absolute targets, but what if it's a bug in someone else's code. How do you prevent others from being able to annotate your Modules without your permission? In fact, this issue is exacerbated by the fact that even if a library writer makes targets absolute, what if their elaborated circuit gets injected into a larger circuit? Now all of the sudden these absolute targets are no longer absolute. I suspect in practice this won't be a problem but if it comes up, it's incredibly hard to debug.

To summarize the con in this change: it creates a subtle yet important distinction between absolute and non-absolute targets. Unless I am mistaken, there is no distinction between them currently.

### Alternative approach

The alternative approach that @azidar and I have discussed in the past is that annotations don't prevent dedup and will disambiguate. This means that annotations that differ between deduping modules will become absolute (or at least an instance annotation to the least common ancestor), while annotations that are the same between deduping modules should be able to stick with a ModuleTarget. From an implementation perspective, this is actually pretty tricky and requires extremely specialized renaming logic.

I actually do think there is a way with this approach to express annotations that need to propagate to all targets that dedup with the given target: you write a custom `update` method on the annotation that just refuses to rename to an instance target. This could be a mixin for annotations that need it.

seldridge:
I'm in agreement that annotations should have no effect on duplication. However, I agree with @azidar that these need to be renamed to `InstanceTarget`s.

tl;dr: hierarchy mainpulations (e.g., `DedupModules`, `EliminateTargetPaths`) should never alias annotations.

> That sort of changes the meaning of annotations though. If we turn a `ModuleTarget` anno into an absolute `InstanceTarget` annotation and a later transform duplicates the target module then it would not automatically inherit the annotations of the target module. 

In this situation, you go from a `ModuleTarget` to an `InstanceTarget`. The duplication is just the reverse, unless I'm missing something.

I think the problem stems from `Dedup` and `EliminateTargetPaths` not doing the correct things right now and I'm proposing they be modified to do the right thing.

(What is the right thing?) I think we can implement @jackkoenig's "Alternative Approach" but solely by fixing the generated rename map. (And possibly adding more methods to the rename map to help generate these types of renames.)

Starting with a toy example:

```
circuit Top:
  module Foo:
    skip
  module Bar:
    skip
  module Top:
    inst foo of Foo
    inst bar of Bar
```

Here, `Bar` gets deduplicated to `Foo`. This currently generates the following rename map:

```
  ~Top|Top/bar:Bar -> ~Top|Top/bar:Foo
  ~Top|Bar         -> ~Top|Foo
```

Here, `Bar` is aliasing to `Foo` incorrectly. I'm arguing that this should be:

```diff
  ~Top|Top/bar:Bar -> ~Top|Top/bar:Foo
- ~Top|Bar         -> ~Top|Foo
+ ~Top|Foo         -> ~Top|Top/foo:Foo
+ ~Top|Bar         -> ~Top|Top/bar:Foo
```

Note that, at this point, there are no `ModuleTarget`s pointing at `Foo`. This may be fine enough for a solution. `ModuleTarget`s *may* be generated after this point, however.

If we then duplicate these by resolving the paths of both instances you wind up with the following renames:

```
  ~Top|Top/bar:Foo -> ~Top|Foo___Top_bar
  ~Top|Top/foo:Foo -> ~Top|Foo___Top_foo
```

I'm arguing that this should be:
```diff
  ~Top|Top/bar:Foo -> ~Top|Foo___Top_bar
  ~Top|Top/foo:Foo -> ~Top|Foo___Top_foo
+ ~Top|Foo         -> [~Top|Foo___Top_bar, ~Top|Foo___Top_foo]
```

Note that there's no change to how the renaming actually works, just fixes to correct the generated rename maps.

I'm skipping over @jackkoenig's suggestion of then combining `InstanceTarget` annotations together if possible. This should be doable (either in the renaming logic or with an additional `SimplifyAnnotations` transform/phase/mixin), but I think it may be a feature and not required.

> Same for relative `InstanceTarget`s.

Yeah, this is where it gets complicated... The toy example above doesn't get into this and I haven't thought about the implications. I do think this is tractable, however.

> It would also make renaming more expensive because of the extra absolute paths introduced and because `Dedup` runs pretty early.

It makes generating the rename maps more complicated, but I'm not sure it drives up the expense enormously. `Dedup` does run early, but I expect the existence of `EliminateTargetPaths` means that transforms may frequently duplicate, run, then dedup. (Or: this is a pattern that I want to use...)

albert-magyar:
> It makes generating the rename maps more complicated, but I'm not sure it drives up the expense enormously. `Dedup` does run early, but I expect the existence of `EliminateTargetPaths` means that transforms may frequently duplicate, run, then dedup. (Or: this is a pattern that I want to use...)

This is also a pattern I want to use more.

I agree that this change could break some expectations in the short term, but I think that it would really help in the long run if it were easier to rely on a combo of `Dedup` and `EliminateTargetPaths`. Right now, how a transform handles being applied to instances vs. modules leads to a lot of fragility.

I think the downside is that injudicious use of that pattern could lead to crazy compiler performance pathologies.

Issue #1090, opened 2019-05-30
Let user specify the destination filename for "firrtl_black_box_resource_files.f"


jerryz123:
**Type of issue**: feature request 

  * **What is the current behavior?** The BlackBoxResourceHelper always emits the filelist into the file "firrtl_black_box_resource_files.f". 
  * **What is the expected behavior?** Allow the destination file of the BlackBoxResouceHelper to be manually specified. 

**What is the use case for changing the behavior?** We want to avoid clobbering the resource_files.f when running multiple invocations of FIRRTL in one directory.

**Impact**: API addition (no impact on existing code)

**Development Phase**: request

albert-magyar:
Closed by #1129

PR #1129, opened 2019-07-23
Allow name of blackbox resource .f file to change from static value

albert-magyar:
Currently, the `.f` resource file gets generated with the constant filename of `firrtl_black_box_resource_files.f`. This PR keeps that as a default, but allows the user to override this filename with an optional annotation. The only other visibility outside of the transform is that a new parameter is added to `verilogToCpp` to enable sourcing a renamed file (but the default filename is used by default).

This is helpful to manage multiple `.f` files, which has come up as an issue in Chipyard.

### DIFF
--- a/src/main/scala/firrtl/transforms/BlackBoxSourceHelper.scala
+++ b/src/main/scala/firrtl/transforms/BlackBoxSourceHelper.scala
@@ -172,8 +172,8 @@ object BlackBoxSourceHelper {
       // If the path isn't canonical, when make tries to determine dependencies based on the *__ver.d file, we end up with errors like:
       //  make[1]: *** No rule to make target `test_run_dir/examples.AccumBlackBox_PeekPokeTest_Verilator345491158/AccumBlackBox.v', needed by `.../chisel-testers/test_run_dir/examples.AccumBlackBox_PeekPokeTest_Verilator345491158/VAccumBlackBoxWrapper.h'.  Stop.
       //  or we end up including the same file multiple times.
-      val outputFile = outputFileName.getOrElse(defaultFileListName)
-      writeTextToFile(files.map(_.getCanonicalPath).mkString("\n"), new File(targetDir, outputFile))
+      val outputFile = outputFileName.map(new File(_)).getOrElse(new File(targetDir, defaultFileListName))
+      writeTextToFile(files.map(_.getCanonicalPath).mkString("\n"), outputFile)
     }
   }
 
--- a/src/main/scala/firrtl/transforms/BlackBoxSourceHelper.scala
+++ b/src/main/scala/firrtl/transforms/BlackBoxSourceHelper.scala
@@ -71,8 +71,8 @@ class BlackBoxSourceHelper extends firrtl.Transform {
           val targetDir = new File(dir)
           if (!targetDir.exists()) { FileUtils.makeDirectory(targetDir.getAbsolutePath) }
           (acc, targetDir, flistName)
-        case a: BlackBoxHelperAnno => (acc + a, tdir, flistName)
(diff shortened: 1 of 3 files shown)


albert-magyar:
@colinschmidt I'm going to try using this in Chipyard before I take the DO NOT MERGE off

src/main/scala/firrtl/transforms/BlackBoxSourceHelper.scala
```diff
@@ -150,17 +159,18 @@ object BlackBoxSourceHelper {
     out.close()
   }
 
-  val fileListName = "firrtl_black_box_resource_files.f"
```
jackkoenig: I think it'd be good to have `fileListName` as a deprecated `def` just in case someone is using it

albert-magyar: Good call

jerryz123:
Resolves #1090

albert-magyar:
@colinschmidt, this works to resolve the current `.f` issues in chipyard.

The one question left is whether to make the string filename in the annotation a) absolute or b) relative to `targetDir`. It seems like more existing FIRRTL things treat overridden filenames as (a) than (b), but maybe @jackkoenig or @azidar could chime in.

seldridge:
Have you tried using `firrtl.options.StageOptions.getBuildFileName`? That will take a string and do absolute if it's absolute (in Linux if it starts with `/`) and relative to the target directory otherwise. 

Something like the following would likely work (untested, though!):

```scala
import firrtl.AnnotationSeq
import firrtl.options.{StageOptions, Viewer}

abstract class Foo {
  /** Given some annotations */
  val annotations: AnnotationSeq

  /** Construct [[StageOptions]] from the annotations */
  val sopts = Viewer[StageOptions].view(annotations)

  /** An absolute path */
  sopts.getBuildFileName("foo/bar/baz")

  /** A path relative to the target directory as determined by the annotations */
  sopts.getBuildFileName("/foor/bar/baz")
}
```

jackkoenig:
I think @seldridge's suggestion is good; supporting both use styles of paths is the most flexible approach.

I do want to make a point about avoiding absolute paths to whatever extent possible. Historically we (myself especially) have done a poor job of this, because absolute paths seem better at first blush, but they're actually pretty problematic. In particular, it is incredibly difficult to make a build relocatable and reproducible if you use absolute paths. 

### Relocatability

Imagine if your annotations are full of absolute file paths, but then you want to farm out part of your compute to some other machine in the cloud. If the machine's directory structure doesn't match the directory structure you used to create the paths in the annotations, you're hosed. 

### Caching

Many tools now use hash based caching (including sharing caching between users) that can enable powerful short-circuiting builds when some intermediate artifact is actually the same as a previous run (see bazel, pants, wake). Absolute paths make this nigh on impossible because they differ between users or even for a single user when you have multiple clones.

**TLDR** We should use relative paths.

jwright6323:
@seldridge are these backwards?

>   /** An absolute path */
>   val absolutePath: String = sopts.getBuildFileName("foo/bar/baz")
> 
>   /** A path relative to the target directory as determined by the annotations */
>   val relativePath: String = sopts.getBuildFileName("/foor/bar/baz")

seldridge:
@jwright6323: Yep, backwards.

albert-magyar:
@seldridge that is a good idea. Unfortunately, `BlackBoxSourceHelper` has it's _own distinct_ `targetDir` option (so blackbox resources could end up in a different place), so I basically copied the functionality of `getBuildFileName`.

This now supports absolute and relative paths.

seldridge:
That sounds fine. At some point I'd like to revisit and get things unified under `TargetDirAnnotation`. Can happen in the future...

PR #1134, opened 2019-07-24
Filter out more filename extensions for blackbox source headers

albert-magyar:
Since it's basically always bad for header files to be listed in the `.f` files, this makes filtering them out a bit more general.


jwright6323:
Is there ever a reason to `addResource("something.h")`? Isn't that the only way they would end up in the .f?

albert-magyar:
I believe this was an issue that @abejgonzalez was currently fixing with grep. Maybe there's a better solution by just avoiding the situation as you say. Abe?

abejgonzalez:
Yes. In Rocket Chip they add a `.h` for remote bitbang (https://github.com/chipsalliance/rocket-chip/blob/2eafc09ca30f932253c8d5828a0f34597fae0b92/src/main/scala/devices/debug/Periphery.scala#L197). I think the main reason they add this is to move the header to the build directory. However, this also adds it to the `.f` file which could cause issues.

jwright6323:
I guess I should rephrase my question- is there a reason someone might want the header file in the `.f`? Is `.f` explicitly meant to be consumed by a tool like VCS, or is it an unregulated "list of files" ? It seems reasonable to me that firrtl would emit a list of all resources and expect you to process them into the files you actually want, but I could see the other way as well.

abejgonzalez:
My assumption is that it is an unregulated list of files that need to be filtered/changed to be put in the simulation/CAD environment. However, I think one of the original authors of this pass should chime in.

albert-magyar:
Currently, `.vh` files are already filtered out under the assumption that anything that is a 'header' will be double-included if listed in the filelist. The idea here is that this assumption holds true for other kinds of headers (`.h`, `.svh`), so it just makes it a bit more consistent.

PR #1139, opened 2019-07-30
Fix "since" deprecation, should be "1.3" not "1.2"

seldridge:
The deprecation here should be since `1.2` and not since `1.3`, I believe.

This look okay @albert-magyar?


albert-magyar:
Actually maybe change the commit message :)

PR #1142, opened 2019-07-31
io.Source is not closed when used in most common text reading idiom

chick:
- Provide new tools for reading in text
  - from a file
  - from a string file name
  - from a resource file
- text can be read in as
  - a single string with newlines
  - a seq of strings, one string per line
- FileUtils put in its own file (was in `Driver`)
  - in same package to keep existing API the same

Hopefully this will protect users from creating resource leaks that
may explain sbt crashing.
Also simpler interface should make code more streamlined, existing
uses of io.Source seem often to be converting between text and seq
unnecessarily.

PR #1143, opened 2019-08-01
Followup to PR #1142

chick:
Fixes a threading bug in where lazy reading of file
caused a problem for multithreaded access to the that was read.

Changes all uses of io.Source to use new API
getText
getLines
getTextResource
getLinesResouce

Make style to only import FileUtils and not its methods
So code is more explicit as e.g. FileUtils.getText()

### DIFF
--- a/src/main/scala/firrtl/FileUtils.scala
+++ b/src/main/scala/firrtl/FileUtils.scala
@@ -149,6 +149,7 @@ object FileUtils {
   def getLinesResource(resourceName: String): Seq[String] = {
     val inputStream = getClass.getResourceAsStream(resourceName)
     val text = io.Source.fromInputStream(inputStream).getLines().toSeq
+    text.length  // This forces lazy buffer to reify, please suggest a better solution
     inputStream.close()
     text
   }
--- a/src/main/scala/firrtl/FileUtils.scala
+++ b/src/main/scala/firrtl/FileUtils.scala
@@ -98,7 +98,7 @@ object FileUtils {
     * @param fileName The file to read
     */
   def getLines(fileName: String): Seq[String] = {
-    val source = io.Source.fromFile(fileName)
+    val source = scala.io.Source.fromFile(fileName)
     val lines = source.getLines()
     source.close()
     lines.toSeq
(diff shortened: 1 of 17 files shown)


src/main/scala/firrtl/FileUtils.scala
```diff
@@ -149,6 +149,7 @@ object FileUtils {
   def getLinesResource(resourceName: String): Seq[String] = {
     val inputStream = getClass.getResourceAsStream(resourceName)
     val text = io.Source.fromInputStream(inputStream).getLines().toSeq
+    text.length  // This forces lazy buffer to reify, please suggest a better solution
```
seldridge: While I can't speak to correctness, I think using `toList` instead of `toSeq` should have a similar effect.

src/main/scala/firrtl/FileUtils.scala
```diff
@@ -149,6 +149,7 @@ object FileUtils {
   def getLinesResource(resourceName: String): Seq[String] = {
     val inputStream = getClass.getResourceAsStream(resourceName)
     val text = io.Source.fromInputStream(inputStream).getLines().toSeq
```
seldridge: While I realize that this is out scope of this PR, but I think it's in the spirit of this PR... can you switch all of these to use `scala.io.Source`? Annoyingly `io` can be overridden by other libraries/plugins and result in no such method errors if used this way.
```suggestion
     val text = scala.io.Source.fromInputStream(inputStream).getLines().toSeq
```

src/test/scala/firrtlTests/ProtoBufSpec.scala
```diff
@@ -30,8 +30,11 @@ class ProtoBufSpec extends FirrtlFlatSpec {
 
   for (FirrtlResourceTest(name, dir) <- firrtlResourceTests) {
     s"$name" should "work with Protobuf serialization and deserialization" in {
-      val stream = getClass.getResourceAsStream(s"$dir/$name.fir")
-      val circuit = parse(scala.io.Source.fromInputStream(stream).getLines.mkString("\n"))
+//      val stream = getClass.getResourceAsStream(s"$dir/$name.fir")
+//      val circuit = parse(scala.io.Source.fromInputStream(stream).getLines.mkString("\n"))
+//      stream.close()
```
seldridge: Remove comment.

src/main/scala/firrtl/FileUtils.scala
chick: Done!

src/test/scala/firrtlTests/ProtoBufSpec.scala
chick: Done!

PR #1145, opened 2019-08-02
FileUtils#getLines was broken in some multi-threading cases

chick:
Fix by forcing file to be processed before returning
Add tests that seems to show if fixed


ucbjrl:
Can we integrate this and #1144 or are they independent?

seldridge:
@ucbjrl: Yeah, they can be integrated. I can change the base branch of #1144 to this PR. @chick's fix here is equivalent, but I think it will cause the file to be iterated over twice this way. 

I'd like to add some more of the simple tests from #1144, too.

chick:
I think PR #1144 handles everything this does

Issue #1146, opened 2019-08-05
Disallow Zero Latency Memories


ekiwi:
When talking about "zero latency memories" I am referring to memories with `read-latency => 0` and `write-latency => 0`.

These memories are allowed by the current spec (from _5.10 Memories_):
> 4. A non-negative integer indicating the read latency
> 5. A non-negative integer indicating the write latency

These memories are problematic for multiple reasons:
1. they can lead to combinatorial loops that depend on the read/write address which may not be known at compile time
2. they are currently miscompiled by the Verilog Backend

Let us look at the following examples:
```.fir
circuit zero_latency_mem :
  module zero_latency_mem :
    input clock: Clock
    input reset: UInt<1>
    input raddr : UInt<1>
    input waddr : UInt<1>
    output rdata : UInt<4>
    mem m :
      data-type => UInt<4>
      depth => 2
      reader => r
      writer => w
      read-latency => 0
      write-latency => 0
      read-under-write => undefined

    rdata <= m.r.data
    m.r.addr <= raddr
    m.r.en <= UInt(1)
    m.r.clk <= clock

    m.w.addr <= waddr
    m.w.data <= mux(reset, UInt(0), add(rdata, UInt(1)))
    m.w.en <= UInt(1)
    m.w.clk <= clock
    m.w.mask <= UInt(1)
```

To see how this is miscompiled, run : `./utils/bin/firrtl -i zero_latency_mem.fir`
The relevant snippet from the output is:
```.v
  always @(posedge clock) begin
    if(m_w_en & m_w_mask) begin
      m[m_w_addr] <= m_w_data;
    end
  end
```

This seems to implement a write latency of `1` and not of `0`.

Now assuming that the `firrtl` would actually compile correctly, we would have a combinatorial loop whenever `eq(raddr, waddr)` which is not detected by the combinatorial loop checker.

For these reasons I would like to propose changing the spec and the compiler to require a minimum write latency of `1`.

ekiwi:
fixed in #1147

PR #1147, opened 2019-08-05
Check mems for legal latencies; ban zero write latency.

albert-magyar:
@azidar and @ekiwi, I looked at it after we were talking at the meeting, and it turns out that write latencies were not specified to be positive in the spec. I changed this and also added a check to `CheckHighForm`.


ekiwi:
Nice! See also: #1146

albert-magyar:
I'm going to wait for @azidar to sign off, since this has changes to `spec.tex`.

PR #1148, opened 2019-08-06
DRY check chirrtl

albert-magyar:
This streamlines `CheckChirrtl` and `CheckHighForm` to be implemented by one common trait, while also adding an explicit check that no Chirrtl constructs exist in a circuit during `CheckHighForm`.

Currently, `CheckChirrtl` does not actually check any Chirrtl-related features. The primary difference between `CheckChirrtl` and `CheckHighForm` is that they operate on IR and WIR, respectively. However, there are a number of things about this setup that are non-ideal:

- There is significant code duplication
- The current name of `CheckChirrtl` could lead people to infer that it checks `CDefMemories` and their ports, which is misleading.
- `CheckChirrtl` checks a semi-random subset of the things that `CheckHighForm` does, even though all of the checks in `CheckHighForm` are applicable to Chirrtl.
- Even though `CheckChirrtl` works on IR and `CheckHighForm` works on WIR, neither actually checks that that the circuit is actually in IR or WIR, which makes the distinction somewhat useless.
- `CheckHighForm` currently allows Chirrtl statements through, anyway.

Here, I extend `CheckHighForm` to *actually* check that there are no Chirrtl-only constructs in the circuit (which it does not currently do).

I dropped `InternalErrorSpec.scala` because it is actually a special-case check of one specific internal error; since this input now causes a (more desirable) `PassException`, the test is no longer meaningful.


src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -219,6 +232,18 @@ object CheckHighForm extends Pass {
   }
 }
 
+object CheckHighForm extends Pass with CheckHighFormLike {
+  class IllegalChirrtlMemException(info: Info, mname: String, name: String) extends PassException(
+    s"$info: [module $mname] Memory $name has not been properly lowered from Chirrtl IR.")
```
seldridge: @albert-magyar: This is defined in both `CheckHighForm` and `CheckHighFormLike`. I think this is just oversight?

albert-magyar: My idea was to move it from CheckHighFormLike to CheckHighForm so that there wouldn't be a paradoxically-named `CheckChirrtl.IllegalChirrtlMemException` in the class hierarchy. Looks like I left the dead code in :(

It's unfortunate that "overriding" inner classes doesn't give me nice goof-proof messages like defs.

Issue #1033, opened 2019-02-25
Missed Optimization Opportunity with Mux and Invalid


jackkoenig:
The following Chisel:
```scala
class DontCareMux extends Module {
  val io = IO(new Bundle {
    val en = Input(Bool())
    val out = Output(UInt(8.W))
  })
  io.out := Mux(io.en, 123.U, DontCare)
}
```
Gives the following FIRRTL:
```
circuit DontCareMux : 
  module DontCareMux : 
    input clock : Clock
    input reset : UInt<1>
    output io : {flip en : UInt<1>, out : UInt<8>}
    
    wire _T_1 : UInt<7>
    _T_1 is invalid
    node _T = mux(io.en, UInt<7>("h07b"), _T_1)
    io.out <= _T
```
Which results in the following Verilog:
```verilog
module DontCareMux(                                                                                                                                                                                                                                                                                                         
  input        clock,
  input        reset,
  input        io_en,
  output [7:0] io_out
);
  wire [6:0] _T;
  assign _T = io_en ? 7'h7b : 7'h0;
  assign io_out = {{1'd0}, _T};
endmodule
```

Unfortunately, because FIRRTL introduced the `0` instead of optimizing, synthesis cannot remove this mux.

**Type of issue**: other enhancement

**If the current behavior is a bug, please provide the steps to reproduce the problem:**
  * **If possible, please directly provide the FIRRTL file or a relevant fragment if the file is large**
  * **What is the current behavior?**
  * **What is the expected behavior?**
  * **Please tell us about your environment:**
    (examples)
    - version: `3.0-SNAPSHOT`
    - OS: `Linux knight 4.4.0-92-generic #115-Ubuntu SMP Thu Aug 10 09:04:33 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux`
    - verilator version (if relevant)

**What is the use case for changing the behavior?**

Improved QoR

**Impact**: no functional change

**Development Phase**: request

**Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. Stack Overflow, gitter, etc)

jackkoenig:
Note, handling muxes the same way `when`s are handled in `ExpandWhens` is not sufficient, this same issue can happen to `when`s:
```
    wire w : UInt<8>
    w is invalid
    io.out <= w
    when io.en :
      io.out <= UInt<7>("h07b")
      skip
```
This keeps the mux to Verilog.

[...]

PR #1150, opened 2019-08-07
Improve RemoveReset on Invalid Inits

### LINKED ISSUE #1149: Invalid should be respected by reset
When fed into a reset, invalid becomes 0, rather than disappearing.  This isn't incorrect, but it is an important missed opportunity for optimization.

This construct is particularly useful in the aggregate case, e.g., if I want to reset some fields of an aggregate but not others.  As it is, there is no way to do that besides `when reset` (right?), which is not ideal for async reset, since there, reset is not isomorphic to a mux.

Here's an example:

```
circuit Example :
  module Example :
    input clock : Clock
    input reset : UInt<1>
    input in : UInt<1>
    output out : UInt<1>

    wire whatever : UInt<1>
    whatever is invalid

    reg foo : UInt<1>, clock with : (reset => (reset, whatever))
    foo <= in
    out <= foo
```

```
  always @(posedge clock) begin
    if (reset) begin
      foo <= 1'h0;
    end else begin
      foo <= in;
    end
  end
```

**Type of issue**: feature request

**Impact**:  unknown

seldridge:
Adds a case for `RemoveReset` where an invalid register initialization will result in a self connection without mux generation. This does a preprocessing pass to build up a set of all invalidations in a module.

I've identified one further optimization ([detailed in a comment below](https://github.com/freechipsproject/firrtl/pull/1150#issuecomment-518968556)) that removes the additional pass to statefully process the module, but requires a guarantee that invalidations happen before the mux insertion.

For the @aswaterman provided example in #1149, this will produce:

```verilog
module Example(
  input   clock,
  input   rst,
  input   in,
  output  out
);
  reg  foo;
  assign out = foo;
  always @(posedge clock) begin
    foo <= in;
  end
endmodule

```

## Todo
- [x] A real test
- [x] Rebase to clean things up

## Metadata
- (Should) fix #1149.

### DIFF
--- a/src/main/scala/firrtl/transforms/RemoveReset.scala
+++ b/src/main/scala/firrtl/transforms/RemoveReset.scala
@@ -5,8 +5,10 @@ package transforms
 
 import firrtl.ir._
 import firrtl.Mappers._
+import firrtl.traversals.Foreachers._
+import firrtl.WrappedExpression.we
 
-import scala.collection.mutable
+import scala.collection.{immutable, mutable}
 
 /** Remove Synchronous Reset
   *
@@ -18,10 +20,30 @@ class RemoveReset extends Transform {
 
   private case class Reset(cond: Expression, value: Expression)
 
+  /** Return an immutable set of all invalid expressions in a module
+    * @param m a module
+    */
+  private def computeInvalids(m: DefModule): immutable.Set[WrappedExpression] = {
+    val invalids = mutable.HashSet.empty[WrappedExpression]
+
+    def onStmt(s: Statement): Unit = s match {
+      case IsInvalid(_, expr)                                 => invalids += we(expr)
+      case Connect(_, lhs, rhs) if invalids.contains(we(rhs)) => invalids += we(lhs)
+      case other                                              => other.foreach(onStmt)
+    }
+
+    m.foreach(onStmt)
+    invalids.toSet
+  }
+
   private def onModule(m: DefModule): DefModule = {
     val resets = mutable.HashMap.empty[String, Reset]
+    val invalids = computeInvalids(m)
     def onStmt(stmt: Statement): Statement = {
       stmt match {
+        /* A register is initialized to an invalid expression */
+        case reg @ DefRegister(_, _, _, _, _, init) if invalids.contains(we(init)) =>
+          reg.copy(reset = Utils.zero, init = WRef(reg))
         case reg @ DefRegister(_, rname, _, _, reset, init)
             if reset != Utils.zero && reset.tpe != AsyncResetType =>
           // Add register reset to map
--- a/src/test/scala/firrtlTests/transforms/RemoveResetSpec.scala
+++ b/src/test/scala/firrtlTests/transforms/RemoveResetSpec.scala
@@ -1,0 +1,115 @@
+// See LICENSE for license details.
+
+package firrtlTests.transforms
+
+import org.scalatest.GivenWhenThen
+
+import firrtlTests.FirrtlFlatSpec
+import firrtlTests.FirrtlCheckers._
+
+import firrtl.{CircuitState, WRef}
+import firrtl.ir.{Connect, Mux}
+import firrtl.stage.{FirrtlCircuitAnnotation, FirrtlSourceAnnotation, FirrtlStage}
+
+class RemoveResetSpec extends FirrtlFlatSpec with GivenWhenThen {
+
+  private def toLowFirrtl(string: String): CircuitState = {
+    When("the circuit is compiled to low FIRRTL")
+    (new FirrtlStage)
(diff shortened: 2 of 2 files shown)


aswaterman:
Awesome!  I also checked that it works as expected when resetting only a portion of an aggregate.

```
circuit Example :
  module Example :
    input clock : Clock
    input reset : UInt<1>
    input in :  {a : UInt<1>[2], b : UInt<1>}
    output out :  {a : UInt<1>[2], b : UInt<1>}

    wire whatever : {a : UInt<1>[2], b : UInt<1>}
    whatever is invalid
    whatever.a[1] <= UInt<1>(0)

    reg foo :  {a : UInt<1>[2], b : UInt<1>}, clock with : (reset => (reset, whatever))
    foo <= in
    out <= foo
```

aswaterman:
```
  always @(posedge clock) begin
    foo_a_0 <= in_a_0;
    if (reset) begin
      foo_a_1 <= 1'h0;
    end else begin
      foo_a_1 <= in_a_1;
    end
    foo_b <= in_b;
  end
```

seldridge:
Note: there is a further optimization possible that removes the additional module traversal if there's a guarantee that the invalidation happens before the connection that will be converted to a mux. If that lemma holds, then you can remove the pending mux connection encoded in the `resets` field. I can't currently reason through if this holds... Also, there's a cognitive load as this is statefully processing the module.

This improved implementation is on [branch `issue-1149-stateful`](https://github.com/freechipsproject/firrtl/compare/master..issue-1149-stateful) for this approach. I'm fine with either one, so long as the lemma above holds.

aswaterman:
I've found a case where this doesn't do what I'd expect:

```
circuit Example : 
  module Example : 
    input clock : Clock
    input reset : UInt<1>
    input in : {cfg : {l : UInt<1>, res : UInt<2>, a : UInt<2>, x : UInt<1>, w : UInt<1>, r : UInt<1>}, addr : UInt<30>}
    output out : {cfg : {l : UInt<1>, res : UInt<2>, a : UInt<2>, x : UInt<1>, w : UInt<1>, r : UInt<1>}, addr : UInt<30>}

    wire _T_84 : {cfg : {l : UInt<1>, res : UInt<2>, a : UInt<2>, x : UInt<1>, w : UInt<1>, r : UInt<1>}, addr : UInt<30>}
    _T_84 is invalid
    _T_84.cfg.a <= UInt<2>(0)


    wire _T_92 : {cfg : {l : UInt<1>, res : UInt<2>, a : UInt<2>, x : UInt<1>, w : UInt<1>, r : UInt<1>}, addr : UInt<30>}
    _T_92 is invalid
    _T_92 <- _T_84

    reg reg_pmp : {cfg : {l : UInt<1>, res : UInt<2>, a : UInt<2>, x : UInt<1>, w : UInt<1>, r : UInt<1>}, addr : UInt<30>}, clock with : (reset => (reset, _T_92))
    reg_pmp <= in
    out <= reg_pmp
```

The result is that all fields of `reg_pmp` get reset.

The extra wire in the middle appears to foil things.  If, in the `reg reg_pmp` declaration, I replace `_T_92` with `_T_84`, it appears to work how I'd expect.

seldridge:
@aswaterman: Thanks for the test case. I've simplified that and rebased to include the failure. Let me see if I can fix this inside `RemoveReset` by doing an improved tracking of the connectivity. Alternatively, and perhaps more cleanly, this may beg for earlier constant prop.

seldridge:
Updated to (in the data structure) propagate invalidates across connects (and only connects). This is very narrowly addressing the @aswaterman situation above. The resulting Verilog is now:

```verilog
always @(posedge clock) begin
  reg_pmp_cfg_l <= in_cfg_l;
  reg_pmp_cfg_res <= in_cfg_res;
  if (reset) begin
    reg_pmp_cfg_a <= 2'h0;
  end else begin
    reg_pmp_cfg_a <= in_cfg_a;
  end
  reg_pmp_cfg_x <= in_cfg_x;
  reg_pmp_cfg_w <= in_cfg_w;
  reg_pmp_cfg_r <= in_cfg_r;
  reg_pmp_addr <= in_addr;
end
```

seldridge:
Thinking about this more, the @jackkoenig https://github.com/freechipsproject/firrtl/issues/1033#issuecomment-467177030 is likely a better and more general solution to the underlying problem here.

aswaterman:
Cool!

I agree that constpropping ValidIf is a good direction. We probably should
not go wild with those optimizations at first, as they are certain to
reveal bugs in people’s designs, particularly old chisel2 code.

seldridge:
Currently blocked waiting on the results of SiFive regressions.

seldridge:
I have confirmation that SiFive regressions pass with this PR cherry-picked on the current Rocket Chip FIRRTL commit. Going to go ahead and merge.

PR #1152, opened 2019-08-09
Make "build" the default make target rather than "clean"

aswaterman:
It's a little surprising that typing "make" in the firrtl directory deletes firrtl :)


ucbjrl:
We should review this (and any other chisel eco-system Makefiles) and clean them up, make them consistent, or delete them.

PR #1157, opened 2019-08-14
Refactor exceptions to remove stack trace from user errors

jackkoenig:
This has driven me crazy for a long time so I finally decided to fix it.

Refactored `FIRRTLException` into `FirrtlInternalException` and added `FirrtlUserError` which does not create a stack trace. I think further refinement would be good (like there isn't really a need to say which exception fired), but this is *much* better and I hope can make it into 1.2

This now means that for user errors you will see something like:
```
Exception in thread "main" firrtl.passes.CheckTypes$InvalidConnect: : [module test]  Type mismatch in 'y <= x'.
  y: SInt<1>
  x: UInt<1>
```

Instead of:
```
Exception in thread "main" firrtl.passes.CheckTypes$InvalidConnect: : [module test]  Type mismatch in 'y <= x'.
  y: SInt<1>
  x: UInt<1>
	at firrtl.passes.CheckTypes$.check_types_s$1(Checks.scala:489)
	at firrtl.passes.CheckTypes$.$anonfun$run$55(Checks.scala:538)
	at firrtl.passes.CheckTypes$.$anonfun$run$55$adapted(Checks.scala:538)
	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:58)
	at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:51)
	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:47)
	at firrtl.ir.Block.foreachStmt(IR.scala:363)
	at firrtl.traversals.Foreachers$StmtForMagnet$$anon$1.foreach(Foreachers.scala:18)
	at firrtl.traversals.Foreachers$StmtForeach$.foreach$extension(Foreachers.scala:35)
	at firrtl.passes.CheckTypes$.check_types_s$1(Checks.scala:538)
	at firrtl.passes.CheckTypes$.$anonfun$run$58(Checks.scala:541)
	at firrtl.passes.CheckTypes$.$anonfun$run$58$adapted(Checks.scala:541)
	at firrtl.ir.Module.foreachStmt(IR.scala:689)
	at firrtl.traversals.Foreachers$ModuleForMagnet$$anon$11.foreach(Foreachers.scala:79)
	at firrtl.traversals.Foreachers$ModuleForeach$.foreach$extension(Foreachers.scala:92)
	at firrtl.passes.CheckTypes$.$anonfun$run$57(Checks.scala:541)
	at firrtl.passes.CheckTypes$.$anonfun$run$57$adapted(Checks.scala:541)
	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:58)
	at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:51)
	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:47)
	at firrtl.passes.CheckTypes$.run(Checks.scala:541)
	at firrtl.passes.Pass.execute(Passes.scala:23)
	at firrtl.passes.Pass.execute$(Passes.scala:21)
	at firrtl.passes.CheckTypes$.execute(Checks.scala:271)
	at firrtl.Transform.$anonfun$runTransform$2(Compiler.scala:216)
	at firrtl.Utils$.time(Utils.scala:183)
	at firrtl.Transform.runTransform(Compiler.scala:216)
	at firrtl.SeqTransformBased.$anonfun$runTransforms$1(Compiler.scala:281)
	at scala.collection.LinearSeqOptimized.foldLeft(LinearSeqOptimized.scala:122)
	at scala.collection.LinearSeqOptimized.foldLeft$(LinearSeqOptimized.scala:118)
	at scala.collection.immutable.List.foldLeft(List.scala:85)
	at firrtl.SeqTransformBased.runTransforms(Compiler.scala:281)
	at firrtl.SeqTransformBased.runTransforms$(Compiler.scala:280)
	at firrtl.SeqTransform.runTransforms(Compiler.scala:285)
	at firrtl.SeqTransform.execute(Compiler.scala:291)
	at firrtl.Transform.$anonfun$runTransform$2(Compiler.scala:216)
	at firrtl.Utils$.time(Utils.scala:183)
	at firrtl.Transform.runTransform(Compiler.scala:216)
	at firrtl.Compiler.$anonfun$compile$2(Compiler.scala:483)
	at scala.collection.LinearSeqOptimized.foldLeft(LinearSeqOptimized.scala:122)
	at scala.collection.LinearSeqOptimized.foldLeft$(LinearSeqOptimized.scala:118)
	at scala.collection.immutable.List.foldLeft(List.scala:85)
	at firrtl.Compiler.$anonfun$compile$1(Compiler.scala:481)
	at firrtl.Utils$.time(Utils.scala:183)
	at firrtl.Compiler.compile(Compiler.scala:481)
	at firrtl.Compiler.compile$(Compiler.scala:477)
	at firrtl.VerilogCompiler.compile(LoweringCompilers.scala:164)
	at firrtl.stage.phases.Compiler.f$1(Compiler.scala:91)
	at firrtl.stage.phases.Compiler.$anonfun$internalTransform$2(Compiler.scala:95)
	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:233)
	at scala.collection.mutable.ResizableArray.foreach(ResizableArray.scala:58)
	at scala.collection.mutable.ResizableArray.foreach$(ResizableArray.scala:51)
	at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:47)
	at scala.collection.TraversableLike.map(TraversableLike.scala:233)
	at scala.collection.TraversableLike.map$(TraversableLike.scala:226)
	at scala.collection.AbstractTraversable.map(Traversable.scala:104)
	at firrtl.stage.phases.Compiler.internalTransform(Compiler.scala:95)
	at firrtl.stage.phases.Compiler.internalTransform(Compiler.scala:45)
	at firrtl.options.Translator.transform(Phase.scala:140)
	at firrtl.options.Translator.transform$(Phase.scala:140)
	at firrtl.stage.phases.Compiler.transform(Compiler.scala:45)
	at firrtl.options.phases.DeletedWrapper.internalTransform(DeletedWrapper.scala:32)
	at firrtl.options.phases.DeletedWrapper.internalTransform(DeletedWrapper.scala:15)
	at firrtl.options.Translator.transform(Phase.scala:140)
	at firrtl.options.Translator.transform$(Phase.scala:140)
	at firrtl.options.phases.DeletedWrapper.transform(DeletedWrapper.scala:15)
	at firrtl.stage.FirrtlStage.$anonfun$run$1(FirrtlStage.scala:27)
	at scala.collection.LinearSeqOptimized.foldLeft(LinearSeqOptimized.scala:122)
	at scala.collection.LinearSeqOptimized.foldLeft$(LinearSeqOptimized.scala:118)
	at scala.collection.immutable.List.foldLeft(List.scala:85)
	at firrtl.stage.FirrtlStage.run(FirrtlStage.scala:27)
	at firrtl.options.Stage$$anon$1.transform(Stage.scala:43)
	at firrtl.options.Stage$$anon$1.transform(Stage.scala:43)
	at firrtl.options.phases.DeletedWrapper.internalTransform(DeletedWrapper.scala:32)
	at firrtl.options.phases.DeletedWrapper.internalTransform(DeletedWrapper.scala:15)
	at firrtl.options.Translator.transform(Phase.scala:140)
	at firrtl.options.Translator.transform$(Phase.scala:140)
	at firrtl.options.phases.DeletedWrapper.transform(DeletedWrapper.scala:15)
	at firrtl.options.Stage.$anonfun$transform$5(Stage.scala:46)
	at scala.collection.LinearSeqOptimized.foldLeft(LinearSeqOptimized.scala:122)
	at scala.collection.LinearSeqOptimized.foldLeft$(LinearSeqOptimized.scala:118)
	at scala.collection.immutable.List.foldLeft(List.scala:85)
	at firrtl.options.Stage.$anonfun$transform$3(Stage.scala:46)
	at logger.Logger$.$anonfun$makeScope$2(Logger.scala:168)
	at scala.util.DynamicVariable.withValue(DynamicVariable.scala:58)
	at logger.Logger$.makeScope(Logger.scala:166)
	at firrtl.options.Stage.transform(Stage.scala:46)
	at firrtl.options.Stage.execute(Stage.scala:57)
	at firrtl.options.StageMain.main(Stage.scala:70)
	at firrtl.stage.FirrtlMain.main(FirrtlStage.scala)
```

### DIFF
--- a/src/main/scala/firrtl/FirrtlException.scala
+++ b/src/main/scala/firrtl/FirrtlException.scala
@@ -1,0 +1,43 @@
+// See LICENSE for license details.
+
+package firrtl
+
+import scala.util.control.NoStackTrace
+
+@deprecated("External users should use either FirrtlUserException or their own hierarchy", "1.2")
+object FIRRTLException {
+  def defaultMessage(message: String, cause: Throwable) = {
+    if (message != null) {
+      message
+    } else if (cause != null) {
+      cause.toString
+    } else {
+      null
+    }
+  }
+}
+@deprecated("External users should use either FirrtlUserException or their own hierarchy", "1.2")
+class FIRRTLException(val str: String, cause: Throwable = null)
+  extends RuntimeException(FIRRTLException.defaultMessage(str, cause), cause)
+
+/** Exception indicating user error
+  *
+  * These exceptions indicate a problem due to bad input and thus do not include a stack trace.
+  * This can be extended by custom transform writers.
+  */
+class FirrtlUserException(message: String, cause: Throwable = null)
+  extends RuntimeException(message, cause) with NoStackTrace
+
+/** Wraps exceptions from CustomTransforms so they can be reported appropriately */
+case class CustomTransformException(cause: Throwable) extends Exception("", cause)
+
+/** Exception indicating something went wrong *within* Firrtl itself
+  *
+  * These exceptions indicate a problem inside the compiler and include a stack trace to help
+  * developers debug the issue.
+  *
+  * This class is private because these are issues within Firrtl itself. Exceptions thrown in custom
+  * transforms are treated differently and should thus have their own structure
+  */
+private[firrtl] class FirrtlInternalException(message: String, cause: Throwable = null)
+  extends Exception(message, cause)
--- a/src/main/scala/firrtl/Compiler.scala
+++ b/src/main/scala/firrtl/Compiler.scala
@@ -43,7 +43,7 @@ case class CircuitState(
   def getEmittedCircuit: EmittedCircuit = emittedCircuitOption match {
     case Some(emittedCircuit) => emittedCircuit
     case None =>
-      throw new FIRRTLException(s"No EmittedCircuit found! Did you delete any annotations?\n$deletedAnnotations")
+      throw new FirrtlInternalException(s"No EmittedCircuit found! Did you delete any annotations?\n$deletedAnnotations")
   }
 
   /** Helper function for extracting emitted components from annotations */
@@ -314,9 +314,6 @@ trait Emitter extends Transform {
   /** An output suffix to use if the output of this [[Emitter]] was written to a file */
   def outputSuffix: String
 }
-
-/** Wraps exceptions from CustomTransforms so they can be reported appropriately */
-case class CustomTransformException(cause: Throwable) extends Exception("", cause)
 
 object CompilerUtils extends LazyLogging {
   /** Generates a sequence of [[Transform]]s to lower a Firrtl circuit
--- a/src/main/scala/firrtl/Parser.scala
+++ b/src/main/scala/firrtl/Parser.scala
@@ -9,7 +9,7 @@ import firrtl.ir._
 import firrtl.Utils.time
 import firrtl.antlr.{FIRRTLParser, _}
 
-class ParserException(message: String) extends FIRRTLException(message)
+class ParserException(message: String) extends FirrtlUserException(message)
 
 case class ParameterNotSpecifiedException(message: String) extends ParserException(message)
 case class ParameterRedefinedException(message: String) extends ParserException(message)
--- a/src/main/scala/firrtl/Utils.scala
+++ b/src/main/scala/firrtl/Utils.scala
@@ -13,20 +13,6 @@ import scala.util.matching.Regex
 import firrtl.annotations.{ReferenceTarget, TargetToken}
 import _root_.logger.LazyLogging
 
-object FIRRTLException {
-  def defaultMessage(message: String, cause: Throwable) = {
-    if (message != null) {
-      message
-    } else if (cause != null) {
-      cause.toString
-    } else {
-      null
-    }
-  }
-}
-class FIRRTLException(val str: String, cause: Throwable = null)
-  extends RuntimeException(FIRRTLException.defaultMessage(str, cause), cause)
-
 object seqCat {
   def apply(args: Seq[Expression]): Expression = args.length match {
     case 0 => Utils.error("Empty Seq passed to seqcat")
@@ -434,7 +420,7 @@ object Utils extends LazyLogging {
   }
    
 // =================================
-  def error(str: String, cause: Throwable = null) = throw new FIRRTLException(str, cause)
+  def error(str: String, cause: Throwable = null) = throw new FirrtlInternalException(str, cause)
 
 //// =============== EXPANSION FUNCTIONS ================
   def get_size(t: Type): Int = t match {
@@ -632,7 +618,7 @@ object Utils extends LazyLogging {
     case EmptyExpression => root
   }
 
-  case class DeclarationNotFoundException(msg: String) extends FIRRTLException(msg)
+  case class DeclarationNotFoundException(msg: String) extends FirrtlUserException(msg)
 
   /** Gets the root declaration of an expression
     *
--- a/src/main/scala/firrtl/annotations/AnnotationUtils.scala
+++ b/src/main/scala/firrtl/annotations/AnnotationUtils.scala
@@ -11,13 +11,13 @@ import firrtl.annotations.AnnotationYamlProtocol._
 
 import firrtl.ir._
 
-case class InvalidAnnotationFileException(file: File, cause: Throwable = null)
-  extends FIRRTLException(s"$file, see cause below", cause)
-case class InvalidAnnotationJSONException(msg: String) extends FIRRTLException(msg)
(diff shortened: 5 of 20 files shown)


src/main/scala/firrtl/FirrtlException.scala
```diff
@@ -0,0 +1,43 @@
...
+class FIRRTLException(val str: String, cause: Throwable = null)
+  extends RuntimeException(FIRRTLException.defaultMessage(str, cause), cause)
+
+/** Exception indicating user error
+  *
+  * These exceptions indicate a problem due to bad input and thus do not include a stack trace.
+  * This can be extended by custom transform writers.
+  */
+class FirrtlUserException(message: String, cause: Throwable = null)
+  extends RuntimeException(message, cause) with NoStackTrace
```
seldridge: User error is something that's common across Stages. I think it makes sense to change this to `UserException` and move it under the `firrtl.options` package.

Going one step further, what should be the behavior if a user runs a `StageMain` from the command line and gets a `UserException`?

For an `OptionsException` (which I think should probably be changed going forward to a `UserException`), this is [caught by the `StageMain`, the message is printed as a "dramatic error", and you get a `System.exit(1)`](https://github.com/freechipsproject/firrtl/blob/master/src/main/scala/firrtl/options/Stage.scala#L72). I think that's the ideal behavior for a `UserException` generated from the FIRRTL fat jar. That may necessitate more verbose exception messages that tell which transform errored as you lose that information from the exception message in this case.

And for more explanation: the reason why this happens for `StageMain` and not for `Stage` is that if you're running a `StageMain` it's clear that this is a fat jar the user called from the command line. If it's a `Stage`, then it's being run by someone else (likely a different `StageMain`) and it should be the responsibility of the caller to figure out how to print this.

chick:
This looks great.
Have you considered adding an annotation that will change the behavior to deliver the stack trace. 
I think this is the far better default, but there are times where you might want to see the stack trace to figure out the etiology of the error.

src/main/scala/firrtl/FirrtlException.scala
jackkoenig: Yeah I see what you mean that `OptionsException` is a `UserException`. I'm not sure if it makes sense to have only `firrtl.options.UserException` since that feels specific to arguments while `firrtl.FirrtlUserException` seems more generic (for example, erroring on a type in the FIRRTL AST doesn't feel like its exception supertype belongs in the `options` package). That being said, perhaps `firrtl.options.UserException` should extend `firrtl.FirrtlUserException`.

Also perhaps I should renaming it from `firrtl.FirrtlUserException` to `firrtl.UserException`

jackkoenig:
Interesting idea @chick. The easiest way I see to support such functionality is to make `FirrtlUserException` still capture stack traces (get rid of the `NoStackTrace` mixin), but then catch and suppress the stack trace. This seems related to @seldridge's comment `I'm thinking those should strip the exception message and color print the message.`.

The only con here other `StageMains` *knowing* to suppress the stack traces (by never having a stack trace, it makes it very clear that the stack trace is not usually useful). That being said, maybe just overridding `.toString` to exclude the stack trace will do.

Thoughts?

src/main/scala/firrtl/FirrtlException.scala
seldridge: I think `firrtl.options` was just named too early... one facet of this is option handling, but it's really a library for Stages and Phases and, with a few exceptions (like `AnnotationSeq`, `DiGraph`), is independent of FIRRTL.

Basically, I think a "user exception" is something that any Stage or Phase should be able to throw. Making this `firrtl.UserException` or `firrtl.options.UserException` communicates that better.

seldridge:
@jackkoenig: your idea of catching the exception and printing the stack trace (in a format controlled by presence/absence of annotations) seems like the best approach. That does need an additional `Throwable` that `StageMain`s are looking for. #1158 adds `StageError` which will cause a `StageMain` to `System.exit(code)`. https://github.com/freechipsproject/chisel3/pull/1160 is an implementation of doing it this way for Chisel's stack trace truncation.

This puts the onus of dealing with the exception as local as possible as opposed to relying on some global `Stage` or `StageMain` behavior.

jackkoenig:
So the immediate solution for when a developer needs a stack trace from a user error is to just change the code to remove the `with NoStackTrace`. A better long term solution is desirable that unifies with stage/phase, but it seems that `FirrtlUserError` should just extend it so I don't think there's any worry. This will make errors much better for users in the meantime.

PR #1158, opened 2019-08-14
Programmed Stage Death

seldridge:
This is part of a fix for https://github.com/freechipsproject/chisel3/issues/1159.

This provides a solution to the following problem: assume you have a deeply nested Stage/Phase system. Stage `foo.A` calls Stage `bar.B` calls Stage `baz.C` where `{foo, bar, baz}` are all different third-party libraries. A Phase is Stage `C` throws an exception indicating that it cannot continue. How do you manipulate the exception message in such a way that it's presentable to users?

This processing (pretty printing, stack trace trimming) needs to happen inside Stage `C` as only it understands that exception; `foo` and `bar` are other libraries you're composing with it. However, after pretty printing, it can do one of two things:

1. Rethrow the exception
2. Swallow the exception

For (1) then you wind up with an ugly, unprocessed stack trace in the top-level Stage `A`. For (2) then Stages `{B, C}` continue on thinking everything is working fine when it's not. (2) is the issue of https://github.com/freechipsproject/chisel3/issues/1159.

This PR provides a third option. This adds a `private[options] class StageError extends Error` and a public factory method to generate it, `Stage.apoptosis`. `StageMain` is modified to catch `StageError` and do a hard `System.exit` with the provided exit code.

This effectively adds a third option to the two above: 

3. Throw an exception that only `StageMain` can catch

So long as there's a top-level `StageMain` (which there is supposed to be), then you can both allow for local processing of an exception as well as hard exit behavior.

This assumes that there is always a top-level `StageMain` which, in the post-Stage/Phase world is supposed to be the case. During the deprecation of `Driver`s, they will need to kludgily catch all `Error`.


seldridge:
Thinking about this more, it may be overly restrictive to make `StageError` private. With this private, it follows that any Scala program that wants to use Chisel or FIRRTL should be a `StageMain`.

**PR text above is updated to reflect this.**

schoeberl:
OT to @jackkoenig: I don’t know if I have your email address, so could you please drop me an email to masca@dtu.dk <mailto:masca@dtu.dk> as I would like to contact you with a non-public question.

Sorry for misusing this issue.

Cheers,
Martin

PR #1164, opened 2019-08-22
allow GroupComponents to handle invalids

abejgonzalez:
Small fix to allow grouping with components that use `invalid`.


colinschmidt:
This smells like something I've seen before: https://github.com/freechipsproject/firrtl/tree/bugfix-grouping-isinvalids

abejgonzalez:
So we should merge that in instead of this? Is one solution better than the other?

colinschmidt:
@azidar I think you worked on or were consulted on both of these fixes, do you have an opinion?

albert-magyar:
This just avoids crashing on invalidated stuff in some cases, while #1276 actually pulls in the invalidates in a manner resembling connects. I'm going to close this for now.

PR #1172, opened 2019-09-04
Attributes

grebe:
Adds verilog attributes to firrtl. Re-uses much of the DocString infrastructure in the emitter.

Supersedes #1040, which also added ifdefs. That's a big can of worms, so let's start with attributes first.


src/main/scala/firrtl/AddDescriptionNodes.scala
```diff
@@ -6,7 +6,11 @@ import firrtl.ir._
 import firrtl.annotations._
 import firrtl.Mappers._
 
-case class DescriptionAnnotation(named: Named, description: String) extends Annotation {
+sealed trait DescriptionType
+case object DocStringDescription extends DescriptionType
+case object AttributeDescription extends DescriptionType
+
+case class DescriptionAnnotation(named: Named, description: String, tpe: DescriptionType) extends Annotation {
```
chick: This seems to be a visible API feature. Shouldn't this at least have a default value for `tpe` so it won't break potential uses in the wild?

johnsbrew:
This is a nice feature, do you plan to come with a standardized chisel API as well?  Or is it up to the user to add annotations depending on its own use case ?

As @chick mentioned I think it is worth a bit more documentation :)

seldridge:
@johnsbrew: Chisel API would (should?) follow. Those tend to lag for two reasons: (1) once the annotations are in FIRRTL, the user that added it may not be super motivated to expose it as they are likely already using it and (2) the Chisel APIs tend to have a fair amount of back-and-forth to get right. See https://github.com/freechipsproject/chisel3/pull/838 for an example.

All that said, once this lands, then the anybody can start prototyping a Chisel API. Definitely needed. :)

src/main/scala/firrtl/AddDescriptionNodes.scala
grebe: Fair point. I'm now wondering if it would be better to have different annotations for each `DescriptionType` rather than a `tpe` field. But in that case, I think I'd want to change the name of `DescriptionAnnotation`

grebe:
@johnsbrew: Yeah, a chisel API PR will follow soon unless I get negative feedback here. I don't think it will be very complicated, I would think it's just something like
```scala
object Attribute {
  def apply[T <: Data](obj: T, attr: String): T = {
    // validate attr
    // add annotation
    return obj
  }
}
```

kammoh:
Thank you @grebe for this PR! IMO this is a crucial mechanism for clean injection of attributes and comments down to Verilog. I'm using it to allow propagation of `dontTouch` annotations to Verilog `DONT_TOUCH` attributes. The resulting `Transform` is quite nifty and I'd be more than happy to make a PR to have it as an optional standard transform. But first I need this PR to be committed. 
I also agree with @grebe in using different annotations for each DescriptionType (instead of `tpe` arguemnt), e.g. `AttributeAnnotation(named: Named, description: String)`
Any chance we can see this in master soon? :)

kammoh:
Hopefully we're going to let go of the `tpe` field, but just wanna add that it currently serializes into blanks in anno.json file, no matter what the value is.

grebe:
@kammoh it looks to me like it no longer serializes to blanks, but I'm not really sure why it would have been doing so before.

grebe:
@chick I added some low-level scaladoc. I think more user-facing documentation would belong with the chisel API.

src/main/scala/firrtl/AddDescriptionNodes.scala
```diff
@@ -6,31 +6,79 @@ import firrtl.ir._
...
 
-case class DescriptionAnnotation(named: Named, description: String) extends Annotation {
-  def update(renames: RenameMap): Seq[DescriptionAnnotation] = {
+/**
+  * A base trait for `Annotation`s that describe a `FirrtlNode`.
+  * Usually, we would like to emit these descriptions in some way.
+  */
+sealed trait DescriptionAnnotation extends Annotation {
+  val named: Named
+  val description: String
```
jackkoenig: Is there a reason these are vals instead of defs?

src/main/scala/firrtl/AddDescriptionNodes.scala
```diff
@@ -71,44 +125,85 @@ class AddDescriptionNodes extends Transform {
...
+    * Similarly, multiple attributes on the same object get merged into one attribute with attributes separated by
+    * commas.
+    * @param descs List of `Description`s that are modifying the same object
+    * @return List of `Description`s with some descriptions merged
+    */
+  def mergeDescriptions(descs: Seq[Description]): Seq[Description] = {
+    val (docs: Seq[DocString], nodocs) = descs.partition(_ match {
+      case _: DocString => true
+      case _ => false
+    })
```
jackkoenig: This is a minor stylistic nit, but `_ match {` is never (? I think) necessary, you can just write:
```scala
    val (docs: Seq[DocString], nodocs) = descs.partition {
      case _: DocString => true
      case _ => false
    }
```
Also I really wish Scala had like a `collectPartition` where you could just provide the `true` case and the `false` would go to the other. Alas.

src/main/scala/firrtl/AddDescriptionNodes.scala
```diff
@@ -71,44 +125,85 @@ class AddDescriptionNodes extends Transform {
...
 
     val compMap = annos.collect {
-      case DescriptionAnnotation(ComponentName(comp, ModuleName(mod, CircuitName(circ))), desc) =>
-        (mod, comp, desc)
-    }.groupBy(_._1).mapValues(_.groupBy(_._2).mapValues(_.map(_._3)))
+      case DocStringAnnotation(ComponentName(c, ModuleName(m, _)), desc) =>
+        (m, c, DocString(StringLit.unescape(desc)))
+      case AttributeAnnotation(ComponentName(c, ModuleName(m, _)), desc) =>
+        (m, c, Attribute(StringLit.unescape(desc)))
+    }.groupBy(_._1).mapValues(_.groupBy(_._2).mapValues(_.map(_._3)).mapValues(mergeDescriptions))
```
jackkoenig: 👀 Is this something we could make more clear? Maybe just in a function?

src/main/scala/firrtl/AddDescriptionNodes.scala
grebe: Thanks, this is a good habit I should get into and looks much nicer.

src/main/scala/firrtl/AddDescriptionNodes.scala
grebe: Scala 2.13 has `groupMap`, which seems like it's defined to deal with some of this.

I broke it down a bit and added some comments, which hopefully makes it a bit easier to understand.

grebe:
Ugh, the history is kind of messed up now b/c I left a file from something else in history before merging master. I rebased to remove it and now the history is ugly. Sorry!

src/main/scala/firrtl/AddDescriptionNodes.scala
```diff
@@ -7,31 +7,79 @@ import firrtl.annotations._
...
 import firrtl.options.{Dependency, PreservesAll}
 
-case class DescriptionAnnotation(named: Named, description: String) extends Annotation {
-  def update(renames: RenameMap): Seq[DescriptionAnnotation] = {
+/**
+  * A base trait for `Annotation`s that describe a `FirrtlNode`.
+  * Usually, we would like to emit these descriptions in some way.
+  */
+sealed trait DescriptionAnnotation extends Annotation {
+  def named: Named
```
jackkoenig, line 15: Should we be using `Named` or `Target`?

src/main/scala/firrtl/AddDescriptionNodes.scala
```diff
@@ -89,44 +143,97 @@ class AddDescriptionNodes extends Transform with PreservesAll[Transform] {
...
-    stmt.map(onStmt(compMap)) match {
-      case d: IsDeclaration if compMap.contains(d.name) =>
-        DescribedStmt(DocString(StringLit.unescape(compMap(d.name).mkString("\n\n"))), d)
-      case other => other
+  def onStmt(compMap: Map[String, Seq[Description]])(stmt: Statement): Statement = {
+    val s = stmt.map(onStmt(compMap))
+    val sname = s match {
+      case d: IsDeclaration => Some(d.name)
+      case _ => None
+    }
```
jackkoenig, line 151: Minor style nit but this could just be
```suggestion
    s.collect { case d: IsDeclaration => d.name }
```

src/main/scala/firrtl/AddDescriptionNodes.scala
```diff
@@ -89,44 +143,97 @@ class AddDescriptionNodes extends Transform with PreservesAll[Transform] {
...
+    val descs = sname.flatMap({ case name =>
+      compMap.get(name)
+    })
```
jackkoenig, line 154:
```suggestion
    val descs = sname.flatMap(compMap.get)
```

src/main/scala/firrtl/AddDescriptionNodes.scala
```diff
@@ -89,44 +143,97 @@ class AddDescriptionNodes extends Transform with PreservesAll[Transform] {
...
-        case p @ Port(_, name, _, _) if compMap.contains(name) =>
-          name -> DocString(StringLit.unescape(compMap(name).mkString("\n\n")))
-      }.toMap)
-    }
+    val compMap = compMaps.getOrElse(mod.name, Map())
+    val newMod = mod.mapStmt(onStmt(compMap))
+    val portDesc = mod.ports.collect {
+      case p @ Port(_, name, _, _) if compMap.contains(name) =>
+        name -> compMap(name)
+    }.toMap
```
jackkoenig, line 169: I'm not convinced this is actually better, kind of fun though 😜 
```suggestion
    val portDesc = mod.ports.flatMap { case Port(_, name, _, _) => compMap.get(name).map(name -> _) }.toMap
```

src/main/scala/firrtl/AddDescriptionNodes.scala
```diff
@@ -89,44 +143,97 @@ class AddDescriptionNodes extends Transform with PreservesAll[Transform] {
...
 
-    val modDesc = modMap.get(newMod.name).map {
-      desc => DocString(StringLit.unescape(desc.mkString("\n\n")))
-    }
+    val modDesc = modMap.get(newMod.name).getOrElse(Seq())
```
jackkoenig, line 171:
```suggestion
    val modDesc = modMap.getOrElse(newMod.name, Seq())
```

src/test/scala/firrtlTests/VerilogEmitterTests.scala
```diff
@@ -835,27 +835,43 @@ class VerilogDescriptionEmitterSpec extends FirrtlFlatSpec {
...
     // We don't use executeTest because we care about the spacing in the result
     val modName = ModuleName("Test", CircuitName("Test"))
+    def altFunc(alternative: ir.FirrtlNode): String = alternative match {
+      case ir.Port(_, name, dir, _) =>
+        s"real ${dir.serialize} $name"
+      case d: ir.IsDeclaration=>
+        s"real wire ${d.name};"
+      case m: ir.DefModule =>
+        s"module ${m.name}()"
+    }
```
jackkoenig, line 863: Is this used?

jackkoenig:
Note that kind of the nuclear option for cleaning the history is to create a diff between master and this branch, commit it, then force the branch to point to that commit.

sequencer:
any progress on this PR?
Really need this for verilator hacking:)

kammoh:
Hi @grebe! Any updates on this? I really need to push down Verilog attributes, but keeping this PR synched with master is becoming more difficult every day. Hope you can get it merged soon. Thanks for the great work!

sequencer:
according to this dev-meeting, maybe I can rebase it to master and create a new PR.

sequencer:
A new PR #1643 is created.

grebe:
Sorry, I've been MIA for a while due to thesis writing taking up all my time.

sequencer:
No worries, if you are not mind, I rebased your great job on to master. I will fix the CI issue today or tomorrow, if you have any concerns please just tells me.

sequencer:
closed since #1643 is merged

PR #1180, opened 2019-09-13
Update Travis stage names to match new versions

jackkoenig:
Minor whoopsie in my last PR


jackkoenig:
Yeah I thought about that, I'll drop the minor versions

PR #1181, opened 2019-09-16
Bump sbt to 1.3.0

ucbjrl:


ucbjrl:
The nightmare that is sbt continues. 1.3.0 has bugs that cause sequences of `sbt +publishSigned` to fail (after a successful publish) with `[error] java.lang.NoClassDefFoundError: com/typesafe/sbt/SbtPgp$`.
1.3.1 supposedly has a fix.

jackkoenig:
Looks like 1.3.2 has been released, we should move everything to it maybe?

ucbjrl:
I don't think it has the fix (sbt/sbt#5081, sbt/sbt#5082) which should address the issue (sbt/sbt#5088). I'm happy to wait.

jackkoenig:
https://github.com/sbt/sbt/releases/tag/v1.3.2 mentions
```
Fixes NoClassDefFoundError at the end of run when thread keeps running after the task. 
Use allowZombieClassLoaders := true to opt into using classes after the task. #5081 / #5083 by @eatkins
```

Is that not the fix?

ucbjrl:
Hmmm. I'm used to seeing the tags containing the commit along with the commit(i.e. https://github.com/freechipsproject/firrtl/commit/c1feef14e30802b460976197479cb6b967f31324), and since I didn't see an explicit reference to v1.3.1 or v1.3.2 I assumed they weren't included. I notice that if you list the commits for v1.3.2, it does include sbt/sbt#5082. I'll give it a shot next week.

ucbjrl:
1.3.2 does appear to resolve the `publishSigned` issue noted above. #1187 bumps sbt to 1.3.2.

PR #1182, opened 2019-09-17
Speed up InlineInstances

jackkoenig:
Create instance maps once for each Module

Performance regression introduced in https://github.com/freechipsproject/firrtl/pull/1073

This doesn't fix the entire performance regression (there's still an issue with lots and lots of rename maps), but for a large SiFive design, this reduced Inline Instance runtime from 94 minutes to 86 seconds (out of a total of 534 seconds). There is still a lot of improvement to be had considering it takes 6 seconds when I revert https://github.com/freechipsproject/firrtl/pull/1073.

I think we ought to make the `instMaps` code a utility on `InstanceGraph`, but I'm unsure of the right API (but note that this code is duplicated, it also exists in `ConstantPropagation`).

### DIFF
--- a/src/main/scala/firrtl/passes/Inline.scala
+++ b/src/main/scala/firrtl/passes/Inline.scala
@@ -131,21 +131,11 @@ class InlineInstances extends Transform with RegisteredTransform {
     val flatInstances = instsToInline.map(i => i.module.name + "." + i.name) ++ getInstancesOf(c, flatModules)
     val iGraph = new InstanceGraph(c)
     val namespaceMap = collection.mutable.Map[String, Namespace]()
-
-    def getInstMap(mod: DefModule): Map[String, String] = mod match {
-      case m: Module => getInstMapBody(m.body)
-      case _ => Map.empty[String, String]
-    }
-    def getInstMapBody(stmt: Statement): Map[String, String] = {
-      val instMap = mutable.Map.empty[String, String]
-      def onStmt(s: Statement): Statement = s.map(onStmt) match {
-        case wDef@ WDefInstance(_, instName, modName, _) =>
-          instMap += (instName -> modName)
-          wDef
-        case other => other
-      }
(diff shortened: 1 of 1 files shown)


albert-magyar:
`getChildrenInstancesOfModule` is the closest thing, but unfortunately it returns a `Map[String, Set[(Instance, OfModule)]]`

The idea was to try to make more APIs use strings that are boxed with annotation / target utilities, but that never really took off. It might be good to wait for the Named / Target stuff to all be past the deprecation stage, but it also might be worth just adding this as a `Map[String, (String, String)]` for convenience.

jackkoenig:
> views can be good.

Note that the views here are only to prevent additional traversals, you can do the `map` and `.toMap` (ie. new collection construction) at the same time. It's the replacement for `collection.breakOut` in Scala 2.13 (and I'm pretty sure the performance is roughly equivalent in Scala 2.12)

albert-magyar:
@jackkoenig, how confident are you that the `Map` you get from this would have stable iteration order?

jackkoenig:
@albert-magyar I am 100% confident it will not have stable iteration order, it's not intended to be iterated on.

albert-magyar:
Then it doesn't automatically make sense to have that pattern appear as an API, since it will create the risk of someone using it and iterating over it.

seldridge:
@albert-magyar: I agree with you that a public `InstanceGraph` API doing the same thing as `instMaps` must be stable (and can be made stable), but the usage of `instMaps` in `InlineInstances` is private and is not iterated over. Is your concern about a future public API or the fact that this PR introduces a private `Map` with an unstable iteration order?

Basically, do you see a reason to hold this up? (As this is a blocker on 1.2, I'd like to get this in!)

albert-magyar:
I think we should just merge this now and add the API / delete code later since just copying over what's in `InlineInstances` isn't ideal for use everywhere.

PR #1183, opened 2019-09-18
Implement read-first memory behavior in Verilog

albert-magyar:
This PR generally improves support for different read-under-write behaviors

* FIRRTL proto format now supports read-under-write as an enumeration
* VerilogMemDelays can correctly implement read-first (“old”) behavior by pipelining read data

There is some slight refactoring of VerilogMemDelays, but I generally tried to leave it alone where possible to avoid surprises.

Since all chisel memories have undefined write collision behavior by default, there is a companion chisel3 PR.

I tested this with rocket-chip by adding a pass to coerce undefined write collision behavior memories to be read-first and simulating the resulting Verilog, and it ran fine. However, since that was just a custom transform for testing, this PR will have essentially no effect on existing code, since almost nothing uses read-first memories.

### DIFF
--- a/src/main/scala/firrtl/ir/IR.scala
+++ b/src/main/scala/firrtl/ir/IR.scala
@@ -285,6 +285,13 @@ case class DefInstance(info: Info, name: String, module: String) extends Stateme
   def foreachString(f: String => Unit): Unit = f(name)
   def foreachInfo(f: Info => Unit): Unit = f(info)
 }
+
+object ReadUnderWrite extends Enumeration {
+  val Undefined = Value("undefined")
+  val Old = Value("old")
+  val New = Value("new")
+}
+
 case class DefMemory(
     info: Info,
     name: String,
@@ -296,7 +303,7 @@ case class DefMemory(
     writers: Seq[String],
     readwriters: Seq[String],
     // TODO: handle read-under-write
-    readUnderWrite: Option[String] = None) extends Statement with IsDeclaration {
+    readUnderWrite: ReadUnderWrite.Value = ReadUnderWrite.Undefined) extends Statement with IsDeclaration {
   def serialize: String =
     s"mem $name :" + info.serialize +
     indent(
@@ -307,7 +314,7 @@ case class DefMemory(
           (readers map ("reader => " + _)) ++
           (writers map ("writer => " + _)) ++
           (readwriters map ("readwriter => " + _)) ++
-       Seq("read-under-write => undefined")) mkString "\n")
+       Seq(s"read-under-write => ${readUnderWrite}")) mkString "\n")
   def mapStmt(f: Statement => Statement): Statement = this
   def mapExpr(f: Expression => Expression): Statement = this
   def mapType(f: Type => Type): Statement = this.copy(dataType = f(dataType))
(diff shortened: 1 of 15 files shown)


albert-magyar:
@jackkoenig @ucbjrl @seldridge 

One thing I though about after our discussion at the meeting is that this PR (unlike the Chisel PR) has absolutely zero detectable changes on existing code, either in FIRRTL compilation or in simulation behavior. It changes the CHIRRTL grammar, but only by adding an optional `read-under-write` parameter to `smem`, so existing CHIRRTL files are still legal and infer "undefined."

Since the commits are all split up into different clean changes, I would be tempted to label the one that changes `VerilogMemDelays` as a pure bugfix.

On the other hand, the [Chisel PR](https://github.com/freechipsproject/chisel3/pull/1183) emits the new flavor of `smem`, so it is backwards-incompatible with existing releases of FIRRTL. That should definitely go in some major release.

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -285,6 +285,13 @@ case class DefInstance(info: Info, name: String, module: String) extends Stateme
   def foreachString(f: String => Unit): Unit = f(name)
   def foreachInfo(f: Info => Unit): Unit = f(info)
 }
+
+object ReadUnderWrite extends Enumeration {
+  val UNDEFINED = Value("undefined")
+  val OLD = Value("old")
+  val NEW = Value("new")
```
ekiwi: Should these be CamelCase?

See e.g., [PrimOps](https://github.com/freechipsproject/firrtl/blob/f08f8dbb3c480220f92923a7f3242fcbb644b65e/src/main/scala/firrtl/PrimOps.scala) which uses CamelCase for enum like constans.

src/main/scala/firrtl/passes/memlib/ToMemIR.scala
```diff
@@ -11,14 +11,15 @@ import firrtl.ir._
   *   - read latency and write latency of one
   *   - only one readwrite port or write port
   *   - zero or one read port
+  *   - undefined read-under-write behavior
   */
 object ToMemIR extends Pass {
   /** Only annotate memories that are candidates for memory macro replacements
-    * i.e. rw, w + r (read, write 1 cycle delay)
+    * i.e. rw, w + r (read, write 1 cycle delay) and read-under-write "undefined."
```
ekiwi: Are there no memory compilers that support a defined behavior for this case? I could see a memory compiler generating logic to make this case defined, but I am lacking experience with memory compilers here.

ekiwi:
Just saw that you have a test case in your Chisel pull request (`SyncReadMemWriteCollisionTester`).
That might be enough if it is hard to run things through Verilator from `firrtl`.

albert-magyar:
Chisel does make it a quite a bit easier to run simple execution-driven tests.

src/main/scala/firrtl/passes/memlib/ToMemIR.scala
albert-magyar: Currently we lack the ability to enforce any downstream behavior, so Jack requested I put this in as an extra safety check. In practice, memory IPs would implement one of the three options or expose it as a flag.

src/main/scala/firrtl/ir/IR.scala
albert-magyar: Yeah, I agree. I was copying the pattern from `Gender`, which used all-caps, but it looks like the switch to `Flow` has eliminated the all-caps.

albert-magyar:
Thanks, I am changing the `UNDEFINED` -> `Undefined` and so on.

PR #1184, opened 2019-09-19
Faster inline renaming

albertchen-sifive:
This PR makes some optimizations to `InlineInstances`. Mainly by reducing the number of chained `RenameMap`s. The number of chained `RenameMap`s should now be equal to the length of the longest path in the subgraph of inlined instances whereas before it was equal to the number of inlined instances.

### DIFF
--- a/src/main/scala/firrtl/passes/Inline.scala
+++ b/src/main/scala/firrtl/passes/Inline.scala
@@ -141,6 +141,7 @@ class InlineInstances extends Transform with RegisteredTransform {
     /** Add a prefix to all declarations updating a [[Namespace]] and appending to a [[RenameMap]] */
     def appendNamePrefix(
       currentModule: IsModule,
+      nextModule: IsModule,
       prefix: String,
       ns: Namespace,
       renames: mutable.HashMap[String, String],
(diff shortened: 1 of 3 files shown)


albertchen-sifive:
on a large SiFive design inline runtime is reduced to 3s out of 633s compared to 89s out of 539s without the PR.

jackkoenig:
> on a large SiFive design inline runtime is reduced to 3s out of 633s compared to 89s out of 539s without the PR

@albertchen-sifive's numbers alarmed me a little because the overall runtime still got way worse, so I ran the same experiment.

I got InlineInstances runtime reduced to 3s out of 457s compared to 94s out of 550s. So for me this is 100% a win, I think we should get this in and crack out 3.2-RC2 @ucbjrl @azidar @seldridge

albertchen-sifive:
> Do people generally think it's worthwhile to use OfModule and InstanceOf to box strings in an API like that?

I like it. I changed a bunch of the strings to `OfModule`/`Instance` while working on this PR because it was really hard keeping track of what strings were modules and which were instances. and `WDefInstance` has extra unused fields that might lead to hard-to-catch issues, like if someone accidentally uses a `WDefInstance` with a different `Info` to index into the edge map. and if we can add helpers to let us do things like `"TopMod".ofMod` and `"topInst".inst` that would make using boxed strings less annoying I think.

albert-magyar:
> I like it. I changed a bunch of the strings to `OfModule`/`Instance` while working on this PR because it was really hard keeping track of what strings were modules and which were instances. and `WDefInstance` has extra unused fields that might lead to hard-to-catch issues, like if someone accidentally uses a `WDefInstance` with a different `Info` to index into the edge map. and if we can add helpers to let us do things like `"TopMod".ofMod` and `"topInst".inst` that would make using boxed strings less annoying I think.

That's pretty much exactly how I feel. I just rewrote some of the "instance analysis" stuff as an API in `InstanceGraph`, and I'm testing whether there's any meaningful performance difference from using `OfModule` and so on when porting `ConstantPropagation` to use these new library methods.

jackkoenig:
It would be good to benchmark, we should also experiment with Java 11 and GraalVM because both are better for functional code and might be able to optimize away the boxing in practice (stress the *might*)

jackkoenig:
Some notes on Scala 2.13 performance: https://github.com/scala/bug/issues/11627#issuecomment-514619316

Worth keeping in mind as we make our benchmarking more robust.

albert-magyar:
And some day we can stop supporting 2.12 and make lots of "dependency analysis"-type of things (instances, etc) much faster with `groupMapReduce` :)

jackkoenig:
We can consider making it faster in 2.13 anyway

PR #1186, opened 2019-09-20
Absorb some instance analysis into InstanceGraph, use safer boxed Strings

albert-magyar:
@jackkoenig @albertchen-sifive 

This absorbs some repetitive instance hierarchy analysis from `InlineInstances` and `ConstantPropagation` into `InstanceGraph`. In this version, strings are hidden with boxed types that have some potential to help with type safety when maintaining passes.

`InlineInstances` already uses the `TargetToken`s to hide strings, so I just went through and ported `ConstantPropagation` to get rid of raw `String` as much as possible. It does get rid of some scarier type signatures like `(NodeMap, Map[String, String], Map[String, Map[String, Literal]]) => ...`. I added some implicit classes to make it possible to write things like `inst.OfModule` or `inst.Instance`, which made the code a bit less cluttered in the few places where conversions happen.

Absorbing the instance analysis into API calls reduces both complexity (for example, `staticInstanceCount` is more efficient than existing code) and code size a bit, but this is offset by the overhead of boxing. I could alternatively add these APIs as versions that work with raw `String`s, which might be faster in theory, though I will be testing if that is the case in practice. This PR has no measurable performance impact on the integration test benchmark performance, but I am still testing bigger designs.

The new code in `InstanceGraph` returns collections with stable iteration orders in all cases.


jackkoenig:
From my benchmarking, this looks like a nice speedup for small designs and a wash to perhaps minor speedup for small designs. I'm a little surprised by the size of the speedup for the small designs so I'm rerunning the benchmarks.

Design | master runtime (SD) | this runtime (SD) | this/master (avg) | this/master (worst) | this/master (best)
-- | -- | -- | -- | -- | --
FPU | 13.71 (3.56) | 4.6 (0.27) | 0.34 | 0.48 | 0.25
HwachaSequencer | 39.36 (10.74) | 13.02 (0.75) | 0.33 | 0.48 | 0.24
Icache | 5.68 (0.99) | 2.54 (0.1) | 0.45 | 0.56 | 0.37
Rob | 52.24 (14) | 29.45 (2.56) | 0.56 | 0.84 | 0.41
RocketCore | 11.05 (4.44) | 5.09 (0.38) | 0.46 | 0.83 | 0.30
SiFive A | 328.58 (65.32) | 329.5 (38) | 1.00 | 1.40 | 0.74
SiFive B | 80.75 (2.81) | 78.53 (2.51) | 0.97 | 1.04 | 0.91
SiFive C | 219.78 (13.15) | 219.26 (11.2) | 1.00 | 1.12 | 0.89
SiFive D | 90.75 (3.09) | 83.64 (1.58) | 0.92 | 0.97 | 0.87

Notes:
* "avg" - simply dividing the runtimes
* "worst" - (this + SD) / (master - SD)
* "best" - (this - SD) / (master + SD)

albert-magyar:
That is a little .. too good. I was seeing a speedup in some smaller designs, but 3x seems kind of too high.

jackkoenig:
Updated after rerunning the smaller configs in `regress/`. A bit more reasonable and the high variance makes it obvious not to read too much into the numbers.


Design | master runtime (SD) | this runtime (SD) | this/master (avg) | this/master (worst) | this/master (best)
-- | -- | -- | -- | -- | --
FPU | 6.98 (3.66) | 4.28 (0.28) | 0.61 | 1.37 | 0.38
HwachaSequencer | 16.73 (5.49) | 11.78 (0.6) | 0.70 | 1.10 | 0.50
Icache | 3.74 (1.4) | 2.17 (0.15) | 0.58 | 0.99 | 0.39
Rob | 27.39 (11.52) | 25.81 (2.54) | 0.94 | 1.79 | 0.60
RocketCore | 4.78 (0.64) | 4.84 (0.24) | 1.01 | 1.23 | 0.85
SiFive A | 328.58 (65.32) | 329.5 (38) | 1.00 | 1.40 | 0.74
SiFive B | 80.75 (2.81) | 78.53 (2.51) | 0.97 | 1.04 | 0.91
SiFive C | 219.78 (13.15) | 219.26 (11.2) | 1.00 | 1.12 | 0.89
SiFive D | 90.75 (3.09) | 83.64 (1.58) | 0.92 | 0.97 | 0.87

jackkoenig:
It's weird that this PR makes the standard deviation so much lower

albert-magyar:
It's also funny, since the point of the changes were to make it "safer" by relying more on the type system to distinguish types of strings without slowing things down, but if the `InstanceGraph` API calls are a bit faster that's good :)

albert-magyar:
I can take them out, it's not a huge deal.

I figure the real risk would be actually having an implicit conversion `T => OfModule`. This still requires you to call a helper method of the implicit class to convert, so it's basically just syntactic sugar compared to making it `OfModule.apply(Module)` or something.

PR #1188, opened 2019-09-25
[RFC] Define read-write collison for independently clocked mem ports

albert-magyar:
H/t @seldridge for bringing up this recent [discussion](https://twitter.com/whitequark/status/1176622348222398464) about FIRRTL memories.

I think it's fair to say that the behavior of collisions across independently clocked ports is not well specified. I think that there are two opposite traps to avoid:
- Baking in the specifics of present implementations.
- Tying our hands by specifying overly general behaviors that *have not been implemented* or create "nearly dead" code paths in the compiler.

With that said, it would be good to hear what people think about making read-write collisions across multiple clock domains "undefined" by default.

As an aside, this conversation seems to have come up with some FPGA-related context (as other read-under-write questions have in the past), so it's worth noting that both Xilinx and Intel make the value read by a read port when colliding with an independently clocked write port undefined.


spec/spec.tex
```diff
@@ -829,6 +829,15 @@ \subsubsection{Read Under Write Behaviour}
 
 If the read-under-write flag is set to \verb|undefined|, then the value held by the read port after the appropriate read latency is undefined.
 
+
+For the purpose of defining such collisions, an ``active write port'' is either a write port or readwrite port that is used to initiate a write operation on a given clock edge.
+An ``active read port'' is a read port or readwrite port that is used to initiate a read operation on a given clock edge.
```
ekiwi: Maybe add that this means that `en=1 and wmode=1`.

spec/spec.tex
```diff
@@ -829,6 +829,15 @@ \subsubsection{Read Under Write Behaviour}
...
+Each operation is defined to be ``active'' for the number of cycles set by its corresponding latency, starting from the cycle where its inputs were provided to its associated port.
```
ekiwi: I always thought that a read/write was only active in the cycle when it is actually executed, i.e. after `N` (=latency) cycles.

spec/spec.tex
```diff
@@ -829,6 +829,15 @@ \subsubsection{Read Under Write Behaviour}
...
+Note that this excludes combinational reads, which are simply modeled as combinationally selecting from stored values
+
+For memories with independently clocked ports, a collision between a read operation and a write operation with independent clocks is defined to occur when the address of an active write port and the address of an active read port are the same for overlapping clock periods, or when any portion of a read operation overlaps part of a write operation with a matching addresses.
+In such cases, the data that is read out of the read port is undefined.
```
ekiwi: Could the data that is written also be undefined (i.e. the memory is corrupted at the addressed location)? Do we want to allow that behavior/

albert-magyar: Oops, I deleted my comments. This would have the effect of making read operations have side effects, which would be bad. There's no reason I can see to make the contents of memory undefined.

spec/spec.tex
albert-magyar: That's a good idea.

spec/spec.tex
albert-magyar: Oops, I deleted my comments. The current read-under-write specification implies that a read can be active before that. The definition in the this current spec is meant to be conservative to avoid the other alternative of a huge matrix of behaviors.

whitequark:
What does it mean exactly to have "independently clocked" ports? Is this property purely syntactical (two ports using a different clock signal are always independent), or is it semantical (two ports clocked with two clock signals that are driven to the exact same value would not be independent)? This is what was causing issues with RTLIL.

ekiwi:
> Is this property purely syntactical (two ports using a different clock signal are always independent), or is it semantical (two ports clocked with two clock signals that are driven to the exact same value would not be independent)?

We don't have an automated check (and as far as I know we never will), so you could assume it to be semantical.

This addition to the spec was motivated by the undefinedness of write-write or write-read collisions on common FPGA memories when clocking ports independently.


> This is what was causing issues with RTLIL.

Could you link to the RTLIL issue? Might be interesting to see if we can learn from it.

albert-magyar:
For the purposes of the spec, “independently clocked” means any ports that do not have identical syntactic expressions for their clock connections.

There are a few key factors here:
* Two clocks entering a circuit could be tied together outside the circuit.
* FIRRTL behavior is predominantly defined in terms of declared signals, not structural wires.
* The compiler is generally structured around syntactic, source-to-source transformations with limited exceptions. Behaviors based on logic analysis are not guaranteed by any other part of the spec.
* A downstream pass deciding that two clocks are actually connected would merely produce a legal refinement of the behavior.

whitequark:
> For the purposes of the spec, “independently clocked” means any ports that do not have identical syntactic expressions for their clock connections.

Thanks, this makes perfect sense.

PR #1191, opened 2019-09-29
Add check for multiple sources for same wiring pin

jackkoenig:
Can lead to some very weird errors


jackkoenig:
Revisiting this,
> This may have an impact on situations where wiring fails because deduplication didn't succeed. I would think about how this would affect

@seldridge, do you think there's something we should do about this issue or punt on that?

seldridge:
I'm okay to merge. Punt.

azidar:
@Mergifyio backport 1.2.x

PR #1197, opened 2019-10-03
Add Block factory from argument list of Statements

albert-magyar:
This can eliminate boilerplate `Block(Seq(...))`. The 2-ary function pattern is necessary to disambiguate it from the default constructor, since a variadic parameter list is indistinguishable from a `Seq` after erasure.

### DIFF
--- a/src/main/scala/firrtl/ir/IR.scala
+++ b/src/main/scala/firrtl/ir/IR.scala
@@ -360,6 +360,11 @@ case class Conditionally(
   def foreachString(f: String => Unit): Unit = Unit
   def foreachInfo(f: Info => Unit): Unit = f(info)
 }
+
+object Block {
+  def apply(head: Statement, tail: Statement*): Block = Block(head +: tail)
+}
+
 case class Block(stmts: Seq[Statement]) extends Statement {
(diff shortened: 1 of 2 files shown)


src/main/scala/firrtl/ir/IR.scala
```diff
@@ -360,6 +360,11 @@ case class Conditionally(
   def foreachString(f: String => Unit): Unit = Unit
   def foreachInfo(f: Info => Unit): Unit = f(info)
 }
+
+object Block {
+  def apply(head: Statement, tail: Statement*): Block = Block(head +: tail)
```
jackkoenig: Word of caution, varargs are implemented in Scala by WrappedArrays, so this is going to have to copy the entire array

jackkoenig: In response to my own point, this is only usable by static code (ie. you can't splat to it), so I doubt anyone is going to have large `tails` here so it probably doesn't matter in practice.

albert-magyar: The `Seq` factory that you avoid uses varargs, too :)

PR #1201, opened 2019-10-14
Fix PrimOp of Literals reported as NonLiteralAsyncResetValue

johnsbrew:
**Type of issue**: bug report

### FIRRTL 

```scala
input clock : Clock
input reset : AsyncReset
output o : {field0 : UInt<1>, field1 : UInt<1>}
wire bundleInit : {field0 : UInt<1>, field1 : UInt<1>}
wire uintInit : UInt<2>
uintInit <= UInt<1>("h00")
node field0Init = bits(uintInit, 0, 0)
bundleInit.field0 <= field0Init
node field1Init = bits(uintInit, 1, 1)
bundleInit.field1 <= field1Init

reg bundle : {field0 : UInt<1>, field1 : UInt<1>}, clock with : (reset => (reset, bundleInit)) 
o <= bundle
```

### Chisel
Chisel use-case triggering the bug is the following:
```scala
val bundle = new Bundle {
  val field0 = Bool()
  val field1 = Bool()
}
val r = RegInit(0.U.asTypeOf(bundle)) 

```

**What is the current behavior?**
Any `PrimOp` of `Literal` (typically bit select `bits`) used as `AsyncReset` Init Value throw an `checks.CheckResets.NonLiteralAsyncResetValueException`

**What is the expected behavior?**
No error

**Please tell us about your environment:**
    - version: `master`
    - OS: `Darwin Kernel Version 17.7.0: Sun Jun  2 20:31:42 PDT 2019; root:xnu-4570.71.46~1/RELEASE_X86_64`

Better support of `AsyncReset` init values towards #1050 =)

### DIFF
--- a/src/main/scala/firrtl/LoweringCompilers.scala
+++ b/src/main/scala/firrtl/LoweringCompilers.scala
@@ -69,7 +69,6 @@ class HighFirrtlToMiddleFirrtl extends CoreTransform {
     passes.ResolveKinds,
     passes.InferTypes,
     passes.CheckTypes,
-    new checks.CheckResets,
     passes.ResolveFlows,
     new passes.InferWidths,
     passes.CheckWidths,
@@ -94,6 +93,7 @@ class MiddleFirrtlToLowFirrtl extends CoreTransform {
     passes.Legalize,
     new firrtl.transforms.RemoveReset,
     new firrtl.transforms.CheckCombLoops,
+    new checks.CheckResets,
     new firrtl.transforms.RemoveWires)
 }
 
--- a/src/main/scala/firrtl/checks/CheckResets.scala
+++ b/src/main/scala/firrtl/checks/CheckResets.scala
@@ -39,7 +39,11 @@ class CheckResets extends Transform {
           case prim @ DoPrim(_, args, _, _) => drivers += we(prim) -> args
           case _ => // Do nothing
         }
-      case Connect(_, lhs, rhs) => drivers += we(lhs) -> Seq(rhs)
+      case Connect(_, lhs, rhs) => 
+        lhs match {
+          case WRef(_,_,RegKind, _) => // registers are never to be literals
+          case _ => drivers += we(lhs) -> Seq(rhs)
+        }
       case reg @ DefRegister(_,_,_,_, reset, init) if reset.tpe == AsyncResetType =>
         regCheck += init -> reg
       case _ => // Do nothing
--- a/src/main/scala/firrtl/checks/CheckResets.scala
+++ b/src/main/scala/firrtl/checks/CheckResets.scala
@@ -18,7 +18,7 @@ object CheckResets {
   private type RegCheckList = mutable.ListBuffer[(Expression, DefRegister)]
   // Record driving for literal propagation
   // Indicates *driven by*
-  private type DirectDriverMap = mutable.HashMap[WrappedExpression, Expression]
+  private type DirectDriverMap = mutable.HashMap[WrappedExpression, Seq[Expression]]
 
 }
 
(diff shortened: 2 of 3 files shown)


johnsbrew:
@jackkoenig

johnsbrew:
@jackkoenig I guess this test-case must be updated ?!
https://github.com/freechipsproject/chisel3/blob/master/src/test/scala/chiselTests/AsyncResetSpec.scala#L132-L139

```scala
  it should "NOT be allowed with non-literal reset values" in {
    a [NonLiteralAsyncResetValueException] shouldBe thrownBy {
      compile(new BasicTester {
        val x = WireInit(123.U + 456.U)
        withReset(reset.asAsyncReset)(RegInit(x))
      })
    }
  }
```
`val x = WireInit(123.U + 456.U)` is a literal (or should be seen as one!)

azidar:
@johnsbrew Why did the `checks.CheckResets` move from `HighFirrtlToMiddleFirrtl` to `MiddleFirrtlToLowFirrtl`?

johnsbrew:
> @johnsbrew Why did the `checks.CheckResets` move from `HighFirrtlToMiddleFirrtl` to `MiddleFirrtlToLowFirrtl`?

@azidar because the check will fail with a jvm stackoverflow if it encounters a combinational loop, hence I moved it after the `CheckCombLoops` check.

jackkoenig:
Ah thanks for this @johnsbrew. Sorry I've been on vacation but am back now and will take a look soon.

I would prefer to have `CheckResets` in `HighFirrtlToMiddleFirrtl`, but yeah it definitely needs to not hang forever if there's a combinational loop... I will think about it. Moving it is probably a fine workaround if a solution isn't simple.

jackkoenig:
This is a really clever implementation (I'm impressed you were able to enhance `InferResets` so cleanly, nice!). We discussed this in the developers meeting today and are thinking about fixing this issue via tweaking Chisel's FIRRTL-generation rather than opening up the spec in FIRRTL (see https://github.com/freechipsproject/chisel3/pull/1225).

Our (My) hesitation to take this approach is that is opens up the definition of correctness to some application of constant propagation. My suspicion is that synthesis tools are not going to perform such constant propagation so we will have to do it for them. If you combine this PR with reseting an async reset register to some primop that we don't yet propagate (like division or multiplication, we need to fix that!), I suspect that synthesis tools will error on our Verilog.

Fundamentally, we want to avoid optimizations as part of the definition of correctness. If https://github.com/freechipsproject/chisel3/pull/1225 turns out to be insufficient (there may be cases missing), we should reconsider this approach. We could also consider a more restricted form of it with some whitelist of allowed primops (bit select, concatenation, and pad seem like good candidates), and then ensure we include this optimization separately from ConstantPropagation so that running without optimizations includes it.

Other frontends may also necessitate revisiting this PR as well.

jackkoenig:
The other fix in this PR, moving `CheckResets` after `CheckCombLoops` should be incorporated. I would like to fix `CheckResets` to not need this move, but I think it's a fine fix in the meantime.

johnsbrew:
Hi @jackkoenig, thanks for your feedbacks

If I understand properly your point, you want to enforce a strict FIRRTL spec, allowing only explicit literals (no lookup through primop).
This requires that constant propagation has been handled prior to any passes on this strict spec.
You chose to have this constant propagation done within chisel emission with https://github.com/freechipsproject/chisel3/pull/1225 

My question is : why not doing constant propagation as the first step of `HighToMiddleFirrtl`? 
IMHO constant propagation is a concern shared by all front-ends and it would make sense to provide this transform as part of FIRRTL.

jackkoenig:
Hi @johnsbrew,

The idea isn't that we want constant propagation to be done by the frontend; it's that we don't want constant propagation at all. I would not actually characterize https://github.com/freechipsproject/chisel3/pull/1225 as performing constant propagation. Rather, I'd argue that in Chisel `someLiteral.asTypeOf(someType)` should always return a literal since it's just a cast. The issue is that FIRRTL does not have any construct to represent aggregate literals, so the only way we can express them is with a wire and connections. All I did in https://github.com/freechipsproject/chisel3/pull/1225 is eliminate an intermediate wire so that Chisel is emitting FIRRTL's representation of an aggregate literal.

We want to avoid any optimizations in the definition of correctness. The analogy I usually point to is that Java Unreachable Statement error (https://docs.oracle.com/javase/specs/jls/se7/html/jls-14.html#jls-14.21) which is generally considered a bad idea. If someone wanted to disable optimizations yet we baked constant propagation into the definition of correctness for async reset values, we would have to ignore the disabling and perform some constant propagation anyway.

I recognize there could be some utility in doing a restricted amount of constant propagation, but I'm not yet convinced it's worth the cost of optimizations leaking into the spec.

jackkoenig:
I'm going to close this for now. I want to see how far we can go with the current semantics and revisit if we need to.

PR #1206, opened 2019-10-22
Add separate Issue and PR templates

seldridge:
First crack at separate issue and PR templates. GitHub lets you add multiple issue templates which users can select from. For each of these, specific labels can be set (so maybe we should go with a lot of templates?). Also, reviewers can be assigned (though this PR doesn't use that feature). Pull request templates do not appear to support user selection of a specific template, but there is a web API to do it. See: https://github.community/t5/How-to-use-Git-and-GitHub/Multiple-Pull-Request-Templates/td-p/20909

More info: 

  - Use of the `labels` yaml field will result in that label being set (so long as that label already exists)
  - Users get an option of two different types of issues to file
  - Users have a checklist of things they're supposed to go over
  - GitHub documentation: https://help.github.com/en/github/building-a-strong-community/about-issue-and-pull-request-templates

You can play around with this here: https://github.com/seldridge/issue-pr-template-testing.


azidar:
@Mergifyio backport 1.2.x

Issue #1203, opened 2019-10-16
Improve error message for combinational loops


oharboe:
The error message below leaves something to be desired when it comes to chasing down a combinational loop.

The trick that I've found is to bring up the generated .fir file and from there, I search for e.g. _T_659 and there will be an annotation in the .fir file that tells me where in the .scala file to look: Thingamabob.scala line 177

Obviosuly it would improve the error message if the references back to Chisel code lines was directly in the error message.

```
node _T_659 = eq(_T_589, UInt<1>("h00")) @[Thingamabob.scala 177:69]
```

```
[deleted]
[info]   firrtl.transforms.CheckCombLoops$CombLoopException: : [module Thingamabob] Combinational loop detected:
[info] Thingamabob._T_589
[info] Thingamabob._T_659
[info] Thingamabob._T_660
[info] Thingamabob._GEN_13
[info] Thingamabob._GEN_43
[deleted]
```

albert-magyar:
That's a good idea, I will add the info printing to the error.

oharboe:
@albert-magyar Thanks!

PR #1208, opened 2019-10-23
Enhance CheckCombLoops errors with connection info

albert-magyar:
A good suggestion was made by @oharboe in #1203 to add source locators to combinational loop errors.

I added this feature to `CheckCombLoops` using a new `EdgeData` mixin for graphs and moved some of the internal code around a bit.

I added Scaladoc for the `EdgeData` traits, since they are fairly reusable.

### DIFF
--- a/src/main/scala/firrtl/graph/EdgeData.scala
+++ b/src/main/scala/firrtl/graph/EdgeData.scala
@@ -1,0 +1,126 @@
+// See LICENSE for license details.
+
+package firrtl.graph
+
+import scala.collection.mutable
+
+/**
+  * An exception that indicates that an edge cannot be found in a graph with edge data.
+  * 
+  * @note the vertex type is not captured as a type parameter, as it would be erased.
+  */
+class EdgeNotFoundException(u: Any, v: Any)
+    extends IllegalArgumentException(s"Edge (${u}, ${v}) does not exist!")
+
+/**
+  * Mixing this trait into a DiGraph indicates that each edge may be associated with an optional
+  * data value. The EdgeData trait provides a minimal API for viewing edge data without mutation.
+  *
+  * @tparam V the vertex type (datatype) of the underlying DiGraph
+  * @tparam E the type of each edge data value
+  */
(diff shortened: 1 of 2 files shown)


src/main/scala/firrtl/transforms/CheckCombLoops.scala
```diff
@@ -11,14 +11,51 @@ import firrtl.passes.{Errors, PassException}
...
+ * A case class that represents a net in the circuit. This is
+ * necessary since combinational loop checking is an analysis on the
+ * netlist of the circuit; the fields are specialized for low
+ * FIRRTL. Since all wires are ground types, a given ground type net
+ * may only be a subfield of an instance or a memory
+ * port. Therefore, it is uniquely specified within its module
+ * context by its name, its optional parent instance (a WDefInstance
+ * or WDefMemory), and its optional memory port name.
+ */
+case class LogicNode(name: String, inst: Option[String] = None, memport: Option[String] = None)
```
azidar: Is the plan to eventually use `Target`? Or is there a reason to keep `LogicNode` around?

src/main/scala/firrtl/graph/EdgeData.scala
```diff
@@ -0,0 +1,123 @@
...
+  * @tparam V the vertex type (datatype) of the underlying DiGraph
+  * @tparam E the type of each edge data value
+  */
+trait EdgeData[V, E] {
+  this: DiGraph[V] =>
+  protected val edgeDataMap: collection.Map[(V, V), E]
+
+  protected def assertEdgeExists(u: V, v: V): Unit = {
+    if (!contains(u) || !getEdges(u).contains(v)) {
+      throw EdgeNotFoundException
```
azidar: Can this error also report the vertices of the missing edge?

albert-magyar: That's probably a good idea.

albert-magyar: Added

Issue #727, opened 2018-01-09
Eliminate CHIRRTL


azidar:
* **Type of issue**
  - [ ] Bug report
  - [ ] Feature request
  - [x ] Other enhancement

* **What is the use case for changing the behavior?**
#702 exposes a problem with the current way we are handling memory ports with CHIRRTL. The plan of record is to remove CHIRRTL entirely, and add MPort as first-class node in the FIRRTL IR.

This issue is to track discussions about how to make this possible.

* **Impact**
  - [ ] no functional change
  - [ ] API addition (no impact on existing code)
  - [ x] API modification
  - [ ] unknown

* **Development Phase**
  - [ ] request
  - [x] proposal

The advantages of this approach is that CHIRRTL is a wart, bridging how FIRRTL handles memories, and how Chisel handles memories. See #702 for more detail.

azidar:
The current proposal is to remove MPort's after running ExpandWhens. To handle mport enables, we will need to add an explicit enable field in MPorts.

One problem I haven't solved yet is how to handle the read enable for non-combinational read ports (for combinational reads, the enable can always be true without loss of QOR). The current CHIRRTL and Chisel semantics is that the enable is true wherever the read port is **referenced**, not where it is **declared**. Thus:

```
mem m:
  readLatency => 1
  ...
read mport r = m[addr], clk
when p:
  out <= r
```

should be compiled to:

```
mem m:
  readLatency => 1
  reader => r
m.r.en <= p
m.r.addr <= addr
out <= m.r.data
```

This would just require more special casing in ExpandWhen's to keep track of wherever these mports are referenced, and then add them to the enable field. If feels very inelegant.

azidar:
Another thought is to change mports to have two fields, an enable and a data field:

```
mem m:
  readLatency => 1
  ...
read mport r = m[addr], clk
r.en <= UInt(0)
when p:
  out <= r.data
  r.en <= UInt(1)
```

should be easily expanded after ExpandWhens:

```
mem m:
  readLatency => 1
read mport r = m[addr], clk
r.en <= mux(p, UInt(1), UInt(0))
out <= validif(p, r.data)
```

albert-magyar:
@azidar, one thing to note is that `RemoveCHIRRTL` currently define that the enable is true wherever the address used for the read port is assigned -- from looking at the code, this seems like it was an intentional move, not an implementation bug.

[...]

PR #1210, opened 2019-10-24
Remove CHIRRTL

azidar:
Triage of problems poised by #727 

Questions to discuss:
- [ ] Removing WIR. Removing CHIRRTL requires handling memories and the WIR stuff. I think since they are separable, this PR should only handle the memory stuff.

TODO List:
- [ ] Infer memory port direction (done in ResolveFlows?)
- [ ] Check memory port direction (done in CheckFlows?)
- [ ] Add to ir: DefMemoryPort, MDir
- [ ] Update InferTypes to know "hidden" fields of memory port: en and data (? and mask)
- [ ] Handle masks (are they also a secret hidden field?)
- [ ] Update what Chisel generates (needs to assign to hidden fields)
- [ ] Add support for DefMemoryPort in all HighForm passes and any passes prior to expand whens
  - [ ] Dedup?
  - [ ] passes.CheckHighForm,
  - [ ] passes.ResolveKinds,
  - [ ] passes.InferTypes,
  - [ ] passes.CheckTypes,
  - [ ] passes.Uniquify,
  - [ ] passes.ResolveKinds,
  - [ ] passes.InferTypes,
  - [ ] passes.ResolveFlows,
  - [ ] passes.CheckFlows,
  - [ ] new passes.InferBinaryPoints(),
  - [ ] new passes.TrimIntervals(),
  - [ ] new passes.InferWidths,
  - [ ] passes.CheckWidths,
  - [ ] new firrtl.transforms.InferResets
  - [ ] passes.PullMuxes,
  - [ ] passes.ReplaceAccesses,
  - [ ] passes.ExpandConnects,
  - [ ] passes.RemoveAccesses,
  - [ ] passes.ExpandWhens,
  - [ ] passes.CheckInitialization,
  - [ ] new checks.CheckResets,


albert-magyar:
With the idea being that any transform will declare its dependencies, would it be reasonable to change `RemoveCHIRRTL` to `RemoveInferredMems` or something like that, and just have every `HighForm` pass that doesn't handle those statements declare it as a dependency?

seldridge:
@albert-magyar: yeah, you can do it that way.

More info: The dependency-api-2 branch is simpler in that almost every later transform is specified as depending on all earlier transforms (primarily to force the old order). However, it makes more sense to relax this to accurately specify the true dependencies. Divining the true dependencies may take some work, however.

albert-magyar:
@seldridge that set of baseline dependencies does sound very useful for "blessing" `CDefMemory` and `CDefMPort`. I imagine the following would work and remove the burden of supporting inferred memories in any passes declaring the new name of `RemoveCHIRRTL` as a dependency.

1. Compiler starts with input FIRRTL file that may contain inferred memories, inferred ports, and inferred read/write enables.
2. `CheckHighForm` (already essentially identical to `CheckCHIRRTL` since #1148).
3. `InferTypes` (would need to be merged with `CInferTypes`).
4. `RemoveCHIRRTL` (bugfixed and renamed to `RemoveInferredMems` or something). It would be possible to fix all buggy cases, but the read-enable semantics would need to be set in stone in the spec first.
5. Rest of compiler proceeds as normal.

This would also avoid any changes or special-casing in `ExpandWhens`.

spec/spec.tex
```diff
@@ -842,6 +842,85 @@ \subsubsection{Write Under Write Behaviour}
...
+write mport w = syncmem[cmd.addr], clk
+when (cmd.wen) :
+  w <= cmd.wdata ; conditional write operation
+
+; last-connect semantics apply to inferred writes
+when (cmd.wen && cmd.isClear) :
+  w <= UInt<32>(``h00'') ; different condition
+\end{lstlisting}
+
+Inferred write ports do not support masked writes; therefore, all write operations inherently have
```
azidar, line 918: Aren't masked writes inferred though?

src/main/antlr4/FIRRTL.g4
```diff
@@ -94,8 +94,6 @@ stmt
   : 'wire' id ':' type info?
   | 'reg' id ':' type exp ('with' ':' reset_block)? info?
   | 'mem' id ':' info? INDENT memField* DEDENT
-  | 'cmem' id ':' type info?
-  | 'smem' id ':' type ruw? info?
```
azidar, line 98: I thought we were going to keep around the CHIRRTL syntax for backwards compatibility?

src/main/antlr4/FIRRTL.g4
```diff
@@ -145,8 +143,7 @@ info
   ;
 
 mdir
-  : 'infer'
```
azidar, line 148: Why are we getting rid of `infer`? Doesn't Chisel generate this?

Issue #1207, opened 2019-10-23
Type-parameterized annotations are not deserializable


jwright6323:
**Type of issue**: bug report

**What is the use case for changing the behavior?**
If a user tries to build a type-parameterized annotation, chisel will emit it without complaint, but it will not be deserializable by firrtl.

**Impact**: API modification

**Development Phase**: request

**Other information**

For example:
```
trait MyCoolTrait {
  val foo: Int
}

case class MyConcreteClassA(foo: Int) extends MyCoolTrait
case class MyConcreteClassB(foo: Int, bar: String) extends MyCoolTrait

case class MyAnnotation[T <: MyCoolTrait](target: ModuleTarget, data: T)
  extends SingleTargetAnnotation[ModuleTarget] { /* snipped boilerplate */ }
```

This is serialized by chisel without issue, but cannot be deserialized by firrtl because no type hints for `data` are provided. I think there are basically two options:

1. Figure out a smart way to enable this
2. Explicitly disallow it somehow so that chisel errors out when building such a thing

But... I currently have no suggestions on how to do either of these. :)

seldridge:
You may want to investigate a custom serializer/deserializer. You get magic see/Des for case classes through json4s but anything more complicated will start to break down. I have not hit the type parametric case before, though! I was aware that function parameters currently don't work either.

jwright6323:
Currently I have a custom json4s serializer/deserializer for my case classes, and my annotation takes the serialized string as a parameter, but it seems a bit clunky. I wasn't able to figure out how to get the annotation serializer to correctly (de-)serialize my case classes without modifying firrtl itself.

seldridge:
Your use case is custom transforms with custom annotations that need to define how they should be serialized/deserialized without modifying FIRRTL? 

I think we could do this with the `RegisteredTransform` approach where additional serializers/deserializers are defined in the transform and FIRRTL then adds registered serializers/deserializers.

I'll provide a link to the function serialization stuff (https://github.com/freechipsproject/firrtl/issues/847) as these are somewhat related.

jwright6323:
> Your use case is custom transforms with custom annotations that need to define how they should be serialized/deserialized without modifying FIRRTL?

That's a good way of describing it, but perhaps we could relax it a bit and say that I just want to specify how some of the fields in the custom annotation are serialized/deserialized rather than the entire annotation.

davidbiancolin:
[...]

PR #1211, opened 2019-10-24
Supply a trait to allow user annotations to provide JsonProtocol type hints

davidbiancolin:
This is one sketch of a solution to permit more flexible serialization and deserialization of user annotations defined outside of FIRRTL. Currently in FireSim, we use a hacked up version of the JsonProtocol to do precisely what's included in this PR.

I think it would be very easy to add an extra member to supply a custom serializer/deserializer but Json4s does a pretty good job by itself assuming you give it sufficient type hints. It can for instance handle type-parameterized case classes (see #1207), without needing to write a serializer.

### DIFF
--- a/src/main/scala/firrtl/annotations/JsonProtocol.scala
+++ b/src/main/scala/firrtl/annotations/JsonProtocol.scala
@@ -113,14 +113,17 @@ object JsonProtocol {
     }
     // Recursively gather typeHints by pulling the "class" field from JObjects
     // Json4s should emit this as the first field in all serialized classes
-    def findTypeHints(classInst: Seq[JValue]): Seq[String] = classInst.flatMap({
+    // Setting requireClassField mandates that all JObjects must provide a typeHint,
+    // this used on the first invocation to check all annotations do so
+    def findTypeHints(classInst: Seq[JValue], requireClassField: Boolean = false): Seq[String] = classInst.flatMap({
       case JObject(("class", JString(name)) :: fields) => name +: findTypeHints(fields.map(_._2))
-      case obj: JObject => throw new InvalidAnnotationJSONException(s"Expected field 'class' not found! $obj")
+      case obj: JObject if requireClassField => throw new InvalidAnnotationJSONException(s"Expected field 'class' not found! $obj")
+      case JObject(fields) => findTypeHints(fields.map(_._2))
       case JArray(arr) => findTypeHints(arr)
       case oJValue => Seq()
     }).distinct
 
-    val classes = findTypeHints(annos)
+    val classes = findTypeHints(annos, true)
     val loaded = classes.map(Class.forName(_).asInstanceOf[Class[_]])
     implicit val formats = jsonFormat(loaded)
     read[List[Annotation]](in)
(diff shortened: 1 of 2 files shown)


src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -9,6 +9,20 @@ import org.json4s.native.JsonMethods._
...
+  def typeHints(): Seq[Class[_]]
+}
+
+/**
+  * An annotation carrying fully qualified class names for which we'd like the
+  * deserialization protocol to generate type hints for.
+  *
+  * @param typeTags the the additional class names
+  */
+case class DeserializationTypeHintsAnnotation(typeTags: Seq[String]) extends NoTargetAnnotation
```
azidar: Is it possible to have the typeTag argument be `Seq[Class[_]]` instead of `Seq[String]`? Or does that defeat the purpose of having this annotation be serializable?

src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -74,7 +88,7 @@ object JsonProtocol {
   ))
 
   /** Construct Json formatter for annotations */
-  def jsonFormat(tags: Seq[Class[_ <: Annotation]]) = {
+  def jsonFormat(tags: Seq[Class[_]]) = {
```
azidar: I don't fully understand why this change is necessary?

src/main/scala/firrtl/annotations/JsonProtocol.scala
davidbiancolin: I don't think it's serializable otherwise (but i'd have to check).

src/main/scala/firrtl/annotations/JsonProtocol.scala
davidbiancolin: The classes for which we want to provide tags  for may no longer be a subclass of Annotation, since there may be arbitrarily nested class instances _within_ the annotation we're trying to serialize.

davidbiancolin:
Naturally this breaks all tests that expected some fixed number of annotations...

src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -9,6 +9,20 @@ import org.json4s.native.JsonMethods._
 import org.json4s.native.Serialization
 import org.json4s.native.Serialization.{read, writePretty}
 
+trait HasSerializationHints {
+  // For serialization of complicated constuctor arguments, let the annotation
+  // writer specify additional type hints for relevant classes that might be
+  // contained within
+  def typeHints(): Seq[Class[_]]
```
jackkoenig:
```suggestion
  def typeHints: Seq[Class[_]]
```

src/main/scala/firrtl/annotations/JsonProtocol.scala
jackkoenig: It could be serializable by providing custom serializers!

jackkoenig:
> I think it would be very easy to add an extra member to supply a custom serializer/deserializer but Json4s does a pretty good job by itself assuming you give it sufficient type hints. It can for instance handle type-parameterized case classes (see #1207), without needing to write a serializer.

I didn't realize that Json4s does such a good job, that's cool.

This looks mostly good. I'm not a huge fan of `DeserializationTypeHintsAnnotation` since this provides additional annotations that must be handled.

An alternative (that might not work) could be to put `HasSerializationHints` on the companion object of Annotations, then the JSONProtocol could (reflectively, ew) load the companion object of each annotation class, see if it mixes in that trait, and include the type hints if it does. Thoughts?

azidar:
@Mergifyio backport 1.2.x

Issue #1215, opened 2019-10-29
Port InstanceGraph APIs to use Target


albert-magyar:
Most of the APIs in `InstanceGraph` use an ad-hoc mix of product types to box things. This could all switch to `Target`, but it would break people's code.

I propose that a deprecation warning appears in 1.3 on the old methods, and that they go away for 1.4.

PR #1216, opened 2019-10-29
Change findInstancesInHierarchy to return implicit top instance

albert-magyar:
* Asking for instances of `Top` now returns implicit top instance
* Make API more "DWIM" and consistent with other methods
* Add test cases for `findInstancesInHierarchy`
* Update Scaladoc

All of `InstanceGraph` should definitely be ported to use `Target`. It was written before instance annotations were a thing, so it uses `WDefInstance`s to represent products of data, but it would be better to use `InstanceTarget`. I have opened #1215, but that is longer-term, since it will break lots of people's code.

### DIFF
--- a/src/main/scala/firrtl/analyses/InstanceGraph.scala
+++ b/src/main/scala/firrtl/analyses/InstanceGraph.scala
@@ -62,13 +62,12 @@ class InstanceGraph(c: Circuit) {
     */
   lazy val fullHierarchy: mutable.LinkedHashMap[WDefInstance,Seq[Seq[WDefInstance]]] = graph.pathsInDAG(trueTopInstance)
 
-  /** A count of the *static* number of instances of each module. For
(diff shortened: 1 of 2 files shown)


jackkoenig:
@albertchen-sifive or @azidar have thoughts on this?

jackkoenig:
Is this `1.2.1` or `1.3.0`?

albert-magyar:
I would say this is mostly a bugfix (the `instantiated(_)` check actually made it return incorrect results), so 1.2.1.

jackkoenig:
I agree

jackkoenig:
@Mergifyio backport 1.2.x

jackkoenig:
> Hey, we reacted but our real name is @Mergifyio

That's literally how I spelled your name

albert-magyar:
Rude. And they won't even sign the CLA...

PR #1218, opened 2019-10-30
Ignore extmodule instances in Flatten

### LINKED ISSUE #1162: Flatten tries to flatten ExtModules
**Type of issue**: bug report

Small example where you annotate the `Top.i` with the `FlattenAnnotation` (adapted from the `FlattenTests.scala` file).
```
circuit Top :
  module Top :
    input a : UInt<32>
    input na : UInt<32>
    output b : UInt<32>
    output nb : UInt<32>
    inst i of Inline1
    inst ni of NotInline1
    i.a <= a
    b <= i.b
    ni.a <= na
    nb <= ni.b
  module NotInline1 :
    input a : UInt<32>
    output b : UInt<32>
    inst i of Inline2
    i.a <= a 
    b <= i.a 
  module Inline1 :
    input a : UInt<32>
    output b : UInt<32>
    inst i of ExternalMod
    i.a <= a 
    b <= i.a 
  extmodule ExternalMod :
    input a : UInt<32>
    output b : UInt<32>
    defname = ExternalMod
```

When flattening a module with an `ExtModule` within it, the `Inline/Flatten` passes try to flatten/inline the `ExtModule` causing an error:

```
firrtl.passes.PassExceptions: firrtl.passes.PassException: Annotated module cannot be an external module: nameOfExtModule_TO_FLATTEN
```

It seems like the `Flatten` xform should not try and flatten `ExtModule`. Rather everything else should be flattened and the `ExtModule` will be kept the same.

Firrtl version: `9fbe7e4d67f9039439e8b49ba7d36296d0dd91fb` 

Allow `Module`s to be flattened with `ExtModule`s within them.

**Impact**: unknown

**Other information**

albert-magyar:
* Closes #1162
* Instances of extmodules remain in the final hierarchy
* Extmodule definitions are not renamed or duplicated
* The rest of the pass may proceed as normal

@abejgonzalez I also added a simpler extmodule-containing example as a test case.

This is not an API change, since previously the invocation of `InlineInstances` in `Flatten` would just error out in the case where there was an `extmodule` in the hierarchy to be flattened.

### DIFF
--- a/src/main/scala/firrtl/transforms/Flatten.scala
+++ b/src/main/scala/firrtl/transforms/Flatten.scala
@@ -15,16 +15,19 @@ case class FlattenAnnotation(target: Named) extends SingleTargetAnnotation[Named
 }
 
 /**
- * Takes flatten annotations for module instances and modules and inline the entire hierarchy of modules down from the annotations.
- * This transformation instantiates and is based on the InlineInstances transformation.  
- * Note: Inlining a module means inlining all its children module instances      
- */
+  * Takes flatten annotations for module instances and modules and inline the entire hierarchy of
+  * modules down from the annotations. This transformation instantiates and is based on the
+  * InlineInstances transformation.
+  *
+  * @note Flattening a module means inlining all its fully-defined child instances
+  * @note Instances of extmodules are not (and cannot be) inlined
+  */
 class Flatten extends Transform {
    def inputForm = LowForm
    def outputForm = LowForm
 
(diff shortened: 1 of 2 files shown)


abejgonzalez:
This LGTM. Thanks for the help!

albert-magyar:
@jackkoenig you have any thoughts on if this is good to label as 1.2.1?

azidar:
@Mergifyio backport 1.2.x

PR #1219, opened 2019-10-31
Guard initial blocks in emitted Verilog with `ifndef SYNTHESIS

### LINKED ISSUE #1213: Async Register initial blocks should be guarded by non-synthesis macro
**Type of issue**: bug report 

Emit an async reset register w/ FIRRTL

https://github.com/freechipsproject/firrtl/blob/f5e6e2bc201c6206a705244d8977bda4b83e6efd/src/main/scala/firrtl/Emitter.scala#L492

The code emits async register with an initial block as explained in the comment pointed to by the link above. However, as the comment describes, this initial code is for simulation only and isn't usually part of the canonical way of describing an asynchronously reset register. Therefore, this code should be guarded by non-synthesis macros. Examples include `ifndef SYNTHESIS` or `//synopsys translate off`. 

Code should be emitted like this or with some indication that this is not synthesizable code:
https://github.com/chipsalliance/rocket-chip/blob/master/src/main/resources/vsrc/AsyncResetReg.v#L56

N/A

Synthesis tools will be displeased about initial blocks that they don't know how to synthesize and throw a warning for every single async reset flop in the design. Depending on who is running the tool, this is unacceptable in some cases and can require additional post-processing steps of the emitted verilog.

**Impact**:  API modification

jackkoenig:
Fixes #1213 

With bump in rocket-chip, fixes https://github.com/chipsalliance/rocket-chip/issues/2168

Note that this is branched from the common ancestor of master and v1.2.0 so that it can be merged into both (rather than cherry-picked, although I don't know how much it matters).

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -878,6 +878,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
           emit(Seq(s"  reg [$width:0] initvar;"))
         }
         emit(Seq("`endif"))
+        emit(Seq("`ifndef SYNTHESIS"))
         emit(Seq("initial begin"))
         emit(Seq("  `ifdef RANDOMIZE"))
         emit(Seq("    `ifdef INIT_RANDOM"))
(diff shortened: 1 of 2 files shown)


jackkoenig:
Oh wait, this repo requires branches to be up to date, so never mind on being merged into both.

jackkoenig:
@Mergifyio backport 1.2.x

PR #1202, opened 2019-10-16
Fix handling of read enables for write-first (default) memories in VerilogMemDelays

albert-magyar:
Kudos to @chick for exposing a lurking bug in his issue #1179.

By default, `VerilogMemDelays` has always generated write-first memories with the pipeline registers it adds to the memory access commands. The way this was being handled had two issues:

- For normal sync-read mems (`read-latency` and `write-latency` both 1), feeding back read data to an enable field would result in a combinational loop. This came up for the cache of riscv-mini in #1179, but there is a much smaller non-working example that is now fixed in this PR (and included as a test).

- For any memories with `read-latency` greater than 1, the enable signal does not behave correctly, and can lead to incorrect read data. This space was not covered very well by existing Verilog tests.

I added some tests to try to cover some of this space. Several of of these tests (the ones with write-first, two-cycle-read memories) fail without the fix to `VerilogMemDelays`.

Some test infrastructure for getting boilerplate out of these tests ended up in a library exposing a chisel-testers2-like API (poke, expect, step) for simple Verilog execution tests. I tried to write some Scaladoc for the parts that might be reusable, and I put them in the `firrtlTests.execution` package.

Minimal example that caused combinational loops before this PR:
```
circuit loop :
  module loop :
    input clk : Clock
    input raddr : UInt<5>
    mem m :
      data-type => UInt<1>
      depth => 32
      reader => r
      read-latency => 1
      write-latency => 1
      read-under-write => undefined
    m.r.clk <= clk
    m.r.addr <= raddr
    m.r.en <= m.r.data
```

PR #870, opened 2018-08-16
Upstream intervals

azidar:
Major Changes:

- Added intervals to the IR
- Refactored width inferencer into a separate firrtl-independent constraint solver
- Added transforms to infer interval bounds
- Tests for said features

TODO
- [x] Add comments for InferBinaryPoint
- [x] Add comments for InferWidths
- [x] Add comments for TrimIntervals
- [x] Add comments for RemoveIntervals
- [x] Add comments to IntervalType vals
- [x] Passes @jackkoenig tests

Future PR: update spec

META:
- Fixes #373

PR #1223, opened 2019-11-05
Revert binary-compatibility breaking changes since 1.2.0 and enforce checking

jackkoenig:
This now does 3 things:
* Revert Intervals (https://github.com/freechipsproject/firrtl/pull/870)
* Revert VerilogMemDelays bugfix (https://github.com/freechipsproject/firrtl/pull/1202)
* Add MiMa binary-compatibility checking against `1.2.0`

This includes a couple of minor waivers, apparently removing private inner classes technically breaks binary compatibility but from Scala's perspective it's acceptable.


jackkoenig:
Note the base branch is `1.2.x`

jackkoenig:
This failed the chisel3 test because chisel3 master requires intervals...

I pushed an additional commit that has the type of checking I have in mind for `1.2.x`. It will fail the new Travis test I added due to issues discussed in https://github.com/freechipsproject/firrtl/pull/1202#issuecomment-549616153

jackkoenig:
We have a bit of a chicken-and-egg problem with the CI. I think we should first fix chisel3's `3.2.x` CI, revert stuff there, then fix the CI here (to build Chisel from `3.2.x`), and then revert stuff, then add the new binary compatibility checks.

jackkoenig:
This is larger than the original title (and branch title), please see revised top comment

jackkoenig:
Notably, the goal is to include Revert VerilogMemDelays bugfix (#1202) eventually, but there were lots of bin-compat breaking changes so I reverted it for now.

PR #1224, opened 2019-11-05
Move CheckResets after CheckCombLoops

jackkoenig:
Recursive literal lookup needs to be guarded against combinational loops
Added a test-case to illustrate the issue when CheckResets is run before CheckCombLoops

Cherry-pick of patch from https://github.com/freechipsproject/firrtl/pull/1201

As stated in #1201, we should fix `CheckResets` to not require this, but the patch fixes the problem and the test is great to include.


albert-magyar:
Yeah and none of those transforms in question that `CoreTransform` actually set the form of the output `CircuitState` to `outputForm`.

azidar:
@Mergifyio backport 1.2.x

PR #1228, opened 2019-11-07
getSimpleName considered harmful

seldridge:
`getSimpleName` can throw stupid exceptions in Java 8. This means that if you define a `Transform` inside an object (or in an environment like a REPL that is wrapping things in objects), if you call the `name` method of a `Transform` (delete an annotation, run with `-ll info`) then you'll wind up with a malformed class name exception.

This changes all usages of `getSimpleName` to use `getName`. This is not super desirable from a readability perspective, but it's actually safe.

Alternatively, we could define a custom method like https://github.com/freechipsproject/chisel3/pull/1224. However, I don't think this is really necessary here as a `Transform` name is intended for printing and debugging, not for Verilog naming.

### Checklist

### Type of Improvement

   - bug fix                            

### API Impact

This should have no API impact, but does change the default `name` of all `Transform`s. This may have impact for things relying on `DeletedAnnotation`s or on parsing `-ll info` output. The former matters as some of the Driver Stage/Phase backwards compatibility is relying on `DeletedAnnotation`s. However, Chisel does not use hard-coded strings. Anything that does use a hard-coded string will break due to this PR.

No impact.

### Desired Merge Strategy

### DIFF
--- a/src/test/scala/firrtlTests/CustomTransformSpec.scala
+++ b/src/test/scala/firrtlTests/CustomTransformSpec.scala
@@ -6,6 +6,8 @@ import firrtl.ir.Circuit
 import firrtl._
 import firrtl.passes.Pass
 import firrtl.ir._
+import firrtl.annotations.{Annotation, NoTargetAnnotation}
+import firrtl.stage.{FirrtlSourceAnnotation, FirrtlStage, RunFirrtlTransformAnnotation}
 
 class CustomTransformSpec extends FirrtlFlatSpec {
   behavior of "Custom Transforms"
@@ -71,5 +73,28 @@ class CustomTransformSpec extends FirrtlFlatSpec {
       Driver.execute(optionsManager)
     }).getMessage should include (errorString)
   }
-}
 
+  object Foo {
+    class A extends Transform {
+      def inputForm = HighForm
+      def outputForm = HighForm
+      def execute(s: CircuitState) = {
+        println(name)
+        s
+      }
+    }
+  }
+
+  they should "work if placed inside an object" in {
+    val input =
+      """|circuit Foo:
+         |  module Foo:
+         |    node a = UInt<1>(0)
+         |""".stripMargin
+    val annotations = Seq(
+      RunFirrtlTransformAnnotation(new Foo.A),
+      FirrtlSourceAnnotation(input)
+    )
+    (new FirrtlStage).execute(Array.empty, annotations)
+  }
+}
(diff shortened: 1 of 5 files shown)


src/test/scala/firrtlTests/CustomTransformSpec.scala
```diff
@@ -71,5 +73,28 @@ class CustomTransformSpec extends FirrtlFlatSpec {
...
     }).getMessage should include (errorString)
   }
-}
 
+  object Foo {
+    class A extends Transform {
+      def inputForm = HighForm
+      def outputForm = HighForm
+      def execute(s: CircuitState) = {
+        println(name)
```
chick: Seems like this would be a nicer test and make clear to code reader what the new name format looks like
 if you got replaced the `println()` with
`assert(name === "firrtlTests.CustomTransformSpec$Foo$A")`

chick: I did appreciate your XCF structure in PR name

seldridge: I'm reluctant to do that as such a test is really a check on the exact functionality of Java's `getName` method in a Scala environment. That should be allowed to change independent of the FIRRTL code base. The alternative of doing `assert(name == this.getClass.getName)` would seem to be useless...

What do you think?

This test is stripped down from what I originally had (and was exactly what was causing the problem). I had a custom transform inside an object that deleted annotations. The `propagateAnnotations` method calls the `name` method and will except similarly (though more opaquely). This would be a more opaque way of testing the same thing:

```scala
  object Foo {
    case object A extends NoTargetAnnotation
    class B extends Transform {
      def inputForm = HighForm
      def outputForm = HighForm
      private def strip(a: Annotation): Boolean = a match {
        case A => false
        case _ => true
      }
      def execute(s: CircuitState) = s.copy(annotations = s.annotations.filter(strip))
    }
  }

  they should "work if placed inside an object" in {
    val input =
      """|circuit Foo:
         |  module Foo:
         |    node a = UInt<1>(0)
         |""".stripMargin
    val annotations = Seq(
      Foo.A,
      RunFirrtlTransformAnnotation(new Foo.B),
      FirrtlSourceAnnotation(input) )

    (new FirrtlStage).execute(Array.empty, annotatinos)
  }
```

chick: Presumably we would catch such a change some other way, right? Any java changes that we can catch we should because it could affect users

albert-magyar: I can see benefits to both. Generally, I agree with the "conservative test failing as informative signal rather than true bug" philosophy, but I don't know how well that extends to things that only get printed in debug output.

I agree with Schuyler that even though Scala class name mangling and nesting is stable within a major version, it changes a lot across major versions, especially for anonymous values.

azidar:
We are ok with this change being in a minor release, correct? @seldridge

seldridge:
@Mergifyio backport 1.2.x

seldridge:
Yeah, this is fine to go in a minor release. No effect on the output Verilog.

PR #1234, opened 2019-11-12
Formatted 1

chick:
### Checklist

### Type of Improvement
Installed .scalafmt.config file
Reformatting all files in src to see what it looks like
Note: Took two passes to stabilize

### Primary Impact
Could break existing PRs

### API Impact
Should be no API changes. 

No code generation changes

### Desired Merge Strategy
Up in the air

### DIFF
--- a/.scalafmt.config
+++ b/.scalafmt.config
@@ -7,7 +7,7 @@ lineEndings = preserve
 includeCurlyBraceInSelectChains = false
 danglingParentheses = true
 
-align.tokens = [
+align.tokens.add = [
   {
     code = ":"
   }
--- a/.scalafmt.config
+++ b/.scalafmt.config
@@ -1,0 +1,20 @@
+maxColumn = 120
+align = most
+continuationIndent.defnSite = 2
+assumeStandardLibraryStripMargin = true
+docstrings = ScalaDoc
+lineEndings = preserve
+includeCurlyBraceInSelectChains = false
+danglingParentheses = true
+
+align.tokens = [
+  {
+    code = ":"
+  }
+]
+
+newlines.alwaysBeforeCurlyBraceLambdaParams  = false
+
+optIn.annotationNewlines = true
+
+rewrite.rules = [SortImports, PreferCurlyFors, AvoidInfix]
--- a/src/main/scala/firrtl/AddDescriptionNodes.scala
+++ b/src/main/scala/firrtl/AddDescriptionNodes.scala
@@ -31,33 +31,33 @@ private case object EmptyDescription extends Description {
 
 private case class DescribedStmt(description: Description, stmt: Statement) extends Statement with HasDescription {
   def serialize: String = s"${description.serialize}\n${stmt.serialize}"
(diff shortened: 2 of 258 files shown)


.scalafmt.config
```diff
@@ -0,0 +1,20 @@
...
+danglingParentheses = true
+
+align.tokens = [
+  {
+    code = ":"
+  }
+]
+
+newlines.alwaysBeforeCurlyBraceLambdaParams
 = false
```
seldridge: Not sure what this character is in here?

```suggestion
newlines.alwaysBeforeCurlyBraceLambdaParams = false
```

.scalafmt.config
```diff
@@ -0,0 +1,20 @@
+maxColumn = 120
+align = most
+continuationIndent.defnSite = 2
+assumeStandardLibraryStripMargin = true
+docstrings = ScalaDoc
+lineEndings = preserve
+includeCurlyBraceInSelectChains = false
+danglingParentheses = true
+
+align.tokens = [
```
seldridge: I think that this is resulting in *only* `:` being aligned. I think the intent was that `:` would *also* be aligned. Concretely, this seems to be causing `=>` to not be aligned which was one of the things I think we wanted?

```suggestion
align.tokens.add = [
```

chick: I like it. I will make the change. The weird thing I checked a few => and thought they were aligned.

src/main/scala/firrtl/AddDescriptionNodes.scala
```diff
@@ -31,33 +31,33 @@ private case object EmptyDescription extends Description {
...
-  def mapExpr(f: Expression => Expression): Statement = this.copy(stmt = stmt.mapExpr(f))
-  def mapType(f: Type => Type): Statement = this.copy(stmt = stmt.mapType(f))
-  def mapString(f: String => String): Statement = this.copy(stmt = stmt.mapString(f))
-  def mapInfo(f: Info => Info): Statement = this.copy(stmt = stmt.mapInfo(f))
-  def foreachStmt(f: Statement => Unit): Unit = f(stmt)
-  def foreachExpr(f: Expression => Unit): Unit = stmt.foreachExpr(f)
-  def foreachType(f: Type => Unit): Unit = stmt.foreachType(f)
-  def foreachString(f: String => Unit): Unit = stmt.foreachString(f)
-  def foreachInfo(f: Info => Unit): Unit = stmt.foreachInfo(f)
+  def mapStmt(f:       Statement => Statement):   Statement = f(stmt)
```
ekiwi: This alignment seems a bit overzealous.

chick: I think in this case it is a bit, but there are other cases where I think the alignment aids readability.
How overzealous do you think it is.
And would you like to join the formatting meeting at 1:30 on Friday to weigh in?

.scalafmt.config
chick: The change was made and firrtl had been reformatted again and is pushed

chick:
Will be revisited when PR queue diminishes

PR #1235, opened 2019-11-12
Updates for sbt usage of scalafmt

seldridge:
I tried to run the `formatted-1` branch locally, but `sbt` wanted these modifications. Specifically, it wants the file to be called `.scalafmt.conf` not `.scalafmt.config`. Also, the config needs a version string and this adds `scalafmt` as a plugin. I expect that IntelliJ already has all this so it's superfluous there.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,3 +1,4 @@
+version = "2.2.2"
 maxColumn = 120
 align = most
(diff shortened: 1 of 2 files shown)


chick:
Because of the apparent conflict, I have merged this branch into formatted-1 manually.

PR #1238, opened 2019-11-13
Error on nested memory datatypes

abejgonzalez:
### Checklist

### Type of Improvement

code cleanup

### API Impact

No change

This is very similar to what I wrote in the code comment:

This PR catches incorrect memory instantiations when there are nested aggregate types. This is mainly because "repl-seq-mem" MemConf file doesn't support these types of nested datatype memories. Here is an example case:

Here is a memory that I create with a datatype of `Vec(M, new Bundle...)`
```
val memory = SyncReadMem(N, Vec(M, new Bundle {
  val a = Bool()
  val b = UInt(3.W)
}))
```

In this case, the wrapper memory module will have M\*2 mask bits (one bit for each entry in the bundle over each element in the Vec). Additionally, the `DefAnnotatedMemory` will have a `maskGran` (from `ResolveMaskGranularity`) of 4 since the bundle width is 4 bits. Thus, the MemConf file will have a 4 in its `maskGran` field. However, when "vlsi-mem-gen" or Chipyards `MacroCompiler` creates a memory, it will first get the width of the data (here is it M\*4) and divide it by the `maskGran` yielding (M\*4)/4 == M mask bits. This breaks with the M\*2 mask bits created in the FIRRTL wrapper. This is mainly because FIRRTL emits mask bits that align with the structure of the memory datatype. However, MemConf doesn't support this since it assumes that each element of the Vec will only have 1 mask bit associated with it.

The current workaround is to do the following:
```
val memory = SyncReadMem(N, Vec(M, UInt(SIZE_OF_BUNDLE)))
```

Is there a cleaner solution to this (better place to error)? This works on my current test case.

### Desired Merge Strategy

Squash


abejgonzalez:
@albert-magyar @jerryz123

src/main/scala/firrtl/passes/memlib/ReplaceMemMacros.scala
```diff
@@ -36,17 +36,48 @@ class ReplaceMemMacros(writer: ConfWriter) extends Transform {
...
+    *   However, the wrapper created will have M*NUM_BUNDLE_ENTRIES bits.
+    *   In this particular case M != M*2 breaking the ext mem module interface.
+    */
+  private def checkMaskDatatype(mem: DefAnnotatedMemory) = {
+    mem.dataType match {
+      case t: VectorType =>
+        t.tpe match {
+          case t: GroundType =>
+          case _ =>
+            throw new PassException(s"${mem.info} : Unsupported use of nested AggregateType (i.e. Vec(N, new Bundle{...})). Instead, use Vec(N, UInt()).")
```
albert-magyar: Make this error specify that it's _blackboxing_ these memories that isn't supported.

src/main/scala/firrtl/passes/memlib/ReplaceMemMacros.scala
```diff
@@ -36,17 +36,48 @@ class ReplaceMemMacros(writer: ConfWriter) extends Transform {
...
+        }
+      case _ =>
```
albert-magyar: Unless you're totally sure that memories with bundle types are filtered out before this method is called, it would be good to put similar error message here to avoid a `MatchError`.

src/main/scala/firrtl/passes/memlib/ReplaceMemMacros.scala
albert-magyar: It might also be reasonable to define a new flavor of exception (like `ReplaceMemMacros.UnsupportedBlackboxMemory` or something) to get more informative error prompts, but that's up to you.

jackkoenig:
Since this is just erroring rather than emitting bad code, should this be `1.2.X`?

albert-magyar:
Yeah, I think so.

PR #1241, opened 2019-11-18
bug fix for freechipsproject/chisel3#1240

sequencer:
### Type of Improvement

Choose one or more from the following:
- bug fix

bug fix for freechipsproject/chisel3#1240

### Desired Merge Strategy
- Squash


seldridge:
I just want to note that this is an API change as it's a modification of the specification:

2019-11-18-101113_572x181_scrot

The modification proposed in this PR may make writing generators easier, however.

sequencer:
Sorry I missed that spec. The original issue sounds quite reasonable. I wonder if we can propose a modification to the spec?

ucbjrl:
@azidar looks like requested changes have been made. Can you approve?

PR #1245, opened 2019-11-19
[Mergify] Automatically merge backport PRs when ready

albert-magyar:
@jackkoenig 

Mergify can merge backports now. This rule could be generalized for more backport branches in the future, but I thought it would be good to keep it separate from master by checking for a) a backport branch and b) the label.

We could even drop the review requirement for these.

### Desired Merge Strategy: squash

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -43,10 +43,8 @@ pull_request_rules:
   - name: automatic squash-and-merge of backport PRs
     conditions:
       - status-success=continuous-integration/travis-ci/pr
-      - "#approved-reviews-by>=1"
       - "#changes-requested-reviews-by=0"
       - base=1.2.x
-      - label="Please Merge"
       - label="Backport"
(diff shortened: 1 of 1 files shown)


seldridge:
I would be in favor of dropping the review requirement. I doubt that a reviewer would be able to add value or even spot a problem in a backport that merged cleanly and passed all tests. (Or, I certainly wouldn't be able to for 99.9% of PRs!)

jackkoenig:
I'm happy to drop the review requirement, I wish we could put like a 2 hour delay on it or something just in case someone notices something. Even if it messes up for some reason, we can always revert, so I'm happy for automerge (assuming CI passes, as you've required)

albert-magyar:
I think a waiting period would be awesome

albert-magyar:
I updated the YAML to drop the `Please Merge` and review requirements.

PR #1257, opened 2019-11-25
Bump versions for next release

ucbjrl:
### Type of Improvement

1.2.1 release

### API Impact

No API differences from 1.2.0

### Desired Merge Strategy

Squash


ucbjrl:
**NOTE**: If you're doing a `publishLocal`, you should delete any existing copies of this version in your ivy or coursier cache. I believe sbt won't overwrite an existing non-SNAPSHOT version.

PR #1258, opened 2019-11-29
Remove scala-logging fully in favor of our own logger

jackkoenig:
There was some vestigial logging that conflicts with the homebrewed
logger used by most of the codebase

### Checklist

### Type of Improvement

Code cleanup to remove a vestigial dependency. This is definitely 1.3, it should **not** be included in 1.2.X

### API Impact

This is unlikely to conflict with anything, but technically changes the types of `object Parser` and `object PrimOps` in that they are now `logger.LazyLogging` instead of `com.typesafe.scalalogging.LazyLogging`. Furthermore, any codebases relying on depending on FIRRTL to bring in the `scala-logging` dependencies will no longer be getting those dependencies through FIRRTL.

No impact

### Desired Merge Strategy

Don't care

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -55,10 +55,6 @@ javacOptions ++= javacOptionsVersion(scalaVersion.value)
 
 libraryDependencies += "org.scala-lang" % "scala-reflect" % scalaVersion.value
 
-libraryDependencies += "com.typesafe.scala-logging" %% "scala-logging" % "3.9.0"
-
-libraryDependencies += "ch.qos.logback" % "logback-classic" % "1.2.3"
-
 // sbt 1.2.6 fails with `Symbol 'term org.junit' is missing from the classpath`
 // when compiling tests under 2.11.12
 // An explicit dependency on junit seems to alleviate this.
--- a/src/main/scala/firrtl/ExecutionOptionsManager.scala
+++ b/src/main/scala/firrtl/ExecutionOptionsManager.scala
@@ -2,14 +2,14 @@
 
 package firrtl
 
+import logger.LogLevel
+import logger.{ClassLogLevelAnnotation, LogClassNamesAnnotation, LogFileAnnotation, LogLevelAnnotation}
 import firrtl.annotations._
-import firrtl.Parser._
+import firrtl.Parser.{InfoMode, UseInfo, IgnoreInfo, GenInfo, AppendInfo}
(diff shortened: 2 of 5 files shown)


ucbjrl:
Unfortunately, this triggers a latent name resolution bug lurking in our `Logger`. The problem appears to be due to:
```scala
package logger
   ...
trait LazyLogging {
  val logger = new Logger(this.getClass.getName)
}
```
A wildcard import of a class or object that extends this trait, introduces the `val logger` into the namespace, and subsequent attempts to `import logger`, refer to this imported val. Attempting to refer to `logger.LazyLogging`, produces the following confusing error message:
```
[error] ...: type LazyLogging is not a member of logger.Logger
[error]   extends logger.LazyLogging {
```

The fix appears to be:
```scala
trait LazyLogging {
  protected val logger = new Logger(this.getClass.getName)
}
```

I'm not sure this isn't a Scala bug. I thought that changing the accessibility of an object didn't alter it's visibility - you couldn't resolve namespace collisions by making objects `protected` or `private`.

I think the real solution is to avoid object and package name collisions, but that is a larger API change that needs some discussion.

jackkoenig:
I don't think this is a Scala bug, it is totally reasonable Scala to import from any arbitrary object:
```scala
scala> val xs = List(1, 2, 3)
xs: List[Int] = List(1, 2, 3)

scala> import xs._
import xs._

scala> apply(1)
res0: Int = 2

scala> map(_ + 1)
res1: List[Int] = List(2, 3, 4)
```
I completely agree that it's a confusing error message though and you're right that we should avoid object and package name collisions. I think changing `logger` to protected as proposed in https://github.com/freechipsproject/firrtl/pull/1271 is a good bandaid, but it doesn't even fix the whole problem because will still get the same error message if you try to `import logger.<whatever>` within any class that mixes in `LazyLogging`

ucbjrl:
... and the spec says:

> An import clause has the form import $p$.$I$ where $p$ is a stable identifier and $I$ is an import expression. The import expression determines a set of names of importable members of $p$ which are made available without qualification. A member $m$ of $p$ is importable if it is accessible.

which implies that adding an access modifier does determine visibility.

I think you can use:
```scala
import _root_.logger.<whatever>
```
to work around the `logger` name clash in classes that mix in the `LazyLogging` trait.

PR #1259, opened 2019-11-30
Cleanup testing console

jackkoenig:
Built on top of https://github.com/freechipsproject/firrtl/pull/1258 so requires that to be merged first.

Because of the dependency on #1258, it cannot be backported. It doesn't really matter though, this is for testing/debugging which we generally do against master.

### Checklist

### Type of Improvement

Code refactoring/cleanup.

This PR makes the console for running tests *much* cleaner by not printing the shell commands and output of Verilator and stuff by default. Instead, these processes use LogLevel.Info for Stdout and LogLevel.Warn for Stderr. I also lowered the level for printing total FIRRTL compile time from error to warn.

### API Impact

Some people may not see the total FIRRTL compile time by default any more, this is easily fixed by using log-level warn.

No effect

### Desired Merge Strategy

Merge commit, I'll take care of rebasing as necessary.

### DIFF
--- a/src/test/scala/firrtlTests/CustomTransformSpec.scala
+++ b/src/test/scala/firrtlTests/CustomTransformSpec.scala
@@ -79,7 +79,7 @@ class CustomTransformSpec extends FirrtlFlatSpec {
       def inputForm = HighForm
       def outputForm = HighForm
       def execute(s: CircuitState) = {
-        println(name)
+        assert(name.endsWith("A"))
         s
       }
     }
--- a/src/test/scala/firrtlTests/DriverSpec.scala
+++ b/src/test/scala/firrtlTests/DriverSpec.scala
@@ -498,8 +498,8 @@ class VcdSuppressionSpec extends FirrtlFlatSpec {
       val harness = new File(testDir, s"top.cpp")
       copyResourceToFile(cppHarnessResourceName, harness)
 
-      verilogToCpp(prefix, testDir, Seq.empty, harness, suppress).!
-      cppToExe(prefix, testDir).!
+      verilogToCpp(prefix, testDir, Seq.empty, harness, suppress) #&&
+      cppToExe(prefix, testDir) ! loggingProcessLogger
 
       assert(executeExpectingSuccess(prefix, testDir))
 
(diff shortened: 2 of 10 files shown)


jackkoenig:
TODO check performance

src/test/scala/firrtlTests/CustomTransformSpec.scala
```diff
@@ -79,7 +79,6 @@ class CustomTransformSpec extends FirrtlFlatSpec {
       def inputForm = HighForm
       def outputForm = HighForm
       def execute(s: CircuitState) = {
-        println(name)
```
seldridge: Removal of this actually negates what this test is checking. Namely, this is asserting that calling the `name` method doesn't cause a crach. I am in favor of keeping the verbosity down here, so maybe this could be changed to a bare `name` call? (see: https://github.com/freechipsproject/firrtl/pull/1228)

src/test/scala/firrtlTests/DriverSpec.scala
```diff
@@ -76,21 +76,18 @@ class DriverSpec extends FreeSpec with Matchers with BackendCompilationUtilities
       val optionsManager = new ExecutionOptionsManager("test")
 
       optionsManager.parse(Array("--top-name", "dog", "fox", "tardigrade", "stomatopod")) should be(true)
-      println(s"programArgs ${optionsManager.commonOptions.programArgs}")
```
seldridge: These three `println`s are perhaps useful. Consider using:

```scala
info(s"""programArgs: ${optionsManager.commonOptions.programArgs.mkString(", ")}""")
```

To produce the following output:

```
[info] - options include by default a list of strings that are returned in commonOptions.programArgs
[info]   + programArgs: fox, tardigrade, stomatopod 
```

jackkoenig:
@seldridge I resolved the issues with the test printing (I think)

src/test/scala/firrtlTests/CustomTransformSpec.scala
seldridge: Your solution is better than my `println`. Thanks!

ucbjrl:
It looks like the default `state.globalLevel` is `None` which means that any messages other than `Error` (i.e., `printf()` in tests) are suppressed.
I'm not sure that this is a good idea and is likely to further confuse and frustrate users. I think the default level should be `Warn`.

PR #1260, opened 2019-12-02
RFC: Enhance support for singleton transforms

albert-magyar:
Currently, passes and transforms in FIRRTL are a mix of singletons and regular classes. Since I would like to see us move to the robust graph of passes that can be expressed in the Dependency API, I think there are compelling reasons to use singletons.

* The `TransformLike` interface defines a transform as a mathematical operation. Executing its operation with a _static_ execute method provides at least some documentation of programmer intent to adhere to this requirement.
* Transform determinism has been an ongoing issue (more-or-less restating the previous bullet). When the user makes a transform a singleton, they are more likely to be in the mindset of respecting determinism.
* Internally, the `DependencyManager` relies on single reference instance per dependency. This makes the transform ordering analysis robust, but it further incentivizes avoiding multiple instances of a transform.
* Automatically-ordered transforms cannot have constructor arguments.

Many of the early FIRRTL transforms/passes were singletons, but the prevalence dropped for a few reasons:
* Non-referentially transparent mutable state present at the top level of the transform. I would consider this harmful given the idea of "transform as mathematical operation."
* One of the more interesting reasons was in #1057, where a referentially transparent cache was breaking in a threaded context. I fixed this by using thread-safe collections when restoring `RemoveAccesses` to an object.
* APIs like `RunFirrtlTransformAnnotation` support instantiating classes based on command line options.
* The Dependency API is expressed in terms of `Class`es that get instantiated, which currently breaks support for singleton transforms.

I would like to avoid dropping support for singleton transforms in 1.3, since they are prevalent and have nice properties.

The first two commits of this PR fix technical barriers to full support of singleton transforms. These are mostly meant to be a prototype idea for consideration, since they modify some code from @seldridge that would break some downstream PRs.
 
This involves two main changes:
* Changing the type of dependencies to an `Either` that can accept a `Class` or a `Singleton` (Scala built-in type since 2.10 that all singletons comply with). This involves some convenience methods to wrap dependency list creation and object creation, which can be seen in use in the first commit.

### DIFF
--- a/src/main/scala/firrtl/options/DependencyManager.scala
+++ b/src/main/scala/firrtl/options/DependencyManager.scala
@@ -1,4 +1,4 @@
-// See LICENSE for license details.
+// See LICENSE for license detAails.
 
 package firrtl.options
 
@@ -9,6 +9,8 @@ import scala.collection.Set
 import scala.collection.immutable.{Set => ISet}
 import scala.collection.mutable.{ArrayBuffer, HashMap, LinkedHashMap, LinkedHashSet, Queue}
 
+import scala.reflect.{ClassTag, runtime}
+
 /** An exception arising from an in a [[DependencyManager]] */
 case class DependencyManagerException(message: String, cause: Throwable = null) extends RuntimeException(message, cause)
 
@@ -17,7 +19,7 @@ case class DependencyManagerException(message: String, cause: Throwable = null) 
   * @tparam A the type over which this transforms
   * @tparam B the type of the [[firrtl.options.TransformLike TransformLike]]
   */
-trait DependencyManager[A, B <: TransformLike[A] with DependencyAPI[B]] extends TransformLike[A] with DependencyAPI[B] {
+abstract class DependencyManager[A, B <: TransformLike[A] with DependencyAPI[B] : ClassTag] extends TransformLike[A] with DependencyAPI[B] {
 
   /** Requested [[firrtl.options.TransformLike TransformLike]]s that should be run. Internally, this will be converted to
     * a set based on the ordering defined here.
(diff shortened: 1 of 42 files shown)


src/main/scala/firrtl/options/DependencyManager.scala
```diff
@@ -1,4 +1,4 @@
-// See LICENSE for license details.
+// See LICENSE for license detAails.
```
azidar: I approve of your attention to detAail.

src/main/scala/firrtl/options/Phase.scala
```diff
@@ -8,6 +8,32 @@ import logger.LazyLogging
 
 import scala.collection.mutable.LinkedHashSet
 
+object DependencyList {
+  def apply[A](dependencies: DependencyID[A]*): Seq[DependencyID[A]] = {
+    dependencies.toSeq
+  }
+}
+
+case class DependencyID[A](id: Either[Class[_ <: A], A with Singleton]) {
```
jackkoenig: Does it make more sense to have a `case class` of `Either` or should `DependencyID` be a trait with two concrete subclasses, one for `Class[_ <: A]` and one for `A with Singleton`?

albert-magyar: That's a good point. I used `Either` to make the "one or the other" aspect a bit more self-documenting, but case classes do have some nice benefits:

- Dynamic dispatch instead of `Left`/`Right`-casing for `getObject`
- Consistency with pattern that people actually see in the rest of FIRRTL

I guess it also comes down to whether anyone actually seeing `Either` could ever be usefully self-documenting or whether it would just confuse people.

albert-magyar:
Yeah, I can make a new branch that merges this into #1123 this week.

seldridge:
Superseded by #1275 which is now merged into #1123.

PR #1264, opened 2019-12-03
Cleanup verilog emitter casts

jackkoenig:
### Checklist

### Type of Improvement

This is a backend code generation improvement. It removes unnecessary casts that are frequently lint warnings in standard Verilog tools

### API Impact

* Adds new `isCast` utilities to `firrtl.Utils` to check if a `PrimOp` or an `Expression` is a cast
* Adds `InlineCastsTransform` which is used by the `VerilogEmitter` to create complex expressions if one of the expressions is a cast

This is intended to make our generated Verilog more friendly to standard lint tools (eg. Spyglass). Because we emit all wires/regs/ports as unsigned, statements like
```verilog
assign foo = $signed(bar);
```
are pointless, `foo` is unsigned as is `bar`.

Assuming we're then adding `SInt` `foo` to a negative literal, we emit:
```verilog
assign x = $signed(foo) + $signed(-8'sd10);
```
This PR will instead emit the more sensible:
```verilog
assign x = $signed(bar) + -8'sd10;
```

To accomplish these improvements, this PR does the following:
1. Optimize unnecessary casts in ConstantPropagation
2. Inline cast Expressions for cleaner Verilog emission
3. Remove redundant casts in Verilog emission

(1) is straight forward `asUInt` on a `UInt` is replaced with the `UInt` itself (Works for other casts of course)

(2) has 2 forms:
2a)
```
node _T_1 = asSInt(x)
node _T_2 = add(_T_1, y)
```
becomes
```
node _T_2 = add(asSInt(x), y)
```
And
```
node _T_1 = add(x, y)
out <= asUInt(_T_1)
```
becomes
```
out <= asUInt(add(x, y))
```

(3) is perhaps the weirdest part of this PR. Basically, the Verilog emitter is already injecting $signed casts for any signed primitive operations irrespective of the types of the arguments. We could be a bit more principled about this and only inject them if the argument is a signed reference, otherwise emit the cast as appropriate. That's not how I've done it though, at the moment it just ignores all casts, assuming the circuit is in a valid form, and then injects the casts for the operators that need them. This is more in line with the existing implementation of the emitter, but I'm not against small to moderate cleanups.

### Desired Merge Strategy

Merge commit

### Future Work 

It would be good to clean up the Verilog emitter primitive operation emission a bit. This is not necessarily easy because you would like to know if the prim-op is a "top-level" Expression (ie. outer most in the LHS of a Statement), but that information is not available and isn't trivial to add as architected. 

InlineCastsTransform is extremely similar in analysis and structure to ReplaceTruncatingArithmetic. I'd like to chain the transforms manually to (1) improve performance, but (2) illustrate how we might reason about transform chaining (and analysis sharing) in the future. This is definitely beyond the scope of this PR, but something I am thinking about.

### DIFF
--- a/src/main/scala/firrtl/transforms/InlineCasts.scala
+++ b/src/main/scala/firrtl/transforms/InlineCasts.scala
@@ -1,0 +1,70 @@
+package firrtl
+package transforms
+
+import firrtl.ir._
+import firrtl.Mappers._
+
+import firrtl.Utils.{isCast, NodeMap}
+
+object InlineCastsTransform {
+
+  // Checks if an Expression is made up of only casts terminated by a Literal or Reference
+  // There must be at least one cast
+  // Note that this can have false negatives but MUST NOT have false positives
+  private def isSimpleCast(castSeen: Boolean)(expr: Expression): Boolean = expr match {
+    case _: WRef | _: Literal | _: WSubField => castSeen
+    case DoPrim(op, args, _,_) if isCast(op) => args.forall(isSimpleCast(true))
+    case _ => false
+  }
+
+  /** Recursively replace [[WRef]]s with new [[Expression]]s
+    *
+    * @param replace a '''mutable''' HashMap mapping [[WRef]]s to values with which the [[WRef]]
+    * will be replaced. It is '''not''' mutated in this function
+    * @param expr the Expression being transformed
+    * @return Returns expr with [[WRef]]s replaced by values found in replace
+    */
+  def onExpr(replace: NodeMap)(expr: Expression): Expression = {
+    expr.map(onExpr(replace)) match {
+      case e @ WRef(name, _,_,_) =>
+        replace.get(name)
+               .filter(isSimpleCast(castSeen=false))
+               .getOrElse(e)
+      case e @ DoPrim(op, Seq(WRef(name, _,_,_)), _,_) if isCast(op) =>
+        replace.get(name)
+               .map(value => e.copy(args = Seq(value)))
+               .getOrElse(e)
+      case other => other // Not a candidate
+    }
+  }
+
+  /** Inline casts in a Statement
+    *
+    * @param netlist a '''mutable''' HashMap mapping references to [[firrtl.ir.DefNode DefNode]]s to their connected
+    * [[firrtl.ir.Expression Expression]]s. This function '''will''' mutate it if stmt is a [[firrtl.ir.DefNode
+    * DefNode]] with a value that is a cast [[PrimOp]]
+    * @param stmt the Statement being searched for nodes and transformed
+    * @return Returns stmt with casts inlined
+    */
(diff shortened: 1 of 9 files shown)


jackkoenig:
I'd also like to backport this to 1.2.x since it is a pretty nice lint improvement

src/main/scala/firrtl/transforms/InlineCasts.scala
```diff
@@ -0,0 +1,59 @@
...
+    * @return Returns expr with [[WRef]]s replaced by values found in replace
+    */
+  def onExpr(replace: NodeMap)(expr: Expression): Expression = {
+    expr.map(onExpr(replace)) match {
+      case e @ WRef(name, _,_,_) =>
+        replace.get(name)
+               .filter(isCast)
+               .getOrElse(e)
+      case e @ DoPrim(op, Seq(WRef(name, _,_,_)), _,_) if isCast(op) =>
+        replace.getOrElse(name, e)
```
azidar: Why are you removing casts in nested expressions?

azidar: Oh, after reading the PR text, you answered this question: " at the moment it just ignores all casts, assuming the circuit is in a valid form, and then injects the casts for the operators that need them. " I'm not certain that this transformation is always equivalent, but I'm open to being convinced otherwise.

jackkoenig: I think it was a benign mistake. Combined with the behavior of the VerilogEmitter, it's fine to remove the cast, but I didn't actually mean to remove the cast here (in case someone uses this transform in another context), so I"ll fix it.

jackkoenig: Fixed

jackkoenig:
There is a real failure, let's not block https://github.com/freechipsproject/firrtl/pull/1270 on this

jackkoenig:
I will want to rebase this before merging (or I guess we can just squash-and-merge), but I *think* it's now actually good to go

jackkoenig:
I want to highlight the addition of `LegalizeClocksTransform` which I found to be necessary due to https://github.com/freechipsproject/chisel3/blob/954cc41e1349d0df6d2250d6270590340cd36e82/src/test/scala/chiselTests/MultiClockSpec.scala#L182. While this is rare and a strange occurrence that we should optimize away, we need to handle it even when optimizations are disabled (note `MinimumVerilogCompiler` in the tests). It's really a workaround Verilator not supporting what I suspect is legal Verilog, but even lint tools are very unhappy about it so it's a good thing to do regardless.

jackkoenig:
Just the LEC tests on ICache timing out/running out of memory, will manually verify

jackkoenig:
This PR includes `[skip formal checks]`, but I ran them manually. They all pass using using the standard Yosys flow, the ICache just needed 10 GB (Travis provides 3-4 GB).

PR #1265, opened 2019-12-03
Remove Stage-global --no-dedup Option

seldridge:
### Checklist

### Type of Improvement

- bug fix                 

### API Impact

This removes the `--no-dedup` option from all stages which are not FIRRTL. Technically this was leaking in ChiselStage, but it should be safe to remove that. Users can always prevent deduplication with an annotation.

No impact.

### Desired Merge Strategy

- Squash

### DIFF
--- a/src/main/scala/firrtl/options/Shell.scala
+++ b/src/main/scala/firrtl/options/Shell.scala
@@ -3,7 +3,6 @@
 package firrtl.options
 
 import firrtl.AnnotationSeq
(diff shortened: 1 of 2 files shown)


seldridge:
@azidar (and maybe @jackkoenig): Do you have any thoughts on whether or not this is 1.3.0 or 1.2.x? I'm erring on the side of safety with 1.3.0, but this could likely get into 1.2.2 depending on how we define "API".

Issue #1263, opened 2019-12-03
combine bit-wise NOT onto the same line as bit-wise AND/OR


johningalls-sifive:
### Checklist

- [X] Did you write out a description of the feature you want to see?
- [X] Did you look around for any related features?
- [X] Did you specify relevant external information?

Type of issue: feature request | other enhancement
Impact: API modification
Development Phase: request | proposal
Please tell us about your environment: https://github.com/freechipsproject/firrtl/tree/f738fbe8667ed6b76ec00a15960b9c3a42b8654a

### Feature Description

It would be nice to reduce generated code size by reducing this circuit:

```
circuit NestNot :
  module NestNot :
    input a: UInt<1>
    input b: UInt<1>
    output c: UInt<1>
    output d: UInt<1>
    c <= and(a, not(b))
    d <= or(a, not(b))
```

from this emitted Verilog:
```
  wire _GEN_0;
  assign _GEN_0 = b == 1'h0;
  assign c = a & _GEN_0;
  assign d = a | _GEN_0;
```

to this emitted Verilog:
```
  assign c = a & ~b;
  assign d = a | ~b;
```

### Type of Feature
- backend code generation

### Related Features
https://github.com/freechipsproject/firrtl/pull/851 is a similar transform

### External Information
no

PR #1270, opened 2019-12-03
Verilog emitter transform InlineNots

johningalls-sifive:
### Checklist

### Type of Improvement
- backend code generation

### API Impact
none

Verilog example in https://github.com/freechipsproject/firrtl/issues/1263

### Desired Merge Strategy

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -310,15 +310,15 @@ class VerilogEmitter extends SeqTransform with Emitter {
      def c0: Int = doprim.consts.head.toInt
      def c1: Int = doprim.consts(1).toInt
 
-     def checkArgumentLegality(e: Expression) = e match {
+     def checkArgumentLegality(e: Expression): Unit = e match {
        case _: UIntLiteral | _: SIntLiteral | _: WRef | _: WSubField =>
-       case _ if isNot(e) =>
+       case DoPrim(Not, args, _,_) => args.foreach(checkArgumentLegality)
        case _ => throw EmitterException(s"Can't emit ${e.getClass.getName} as PrimOp argument")
      }
 
      def checkCatArgumentLegality(e: Expression): Unit = e match {
        case _: UIntLiteral | _: SIntLiteral | _: WRef | _: WSubField =>
-       case _ if isNot(e) =>
+       case DoPrim(Not, args, _,_) => args.foreach(checkArgumentLegality)
        case DoPrim(Cat, args, _, _) => args foreach(checkCatArgumentLegality)
        case _ => throw EmitterException(s"Can't emit ${e.getClass.getName} as PrimOp argument")
      }
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -237,7 +237,10 @@ class VerilogEmitter extends SeqTransform with Emitter {
         if (e.tpe == AsyncResetType) {
           throw EmitterException("Cannot emit async reset muxes directly")
         }
-        emit(Seq(e.cond," ? ",cast(e.tval)," : ",cast(e.fval)),top + 1)
+        e.cond match {
+          case DoPrim(Not, Seq(sel), _,_) => emit(Seq(sel," ? ",cast(e.fval)," : ",cast(e.tval)),top + 1)
+          case _ => emit(Seq(e.cond," ? ",cast(e.tval)," : ",cast(e.fval)),top + 1)
+        }
       }
       case (e: ValidIf) => emit(Seq(cast(e.value)),top + 1)
       case (e: WRef) => w write e.serialize
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -309,11 +309,13 @@ class VerilogEmitter extends SeqTransform with Emitter {
 
      def checkArgumentLegality(e: Expression) = e match {
        case _: UIntLiteral | _: SIntLiteral | _: WRef | _: WSubField =>
+       case _ if isNot(e) =>
        case _ => throw EmitterException(s"Can't emit ${e.getClass.getName} as PrimOp argument")
      }
 
      def checkCatArgumentLegality(e: Expression): Unit = e match {
        case _: UIntLiteral | _: SIntLiteral | _: WRef | _: WSubField =>
+       case _ if isNot(e) =>
        case DoPrim(Cat, args, _, _) => args foreach(checkCatArgumentLegality)
        case _ => throw EmitterException(s"Can't emit ${e.getClass.getName} as PrimOp argument")
      }
@@ -961,6 +963,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
   def transforms = Seq(
     new BlackBoxSourceHelper,
     new ReplaceTruncatingArithmetic,
+    new InlineNotsTransform,
     new FlattenRegUpdate,
     new DeadCodeElimination,
     passes.VerilogModulusCleanup,
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -378,7 +378,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
          case (_: UIntType) => Seq("{1'b0,", cast(a0), "}")
          case (_: SIntType) => Seq(cast(a0))
        }
-       case Not => Seq("~ ", a0)
+       case Not => Seq("~", a0)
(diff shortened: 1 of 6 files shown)


azidar:
@albert-magyar @jackkoenig this seems fine but yosys is taking too long on formal equivalence of the FPU. Should we merge regardless? Or is there a plan to improve the yosys scripts?

jackkoenig:
The general idea is someone should run it locally, verify it passes and state that here. Then we push a commit with `[skip formal checks]` and then CI will be green: https://github.com/freechipsproject/firrtl/blob/3243f05a69b4b77761699be412f349a9b8b9193f/.run_formal_checks.sh#L15

johningalls-sifive:
834827d check Formal Equivalence FPU passed with Formality, but Yosys still never finishes even when I run it myself

johningalls-sifive:
formal equivalence check results for FPU:
3a2a20d (0th) to 77b7640 (4th): Yosys **HANG**, Formality **Verification SUCCEEDED**
3a2a20d (0th) to 9a7b96c (1st): Yosys **pass**
9a7b96c (1st) to d5b3035 (2nd): Yosys **HANG**
d5b3035 (2nd) to 47859c7 (3rd): Yosys **pass**
47859c7 (3rd) to 77b7640 (4th): Yosys **pass**

johningalls-sifive:
output Verilog line count improvements:
- ICache: -28 (-3%) from 973 to 945
- FPU: -616 (-5%) from 12553 to 11937
- RocketCore: -198 (-2%) from 9375 to 9177

aswaterman:
Nice.

src/main/scala/firrtl/Emitter.scala
```diff
@@ -309,11 +312,13 @@ class VerilogEmitter extends SeqTransform with Emitter {
 
      def checkArgumentLegality(e: Expression) = e match {
        case _: UIntLiteral | _: SIntLiteral | _: WRef | _: WSubField =>
+       case _ if isNot(e) =>
```
jackkoenig: I know you got this from me on https://github.com/freechipsproject/firrtl/pull/1264, but I realized yesterday that it's not quite right. We need to recurse into the not to check that there aren't more complex expressions:
```suggestion
       case DoPrim(Not, args, _,_) => args.foreach(checkArgumentLegality)
```
Same idea below in `checkCatArgumentLegality` which we probably ought to find a way to unify with this one

src/main/scala/firrtl/Utils.scala
```diff
@@ -188,6 +188,17 @@ object Utils extends LazyLogging {
     case sx => sx
   }
 
+  /** Returns true if PrimOp is a Not, false otherwise */
+  def isNot(op: PrimOp): Boolean = op match {
+    case Not => true
+    case _ => false
+  }
```
jackkoenig: I would delete the `PrimOp` version because one could and should just write `op == Not`, the `Expression` version below makes sense though.

src/main/scala/firrtl/Utils.scala
```diff
@@ -188,6 +188,17 @@ object Utils extends LazyLogging {
...
+  /** Returns true if Expression is a Not PrimOp, false otherwise */
+  def isNot(expr: Expression): Boolean = expr match {
+    case DoPrim(op, _,_,_) if isNot(op) => true
```
jackkoenig: Related to the above
```suggestion
    case DoPrim(Not, _,_,_) => true
```

src/main/scala/firrtl/Utils.scala
jackkoenig: The only reason `isCast` applies to PrimOps is that there are several of them. I could solve it by introducing a trait but I decided against it.

src/main/scala/firrtl/transforms/InlineNots.scala
```diff
@@ -0,0 +1,68 @@
...
+    *
+    * @param netlist a '''mutable''' HashMap mapping references to [[firrtl.ir.DefNode DefNode]]s to their connected
+    * [[firrtl.ir.Expression Expression]]s. This function '''will''' mutate it if stmt is a [[firrtl.ir.DefNode
+    * DefNode]] with a value that is a [[PrimOp]] Not
+    * @param stmt the Statement being searched for nodes and transformed
+    * @return Returns stmt with nots inlined
+    */
+  def onStmt(netlist: Netlist)(stmt: Statement): Statement =
+    stmt.map(onStmt(netlist)).map(onExpr(netlist)) match {
+      case node @ DefNode(_, name, value) if name.head == '_' =>
```
jackkoenig:
```suggestion
      case node @ DefNode(_, name, value) if isTemp(name) =>
```
`isTemp` is defined in `firrtl.Utils`

Why is this only constructing a netlist from temporaries though?

src/main/scala/firrtl/transforms/InlineNots.scala
```diff
@@ -0,0 +1,68 @@
...
+  def onExpr(netlist: Netlist)(expr: Expression): Expression = {
+    expr.map(onExpr(netlist)) match {
+      case e @ WRef(name, _,_,_) =>
+        netlist.get(we(e))
+               .filter(isNot)
+               .getOrElse(e)
+      case lhs @ DoPrim(Not, Seq(inv), _,_) =>
+        netlist.getOrElse(we(inv), inv) match {
+          case DoPrim(Not, Seq(rhs), _,_) => rhs
+          case _ => lhs  // Not a candiate
```
jackkoenig: I'm fine with this here since I don't think there's a better place to put it, but these kind of nested primop optimizations should probably be part of ConstantPropagation or a new standard optimization transform.

src/main/scala/firrtl/transforms/InlineNots.scala
```diff
@@ -0,0 +1,68 @@
...
+    * @param netlist a '''mutable''' HashMap mapping references to [[firrtl.ir.DefNode DefNode]]s to their connected
+    * [[firrtl.ir.Expression Expression]]s. It is '''not''' mutated in this function
+    * @param expr the Expression being transformed
+    * @return Returns expr with Nots inlined
+    */
+  def onExpr(netlist: Netlist)(expr: Expression): Expression = {
+    expr.map(onExpr(netlist)) match {
+      case e @ WRef(name, _,_,_) =>
+        netlist.get(we(e))
+               .filter(isNot)
```
jackkoenig: I know you also got this from https://github.com/freechipsproject/firrtl/pull/1264, and I've convinced myself that the bug I hit there cannot manifest with this PR, there is the potential for subtle bugs in the future. As I fixed in https://github.com/freechipsproject/firrtl/pull/1264/commits/1ac62e0930d7b0cd5531ea27206b741fcab666ec, it's probably not sufficient to just filter on `isNot`. If we do further inlining, it's possible that this would cause inlining of complex Expressions that just so happen to have a `Not` wrapping them. I'd follow a similar approach as I did in the linked commit to save a future developer some debugging.

src/test/scala/firrtlTests/VerilogEmitterTests.scala
```diff
@@ -70,6 +70,62 @@ class DoPrimVerilog extends FirrtlFlatSpec {
...
+        |  output  e,
+        |  output  f,
+        |  output  g
+        |);
+        |  wire _GEN_2;
+        |  wire _GEN_4;
+        |  assign d = a & ~b;
+        |  assign e = a | ~b;
+        |  assign _GEN_2 = c[2];
+        |  assign _GEN_4 = _GEN_2;
```
jackkoenig: I'm not going to block this PR over this, but this is something that should be removed by constant propagation. My guess is that it's due to the `not(not(a))` improvement being done in `InlineNotsTransform` instead of `ConstantPropagation`. I'm just noting this, I don't think fixing it is in the scope of this PR.

jackkoenig:
Also a very nice improvement in line count 🙂

johningalls-sifive:
@jackkoenig all good, hanks for propagating the fixes from your other PR to this one too.  I'll do them here.

src/test/scala/firrtlTests/VerilogEmitterTests.scala
johningalls-sifive: Yes, I will fix this (actually already have this fixed) in a dependent PR that in-lines bit extractions.

src/main/scala/firrtl/transforms/InlineNots.scala
johningalls-sifive: > Why is this only inlining temporaries?

I choose to prioritize keeping descriptive signal names over inlining.

johningalls-sifive:
formal equivalence check results for FPU:
- 77b7640 (4th) to efbacce (5th): Yosys pass

PR #1269, opened 2019-12-03
Add DontSerialize / --write-dont-serialize

seldridge:
### Checklist

- [x] Did you specify the type of improvement?
- [x] Did you state the API impact?
- [x] Did you specify the code generation impact?
- [x] Did you request a desired merge strategy?

### Type of Improvement

This adds a `DontSerialize` annotation mix-in that is intended to indicate that an annotation should not be serialized to JSON. This fixes a bug where large annotations (e.g., `FirrtlCircuitAnnotation`) serialized to JSON would cause Java to die (see: #1272).

This behavior can be disabled with `--write-dont-serialize`/`WriteDontSerializeAnnotation`.

`FirrtlCircuitAnnotation`, `FirrtlSourceAnnotation`, and `EmittedAnnotation` now mix-in `DontSerialize`. 

This includes tests that serialization is enabled/disabled with `WriteDontSerializeAnnotation`.

- bug fix

- new feature/API

### API Impact

This adds the `--write-dont-serialize` API to all stages.

This changes the output annotation JSON file to strip `FirrtlCircuitAnnotation` and `FirrtlSourceAnnotation` which reverts to pre-3.2 behavior. **This is an API change if I pedantically define the output annotation JSON as an API.** However, I don't view the output annotation *JSON* as an API. I only view the internal `AnnotationSeq` as an API.

Because this fixes a bug that is hitting large designs (see #1272), I would like to get this into 1.2.2 even though it's pedantically API breaking for someone who is writing a downstream tool/stage that is consuming the annotation JSON (I expect that such users do not exist).

If the API addition is problematic for 1.2.2, I can both (1) remove the `--write-dont-serialize` option and (2) make the `DontSerialize` mixin `private[firrtl]`. If the change to the output annotation *JSON* is problematic, this can be retargeted for 1.3.0.

#### Broader Discussions

This is starting to hint at the following three lemmas:

> (1): An `AnnotationSeq` can be converted to `(annoJSON: File, other: Seq[File])`.

> (2): The information encoded by `annoJSON` and `other` are mutually exclusive.

> (3): The `AnnotationSeq` may contain extra information not included in the files.

What does this mean? 

[...]

PR #1267, opened 2019-12-03
Make WriteEmitted an Identity Phase

seldridge:
### Checklist

- [x] Did you specify the type of improvement?
- [x] Did you state the API impact?
- [x] Did you specify the code generation impact?
- [x] Did you request a desired merge strategy?

### Type of Improvement

The `WriteEmitted` phase, used to write any `EmitedAnnotation`s by the FIRRTL stage, currently deletes any annotations it writes. This changes that such that `WriteEmitted` will preserve the annotation sequence.

Previously, `WriteEmitted` was idempotent in terms of side effects. After this PR, it will repeatedly write the same `EmittedAnnotation` to a file every time its written.

- bug fix
- code cleanup
- backend code generation

Fixes #1261 

### API Impact

This changes the output annotation sequence. Any downstream phases that are expecting to either not see an `EmittedAnnotation` or to see a `DeletedAnnotation` (which arguably is not a supported API) will now see an `EmittedAnnotation`. This has potential API implications for testers, treadle, and firrtl-interpreter if they are looking for a `DeletedAnnotation`.

### Backend Code Generation Impact

This does not affect Verilog, but it does affect the output annotation sequence.

### Desired Merge Strategy

- Squash: The PR will be squashed and merged (choose this if you have no preference.

PR #1277, opened 2019-12-10
File Serialization of Annotations

### LINKED ISSUE #791: Directory for Transform metadata?
* **Type of issue**
  - [ ] Bug report
  - [x] Feature request
  - [ ] Other enhancement

Many transforms output metadata about what actions they performed, such as lists of files produced or Modules modified. Most pass authors' habits seem to be to dump any metadata files in the same directory as the one targeted by their pass, but I'm not sure this is ideal. For example, `BlackBoxSourceHelper` allows you to specify a particular target directory, but puts both the resource `.v` files there along with a manifest `.f`. This requires some downstream parsing to sort out.

Obviously, individual pass authors could allow two separate locations to be specified (for data and metadata), but I wonder whether the annotations/transform API shouldn't provide some sensible default or standardization? I think 99% of users want one directory for the .v files that are actually included in a verilog compilation step, and another directory for storing information about how and why those files came to exist.

Relatedly, I'd like to see more metadata produced from the built-in transforms (e.g. all verilog file names).

* **Impact**
  - [ ] no functional change
  - [x] API addition (no impact on existing code)
  - [ ] API modification
  - [ ] unknown

* **Development Phase**
  - [x] request
  - [ ] proposal

* **Other information** 

https://github.com/freechipsproject/rocket-chip/issues/1346

### LINKED ISSUE #1272: Prune/Don't Emit Output Annotation Sequence
### Checklist

There are reported issues related to `Stage`s always serializing their output annotation sequence. Either there needs to be a way to turn this off or too many or too many large annotations are being serialized. It may also be worth not serializing annotations, unless the user requests it. This is really a method of checkpoint-restart or an interface between stages that are communicating via files.

The result of this is out-of-memory errors in the JVM during annotation serialization.

This may be as simple as using a `DontSerialize` annotation that is attached to large things which are serialized in other ways (e.g., circuits).

FIRRTL shouldn't get out-of-memory errors for normal circuits (Rocket Chip sized?).

### Steps to Reproduce

Unclear at this time. @oharboe may be able to help.

### Your environment

- Chisel 3.2.0
- FIRRTL 1.2.0.

### External Information

https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/chisel-users/qbx_G58dCEY/eJN-IfOzAwAJ

seldridge:
This adds an annotation mix-in, `CustomFileEmission`, that lets an annotation be serialized to a file as opposed to included in the output annotation JSON. An annotation then controls:

1. How it should be serialized to a string (method `toBytes`)
1. Which file to put this string in (methods `baseFileName`, `suffix`, `filename`)
1. How to replace the file-serialized annotation with some *other* annotations that let downstream stages know about the serialized file (method `replacements`)

The stage-global phase `WriteOutputAnnotations` is modified to handle `CustomFileEmission` annotations such that it writes their contents to the requested file and replaces them with their `replacements`. This means that the FIRRTL Stage-specific Phase `WriteEmitted` for writing emitted annotations is no longer necessary. This is thereby deprecated. A consequence of this is that emitted annotations are no longer deleted. This simplifies the reconstruction of the deprecated `FirrtlExecutionResult` because the deleted emitted annotation no longer has to be examined.

This migrates `EmittedAnnotation` and its children, e.g., `EmittedFirrtlCircuitAnnotation` to use this new file serialization approach. Other candidates for changing include blackbox helpers, memlib, and memory initialization.

This includes the following minor updates:

- `WriteOutputAnnotationsSpec` is updated to check the above logic of `CustomFileEmission`
- `VerilogExecution` is modified to use the `execute` method (which includes Stage-global Phases) as opposed to the `run` method (which does not). This is necessary so that `WriteOutputAnnotations` runs.

I would propose that this fixes #791, but would like @hcook feedback on this approach. As a fix for #791, any transform that wants to write a file needs to emit an annotation that mixes in `CustomFileEmission`. The default Phases common to every Stage take care of the rest.

- Fixes #1272. 
- (Supersedes) Closes #1269.
- (Supersedes) Closes #1267 

### Checklist

#### TODO

- [x] Companion Chisel PRs: 
  - https://github.com/freechipsproject/chisel3/pull/1273
  - https://github.com/freechipsproject/chisel3/pull/1405
- [x] Switch to generate a `Stream[_]` instead of a `String`
- [x] Add behavior for ensuring file creation/uniqueness, appending, or auto-generating a new filename (see `ElaborationArtefacts`)

### Type of Improvement

- code refactoring

- new feature/API

### API Impact

Adds the `CustomFileEmission` trait and changes the way file serialization works for `EmittedAnnotation`s. Removes `WriteEmitted`.

No Verilog/FIRRTL generation changes.

### Desired Merge Strategy

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -14,7 +14,7 @@ import firrtl.PrimOps._
 import firrtl.WrappedExpression._
 import Utils._
 import MemPortUtils.{memPortField, memType}
-import firrtl.options.{HasShellOptions, HowToSerialize, ShellOption, StageOptions, PhaseException}
+import firrtl.options.{HasShellOptions, CustomFileEmission, ShellOption, StageOptions, PhaseException}
 import firrtl.options.Viewer.view
 import firrtl.stage.{FirrtlFileAnnotation, FirrtlOptions, RunFirrtlTransformAnnotation, TransformManager}
 // Datastructures
@@ -95,7 +95,7 @@ final case class EmittedFirrtlModule(name: String, value: String, outputSuffix: 
 final case class EmittedVerilogModule(name: String, value: String, outputSuffix: String) extends EmittedModule
 
 /** Traits for Annotations containing emitted components */
-sealed trait EmittedAnnotation[T <: EmittedComponent] extends NoTargetAnnotation with HowToSerialize {
+sealed trait EmittedAnnotation[T <: EmittedComponent] extends NoTargetAnnotation with CustomFileEmission {
   val value: T
 
   override protected val baseFileName: String = value.name
@@ -105,7 +105,7 @@ sealed trait EmittedAnnotation[T <: EmittedComponent] extends NoTargetAnnotation
 }
 sealed trait EmittedCircuitAnnotation[T <: EmittedCircuit] extends EmittedAnnotation[T] {
 
-  override def howToSerialize = Some(new StringBuilder(value.value).map(_.toByte))
+  override def toBytes = Some(new StringBuilder(value.value).map(_.toByte))
 
   override def filename(annotations: AnnotationSeq): File = {
     val sopts = view[StageOptions](annotations)
@@ -116,22 +116,22 @@ sealed trait EmittedCircuitAnnotation[T <: EmittedCircuit] extends EmittedAnnota
 }
 sealed trait EmittedModuleAnnotation[T <: EmittedModule] extends EmittedAnnotation[T] {
 
-  override def howToSerialize = Some(new StringBuilder(value.value).map(_.toByte))
+  override def toBytes = Some(new StringBuilder(value.value).map(_.toByte))
 
-  override def howToResume(file: File): Option[AnnotationSeq] = None
+  override def replacements(file: File): Option[AnnotationSeq] = None
 
 }
 
 case class EmittedFirrtlCircuitAnnotation(value: EmittedFirrtlCircuit)
     extends EmittedCircuitAnnotation[EmittedFirrtlCircuit] {
 
-  override def howToResume(file: File): Option[AnnotationSeq] = Some(Seq(FirrtlFileAnnotation(file.toString)))
+  override def replacements(file: File): Option[AnnotationSeq] = Some(Seq(FirrtlFileAnnotation(file.toString)))
 
 }
 case class EmittedVerilogCircuitAnnotation(value: EmittedVerilogCircuit)
     extends EmittedCircuitAnnotation[EmittedVerilogCircuit] {
 
-  override def howToResume(file: File): Option[AnnotationSeq] = None
+  override def replacements(file: File): Option[AnnotationSeq] = None
 
 }
 case class EmittedFirrtlModuleAnnotation(value: EmittedFirrtlModule)
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -107,12 +107,12 @@ sealed trait EmittedAnnotation[T <: EmittedComponent] extends NoTargetAnnotation
 }
 sealed trait EmittedCircuitAnnotation[T <: EmittedCircuit] extends EmittedAnnotation[T] {
 
-  override def toBytes = Some(new StringBuilder(value.value).map(_.toByte))
+  override def toBytes = Some(value.value.getBytes)
 
 }
 sealed trait EmittedModuleAnnotation[T <: EmittedModule] extends EmittedAnnotation[T] {
 
-  override def toBytes = Some(new StringBuilder(value.value).map(_.toByte))
+  override def toBytes = Some(value.value.getBytes)
 
 }
 
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -14,7 +14,7 @@ import firrtl.PrimOps._
 import firrtl.WrappedExpression._
 import Utils._
 import MemPortUtils.{memPortField, memType}
-import firrtl.options.{HasShellOptions, CustomFileEmission, ShellOption, StageOptions, PhaseException}
+import firrtl.options.{HasShellOptions, CustomFileEmission, ShellOption, PhaseException}
 import firrtl.options.Viewer.view
 import firrtl.stage.{FirrtlFileAnnotation, FirrtlOptions, RunFirrtlTransformAnnotation, TransformManager}
 // Datastructures
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -105,7 +105,7 @@ sealed trait EmittedAnnotation[T <: EmittedComponent] extends NoTargetAnnotation
 }
 sealed trait EmittedCircuitAnnotation[T <: EmittedCircuit] extends EmittedAnnotation[T] {
 
-  override def howToSerialize: Option[String] = Some(value.value)
+  override def howToSerialize: Option[Stream[Char]] = Some(new StringBuilder(value.value).toStream)
 
   override def filename(annotations: AnnotationSeq): File = {
     val sopts = view[StageOptions](annotations)
@@ -116,7 +116,7 @@ sealed trait EmittedCircuitAnnotation[T <: EmittedCircuit] extends EmittedAnnota
 }
 sealed trait EmittedModuleAnnotation[T <: EmittedModule] extends EmittedAnnotation[T] {
 
-  override def howToSerialize: Option[String] = Some(value.value)
+  override def howToSerialize: Option[Stream[Char]] = Some(new StringBuilder(value.value).toStream)
 
   override def howToResume(file: File): Option[AnnotationSeq] = None
 
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -118,22 +118,16 @@ sealed trait EmittedModuleAnnotation[T <: EmittedModule] extends EmittedAnnotati
 
   override def toBytes = Some(new StringBuilder(value.value).map(_.toByte))
 
-  override def replacements(file: File): Option[AnnotationSeq] = None
-
 }
 
 case class EmittedFirrtlCircuitAnnotation(value: EmittedFirrtlCircuit)
     extends EmittedCircuitAnnotation[EmittedFirrtlCircuit] {
 
-  override def replacements(file: File): Option[AnnotationSeq] = Some(Seq(FirrtlFileAnnotation(file.toString)))
+  override def replacements(file: File): AnnotationSeq = Seq(FirrtlFileAnnotation(file.toString))
 
 }
 case class EmittedVerilogCircuitAnnotation(value: EmittedVerilogCircuit)
-    extends EmittedCircuitAnnotation[EmittedVerilogCircuit] {
-
-  override def replacements(file: File): Option[AnnotationSeq] = None
-
-}
+    extends EmittedCircuitAnnotation[EmittedVerilogCircuit]
 case class EmittedFirrtlModuleAnnotation(value: EmittedFirrtlModule)
     extends EmittedModuleAnnotation[EmittedFirrtlModule]
 case class EmittedVerilogModuleAnnotation(value: EmittedVerilogModule)
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -105,7 +105,7 @@ sealed trait EmittedAnnotation[T <: EmittedComponent] extends NoTargetAnnotation
 }
 sealed trait EmittedCircuitAnnotation[T <: EmittedCircuit] extends EmittedAnnotation[T] {
 
-  override def howToSerialize: Option[Stream[Byte]] = Some(new StringBuilder(value.value).toStream.map(_.toByte))
+  override def howToSerialize = Some(new StringBuilder(value.value).map(_.toByte))
 
   override def filename(annotations: AnnotationSeq): File = {
     val sopts = view[StageOptions](annotations)
@@ -116,7 +116,7 @@ sealed trait EmittedCircuitAnnotation[T <: EmittedCircuit] extends EmittedAnnota
 }
 sealed trait EmittedModuleAnnotation[T <: EmittedModule] extends EmittedAnnotation[T] {
 
-  override def howToSerialize: Option[Stream[Byte]] = Some(new StringBuilder(value.value).toStream.map(_.toByte))
+  override def howToSerialize = Some(new StringBuilder(value.value).map(_.toByte))
 
   override def howToResume(file: File): Option[AnnotationSeq] = None
 
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -105,7 +105,7 @@ sealed trait EmittedAnnotation[T <: EmittedComponent] extends NoTargetAnnotation
 }
 sealed trait EmittedCircuitAnnotation[T <: EmittedCircuit] extends EmittedAnnotation[T] {
 
-  override def howToSerialize: Option[Stream[Char]] = Some(new StringBuilder(value.value).toStream)
+  override def howToSerialize: Option[Stream[Byte]] = Some(new StringBuilder(value.value).toStream.map(_.toByte))
 
   override def filename(annotations: AnnotationSeq): File = {
     val sopts = view[StageOptions](annotations)
@@ -116,7 +116,7 @@ sealed trait EmittedCircuitAnnotation[T <: EmittedCircuit] extends EmittedAnnota
 }
 sealed trait EmittedModuleAnnotation[T <: EmittedModule] extends EmittedAnnotation[T] {
 
-  override def howToSerialize: Option[Stream[Char]] = Some(new StringBuilder(value.value).toStream)
+  override def howToSerialize: Option[Stream[Byte]] = Some(new StringBuilder(value.value).toStream.map(_.toByte))
 
   override def howToResume(file: File): Option[AnnotationSeq] = None
 
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -107,12 +107,12 @@ sealed trait EmittedAnnotation[T <: EmittedComponent] extends NoTargetAnnotation
 }
 sealed trait EmittedCircuitAnnotation[T <: EmittedCircuit] extends EmittedAnnotation[T] {
 
-  override def toBytes = Some(value.value.getBytes)
+  override def getBytes = value.value.getBytes
 
 }
 sealed trait EmittedModuleAnnotation[T <: EmittedModule] extends EmittedAnnotation[T] {
 
-  override def toBytes = Some(value.value.getBytes)
+  override def getBytes = value.value.getBytes
 
 }
 
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -14,10 +14,13 @@ import firrtl.PrimOps._
 import firrtl.WrappedExpression._
 import Utils._
 import MemPortUtils.{memPortField, memType}
-import firrtl.options.{HasShellOptions, PhaseException, ShellOption, Unserializable}
-import firrtl.stage.{RunFirrtlTransformAnnotation, TransformManager}
+import firrtl.options.{HasShellOptions, HowToSerialize, ShellOption, StageOptions, PhaseException}
+import firrtl.options.Viewer.view
+import firrtl.stage.{FirrtlFileAnnotation, FirrtlOptions, RunFirrtlTransformAnnotation, TransformManager}
 // Datastructures
 import scala.collection.mutable.ArrayBuffer
+
+import java.io.File
 
 case class EmitterException(message: String) extends PassException(message)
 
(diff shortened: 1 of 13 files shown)


seldridge:
Feedback from @jackkoenig in the last dev meeting was that this would probably be better to serialize to a `Stream`. I think this means, this should have a `toStream(): Stream[Char]` method implemented. @jackkoenig: does that seem reasonable?

Also, what this PR proposes should be sufficient to replace Rocket Chip's `ElaborationArtefacts`. E.g., instead of emitting "something else" that's serialize to a file, you emit an annotation that defines how it should be serialized and what file it should be serialized to. Notably `ElaborationArtefacts` also has the ability to check if the file that someone is trying to generate would already be written to by someone else. The thing generating the artefact can then spin to generate a new name. This motivates having the ability to force uniqueness of generated files.

jackkoenig:
> I think this means, this should have a toStream(): Stream[Char] method implemented. @jackkoenig: does that seem reasonable?

That's the right idea, although I'm not certain if `Stream[Char]` is the right type. It may work perfectly but because `Stream` (which is deprecated in Scala 2.13 in favor of `LazyList`) is effectively a List, it may incur a large performance penalty per element, in this case, `Char`. It might be that a `Stream[String]` is more appropriate.

I have to admit that when I originally said "stream" I was thinking about like Java `OutputStream`, but that's obviously mutable so not the most desirable API. I'm now wondering how this kind of thing is accomplished in the Scala functional world (ZIO, Scalaz, Cats, etc.). Is there a standard approach?

Before we finalize the API, I think we should see if there's an obvious approach out there, and then do some benchmarking. Assuming `Stream`* is the right approach, we should compare `Stream[Char]` and `Stream[String]`. While my intuition is that `Stream[Char]` has a large overhead, I can also see the JIT making that overhead basically zero so it might actually be even better in practice than `Stream[String]`.

*Unfortunately it looks like `LazyList` is missing from the [compatibility collections library](https://github.com/scala/scala-collection-compat))

seldridge:
From the dev meeting discussion:

- The underlying serialization strategy should be `Stream[Byte]`
- Utility methods/mix-ins should exist for making this more friendly, e.g., a serialization defined to `String]`
- `twitter.chill.meatlocker` may provide an easy way to serialize objects

davidbiancolin:
I'm totally bike shedding here but i think it would be good if `HowToSerialize`'s name could better indicate it is used for serializing the annotations independently from the main FAF since there there are annotation mixins which apply to the JsonProtocol which sound similar. Like `IndependentlySerializable` or  `SerializeToDistinctFile` or something.

src/main/scala/firrtl/options/StageAnnotations.scala
```diff
@@ -14,6 +17,23 @@ sealed trait StageOption { this: Annotation => }
   */
 trait Unserializable { this: Annotation => }
 
+trait HowToSerialize { this: Annotation =>
```
ekiwi-sifive, line 20: this needs scaladoc

src/main/scala/firrtl/options/StageAnnotations.scala
```diff
@@ -14,6 +17,23 @@ sealed trait StageOption { this: Annotation => }
...
+
+  protected def baseFileName: String
+
+  protected def suffix: Option[String]
+
+  def howToSerialize: Option[Stream[Byte]]
```
ekiwi-sifive, line 26: scaladoc missing

src/main/scala/firrtl/options/StageAnnotations.scala
```diff
@@ -14,6 +17,23 @@ sealed trait StageOption { this: Annotation => }
...
+
+  def howToResume(file: File): Option[AnnotationSeq]
```
ekiwi-sifive, line 28: also missing scaladoc

Personally I have no idea what `howToResume` is supposed to do

src/main/scala/firrtl/options/phases/WriteOutputAnnotations.scala
```diff
@@ -27,10 +26,21 @@ class WriteOutputAnnotations extends Phase {
...
   def transform(annotations: AnnotationSeq): AnnotationSeq = {
     val sopts = Viewer[StageOptions].view(annotations)
-    val serializable = annotations.filter{
-      case _: Unserializable    => false
-      case _: DeletedAnnotation => sopts.writeDeleted
-      case _                    => true
+    val serializable: AnnotationSeq = annotations.flatMap{
+      case _: Unserializable    => None
+      case a: DeletedAnnotation => if (sopts.writeDeleted) { Some(a) } else { None }
+      case a: HowToSerialize    =>
```
ekiwi-sifive, line 32: I think especially in this context the name `SerializeToFile` would make more sense since other annotations also get serialized, but not to a custom file.

ekiwi-sifive:
@seldridge , this also needs to be rebased and made to work with Scala `2.13`

src/main/scala/firrtl/options/StageAnnotations.scala
jackkoenig: In dev we talked about renaming this, "emission" seemed to be a good word.

What about `CustomEmission` so that you have
```
case class MyAnnotation(...) extends Annotation with CustomEmission
```
Or `CustomFileEmission` or `FileEmitable`

ekiwi-sifive: I like `CustomFileEmission` for its descriptiveness, but I think `FileEmitable` wins my "vote" since it is so much shorter!

src/main/scala/firrtl/options/StageAnnotations.scala
jackkoenig: Testing, can I do a suggestion in response to another comment?
```suggestion
  /** Implementation of custom file emission
    *
    * @example {{{
    *  // In simple cases with just a String, use .getBytes
    *  def howToSerialize: Option[Iterable[Byte]] = Some(myString.getBytes)
    * }}}
    */
  def howToSerialize: Option[Iterable[Byte]]
```
Also, why is this an `Option`? I feel like there's a reason but I don't remember.

ekiwi-sifive: Should the `getBytes` take an encoding like `UTF-8`?

jackkoenig: `getBytes` has a form that accepts a `Charset`: https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#getBytes-java.nio.charset.Charset-

Not sure if that's something we need to provide or if deferring to the platform is usually sufficient? I guess it depends on the data.

src/main/scala/firrtl/options/StageAnnotations.scala
seldridge: Scaladoc added. Names not yet changed.

src/main/scala/firrtl/options/StageAnnotations.scala
seldridge: Scaladoc added, switched to `Iterable`. I've also included Jack's example.

I can't recall why I made this an `Option`... You could use to abuse the emission. E.g., you set `howToSerialize` to `None` and then use a non-empty `howToResume` to serialize this to some other, safer, annotation. Not sure if that makes sense. (And to reiterate, I cannot recall the original use case.)

src/main/scala/firrtl/options/StageAnnotations.scala
seldridge: Scaladoc added. 

For the name, I'm most in favor of maximizing descriptiveness (at the cost of characters). `CustomFileEmission` is pretty clear. I'll think a bit more about this.

src/main/scala/firrtl/options/StageAnnotations.scala
```diff
@@ -14,6 +17,61 @@ sealed trait StageOption { this: Annotation => }
...
+  def toBytes: Option[Iterable[Byte]]
+
+  /** Optionally, a sequence of annotations that will replace this annotation in the output annotation file.
+    *
+    * A non-None implementation of this method is a mechanism for telling a downstream [[firrtl.options.Stage Stage]]
+    * how to deserialize the information that was serialized to a separate file. For example, if a FIRRTL circuit is
+    * serialized to a separate file, this method could include an input file annotation that a later stage can use to
+    * read the serialized FIRRTL circuit back in.
+    */
+  def replacements(file: File): Option[AnnotationSeq]
```
ekiwi-sifive, line 57: I wonder if we could make this a separate trait, maybe one that extends from the `CustomFileEmission` trait.
Or maybe we could have a default implementation that returns `None`?

seldridge: Default implementation seems good to me. One-off traits to provide partial implementations has seemed attractive before (e.g., `PreservesAll[_]`), but after the headaches with `PreservesAll`, I'm not sure this is actually a good approach.

I'm fixing this up in the rebase. It also occurred to me that `Option[AnnotationSeq]` is weird. I'm trying to do two things here: (1) express the property that there may be no replacements and (2) it's nice to have the replacements in a monad as this lets me concisely `flatMap` over the annotation seq and produce zero or more replacement annotations. `AnnotationSeq` is *almost* a monad (if you call the `toSeq` method on it), so I think it makes sense to just switch the type to `AnnotationSeq`.

Does changing the return type seem reasonable?

src/main/scala/firrtl/options/StageAnnotations.scala
```diff
@@ -14,6 +17,61 @@ sealed trait StageOption { this: Annotation => }
...
+    * by overriding the baseFileName and suffix methods. However, if the filename cannot be statically known and is a
+    * function of the content being serialized, then users may need to override this, e.g., if the filename should
+    * include the top module of a FIRRTL circuit.
+    *
+    * @param annotations the annotations at the time of serialization
+    */
+  def filename(annotations: AnnotationSeq): File = {
+    val name = view[StageOptions](annotations).getBuildFileName(baseFileName, suffix)
+    new File(name)
+  }
```
seldridge, line 71: This API is actually a bit weird. Method `filename` depends on the abstract methods `baseFileName` and `suffix`. The user is entirely free to override `filename` and not use the abstract methods. However, they still have to implement those two abstract methods.

Maybe this would be better expressed as the following?

```scala
def filename(annotations: AnnotationSeq, baseFileName: String, suffix: Option[String]): File
```

seldridge: I thought about this some more. I really do not want users to override this method because it guarantees that files are written in the correct location as defined by `StageOptions.getBuildFileName`. 

What's really going on here is that the `baseFileName` should be a function of the full annotation sequence to enable a `FirrtlCircuitAnnotation` to be written to a location controlled by an optional `OutputFileAnnotation`. I've updated this in a subsequent commit.

src/main/scala/firrtl/options/StageAnnotations.scala
```diff
@@ -14,6 +17,63 @@ sealed trait StageOption { this: Annotation => }
...
+  /** A method that can convert this annotation to bytes that will be written to a file.
+    *
+    * If you only need to serialize a string, you can use the `getBytes` method:
+    * {{{
+    *  def toBytes: Option[Iterable[Byte]] = Some(myString.getBytes)
+    * }}}
+    */
+  def toBytes: Option[Iterable[Byte]]
+
+  /** Optionally, a sequence of annotations that will replace this annotation in the output annotation file.
```
ekiwi-sifive, line 57: "that will replace this annotation **after serialization**" ?

seldridge: I actually think this is correct as written. This only replaces the annotation in the output annotation file. It's not replaced in the actual annotation sequence.

Put differently, the `WriteOutputAnnotations` phase is an identity function.

Maybe I'm misunderstanding?

ekiwi-sifive: I was misunderstanding what WriteOutputAnnotations was doing. Yeah, it seems like it it correct as is.

seldridge:
Thanks for the review @ekiwi-sifive.

This was missing some behavior for file collisions. The most recent commit sets the following behavior

1. No two annotations in one `WriteOutputAnnotations` transform may custom serialize to the same file
2. (Corollary:) The first annotation to custom serialize will overwrite the file if it exists

I think this is the sanest behavior. If you run the FIRRTL compiler, it should blow away old stuff, like the output circuit. However, if you have two annotations trying to write the same circuit, _at the same time_, that's an error. 

This also has the benefit of not breaking multiple runs of `sbt test` where tests may overwrite files.

src/main/scala/firrtl/options/StageAnnotations.scala
```diff
@@ -14,6 +17,63 @@ sealed trait StageOption { this: Annotation => }
...
+
+  /** Output filename where serialized content will be written
+    *
+    * The full annotation sequence is a parameter to allow for the location where this annotation will be serialized to
+    * be a function of other annotations, e.g., if the location where information is written is controlled by a separate
+    * file location annotation.
+    *
+    * @param annotations the annotation sequence at the time of emission
+    */
+  protected def baseFileName(annotations: AnnotationSeq): String
```
jackkoenig, line 43: Just a thought, annotations that would serialize to the same filename could use this to disambiguate themselves. That is, they could see that others of the type exist and provide a disambiguation algorithm.

src/main/scala/firrtl/stage/phases/WriteEmitted.scala
```diff
@@ -24,6 +24,8 @@ import java.io.PrintWriter
   *
   * Any annotations written to files will be deleted.
   */
+@deprecated("Annotations that mixin the HowToSerialize trait are automatically serialized by stages." +
```
jackkoenig, line 27:
```suggestion
@deprecated("Annotations that mixin the CustomFileEmission trait are automatically serialized by stages." +
```

src/test/scala/firrtlTests/options/phases/WriteOutputAnnotationsSpec.scala
```diff
@@ -100,9 +109,59 @@ class WriteOutputAnnotationsSpec extends AnyFlatSpec with Matchers with firrtl.t
...
+
   case class BarAnnotation(x: Int) extends NoTargetAnnotation
+
+  case class Custom(value: String) extends NoTargetAnnotation with CustomFileEmission {
+
+    override protected def baseFileName(a: AnnotationSeq): String = "Custom"
+
+    override protected def suffix: Option[String] = Some(".Emission")
+
+    override def toBytes: Option[Stream[Byte]] = Some(new StringBuilder(value).toStream.map(_.toByte))
```
jackkoenig, line 159:
```suggestion
    override def toBytes: Option[Iterable[Byte]] = Some(value.toBytes)
```

src/main/scala/firrtl/options/StageAnnotations.scala
```diff
@@ -14,6 +17,63 @@ sealed trait StageOption { this: Annotation => }
...
+  def toBytes: Option[Iterable[Byte]]
+
+  /** Optionally, a sequence of annotations that will replace this annotation in the output annotation file.
+    *
+    * A non-empty implementation of this method is a mechanism for telling a downstream [[firrtl.options.Stage Stage]]
+    * how to deserialize the information that was serialized to a separate file. For example, if a FIRRTL circuit is
+    * serialized to a separate file, this method could include an input file annotation that a later stage can use to
+    * read the serialized FIRRTL circuit back in.
+    */
+  def replacements(file: File): AnnotationSeq = Seq.empty
```
jackkoenig, line 64: Is the API that the file has been written? I think there is potential here for some awkward interactions.

* Is it considered acceptable practice for implementers of this method to read the file?
* As implemented, this gives the canonical file. Canonical files are absolute so if people use this argument they lose the ability to make the information relocatable

I'm really not certain what the API here should be, but I'm thinking it should be possible to get a relative path, and maybe just have it be a `String` instead of a `File`. Not sure though.

seldridge: This is very tightly coupled to the implementation of `WriteOutputAnnotations`.

> Is the API that the file has been written? I think there is potential here for some awkward interactions.

Yes. At this point, the custom file has been written. The output annotation file has not. (We can change this to not write the file yet, if needed---see next paragraph.)

> - Is it considered acceptable practice for implementers of this method to read the file?

No... it's not generally acceptable. Thought, it is allowable. Two thoughts: (1) the file could be rewritten after this so that the transform doesn't have access to the file contents and (2) we can use a custom security manager that disallows file writes when we call `replacements` (this is extremely draconian, though). I think probably (1) makes sense?

> - As implemented, this gives the canonical file. Canonical files are absolute so if people use this argument they lose the ability to make the information relocatable

Yeah, this won't work if you elaborate on one machine, then run FIRRTL compilation on another. However, I think this is actually an issue of `StageOptions.getBuildFileName` which is returning a canonical path as opposed to just a path. Let me see if changing that produces the correct behavior.

seldridge: The change to relative paths is actually one line... Fixed here: https://github.com/freechipsproject/firrtl/pull/1277/commits/5637e3f8c5221312e0cd9119ea75528aad390e87.

jackkoenig: I'm a little worried that changing that utility might have downstream affects on dependent projects. In particular I think @chick should be aware since (IIRC) he wrote the original code.

I suspect that some of the original intent of the canonicalization there was to do normalization (eg. turn `foo/../bar` into just `bar`. We can do that without constructing an absolute path by doing: `.toPath.normalize.toFile`.

seldridge: Good point. I've been locally building the world with this, but it may have unexpected consequences. 

Nice idea about `.toPath.normalize.toFile`. I've added a co-authored commit with this change.

src/test/scala/firrtlTests/options/phases/WriteOutputAnnotationsSpec.scala
seldridge: Way simpler, thanks! Fixed in a later commit (and also fixed a few other places where I was doing an overly complex conversion to a string).

(Though, I think the method is `getBytes` not `toBytes`?)

Bikeshedding: Do you think that the `CustomFileEmission` method name should be something different? Apparently `getBytes` is the standard. However, that doesn't return an option. Maybe `getBytesOption` or `getOptionBytes`?

src/main/scala/firrtl/stage/phases/WriteEmitted.scala
seldridge: Fixed.

src/test/scala/firrtlTests/options/phases/WriteOutputAnnotationsSpec.scala
jackkoenig: It is indeed `getBytes`

Yeah I think maybe changing the name makes sense. I will note that it would be super helpful if we could remember _why_ we made it an `Option`. I guess one argument is that it lets you distinguish between not wanting to write the file and wanting to write an empty file? We should write that in the ScalaDoc.

seldridge: There's no mention of the reason/logic in the dev meeting discussions (2019 or 2020) nor in the commit message. I will note that *I made this an option* so it may have just been developer error on my part.

The only reason I can think of is that this allows use of `replacements` without outputting a file. Note that a `toBytes = None` will result in no file written, currently. A `toBytes = Some(Array.empty)` should be an empty file (I didn't test this).

In the spirit of keeping this moving, how about we switch to `toBytes: Iterable[Byte]`? 

If the hypothetical case of only using `replacements`, this could be an API expansion to `Unserializable`:

```scala
trait Unserializable { this: Annotation => 

  def replacements: AnnotationSeq = Seq.empty

}
```

I think that this also synergizes with your work on serializing annotations with `-ll trace` to JSON to ignore things that don't serialize. This is an approach that could subsume the try/catch logic there.

jackkoenig: Yeah that sounds neat. If we make `toBytes` an `Iterable[Byte]`, is/should there be a way to make an empty file vs. no file?

ekiwi-sifive: If you don't want the file, can you just not emit the annotation?

ekiwi-sifive:
@seldridge can you rebase again?

I want to get this in ASAP and there are conflicts with my recent PR that was merged: https://github.com/freechipsproject/firrtl/pull/1835

src/test/scala/firrtlTests/options/phases/WriteOutputAnnotationsSpec.scala
seldridge: Changed to `getBytes: Iterable[Byte]` in added commits.

seldridge:
@jackkoenig: This is ready for a re-review. I've updated the companion Chisel PR and it is passing locally for me.

src/main/scala/firrtl/Emitter.scala
```diff
@@ -92,17 +95,35 @@ final case class EmittedFirrtlModule(name: String, value: String, outputSuffix:
 final case class EmittedVerilogModule(name: String, value: String, outputSuffix: String) extends EmittedModule
 
 /** Traits for Annotations containing emitted components */
-sealed trait EmittedAnnotation[T <: EmittedComponent] extends NoTargetAnnotation with Unserializable {
+sealed trait EmittedAnnotation[T <: EmittedComponent] extends NoTargetAnnotation with CustomFileEmission {
```
jackkoenig, line 98: I don't know why, but I didn't notice that this was just using the API here. Pretty slick :)

PR #1284, opened 2019-12-18
Improve Scaladoc

seldridge:
Minor Scaladoc improvement.

(I'm cleaning up #1123 and pulling a few small things out of this.)

### Checklist

### Type of Improvement

- documentation

### API Impact

None.

None.

### Desired Merge Strategy

### DIFF
--- a/src/main/scala/firrtl/options/Phase.scala
+++ b/src/main/scala/firrtl/options/Phase.scala
@@ -73,9 +73,9 @@ trait DependencyAPI[A <: DependencyAPI[A]] { this: TransformLike[_] =>
   def dependents: Seq[Dependency] = Seq.empty
   private[options] lazy val _dependents: LinkedHashSet[Dependency] = new LinkedHashSet() ++ dependents.toSet
 
-  /** A function that, given a transform will return true if this transform invalidates/undos the effects of the input
-    * transform
-    * @note Can a [[firrtl.options.Phase Phase]] ever invalidate itself?
+  /** A function that, given *another* transform (parameter `a`) will return true if this transform invalidates/undos the
+    * effects of the *other* transform (parameter `a`).
+    * @param a transform
     */
   def invalidates(a: A): Boolean = true
 


src/main/scala/firrtl/options/Phase.scala
```diff
@@ -73,9 +73,9 @@ trait DependencyAPI[A <: DependencyAPI[A]] { this: TransformLike[_] =>
   def dependents: Seq[Dependency] = Seq.empty
   private[options] lazy val _dependents: LinkedHashSet[Dependency] = new LinkedHashSet() ++ dependents.toSet
 
-  /** A function that, given a transform will return true if this transform invalidates/undos the effects of the input
-    * transform
-    * @note Can a [[firrtl.options.Phase Phase]] ever invalidate itself?
+  /** A function that, given *another* transform will return true if this transform invalidates/undos the effects of that
```
chick: Seems like to be extra clear *another* should refer to parameter a at least I think that is what *another* means.

src/main/scala/firrtl/options/Phase.scala
```diff
@@ -73,9 +73,9 @@ trait DependencyAPI[A <: DependencyAPI[A]] { this: TransformLike[_] =>
...
+    * transform. This function may return any value if given this transform.
```
chick: by "this" do you mean this as in `this`. And why return any value. Shouldn't it return false, or maybe even throw an exception if given itself?

src/main/scala/firrtl/options/Phase.scala
seldridge: Nice catch. I'll fix this in a rebase.

src/main/scala/firrtl/options/Phase.scala
seldridge: Yeah, I mean `this`

There's a change in the `dependency-api-2` branch that makes things work like this. A `DependencyManager` is keeping track of all transforms that have run to determine if any relowering is needed. To do this, it is applying the invalidation function to the current state. However, the `dependency-api-2` branch applies the invalidation function *before* adding the transform that just ran to the current state. In pseudocode:

```diff
- val nextState = invalidate(currentState ++ thisState)
+ val nextState = invalidate(currentState) ++ thisState
```

Hence, whether this value is true or false when given `this` doesn't matter. It redefines the invalidation function to have no meaning if given `this` and results in a shorter method body.

I tried to think about this a lot and I came to the conclusion that invalidating yourself doesn't make any sense with statically scheduled idempotent transforms. If we were to allow dynamic scheduling (which we don't) some transform could theoretically invalidate itself and then require it to be rerun to clean itself up. We don't allow this, so this Scaladoc follows.

Regardless, this comment is reflecting a change that only exists in the dependency-api-2 branch is not separable as a separate PR. I'll remove this.

PR #1286, opened 2019-12-19
Repl seq mem renaming

jackkoenig:
This does 2 things (thus 2 commits)
1. Consume `NoDedupMemAnnotation`s in `ResolveMemoryReference`
2. Support renaming in `ReplaceMemMacros`

(1) is included because trying to rename mem instances (ComponentNames) to modules causes a runtime error, so we either have to change the `target` type of `NoDedupMemAnnotation`, or consume them. Consuming them makes more sense IMO since they've already been extracted; however, there is a reasonable argument that the type should be changed. This would be neither source, nor binary compatible, so I chose consumption. In the off chance someone is relying on them not being consumed, the targets of these annos no longer exist (the mems that were replaced) so I think it's pretty justifiable to call relying on it buggy behavior.

(2) Obvious bug, necessary for trying to track smems all the way through.

### Checklist

### Type of Improvement
- bug fix

### API Impact

- no effect

- no effect

### Desired Merge Strategy
(Although squashing isn't a big deal)


jackkoenig:
This currently fails binary compatibility check with:
```
error]  * method updateMemStmts(firrtl.Namespace,scala.collection.mutable.HashMap,java.lang.String,scala.collection.mutable.HashMap,scala.collection.mutable.ArrayBuffer,firrtl.ir.Statement)firrtl.ir.Statement in class firrtl.passes.memlib.ReplaceMemMacros does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("firrtl.passes.memlib.ReplaceMemMacros.updateMemStmts")
[error]  * method updateMemMods(firrtl.Namespace,scala.collection.mutable.HashMap,scala.collection.mutable.ArrayBuffer,firrtl.ir.DefModule)firrtl.ir.DefModule in class firrtl.passes.memlib.ReplaceMemMacros does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("firrtl.passes.memlib.ReplaceMemMacros.updateMemMods")
```

jackkoenig:
Okay I fixed the binary compatibility issues and force pushed. It turns out that adding an argument to a method with a default argument does **not** preserve binary compatibility. However, simply overloading the method does. So that's what it's doing now.

azidar:
LGTM; I'd like @seldridge 's review if possible as well prior to merging.

PR #1296, opened 2020-01-07
Verilog emitter transform InlineBitExtractions

johningalls-sifive:
### Checklist

### Type of Improvement
- backend code generation

### API Impact
none

The goal of this PR is to compact the generated Verilog to have fewer intermediate terms.
We accomplish this by
1. inlining bit extraction (similar to Not in #1270), and
2. replacing back-to-back bit extractions with a single bit extraction.

The term "bit extraction" above includes all combinations of the Firrtl ops Bits, Head, Tail, and Shr (shift-right).

For example, this Chisel code
`a := b[1:0] | (c[2:0] >> 1)`
which previously emitted something like this Verilog code
```
assign _GEN_1 = b[1:0];
assign _GEN_2 = c[2:0];
assign _GEN_3 = _GEN_2 >> 1;
assign a = _GEN_1 | _GEN_3;
```
will now emit this Verilog code
`assign a = b[1:0] | c[2:1];`

This reduces the output Verilog line count by between -2% and -20% depending on the logic:
- ICache: -2% from 945 to 919
- FPU: -20% from 11937 to 9459
- RocketCore: -10% from 9177 to 8199

### Desired Merge Strategy
- Squash: The PR will be squashed and merged (choose this if you have no preference).

### DIFF
--- a/src/main/scala/firrtl/transforms/InlineBitExtractions.scala
+++ b/src/main/scala/firrtl/transforms/InlineBitExtractions.scala
@@ -92,8 +92,8 @@ object InlineBitExtractionsTransform {
 
 /** Inline nodes that are simple bits */
 class InlineBitExtractionsTransform extends Transform {
-  def inputForm = LowForm
-  def outputForm = LowForm
+  def inputForm = UnknownForm
+  def outputForm = UnknownForm
 
   def execute(state: CircuitState): CircuitState = {
     val modulesx = state.circuit.modules.map(InlineBitExtractionsTransform.onMod(_))
--- a/src/main/scala/firrtl/transforms/InlineBitExtractions.scala
+++ b/src/main/scala/firrtl/transforms/InlineBitExtractions.scala
@@ -1,0 +1,102 @@
+package firrtl
+package transforms
+
+import firrtl.ir._
+import firrtl.Mappers._
+import firrtl.PrimOps.{Bits, Head, Tail, Shr}
+import firrtl.Utils.{isBitExtract, isTemp}
+import firrtl.WrappedExpression._
+
+import scala.collection.mutable
+
+object InlineBitExtractionsTransform {
+
+  // Checks if an Expression is made up of only Bits terminated by a Literal or Reference.
+  // private because it's not clear if this definition of "Simple Expression" would be useful elsewhere.
+  // Note that this can have false negatives but MUST NOT have false positives.
+  private def isSimpleExpr(expr: Expression): Boolean = expr match {
+    case _: WRef | _: Literal | _: WSubField => true
+    case DoPrim(op, args, _,_) if isBitExtract(op) => args.forall(isSimpleExpr)
+    case _ => false
+  }
+
+  // replace Head/Tail/Shr with Bits for easier back-to-back Bits Extractions
+  private def lowerToDoPrimOpBits(expr: Expression): Expression = expr match {
+    case DoPrim(Head, rhs, c, tpe) if isSimpleExpr(expr) && rhs.forall(isSimpleExpr) =>
+      val msb = bitWidth(rhs.head.tpe) - 1
+      val lsb = bitWidth(rhs.head.tpe) - c.head
+      DoPrim(Bits, rhs, Seq(msb,lsb), tpe)
+    case DoPrim(Tail, rhs, c, tpe) if isSimpleExpr(expr) && rhs.forall(isSimpleExpr) =>
+      val msb = bitWidth(rhs.head.tpe) - c.head - 1
+      DoPrim(Bits, rhs, Seq(msb,0), tpe)
+    case DoPrim(Shr, rhs, c, tpe) if isSimpleExpr(expr) && rhs.forall(isSimpleExpr) =>
+      DoPrim(Bits, rhs, Seq(bitWidth(rhs.head.tpe)-1, c.head), tpe)
+    case _ => expr // Not a candidate
+  }
+
+  /** Mapping from references to the [[firrtl.ir.Expression Expression]]s that drive them */
+  type Netlist = mutable.HashMap[WrappedExpression, Expression]
+
+  /** Recursively replace [[WRef]]s with new [[Expression]]s
+    *
+    * @param netlist a '''mutable''' HashMap mapping references to [[firrtl.ir.DefNode DefNode]]s to their connected
+    * [[firrtl.ir.Expression Expression]]s. It is '''not''' mutated in this function
+    * @param expr the Expression being transformed
+    * @return Returns expr with Bits inlined
+    */
+  def onExpr(netlist: Netlist)(expr: Expression): Expression = {
+    expr.map(onExpr(netlist)) match {
+      case e @ WRef(name, _,_,_) =>
+        netlist.get(we(e))
+               .filter(isBitExtract)
+               .getOrElse(e)
+      // replace back-to-back Bits Extractions
+      case lhs @ DoPrim(lop, ival, lc, ltpe) if isSimpleExpr(lhs) && ival.forall(isSimpleExpr) =>
+        netlist.getOrElse(we(ival.head), ival.head) match {
+          case of @ DoPrim(rop, rhs, rc, rtpe) if isSimpleExpr(of) && rhs.forall(isSimpleExpr) =>
+            (lop, rop) match {
+              case (Head, Head) => DoPrim(Head, rhs, Seq(lc.head min rc.head), ltpe)
+              case (Tail, Tail) => DoPrim(Tail, rhs, Seq(lc.head + rc.head), ltpe)
+              case (Shr,  Shr)  => DoPrim(Shr,  rhs, Seq(lc.head + rc.head), ltpe)
+              case (_,_) => (lowerToDoPrimOpBits(lhs), lowerToDoPrimOpBits(of)) match {
+                case (DoPrim(Bits, _, Seq(lmsb, llsb), _), DoPrim(Bits, _, Seq(rmsb, rlsb), _)) =>
+                  DoPrim(Bits, rhs, Seq(lmsb+rlsb,llsb+rlsb), ltpe)
+                case (_,_) => lhs  // Not a candidate
+              }
+            }
+            case _ => lhs  // Not a candidate
+          }
+      case other => other // Not a candidate
+    }
+  }
+
+  /** Inline bits in a Statement
+    *
+    * @param netlist a '''mutable''' HashMap mapping references to [[firrtl.ir.DefNode DefNode]]s to their connected
+    * [[firrtl.ir.Expression Expression]]s. This function '''will''' mutate it if stmt is a [[firrtl.ir.DefNode
+    * DefNode]] with a value that is a [[PrimOp]] Bits
+    * @param stmt the Statement being searched for nodes and transformed
+    * @return Returns stmt with Bits inlined
+    */
+  def onStmt(netlist: Netlist)(stmt: Statement): Statement =
+    stmt.map(onStmt(netlist)).map(onExpr(netlist)) match {
+      case node @ DefNode(_, name, value) if isTemp(name) =>
+        netlist(we(WRef(name))) = value
+        node
+      case other => other
+    }
+
+  /** Replaces bits in a Module */
+  def onMod(mod: DefModule): DefModule = mod.map(onStmt(new Netlist))
+}
+
+/** Inline nodes that are simple bits */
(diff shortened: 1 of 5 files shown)


src/main/scala/firrtl/transforms/InlineBitExtractions.scala
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+  /** Replaces bits in a Module */
+  def onMod(mod: DefModule): DefModule = mod.map(onStmt(new Netlist))
+}
+
+/** Inline nodes that are simple bits */
+class InlineBitExtractionsTransform extends Transform {
+  def inputForm = LowForm
+  def outputForm = LowForm
```
seldridge: nit: It's probably safer to make this `UnknownForm` -> `UnknownForm`. While there's no harm in specifying this as `LowForm`, forms aren't sufficient to describe the dependencies of this transform (Using LowForm, if you tried to run this with `--run-firrtl-transform` it would run right before the Verilog Emitter which is earlier than desired).

johningalls-sifive: I'm still learning these concepts, so here I just copy-pasted from InlineNots.scala, which was itself copy-pasted from ReplaceTruncatingArithmetic.scala and CombineCats.scala.  For consistency, should we change those as well?

seldridge: Those should change, but I'd suggest doing it after Jack and I get #1123 merged. That should then allow moving *everything* to `inputForm = UnknownForm`/`outputForm = UnknownForm` as dependencies are then specified using the dependency API.

No need to do it now, though. (And I should have suggested this on the `InlineNots` PR, but wasn't being proactive enough.)

src/main/scala/firrtl/transforms/InlineBitExtractions.scala
```diff
@@ -0,0 +1,102 @@
...
+  // Note that this can have false negatives but MUST NOT have false positives.
+  private def isSimpleExpr(expr: Expression): Boolean = expr match {
+    case _: WRef | _: Literal | _: WSubField => true
+    case DoPrim(op, args, _,_) if isBitExtract(op) => args.forall(isSimpleExpr)
+    case _ => false
+  }
+
+  // replace Head/Tail/Shr with Bits for easier back-to-back Bits Extractions
+  private def lowerToDoPrimOpBits(expr: Expression): Expression = expr match {
+    case DoPrim(Head, rhs, c, tpe) if isSimpleExpr(expr) && rhs.forall(isSimpleExpr) =>
```
azidar: `isSimpleExpr` is recursive, in that it calls itself on all child arguments if it sees a DoPrim that is a bit extract. Thus, wouldn't calling `isSimpleExpr(expr)` also call itself on `rhs`, making `rhs.forall(isSimpleExpr)` unnecessary?

src/main/scala/firrtl/transforms/InlineBitExtractions.scala
```diff
@@ -0,0 +1,102 @@
...
+    case _ => expr // Not a candidate
+  }
+
+  /** Mapping from references to the [[firrtl.ir.Expression Expression]]s that drive them */
+  type Netlist = mutable.HashMap[WrappedExpression, Expression]
+
+  /** Recursively replace [[WRef]]s with new [[Expression]]s
+    *
+    * @param netlist a '''mutable''' HashMap mapping references to [[firrtl.ir.DefNode DefNode]]s to their connected
+    * [[firrtl.ir.Expression Expression]]s. It is '''not''' mutated in this function
```
azidar: Is there a reason there is two `DefNode`s, e.g. `[[firrtl.ir.DefNode DefNode]]`. Same for `Expression`?

src/main/scala/firrtl/transforms/InlineBitExtractions.scala
```diff
@@ -0,0 +1,102 @@
...
+    * @return Returns expr with Bits inlined
+    */
+  def onExpr(netlist: Netlist)(expr: Expression): Expression = {
+    expr.map(onExpr(netlist)) match {
+      case e @ WRef(name, _,_,_) =>
+        netlist.get(we(e))
+               .filter(isBitExtract)
+               .getOrElse(e)
+      // replace back-to-back Bits Extractions
+      case lhs @ DoPrim(lop, ival, lc, ltpe) if isSimpleExpr(lhs) && ival.forall(isSimpleExpr) =>
```
azidar: Similar to the earlier comment, `isSimpleExpr` is recursive so I think the `ival.forall(isSimpleExpr)` is unnecessary.

src/main/scala/firrtl/transforms/InlineBitExtractions.scala
```diff
@@ -0,0 +1,102 @@
...
+        netlist.getOrElse(we(ival.head), ival.head) match {
```
azidar: Why do we look in `netlist` here? At this point in the function, we've already called `onExpr(netlist)` on all children expressions of `expr`. Conceptually this is a post-order traversal, meaning we've already called the `case e @ WRef(..)` on line 48 on any children which could have matched in `netlist`. I suspect if you solely matched on `ival.head`, there would be no difference in behavior.

src/main/scala/firrtl/transforms/InlineBitExtractions.scala
```diff
@@ -0,0 +1,102 @@
...
+          case of @ DoPrim(rop, rhs, rc, rtpe) if isSimpleExpr(of) && rhs.forall(isSimpleExpr) =>
```
azidar: Again, potentially duplicate calls to `isSimpleExpr` on `rhs`.

src/main/scala/firrtl/transforms/InlineBitExtractions.scala
```diff
@@ -0,0 +1,102 @@
...
+            (lop, rop) match {
+              case (Head, Head) => DoPrim(Head, rhs, Seq(lc.head min rc.head), ltpe)
+              case (Tail, Tail) => DoPrim(Tail, rhs, Seq(lc.head + rc.head), ltpe)
+              case (Shr,  Shr)  => DoPrim(Shr,  rhs, Seq(lc.head + rc.head), ltpe)
+              case (_,_) => (lowerToDoPrimOpBits(lhs), lowerToDoPrimOpBits(of)) match {
+                case (DoPrim(Bits, _, Seq(lmsb, llsb), _), DoPrim(Bits, _, Seq(rmsb, rlsb), _)) =>
+                  DoPrim(Bits, rhs, Seq(lmsb+rlsb,llsb+rlsb), ltpe)
+                case (_,_) => lhs  // Not a candidate
```
azidar: Comment on 57-64, I like how you are doing your best to preserve the original operator if you can. Cool!

src/main/scala/firrtl/transforms/InlineBitExtractions.scala
johningalls-sifive: no idea: here I just copy-pasted from InlineNots.scala, which was itself copy-pasted from ReplaceTruncatingArithmetic.scala and CombineCats.scala

src/main/scala/firrtl/transforms/InlineBitExtractions.scala
johningalls-sifive: yes: good eye

src/main/scala/firrtl/transforms/InlineBitExtractions.scala
johningalls-sifive: yes: good eye

src/main/scala/firrtl/transforms/InlineBitExtractions.scala
johningalls-sifive: yes: good eye

src/main/scala/firrtl/transforms/InlineBitExtractions.scala
johningalls-sifive: okay.  my small testcase shows no difference in behavior.

src/main/scala/firrtl/transforms/InlineBitExtractions.scala
azidar: Ok, I presume this is a mistake, just remove the second duplicate word?

seldridge: This is so that you link to the first, but show the second. I can never ascertain when Scaladoc will properly resolve it (I think it's a function of the includes) and opt to be explicit with the full link, but not make the Scaladoc unreadable. Best bet is to run it through Scaladoc and see the output.

seldridge: Checking quickly... you need the `[[firrtl.ir.Expression Expression]]` here. Running this through `sbt unidoc`, you wind up with:

```
[warn] /home/schuyler/repos/github.com/freechipsproject/firrtl/src/main/scala/firrtl/transforms/InlineBitExtractions.scala:40:3: Could not find any member to link for "Expression".
[warn]   /** Recursively replace [[WRef]]s with new [[Expression]]s
```

Why Scaladoc can't find this is beyond me...

src/main/scala/firrtl/transforms/InlineBitExtractions.scala
```diff
@@ -0,0 +1,102 @@
...
+    *
+    * @param netlist a '''mutable''' HashMap mapping references to [[firrtl.ir.DefNode DefNode]]s to their connected
+    * [[firrtl.ir.Expression Expression]]s. This function '''will''' mutate it if stmt is a [[firrtl.ir.DefNode
+    * DefNode]] with a value that is a [[PrimOp]] Bits
+    * @param stmt the Statement being searched for nodes and transformed
+    * @return Returns stmt with Bits inlined
+    */
+  def onStmt(netlist: Netlist)(stmt: Statement): Statement =
+    stmt.map(onStmt(netlist)).map(onExpr(netlist)) match {
+      case node @ DefNode(_, name, value) if isTemp(name) =>
```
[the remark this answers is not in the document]
johningalls-sifive: quick reply: I decided that preserving non-temporary names was more important for Verilog readability than inlining removing them

src/main/scala/firrtl/transforms/InlineBitExtractions.scala
```diff
@@ -0,0 +1,102 @@
...
+        node
+      case other => other
+    }
+
+  /** Replaces bits in a Module */
+  def onMod(mod: DefModule): DefModule = mod.map(onStmt(new Netlist))
+}
+
+/** Inline nodes that are simple bits */
+class InlineBitExtractionsTransform extends Transform {
```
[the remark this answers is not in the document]
johningalls-sifive: this is systemic and not unique to the function of this change: can we please discuss this in a separate PR for these transforms?
InlineBitExtractions
InlineCasts
InlineNots
LegalizeClocks
ReplaceTruncatingArithmetic

src/main/scala/firrtl/transforms/InlineNots.scala
```diff
@@ -42,10 +42,18 @@ object InlineNotsTransform {
...
+      // replace bits-of-not with not-of-bits to enable later bit extraction transform
+      case lhs @ DoPrim(op, Seq(lval), lcons, ltpe) if isBitExtract(op) && isSimpleExpr(lval) =>
+        netlist.getOrElse(we(lval), lval) match {
+          case DoPrim(Not, Seq(rhs), rcons, rtpe) =>
+            DoPrim(Not, Seq(DoPrim(op, Seq(rhs), lcons, ltpe)), rcons, ltpe)
+          case _ => lhs  // Not a candiate
+        }
       // replace back-to-back inversions with a straight rename
-      case lhs @ DoPrim(Not, Seq(inv), _,_) if isSimpleExpr(inv) =>
+      case lhs @ DoPrim(Not, Seq(inv), _, invtpe) if isSimpleExpr(lhs) && isSimpleExpr(inv) && (lhs.tpe == invtpe) && (bitWidth(lhs.tpe) == bitWidth(inv.tpe)) =>
```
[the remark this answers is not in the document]
johningalls-sifive: I'm also uncertain about whether any problems could arise by removing back-to-back inversions with changing types or widths, which is why I added these extra safeguards just to make myself feel better about it.

src/main/scala/firrtl/transforms/InlineBitExtractions.scala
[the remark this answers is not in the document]
azidar: Sure, it's fine as is.

PR #1301, opened 2020-01-08
[wip] Move InferResets later

### LINKED ISSUE #1300: InferResets Emits Invalid FIRRTL
## Checklist

The modifications to `InferResets` introduced in #1291 produce invalid FIRRTL that is only cleaned up after `ExpandWhens`. The implication of this is that any transform between `InferResets` and `ExpandWhens` is operating on invalid FIRRTL and has the potential to break. This impacts custom `HighForm` transforms. Additionally, this breaks #1123 as this adds a `CheckTypes` after `InferResets`.

@ucbjrl: This should block the 1.3 RC from going out today. :sob:

You should be able to run `CheckTypes` after any transform and it should pass. If it doesn't pass, then it should pass with necessary resolution transforms, e.g., `InferTypes`. If it still doesn't pass, then the transform is broken (or there's a problem with resolution/checks!).

### Steps to Reproduce

Concretely, if you try to run `CheckTypes` after `InferResets` with #1291 applied, you wind up with an error. You can test this with the following diff:

```diff
diff --git a/src/main/scala/firrtl/LoweringCompilers.scala b/src/main/scala/firrtl/LoweringCompilers.scala
index 0d9b971b..9cbe6139 100644
--- a/src/main/scala/firrtl/LoweringCompilers.scala
+++ b/src/main/scala/firrtl/LoweringCompilers.scala
@@ -49,7 +49,9 @@ class ResolveAndCheck extends CoreTransform {
     new passes.TrimIntervals(),
     new passes.InferWidths,
     passes.CheckWidths,
-    new firrtl.transforms.InferResets)
+    new firrtl.transforms.InferResets,
+    passes.InferTypes,
+    passes.CheckTypes)
 }
 
 /** Expands aggregate connects, removes dynamic accesses, and when

```

### Your environment

- FIRRTL Versions failing: a4f2eda0 (current master), 3e096a73 (#1291 merge)
- FIRRTL Version passing: 70088cd2 (#1291 base)

### External Information

This was discussed between @seldridge and @jackkoenig on Slack.

### Discussion

This needs more discussion, but there's a question about whether or not the following should actually be legal (the #1291 example):

```
input a : UInt<1>
input b : AsyncReset
output z : Reset
z <= a
z <= b
```

Problematically, from the perspective of `CheckTypes` this is wrong as `z` can't have two types. This is similar to the following, which should continue to error:

```
input a: UInt<1>
input b: SInt<1>
output z: SInt<1>
z <= a
z <= b
```

It's not clear to me why this should be treated differently for resets as opposed to other ground types.

What I think needs to happen here is that `InferResets` needs to run on middle FIRRTL (where the IR is SSA and last-connect semantics are resolved) and not high FIRRTL. 

The current placement of `InferResets` is done, at least, to prevent deduplication of two modules with abstract reset that will later have resets inferred differently. One approach could involve running `EliminateTargetPaths` before a later `InferResets` to duplicate modules with abstract resets.

seldridge:
### Dependencies

- fixes #1300 
- prerequisites: [#1302, #1370, #1388, #1391, #1392, #1394]

### Checklist

### Type of Improvement

- bug fix                            

### API Impact

No API impact.

Module names may change due to to deduplication/duplication to resolve `InferResets`.

### Desired Merge Strategy

### DIFF
--- a/src/main/scala/firrtl/transforms/InferResets.scala
+++ b/src/main/scala/firrtl/transforms/InferResets.scala
@@ -3,12 +3,17 @@
 package firrtl.transforms
 
 import firrtl._
+import firrtl.analyses.InstanceGraph
 import firrtl.ir._
 import firrtl.Mappers._
 import firrtl.traversals.Foreachers._
-import firrtl.annotations.{ReferenceTarget, TargetToken}
+import firrtl.annotations.{
+  InstanceTarget,
+  ReferenceTarget,
+  TargetToken}
+import firrtl.annotations.TargetToken.{fromWDefInstanceToTargetToken, Instance, OfModule}
 import firrtl.Utils.toTarget
-import firrtl.passes.{Pass, PassException, Errors, InferTypes}
+import firrtl.passes.{PassException, Errors}
 
 import scala.collection.mutable
 import scala.util.Try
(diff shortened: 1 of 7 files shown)


src/main/scala/firrtl/transforms/InferResets.scala
```diff
@@ -163,6 +165,12 @@ class InferResets extends Transform {
               val values = conLookup(key).getOrElse(Nil) ++ altLookup(key).getOrElse(Nil)
               map(key) = values
             }
+          case lhs@ DefNode(_, _, rhs) => rhs match {
+            case Mux(_, tval, fval, _) =>
+              markResetDriver(WRef(lhs), tval)
+              markResetDriver(WRef(lhs), fval)
+            case _ => markResetDriver(WRef(lhs), rhs)
+          }
```
jackkoenig: Not having this was probably just an oversight since one could express this in Chisel so it wouldn't work properly with the current transform.

I'll note that moving this after ExpandConnects and ExpandWhens means we can vastly simplify much of the above logic.

seldridge:
This is currently failing three Chisel tests:

1. `chiselTests.AnnotationNoDedup`
1. `chiselTests.LoadMemoryFromFileSpec`
1. `chiselTests.CloneModuleSpec`

The first two are because of duplication (`EliminateTargetPaths`) running before `InferResets` causing names to change. The latter is of unknown cause, but I expect the same thing.

This PR may be good to go if deduplication adds an annotation recording the old module name and (1) duplication uses that when generating the new module and (2) duplication does not change the name if only one instance exists.

seldridge:
@jackkoenig: I'm still cleaning this up, but this is working as far as I can tell. The only failing test on the Chisel side is related to dup/dedup not properly preserving names and tests expecting things to be named a certain way. This is fixable (either with a less brittle Chisel test or with an additional prerequisite PR that adds annotations when something is deduplicated to allow it to be renamed correctly when duplicated), but should be orthogonal to a functional solution for #1300. 

I'll un-draft this after I refactor out one additional PR.

jackkoenig:
@seldridge Is this still needed?

seldridge:
Closed in favor of #1396.

PR #1307, opened 2020-01-10
Change LoggerState.globalLevel to Warn

ucbjrl:
PR #1305 changes the `globalLogLevel` in `LogLevelAnnotation` to from `None` to `Warn`. Update the default `LoggerState.globalLevel` to `Warn` as well.

### Checklist

### Type of Improvement

- bug fix - 3c8e22d inadvertently disabled `printf()` in tests.

### API Impact

This restores default behavior prior to 3c8e22d which had a side-effect of suppressing `printf()` in tests.

This restores generation of Verilog `$fwrite()` which would have been removed by 3c8e22d

### Desired Merge Strategy

### DIFF
--- a/src/main/scala/logger/Logger.scala
+++ b/src/main/scala/logger/Logger.scala
@@ -54,7 +54,7 @@ trait LazyLogging {
   * when used in multi-threaded environments
   */
 private class LoggerState {
-  var globalLevel = LogLevel.None
+  var globalLevel = LogLevel.Warn
   val classLevels = new scala.collection.mutable.HashMap[String, LogLevel.Value]
   val classToLevelCache = new scala.collection.mutable.HashMap[String, LogLevel.Value]
   var logClassNames = false
--- a/src/test/scala/loggertests/LoggerSpec.scala
+++ b/src/test/scala/loggertests/LoggerSpec.scala
@@ -12,6 +12,7 @@ object LoggerSpec {
   val InfoMsg = "message info"
   val DebugMsg = "message debug"
   val TraceMsg = "message trace"
+  val globalLevel = LogLevel.Warn
 }
 
 class Logger1 extends LazyLogging {
@@ -37,7 +38,7 @@ class LogsInfo3 extends LazyLogging {
 class LoggerSpec extends FreeSpec with Matchers with OneInstancePerTest with LazyLogging {
   "Logger is a simple but powerful logging system" - {
     "Following tests show how global level can control logging" - {
-      "only error shows up by default" in {
+      "only warn and error shows up by default" in {
         Logger.makeScope() {
           val captor = new OutputCaptor
           Logger.setOutput(captor.printStream)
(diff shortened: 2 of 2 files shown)


src/main/scala/logger/Logger.scala
```diff
@@ -54,7 +54,7 @@ trait LazyLogging {
   * when used in multi-threaded environments
   */
 private class LoggerState {
-  var globalLevel = LogLevel.None
+  var globalLevel = LogLevel.Warn
```
seldridge: I'm fine with this as is, but you may consider using the default value in the annotation to try to get at one source for ground truth here.

```suggestion
  var globalLevel = LogLevelAnnotation.globalLogLevel
```

ucbjrl: Hmm. The doesn't appear to compile:
```
Error:(57, 40) value globalLogLevel is not a member of object logger.LogLevelAnnotation
  var globalLevel = LogLevelAnnotation.globalLogLevel
```
It seems I would need an instance of the class `LogLevelAnnotation`. Do you have one handy?

seldridge: No, we'd have to construct one (non ideal). Maybe we can add this to the object:

```diff
diff --git a/src/main/scala/logger/LoggerAnnotations.scala b/src/main/scala/logger/LoggerAnnotations.scala
index 323b46e5..f8d01d32 100644
--- a/src/main/scala/logger/LoggerAnnotations.scala
+++ b/src/main/scala/logger/LoggerAnnotations.scala
@@ -14,10 +14,12 @@ sealed trait LoggerOption { this: Annotation => }
   *  - if unset, a [[LogLevelAnnotation]] with the default log level will be emitted
   * @param level the level of logging
   */
-case class LogLevelAnnotation(globalLogLevel: LogLevel.Value = LogLevel.None) extends NoTargetAnnotation with LoggerOption
+case class LogLevelAnnotation(globalLogLevel: LogLevel.Value = LogLevelAnnotation.default) extends NoTargetAnnotation with LoggerOption
 
 object LogLevelAnnotation extends HasShellOptions {
 
+  val default = LogLevel.Warn
+
   val options = Seq(
     new ShellOption[String](
       longOption = "log-level",

```

ucbjrl: I'm slightly concerned about the coupling between logging and annotations. In my mind, logging is a short step up from `println`. I realize it may be convenient to have access to options/annotations for generating logging scopes, but my gut feeling is to invert the dependency and have logging at a lower level (with no awareness of annotations or options).

seldridge: Note: I view this narrowly about defining a single source of truth for the default logging level. I'd be fine if that lives in the logger state and that default is used in the log level annotation.

For stages, it's proved convenient to define defaults in the annotations because those also define the option parsing. And it seemed natural to put parsing and defaults together.

I still approve of the PR as is. :+1:

ucbjrl: Right. I agree that parsing and defaults are connected. I'm not sure I have a clear concept of how `LoggerState` should be managed.

Note that prior to this PR there were two independent defaults for `LogLevelAnnotation` and `LoggerState`. I'd like to leave their merging to a discussion of who's on top and a subsequent PR.

PR #1325, opened 2020-01-25
Create scala.yml

azidar:
Try out new github action for scala CI!


albert-magyar:
Fails because no Verilator :(

PR #1347, opened 2020-01-30
Added product impl for NoTargetAnnotation

azidar:
### Checklist

### Type of Improvement

   - bug fix

### API Impact

I believe this only adds to the current API - it implements the unimplemented `Product` functions for `NoTargetAnnotation`s. Previously, most `NoTargetAnnotation`s are case classes (which implements `Product`), but if it wasn't a case class, you'd have to implement them yourself. The reason `Annotation` extends `Product` is to iterate over `Target`'s.

None.

### DIFF
--- a/src/main/scala/firrtl/annotations/Annotation.scala
+++ b/src/main/scala/firrtl/annotations/Annotation.scala
@@ -43,7 +43,10 @@ trait Annotation extends Product {
 trait NoTargetAnnotation extends Annotation {
   def update(renames: RenameMap): Seq[NoTargetAnnotation] = Seq(this)
   override def productArity: Int = 0
-  override def canEqual(that: Any): Boolean = true
+  override def canEqual(that: Any): Boolean = that match {
+    case _: NoTargetAnnotation => true
+    case _                     => false
+  }
   override def productElement(n: Int): Any = throw new java.lang.IndexOutOfBoundsException(s"Cannot access elements of NoTargetAnnotation: $n")
 }
 
--- a/src/main/scala/firrtl/annotations/Annotation.scala
+++ b/src/main/scala/firrtl/annotations/Annotation.scala
@@ -44,7 +44,7 @@ trait NoTargetAnnotation extends Annotation {
   def update(renames: RenameMap): Seq[NoTargetAnnotation] = Seq(this)
   override def productArity: Int = 0
   override def canEqual(that: Any): Boolean = true
-  override def productElement(n: Int): Any = throw new java.lang.IndexOutOfBoundsException(n)
+  override def productElement(n: Int): Any = throw new java.lang.IndexOutOfBoundsException(s"Cannot access elements of NoTargetAnnotation: $n")
 }
 
 /** An Annotation that targets a single [[Named]] thing */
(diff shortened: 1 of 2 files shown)


src/main/scala/firrtl/annotations/Annotation.scala
```diff
@@ -42,6 +42,9 @@ trait Annotation extends Product {
   */
 trait NoTargetAnnotation extends Annotation {
   def update(renames: RenameMap): Seq[NoTargetAnnotation] = Seq(this)
+  override def productArity: Int = 0
+  override def canEqual(that: Any): Boolean = true
+  override def productElement(n: Int): Any = 0
```
seldridge: What do you think about instead throwing a `java.lang.IndexOutOfBoundsException`? That's what both `case object Foo` or `case class Bar()` behave with `productElement`.

azidar: Yeah, I was thinking about that yesterday. I think you are right, I'll fix it.

src/main/scala/firrtl/annotations/Annotation.scala
```diff
@@ -42,6 +42,9 @@ trait Annotation extends Product {
   */
 trait NoTargetAnnotation extends Annotation {
   def update(renames: RenameMap): Seq[NoTargetAnnotation] = Seq(this)
+  override def productArity: Int = 0
+  override def canEqual(that: Any): Boolean = true
```
seldridge: While I think this is fine as-is, would it be better to do?

```suggestion
  override def canEqual(that: Any): Boolean = that match {
    case _: NoTargetAnnotation => true
    case _                     => false
  }
```

azidar:
Ugh so what should we do, make Annotation not extend Product anymore?

seldridge:
Couple of options:

1. Users have to implement this
1. We add a helper trait that implements these defaults (`trait EmptyAnnotation`?), but it is not mixed into `NoTargetAnnotation`
1. We add some janky `trait MyProduct { val contents: Seq[Any] }` which defines these the right way for a user

I think I'm in favor of (1). (2) is probably okay, too. (3) seems like it is just complicating (1) and would probably not the correct mix-in for `NoTargetAnnotation`.

Re: not extending `Product`... what is the current benefit of extending product? I do think that `Product` is the right way of thinking about annotations, but maybe it's too restrictive?

azidar:
Well actually, if the idea behind "NoTargetAnnotation" is there is nothing to iterate over, I suppose it makes sense that productArity is zero, no?

seldridge:
My reading is that a `NoTargetAnnotation` indicates that there are no _targets_ to iterate over. However, that doesn't mean that the arity is zero as this can contain arbitrary metadata that just doesn't have a target.

Basically, Stage/Phase and "Annotations are Options" makes this harder to reason about.

E.g., currently `TargetDirAnnotation` is defined as:
```scala
case class TargetDirAnnotation(directory: String = ".") extends NoTargetAnnotation with StageOption
```

Clearly, this is a product algebraic data type with arity one.

Maybe the real issue is that `NoTargetAnnotation` needs to be refactored out of `firrtl.options`. I.e., we need to treat annotations as true metadata containers and decouple the association with target.

PR #1349, opened 2020-02-01
Fix conversion of Reference-containing expressions to ReferenceTargets

albert-magyar:
**Type of Improvement:** bug fix
**API impact:** none
**Backend code generation impact:** none in Verilog (only corrects potential for faulty annotations)

The `toTarget` utility currently handles `Reference` wrong by putting the name of the root `Reference` into the `component` path rather than actually using it as the `ref`. This PR fixes that behavior.


jackkoenig:
Is this something that should be backported?

seldridge:
I'd consider this a bugfix and is safe to backport.

albert-magyar:
Thanks, @seldridge. Yeah, this should be backported.

Issue #559, opened 2017-05-03
Enforce CircuitState only contains IR nodes, detect form


azidar:
Mix in working IR trait to transform

PR #1350, opened 2020-02-02
Remove WIR nodes that "shadow" standard IR nodes

### LINKED ISSUE #1097: Support WIR in CHIRRTL Transforms
We can fix this with simply a ToIR transform that runs at the beginning of the CHIRRTL transforms

**Type of issue**: bug report

    (examples)
    - version: `3.0-SNAPSHOT`
    - OS: `Linux knight 4.4.0-92-generic #115-Ubuntu SMP Thu Aug 10 09:04:33 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux`
    - verilator version (if relevant)

albert-magyar:
**Type of improvement:** API simplification + pass reusability improvement
**API impact:** API modification (see below)
**Backend Code Generation Impact:** none
**Desired merge strategy:** rebase
**Release notes:**
_See extended commit message of ba40a5c2cdbe5e485bbcefe1411ab551b78c7cfa for lengthy release notes._

Fixes #1097 
Helps with #559 
Removes incompatibility of WIR with ProtoBuf from #832

This PR absorbs the extra fields of `WRef`, `WSubField`, `WSubIndex`, `WSubAccess`, and `WDefInstance` into the corresponding standard IR nodes. The point of this PR is to show the minimal set of changes to accommodate this, so I haven't included other commits that do things like remove unreachable code and port built-in passes to reduce the use of the `W____` versions.

**Type of improvement:** API simplification + pass reusability improvement
**Backend Code Generation Impact:** None

### API Impact

**TLDR:** Code that used WIR will not change at all. Code that used non-WIR IR will have to update pattern matching signatures to hold the new fields, but this applies to very little (or no) user code.

This uses type aliases and new `unapply` and `apply` methods of objects to act as a compatibility layer, so the API impact is minimal.

There is exactly one code-breaking API change: the signatures of non-WIR `Reference`, `SubField`, et cetera change when pattern matching. There is no way to solve this with an extra `unapply`, so any code that pattern matches against non-WIR nodes will need to change the type signature of the pattern:

```
case Reference(_, _) => // old
case Reference(_, _, _, _) => // new
```

Fortunately, nearly all non-core passes rely on WIR. **Any code that was using WIR before will not have to change at all!**

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -121,7 +121,7 @@ sealed abstract class FirrtlEmitter(form: CircuitForm) extends Transform with Em
       // Use list instead of set to maintain order
       val modules = mutable.ArrayBuffer.empty[DefModule]
       def onStmt(stmt: Statement): Unit = stmt match {
-        case DefInstance(_, _, name) => modules += map(name)
+        case DefInstance(_, _, name, _) => modules += map(name)
         case WDefInstance(_, _, name, _) => modules += map(name)
         case _: WDefInstanceConnector => throwInternalError(s"unrecognized statement: $stmt")
         case other => other.foreach(onStmt)
--- a/src/main/scala/firrtl/Utils.scala
+++ b/src/main/scala/firrtl/Utils.scala
@@ -220,12 +220,12 @@ object Utils extends LazyLogging {
       case WSubAccess(expr, index, _, _) => niceName(depth)(expr) + niceName(depth - 1)(index)
       case WSubField(expr, field, _, _) => niceName(depth)(expr) + "_" + field
       case WSubIndex(expr, index, _, _) => niceName(depth)(expr) + "_" + index
-      case Reference(name, _) if name(0) == '_' => name
-      case Reference(name, _) => "_" + name
-      case SubAccess(expr, index,  _) if depth <= 0 => niceName(depth)(expr)
-      case SubAccess(expr, index,  _) => niceName(depth)(expr) + niceName(depth - 1)(index)
-      case SubField(expr, field, _) => niceName(depth)(expr) + "_" + field
-      case SubIndex(expr, index, _) => niceName(depth)(expr) + "_" + index
+      case Reference(name, _, _, _) if name(0) == '_' => name
+      case Reference(name, _, _, _) => "_" + name
+      case SubAccess(expr, index, _, _) if depth <= 0 => niceName(depth)(expr)
+      case SubAccess(expr, index, _, _) => niceName(depth)(expr) + niceName(depth - 1)(index)
+      case SubField(expr, field, _, _) => niceName(depth)(expr) + "_" + field
+      case SubIndex(expr, index, _, _) => niceName(depth)(expr) + "_" + index
       case DoPrim(op, args, consts, _) if depth <= 0 => "_" + op
       case DoPrim(op, args, consts, _) => "_" + op + (args.map(niceName(depth - 1)) ++ consts.map("_" + _)).mkString("")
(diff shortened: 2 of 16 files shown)


azidar:
I think this is great. I realize this is a 1.4 milestone, is there a reason not to include it in 1.3? Also, if it is 1.4, does that mean we need to wait until we create a 1.3.x branch? (because right now I think that is just the master branch).

seldridge:
As far as I'm concerned 1.3 is wide open until we get the `InferResets` bug sorted out. So, the merge window is open and I'd be happy to get this in (and anything else).

albert-magyar:
I just don't want to make everyone's life harder with the 1.3 release :) I also know that this requires a fair bit of searching for `Reference` et al. in user code, though I do think this is effectively non-existent.

There is still one open question: is it worth moving some of the other non-shadowing types (`WVoid`, etc)? Some of them effectively belong to specific transforms.

With this PR in place, the distinction would be that you have IR, which contains nodes representing real IR, and you have WIR, which exclusively contains "fake" things that don't correspond with real IR. I think that's a reasonable line to draw/

ucbjrl:
I'd prefer to see this remain tagged as 1.4. We're a month later than planned on 1.3, and I would hate to see 1.3 slip another month due to last-minute release busters. There's already a 1.3.x branch, so there's no fundamental reason this can't be merged into master as soon as it's deemed ready. The potential downside is that 1.3 release buster fixes will need to be cherry-picked but we should be prepared to deal with that possibility in any case.

albert-magyar:
I added a (very long) expanded summary to the first commit message that describes the change and associated API impacts.

albert-magyar:
This is ready for review again.

One notable thing that this does **not** do is "banish" the mentions of `WRef`, `WSubIndex`, etc from the core codebase. I intend to follow this PR up with another PR that does that, but that's going to be a huge sed script change, so I didn't want to clog up the diff for now.

albert-magyar:
There is **only one** source incompatibility introduced by this patch: unapply arity for the five "WIR-shadowed" standard IR nodes (`Reference`, `SubField`, `SubAccess`, `SubIndex`, `DefInstance`) increases. This pattern **does not** appear in the Chisel3 source code, so no corresponding patch will be needed for Chisel3.

```diff
-    case Reference(n, t) => ???
+    case Reference(n, t, _, _) => ???

-    case ir.DefInstance(i, n, m) => ???
+    case ir.DefInstance(i, n, m, _) => ???

-    case firrtl.ir.SubField(e, n, t) => ???
+    case firrtl.ir.SubField(e, n, t, _) => ???
```

This scalafix rewriter is an example of how to fix it. Note that it does not semantically interpret the namespace qualifiers of the patterns (a weakness of any scalafix rule), but someone would have to have written some strange stuff for this to break.

```scala
 import scalafix.v1._
 import scala.meta._

 class FirrtlUnapplyOldIR extends SemanticRule("FirrtlUnapplyOldIR") {

   val newArity = Map(
     "Reference" -> 4,
     "SubField" -> 4,
     "SubIndex" -> 4,
     "SubAccess" -> 4,
     "DefInstance" -> 4
   )

   override def fix(implicit doc: SemanticDocument): Patch = {
     doc.tree.collect {
       case t @ Pat.Extract(Term.Name(patName), args) if newArity.contains(patName) =>
         Patch.replaceTree(t, t.copy(args = args.padTo(newArity(patName), Pat.Wildcard())).toString)
       case t @ Pat.Extract(Term.Select(q, Term.Name(patName)), args) if newArity.contains(patName) =>
         Patch.replaceTree(t, t.copy(args = args.padTo(newArity(patName), Pat.Wildcard())).toString)
     }.asPatch
   }

 }
```

PR #1355, opened 2020-02-05
Add constant prop to async regs

azidar:
### Checklist

### Type of Improvement

  - bug fix

### API Impact

Enables constant propping of asynchronously reset registers.


albert-magyar:
Looks like a clean way to reuse the existing code. Good call on renaming `updateNodeMap`.

It might be good to add a comment like "Add a Mux that abstractly represents the reset case so that it can be analyzed as a source of non-constant data"

albert-magyar:
I think it's technically not necessary to track the async reset regs, so it's possible to roll this extra synthetic Mux into the logic cone for any register without breaking the logic cone analysis. Doing this might also smooth the path to HighForm const prop, since it won't cause the pass to break if RemoveResets isn't run (as my current logic cone analysis would).

However, I think your way is better for clarity given the current status and dependencies of the pass.

mwachs5:
what happens if someone don't-touches the async reg that need to be constant propped?

azidar:
If someone dont-touches the register, then constant prop won't happen. Is that not the desired behavior? Would you like to see a test of that?

mwachs5:
Well I fear that that will mean it will end up with latches in the design if people don't touch. 

An alternative would be to emit the `else` clause to the same value as the `if (reset)` clause if they are don't touched.

azidar:
Ah gotcha I understand. I'll update the PR to emit proper registers so latches aren't inferred.

aswaterman:
Stating the obvious, but rather than `if (reset) x <= k; else x <= k;`, can just do `x <= k;`.  This might be nice because it can be done in constprop pass rather than in the emitter.

albert-magyar:
Am I understanding correctly that the proposed issue is:
a) There's an async-reset register that may result in a latch post-synthesis
b) It is protected from constant propagation (which otherwise would fix the problem) with a `dontTouch`?

If so, I don't think this PR will hurt that case, and I think that helping that case is probably best handled outside of `ConstantPropagation` to avoid breaking the invariant that `ConstantPropagation` doesn't change `dontTouch`-ed components.

mwachs5:
Let me be a bit specific the problem I hope we are solving. In Chisel (sorry I don't know FIRRTL) if you do something like:

```
val FooBarBundle extends Bundle {
 val foo = Bool()
 val bar = Bool()
}

val foo_reg = RegInit((new FooBarBundle.fromBits(0.U))
...
// foo_reg.foo never assigned to anything else
foo_reg.bar := true.B
```

If is is synchronously reset, it will turn foo_reg_foo into a wire tied to false, or if the register is dont-touched it will be:

```
always @(posedge clock) 
  if (reset)
    foo_reg_foo <= 1'b0
end
```
which apparently linter tools are OK with. 

But if you have an async flop it is currently (regardless of dont-touching):
```
always @(posedge clock or posedge reset) 
  if (reset)
    foo_reg_foo <= 1'b0
end
```

Which synthesis tools complain is a latch.

My suggestion was if you can't touch the register, you could change the latter into:
```
always @(posedge clock or posedge reset) 
  if (reset)
    foo_reg_foo <= 1'b0
  else 
    foo_reg_foo <= 1'b0
end
```

mwachs5:
@albert-magyar before this PR "b) It is protected from constant propagation (which otherwise would fix the problem) with a dontTouch?" isn't quite true, before this PR the status quo is that no constant prop would happen at all.

albert-magyar:
When I say "would fix the problem", I mean that the improved const-prop from this PR will turn that async-reset register into `wire foo_reg_foo; assign foo_reg_foo = 1'b0;`, which should avoid a latch and satisfies the as-if optimization model we use for const-prop on synchronous registers.

It seems like the remaining issue is that this PR cannot help that situation if the register is `dontTouch`-ed. That seems like a separate issue with how async reset registers are emitted in general, since it should not depend on `ConstantPropagation` to emit useful Verilog.

In other words, I think that this PR helps the issue you point out in the cases where the programmer allows it to?

mwachs5:
"In other words, I think that this PR helps the issue you point out in the cases where the programmer allows it to?" I guess, but the point is programmer shouldn't have to vastly re-write their Chisel code to work with different reset schemes.

jackkoenig:
I think the point is that this PR is improving const prop. Solving the fact that latches can be inferred if such a register is dontTouched should be fixed in a different, follow on PR.

jackkoenig:
Although it looks like @azidar did it here 😝 

@azidar, I wouldn't do this as part of ConstProp. One would hope that we don't emit Verilog that can have latches inferred even if optimizations are disabled.

azidar:
Yeah, I was thinking about that. One thing is that constant prop has some nice functions that @albert-magyar wrote to analyze this case, so its nice to reuse them. But I agree, this should probably be moved to RemoveReset. Emoji for consensus? For now I'll back out the changes for the latch inference to a separate PR.

jackkoenig:
For the logic to emit Verilog that doesn't infer a latch, I think it should be moved to the VerilogEmitter. The FIRRTL is perfectly sensible, it's just stylized Verilog that we're needing to emit.

aswaterman:
I think we should constprop these when don't-touch isn't specified, just for general optimization reasons; and separately make the appropriate changes so we emit better verilog for don't-touch (or for when constprop didn't run).

PR #1359, opened 2020-02-06
Emit 'else' case for trivial-valued async reset regs to avoid latches

albert-magyar:
Type of improvement: backend code generation
API impact: none
Backend code generation impact: emits 'else' case for all async registers when they have trivial non-reset values. This prevents the emission of latch-inferring Verilog in the cases brought up by @mwachs5 in #1355.

The test case I added is a little simplistic, but it fails on master and passes with this PR.

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -563,17 +563,22 @@ class VerilogEmitter extends SeqTransform with Emitter {
             Seq(tabs, "end else ", _falsex.head.tail) +: _falsex.tail
           }
 
-          /* For a Mux assignment, there are five possibilities:
+          /* For a Mux assignment, there are five possibilities, with one subcase for asynchronous reset:
            *   1. Both the true and false condition are self-assignments; do nothing
            *   2. The true condition is a self-assignment; invert the false condition and use that only
-           *   3. The false condition is a self-assignment; skip the false condition
+           *   3. The false condition is a self-assignment
+           *      a) The reset is asynchronous; emit both 'if' and a trivial 'else' to avoid latches
+           *      b) The reset is synchronous; skip the false condition
            *   4. The false condition is a Mux; use the true condition and use 'else if' for the false condition
            *   5. Default; use both the true and false conditions
            */
           (m.tval, m.fval) match {
             case (t, f) if weq(t, r) && weq(f, r) => Nil
             case (t, _) if weq(t, r)              =>  _ifNot +: _false                           :+ _end
-            case (_, f) if weq(f, r)              =>  _if    +: _true                            :+ _end
+            case (_, f) if weq(f, r) => m.cond.tpe match {
+              case AsyncResetType =>                 (_if +: _true     :+ _else)       ++ _true  :+ _end
+              case _ =>                               _if +: _true :+                               _end
+            }
             case (_, _: Mux)                      => (_if    +: _true) ++ _elseIfFalse
             case _                                => (_if    +: _true  :+ _else)       ++ _false :+ _end
           }
--- a/src/test/scala/firrtlTests/AsyncResetSpec.scala
+++ b/src/test/scala/firrtlTests/AsyncResetSpec.scala
@@ -319,6 +319,29 @@ class AsyncResetSpec extends FirrtlFlatSpec {
     )
   }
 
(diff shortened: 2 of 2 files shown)


src/test/scala/firrtlTests/AsyncResetSpec.scala
```diff
@@ -319,6 +319,29 @@ class AsyncResetSpec extends FirrtlFlatSpec {
...
+      |    reg r : UInt<8>, clock with : (reset => (reset, UInt(123)))
+      |""".stripMargin
+    val annos = Seq(dontTouch("m.r")) // dontTouch prevents ConstantPropagation from fixing this problem
+    val result = (new VerilogCompiler).compileAndEmit(CircuitState(parse(withDontTouch), ChirrtlForm, annos))
+    result should containLines (
+       "always @(posedge clock or posedge reset) begin",
+       "if (reset) begin",
+       "r <= 8'h7b;",
+       "end else begin",
+       "r <= r;",
```
jackkoenig: I think what we actually want is the same constant assignment, I would guess that assigning the register to itself here will also infer a latch but I'm not certain

albert-magyar: I can do that. It's a bit const-proppy but if you think that's better I'm down :)

albert-magyar: If they can figure out that it's a self assign and treat it specially, why couldn't they figure out that the original code wasn't a latch 🤣

aswaterman: Really I think we just need to find the thing that makes the linters happy - synthesis tools should constprop any of these (including the old formulation)

And I do know that assigning the constant in both places does make at least one linter happy.

albert-magyar: Does the self-assign also make it happy? The two versions are distinguishable in a waveform, so I'd like to do whatever obeys the principle of least surprise, though I imagine seeing a constant value from a constant register won't surprise people too much.

I will happily do either as long as nobody objects to putting a special form of const-prop in the emitter :)

albert-magyar:
@azidar @jackkoenig, the github traffic storm continues :)

src/test/scala/firrtlTests/AsyncResetSpec.scala
seldridge: I thought the self assignment fixed this as the issue is that there's no "action" if the `posedge clock` path trips. I'd have to actually check though.

albert-magyar: Now it's constant on both sides.

src/main/scala/firrtl/Emitter.scala
```diff
@@ -563,17 +563,24 @@ class VerilogEmitter extends SeqTransform with Emitter {
...
           (m.tval, m.fval) match {
             case (t, f) if weq(t, r) && weq(f, r) => Nil
             case (t, _) if weq(t, r)              =>  _ifNot +: _false                           :+ _end
-            case (_, f) if weq(f, r)              =>  _if    +: _true                            :+ _end
+            case (_, f) if weq(f, r) =>
+              if (m.cond.tpe == AsyncResetType) {
+                (_if +: _true :+ _else) ++ _false :+ _end     // if async reset (case 3a) avoid latches
+              } else {
+                _if +: _true :+ _end                          // else: case 3b
+              }
```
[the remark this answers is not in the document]
albert-magyar: I tried like 10 ways to nest statements to try to minimally mess up the pretty formatting, but I didn't think of that option, which is clearly better :)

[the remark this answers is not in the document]
albert-magyar: I changed it to that style. I think the match statement starting on the same line right after the guard might break up the visual flow, so I could also just curly brace deeper, but that would break it up more vertically.

albert-magyar:
I changed it to symmetric on both sides of the `if`/`else` from the original thread with @aswaterman and @jackkoenig.

There are lots of ways to further refine these patterns (like no `if`/`else` at all, which could even be applied to non-async-reset registers), but the way it is now minimizes the invasiveness of the backend emission changes and the amount of emitter const-prop.

For now, it seems like this emission pattern is the consensus.

PR #1361, opened 2020-02-06
Constant prop binary PrimOps with matching arguments

albert-magyar:
**Type of improvement:** constant prop improvement
**API impact:** none
**Backend code generation impact:** avoids emitting some degenerate operators Verilog

Currently, the case of `<op>(x, x)` for non-literal `x` is not optimized at all. While this could be added to a new pass, it fits pretty nicely into the operator folding of `ConstantPropagation`. The results of all but the `And` case are constant. 

This PR adds this feature and associated test cases.

### DIFF
--- a/src/main/scala/firrtl/transforms/ConstantPropagation.scala
+++ b/src/main/scala/firrtl/transforms/ConstantPropagation.scala
@@ -18,6 +18,11 @@ import annotation.tailrec
 import collection.mutable
 
 object ConstantPropagation {
+  private def litOfType(value: BigInt, t: Type): Literal = t match {
+    case UIntType(w) => UIntLiteral(value, w)
+    case SIntType(w) => SIntLiteral(value, w)
+  }
+
   private def asUInt(e: Expression, t: Type) = DoPrim(AsUInt, Seq(e), Seq(), t)
 
   /** Pads e to the width of t */
@@ -99,14 +104,22 @@ class ConstantPropagation extends Transform with ResolvedAnnotationPaths {
 
   override val annotationClasses: Traversable[Class[_]] = Seq(classOf[DontTouchAnnotation])
 
-  trait FoldCommutativeOp {
+  sealed trait SimplifyBinaryOp {
+    def matchingArgsValue(e: DoPrim, arg: Expression): Expression
+    def apply(e: DoPrim): Expression = {
+      if (e.args.head == e.args(1)) matchingArgsValue(e, e.args.head) else e
+    }
+  }
+
+  sealed trait FoldCommutativeOp extends SimplifyBinaryOp {
     def fold(c1: Literal, c2: Literal): Expression
     def simplify(e: Expression, lhs: Literal, rhs: Expression): Expression
 
-    def apply(e: DoPrim): Expression = (e.args.head, e.args(1)) match {
+    override def apply(e: DoPrim): Expression = (e.args.head, e.args(1)) match {
       case (lhs: Literal, rhs: Literal) => fold(lhs, rhs)
       case (lhs: Literal, rhs) => pad(simplify(e, lhs, rhs), e.tpe)
       case (lhs, rhs: Literal) => pad(simplify(e, rhs, lhs), e.tpe)
+      case (lhs, rhs) if (lhs == rhs) => matchingArgsValue(e, lhs)
       case _ => e
     }
   }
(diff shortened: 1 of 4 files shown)


src/main/scala/firrtl/transforms/ConstantPropagation.scala
```diff
@@ -99,14 +108,22 @@ class ConstantPropagation extends Transform with ResolvedAnnotationPaths {
 
   override val annotationClasses: Traversable[Class[_]] = Seq(classOf[DontTouchAnnotation])
 
-  trait FoldCommutativeOp {
+  sealed trait SimplifyBinaryOp {
```
jackkoenig: Not that renaming a trait is not binary compatible so this isn't backportable

jackkoenig: Oh I see, misread the diff, you didn't remove it, but you did seal it which is probably not binary compatible since someone could extend it. I like sealing it though, maybe on the backport PR we undo the sealing?

jackkoenig: A new virtual method `matchingArgsValue` without a default implementation is definitely not backportable though, so maybe we just hold this improvement off till 1.3

src/main/scala/firrtl/transforms/ConstantPropagation.scala
```diff
@@ -99,14 +108,22 @@ class ConstantPropagation extends Transform with ResolvedAnnotationPaths {
...
+    def matchingArgsValue(e: Expression, arg: Expression): Expression
```
jackkoenig:
```suggestion
    def matchingArgsValue(e: DoPrim, arg: Expression): Expression
```
We know that `e` is always a `DoPrim` right? I think we should use the more precise type if we can.

src/main/scala/firrtl/transforms/ConstantPropagation.scala
```diff
@@ -141,6 +172,7 @@ class ConstantPropagation extends Transform with ResolvedAnnotationPaths {
       case UIntLiteral(v, IntWidth(w)) if v == (BigInt(1) << bitWidth(rhs.tpe).toInt) - 1 => lhs
       case _ => e
     }
+    def matchingArgsValue(e: Expression, arg: Expression) = uIntMax(bitWidth(arg.tpe))
```
jackkoenig: Isn't `A || A == A`?

src/test/scala/firrtlTests/ConstantPropagationTests.scala
```diff
@@ -1393,6 +1393,50 @@ class ConstantPropagationIntegrationSpec extends LowTransformSpec {
...
+    matchingArgs("sub", "SInt<8>", "SInt<8>", """ SInt<8>("h0")  """ )
+    matchingArgs("div", "UInt<8>", "UInt<8>", """ UInt<8>("h1")  """ )
+    matchingArgs("div", "SInt<8>", "SInt<8>", """ SInt<8>("h1")  """ )
+    matchingArgs("rem", "UInt<8>", "UInt<8>", """ UInt<8>("h0")  """ )
+    matchingArgs("rem", "SInt<8>", "SInt<8>", """ SInt<8>("h0")  """ )
+    matchingArgs("and", "UInt<8>", "UInt<8>", """ i              """ )
+    matchingArgs("and", "SInt<8>", "UInt<8>", """ asUInt(i)      """ )
+    // Signedness doesn't matter
+    matchingArgs("or",  "UInt<8>", "UInt<8>", """ UInt<8>("hff") """ )
+    matchingArgs("or",  "SInt<8>", "UInt<8>", """ UInt<8>("hff") """ )
```
jackkoenig: Result should be `i` right?

src/main/scala/firrtl/transforms/ConstantPropagation.scala
albert-magyar: Yes. I think I made it `Expression` just to not differ from `FoldCommutativeOp`, which I couldn't fix for backport reasons, but that's a bad reason.

src/main/scala/firrtl/transforms/ConstantPropagation.scala
albert-magyar: It _does_ have a default implementation in the trait that already existed before (`FoldCommutativeOp`).  The weird diff is not helping :)

src/test/scala/firrtlTests/ConstantPropagationTests.scala
albert-magyar: True, thanks :)

src/main/scala/firrtl/transforms/ConstantPropagation.scala
jackkoenig: Oh yes, I see it, thanks!

Issue #1344, opened 2020-01-29
Emit Correct Zero-Width Reduction Operators


seldridge:
### Checklist

- [x] Did you specify the current behavior?
- [x] Did you specify the expected behavior?
- [x] Did you provide a code example showing the problem?
- [x] Did you describe your environment?
- [x] Did you specify relevant external information?

### What is the current behavior?

Verilog emission will emit `andr`, `orr`, and `xorr` incorrectly for zero-width things.

Consider the following FIRRTL IR:

```
circuit Foo_0 :
  module Foo_0 :
    input in : UInt<0>
    output outAnd : UInt<1>
    output outOr : UInt<1>
    output outXor : UInt<1>

    node _T = andr(in)
    outAnd <= _T
    node _T_1 = orr(in)
    outOr <= _T_1
    node _T_2 = xorr(in)
    outXor <= _T_2
```

This produces the following Verilog:

```
module Foo_0(
  output  outAnd,
  output  outOr,
  output  outXor
);
  assign outAnd = &1'h0;
  assign outOr = |1'h0;
  assign outXor = ^1'h0;
endmodule
```

Above, `or` and `xor` are correct, but `and` is wrong.

### What is the expected behavior?

Any reduction operation on a zero-width thing should be as follows:

- `andr` -> `1`
- `orr` -> `0`
- `xorr` -> `0`

Why is this? 

Even though `andr` is a "reduction", this is actually a `fold` with an initial value of `1`. Similarly, `orr` and `xorr` are `fold`s with initial values of `0`. This should be fixed in the Verilog emitter and clarified in documentation and the spec that this is what is going on.

### Steps to Reproduce

See example above.

### Your environment

`master`

### External Information

https://github.com/freechipsproject/chisel3/pull/1315#issuecomment-579590423

PR #1362, opened 2020-02-06
Change zero-width base case for Andr

albert-magyar:
Type of Improvement: operator semantics
API impact: changes behavior of user code!
Backend code generation impact: andr of zero-width wires now emits `1'b1`.

This fixes the issue from #1344. Since the behavior flagged as undesirable in that issue gets introduced in `ZeroWidth`, and since it's possible to observe it before Verilog emission, I fixed it there rather than in the emitter.

This change pattern-matches an `Andr` with a zero-width argument. It also makes one optimization, as it replaces it with `UIntLiteral(1)` rather than the most literal substitution of `DoPrim(Andr, Seq(UIntLiteral(1)), Nil, BoolType)`.


albert-magyar:
I updated the spec with some commentary on the inductive base cases.

As far as the other reductions go, they get their values as a consequence of the fact that literally all zero-width expressions in the design that are not children of a `Cat` are brute-force replaced with either `UIntLiteral(0)` or `SIntLiteral(0)`.

I agree with you that explicitly putting the behavior of the the other reduction patterns in might help with clarity. However, this argument also applies to the other non-width-preserving operations like `Eq`, `Neq`, so it might make sense to add explicit patterns for all of these if not just for `Andr`.

I imagine the original conception was that substituting 1-bit literals as "singular values" in place of all the zero-width operands was sufficient for those, but it kind of hidden in the implicit behavior of the whole pass.

albert-magyar:
There is another operator that we should think about: `shr` of a signed zero-width value must return a one-bit signal that contains the sign bit. Currently it returns zero 🤷‍♂ 

That one feels like it definitely deserves an explicit pattern in the transform source.

seldridge:
Note: I only actually checked this now... but there is a slight generation difference between `andr` and `orr`/`xorr`:

```verilog
  assign out_0 = 1'h1; 
  assign out_1 = |1'h0;
  assign out_2 = ^1'h0;
```

I think cleaning all of this, and the additional operators you mention @albert-magyar, would make sense in a follow-up PR. This shouldn't be hurting anything as-is, though.

PR #1052, opened 2019-03-11
More constprop on muxes

donggyukim:
Two rules discussed in #1049.

PR #1366, opened 2020-02-07
Better register const prop through speculative de-optimization

### LINKED ISSUE #1240: Registers not constpropped away, perhaps due to Mux optimizations

The register in the following circuit does not disappear, even though it can be replaced with the constant 1 (because at all times it is either unspecified or 1).

I suspect it is due to the mux optimizations merged a while ago.  Register constprop only understands muxes, not other combinational logic.  The mux optimizations turn `r <= en ? 1'b1 : r` into `r <= en | r`, foiling register constprop.

The register constprop is the more important of the two, since the mux constprop is basically just cosmetic.  So disabling the latter might be an acceptable interim workaround if it will be too time-consuming to make register constprop more general.

```
circuit Test :
  module Test :
    input clock : Clock
    input reset : UInt<1>
    input en : UInt<1>
    output out : UInt<1>

    reg r : UInt<1>, clock

    when en :
      r <= UInt<1>(1)

    out <= r
```

We should get `out <= 1'b1;`

### Your environment

Firrtl 3.2.0 (commit f738fbe8667ed6b76ec00a15960b9c3a42b8654a)

albert-magyar:
Type of improvement: const prop
API impact: none
Backend code generation impact: fixes the negative impact on register const prop introduced by #1052 

* Fixes #1240
* Add failing reg const prop test case from #1240

**Summary:**

Some optimization of Mux trees turn 1-bit mux operators into boolean operators. This can stifle register constant propagations, which looks at drivers through value-preserving `Mux`es and `Connect`s only. By speculatively expanding some 1-bit Or and And operations into `Mux`es, we can obtain better insight on the true possible values of the register with a simple peephole de-optimization.

N.B. this de-optimized code is never actually added to the design.

### DIFF
--- a/src/main/scala/firrtl/transforms/ConstantPropagation.scala
+++ b/src/main/scala/firrtl/transforms/ConstantPropagation.scala
@@ -490,6 +490,12 @@ class ConstantPropagation extends Transform with ResolvedAnnotationPaths {
           * NonConstant values.  When encountering a node reference, it expands the node by to its
           * RHS assignment and recurses.
           *
+          * @note Some optimization of Mux trees turn 1-bit mux operators into boolean operators. This
+          * can stifle register constant propagations, which looks at drivers through value-preserving
+          * Muxes and Connects only. By speculatively expanding some 1-bit Or and And operations into
+          * muxes, we can obtain the best possible insight on the value of the mux with a simple peephole
+          * de-optimization that does not actually appear in the output code.
+          *
           * @return a RegCPEntry describing the constant prop-compatible sources driving this expression
           */
           def regConstant(e: Expression): RegCPEntry = e match {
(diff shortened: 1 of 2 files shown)


aswaterman:
Thank you!

src/main/scala/firrtl/transforms/ConstantPropagation.scala
```diff
@@ -498,6 +504,8 @@ class ConstantPropagation extends Transform with ResolvedAnnotationPaths {
             case WRef(nodeName, _, NodeKind, _) if nodeMap.contains(nodeName) =>
               nodeRegCPEntries.getOrElseUpdate(nodeName, { regConstant(nodeMap(nodeName)) })
             case Mux(_, tval, fval, _) => regConstant(tval).resolve(regConstant(fval))
+            case DoPrim(Or, Seq(a, b), Nil, BoolType) => regConstant(Mux(a, one, b, BoolType))
+            case DoPrim(And, Seq(a, b), Nil, BoolType) => regConstant(Mux(a, b, zero, BoolType))
```
jackkoenig: Should we also be checking the alternative forms? eg. `Mux(b, one, a, ...)`?

albert-magyar: It checks these two since mux-const-prop puts them in that way. There's not really a great way to express "it's either one of these" without refactoring how reg const prop works entirely.

I'll add more cases that figure out which argument is a register. Since these are the only "special" arguments, this will make accomplish matching the alternative forms. I agree that this is a good idea since it decouples it from the `Mux` const prop implementation at the cost of two lines of code.

albert-magyar: I refactored the `regConstant` function to make it easier to try multiple orderings. This will also have the side effect of possibly optimizing away some **user** code that has similar `And`s and `Or`s.

PR #1387, opened 2020-02-12
Include Dead Modules in InstanceGraph.staticInstanceCount

seldridge:
This changes the API of `InstanceGraph.staticInstanceCount` to include counts of modules never instantiated (dead modules). Previously, dead modules were not included at all in the count leading to some confusion on my part.

I assumed that this was what the API was supposed to be as this allows a user to differentiate between modules not instantiated (modules with a count of zero) and modules which don't exist (modules which are not in the key list). *Therefore, I'm treating this as a bug-fix.* However, I can see a situation where the old behavior is also reasonable. I hit this when I was trying to use `staticInstanceCount` to find dead modules and was surprised at the current behavior.

@albert-magyar: what is the desired API in your view, here?

### Contributor Checklist

### Type of Improvement

- bug fix

- new feature/API

### API Impact

This changes the behavior of `InstanceGraph.staticInstanceCount` to return counts of modules *not* instantiated. This is technically an API change.

None.

### Desired Merge Strategy

- [x] Did you review?
- ~~[ ] Did you mark as `Please Merge`?~~

### DIFF
--- a/src/main/scala/firrtl/analyses/InstanceGraph.scala
+++ b/src/main/scala/firrtl/analyses/InstanceGraph.scala
@@ -62,16 +62,22 @@ class InstanceGraph(c: Circuit) {
     */
   lazy val fullHierarchy: mutable.LinkedHashMap[WDefInstance,Seq[Seq[WDefInstance]]] = graph.pathsInDAG(trueTopInstance)
 
-  /** A count of the *static* number of instances of each module. For any module
-    * other than the top (main) module, this is equivalent to the number of inst
-    * statements in the circuit instantiating each module, irrespective of the
-    * number of times (if any) the enclosing module appears in the hierarchy.
-    * Note that top module of the circuit has an associated count of 1, even
-    * though it is never directly instantiated.
+  /** A count of the *static* number of instances of each module. For any module other than the top (main) module, this is
+    * equivalent to the number of inst statements in the circuit instantiating each module, irrespective of the number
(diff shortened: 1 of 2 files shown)


albert-magyar:
FWIW, this is just used for `ConstantPropagation` in any current FIRRTL/Chisel/Chipyard/Rocket Chip source, and it is only used to check counts for modules for which it is guaranteed there are instance, so this change will only help with corner cases.

seldridge:
There may be one other oddity here that should maybe be fixed:

```
circuit Top:
  module Dead:
    skip
  module Beef:
    inst dead of Dead
    skip
  module Top
    skip
```

Currently, this will report: 
```
Dead -> 1
Beef -> 0
Top -> 1
```

I think this should be:
```
Dead -> 0
Beef -> 0
Top -> 1
```

Thoughts @albert-magyar?

albert-magyar:
That is by design. The instance count that is returned does not depend on
the number of times the parent is instantiated. The only exception is Top.

seldridge:
I think that makes sense. This is really counting inbound edges. I can certainly see why it shouldn't be multiplicative as that doesn't represent the graph structure. 

I think the exception of `Top` is really that there is an edge from the outer circuit. So, maybe that isn't as much of an exception as it seems.

For more info: I'm abusing this method to find all dead modules, but that's not what it is intended to be used for. 

Do you think that a `unreachableModules` would be an API that `InstanceGraph` should include or would that be better to include/prototype privately in transforms as needed?

PR #1391, opened 2020-02-13
Add InstaceGraph (Un)?Reachable Helpers

seldridge:
Add helper methods for InstanceGraph to query:

1. The set of all modules
1. The set of modules reachable from the main module
1. The set of modules unreachable from the main module

### Contributor Checklist

- [x] Did you add Scaladoc to every public function/method?
- [x] Did you add at least one test demonstrating the PR?
- [x] Did you delete any extraneous printlns/debugging code?
- [x] Did you specify the type of improvement?
- [x] Did you state the API impact?
- [x] Did you specify the code generation impact?
- [x] Did you request a desired merge strategy?

### Type of Improvement

  - new feature/API

### API Impact

None.

### Backend Code Generation Impact

None.

### Desired Merge Strategy

- Rebase: You will rebase the PR onto master and it will be merged with a merge commit.

### Reviewer Checklist (only modified by reviewer)
- [ ] Did you add the appropriate labels?
- [ ] Did you mark the proper milestone (1.2.x, 1.3.0, 1.4.0) ?
- [ ] Did you review?
- [ ] Did you mark as `Please Merge`?

PR #1394, opened 2020-02-13
EliminateTargetPaths and Unreachable Modules

### LINKED ISSUE #1380: Duplication Leaves Dead Modules
### Checklist

If you duplicate a module, you may wind up with a dead module with no references to it. This can cause problems if you use the duplication/deduplication pattern to remove something that is invalid FIRRTL or that a downstream transform cannot handle.

There are two situations to consider here:

1. A module shouldn't be duplicated if it's the only instance in a design
1. If all instances are duplicated, then the original module can be removed (or used as one of the duplicated instances)

Duplication should remove modules without references. However, duplication should not remove modules that previously had no references...

### Steps to Reproduce

```
circuit Foo:
  module Baz:
    skip
  module Bar:
    inst baz of Baz
    skip
  module Foo:
    inst bar of Bar
```

Try to duplicate `~Foo|Foo/bar:Bar/baz:Baz` and you wind up with:

```
circuit Foo:    
  module Baz___Foo_bar_baz :
    skip
  module Baz :
    skip
  module Bar___Foo_bar :
    inst baz of Baz___Foo_bar_baz
  module Foo :
    inst bar of Bar___Foo_bar
```

The module `Baz` is left around when it shouldn't be.

### Your environment

- Chisel Verions: 
- OS: 
- Verilator version: 

### External Information

None.

seldridge:
This modifies `EliminateTargetPaths` (module duplication) such that all newly unreachable modules are removed (See #1380). 

This broadly fixes the problem of using the dup/dedup pattern for `InferResets`. Namely, if `InferResets` duplicates a module which is unreachable that has an abstract reset, this will fail in the `VerilogEmitter` during emission.

### Dependencies

- Prerequisites: [#1391, #1392]
- Fixes #1380 

### Contributor Checklist

### API Impact

None, but this requires the API-extending #1391 (which is 1.3.0). Therefore, this is 1.3.0.

None.

### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/annotations/transforms/EliminateTargetPaths.scala
+++ b/src/main/scala/firrtl/annotations/transforms/EliminateTargetPaths.scala
@@ -5,7 +5,7 @@ package firrtl.annotations.transforms
 import firrtl.Mappers._
 import firrtl.analyses.InstanceGraph
 import firrtl.annotations.ModuleTarget
-import firrtl.annotations.TargetToken.{Instance, OfModule}
+import firrtl.annotations.TargetToken.{Instance, OfModule, fromDefModuleToTargetToken}
 import firrtl.annotations.analysis.DuplicationHelper
 import firrtl.annotations._
 import firrtl.ir._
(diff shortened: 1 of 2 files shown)


azidar:
So the original reason for the code you deleted was to preserve existing modules in circuit which are never referenced, but to delete newly added modules which are ultimately never instantiated in the transform. I think your changes would mean that modules in the input circuit which are not references will be deleted.

seldridge:
I can double check on this, but this is not intending to change any old behavior. 

This is fixing the different problem where the original module is left around when:

1. The only instance is duplicated
1. All instances are duplicated

In either of these situations, you wind up with a dead module.

seldridge:
Specifically, the removed modules are [`newUnreachable diff oldUnreachable`](https://github.com/freechipsproject/firrtl/pull/1394/files#diff-240ff27fedb90ffe7f1d5cf3cf9b32c3R161), which I think is the intended behavior.

jackkoenig:
I'm going to backport https://github.com/freechipsproject/firrtl/pull/1391 then this because it's started coming up a lot. If anyone cares about the new methods in https://github.com/freechipsproject/firrtl/pull/1391 I can make them `private[firrtl]`, but I don't think it really matters that much

seldridge:
Sounds good to me. I am fine with the straight backport of both. They're pure API expansions and I don't think we need to jump through hoops her for the sake of backport purity.

jackkoenig:
@Mergifyio backport 1.2.x

Issue #1376, opened 2020-02-10
RFC: Restructuring InferResets to be an optimization.


albert-magyar:
Many of the most powerful things done in the FIRRTL compiler are optimizations. I believe that this is due to two strong features guaranteed by the definition of an "optimization" pass:
1. A strong as-if correctness model that allows for behavior refinements. In FIRRTL, this as-if model is currently assumed to be "of the whole circuit unless modified by `dontTouch`".
2. The requirement that the unoptimized base case be a functionally correct default implementation for all inputs
3. Some general metric or notion of "improving" an implementation

## Benefits and drawbacks

Implementing `InferResets` as an optimization would allow it to be **repeatedly** run in the compiler and make it easy to move in the transform order, since it would have no obligation to exhaustively classify the resets in the input. Additionally, it would simplify its implementation, since a valid implementation could decide to "punt" on tricky cases until they are simplified with `ConstantPropagation`, `ExpandWhens`, or other passes.

However, converting `InferResets` to an optimization would create a new requirement: that there exist a default implementation that can be passed through all stages of the compiler to generate functional code with (potentially) worse output quality metrics.

In order to fulfill this requirement, there are two main implementation patterns. In each case, any notion of "subtyping" is taken under the model that a `Connect` can be modeled as an assignment and that a `PartialConnect`, `Connect` in a `Conditionally` or `Mux` can be modeled as a conditional assignment. Any signal is assumed to be be a supertype of the least upper bound of all of the types of the signals it is (conditionally) connected to. At any point in compilation, all connections in the circuit **must be legal**.

## Case 1: Retaining abstract reset

In the most direct porting of the current code, making `InferResets` an optimization would involve allowing `Reset` to remain in the design while inferrable resets are selectively optimized to `UInt<1>` or `AsyncReset`. However, this would require defining behavior for `Reset` that allows it to act as a safe default in all cases. Given that it must support all behaviors, the natural behavior for such a "safe default" would mirror that of `AsyncReset`, causing asynchronously reset registers to be inferred conservatively.

Under this scheme, the following code would produce an asynchronously-reset register without running `InferReset`:
```
input clock: Clock
input sr: UInt<1>
wire r: Reset
reg r: UInt<1>, clock with: (reset => (r, UInt<1>(0)))
```

## Case 2: UInt<1> <: AsyncReset

[...]

PR #1396, opened 2020-02-15
Remove last connect semantics from reset inference

jackkoenig:
**TLDR**
* Reverts support for last connect semantics in reset inference (reverts https://github.com/freechipsproject/firrtl/pull/1291, note this was never released)
* Makes reset inference a forward and backward propagation
    * This mainly catches user error in InferResets instead of CheckTypes and can change the type of wires or ports that are **only** invalidated but drive stuff (not necessarily the winning connection though). Note that this case crashes in `1.2.2`.
* Fixes bugs
    * Handling of invalidation on components of type `Reset`
    * Connections or invalidation of submodule ports could override everything done inside the module itself
    * Support AsyncResetType in RemoveValidIfs
* This is orthogonal from possible improvements to the feature like https://github.com/freechipsproject/firrtl/issues/1376. This makes the `1.2.x` branch work in a reasonable way.

**Note** this looks big, but the positive line change is almost entirely new tests

### Contributor Checklist

### Type of Improvement

- bug fix
- new feature/API

### API Impact

This modifies the reset inference API. It reverts support for last connect semantics in inference. This made it to `1.2.x`, but has not yet been released.

This also fixes many bugs present in the current latest release `1.2.2`, especially with regards to invalidation and inference. It does make a change from `1.2.2`, namely that invalidated components that drive another component that itself is driven by something else may get a different type. **While this may seem like cause for concern, as far as I can tell, every case of this crashes on v1.2.2**; not to mention that such components are only invalidated so their types don't really have much impact on anything.

I think the best way to understand this PR is with the new tests:

1. 
```scala
  it should "infer based on what a component *drives* not just what drives it" in {
    val result = compile(s"""
      |circuit top :
      |  module top :
      |    input in : AsyncReset
      |    output out : Reset
      |    wire w : Reset
      |    w is invalid
      |    out <= w
      |    out <= in
      |""".stripMargin)
    result should containTree { case DefWire(_, "w", AsyncResetType) => true }
  }
```
This first one above is the "API change" that technically would differ from `1.2.2`, but it crashes in `1.2.2`. My argument is that it's not an API that means anything because it only matters in the context of components that are **only ever invalidated**, thus the reset should be optimized away anyway, and if optimizations are disabled, the flop will never be reset.

2.
```scala
  it should "infer from connections, ignoring \"winning\" invalidation" in {
    val result = compile(s"""
      |circuit top :
      |  module top :
      |    input in : AsyncReset
      |    output out : Reset
      |    out <= in
      |    out is invalid
      |""".stripMargin)
    result should containTree { case Port(_, "out", Output, AsyncResetType) => true }
  }
```
With the unreleased last connect semantics, `out` would be a `UInt<1>` instead of `AsyncReset`. Note that as above, it is invalidated so it doesn't do anything. Also note that this crashes in `1.2.2`.

3. 
```scala
  // The backwards type propagation constrains `w` to be the same as both `out0` and `out1`
  it should "not allow an invalidated Wire to drive both a UInt<1> and an AsyncReset" in {
    an [InferResets.InferResetsException] shouldBe thrownBy {
      val result = compile(s"""
        |circuit top :
        |  module top :
        |    input in0 : AsyncReset
        |    input in1 : UInt<1>
        |    output out0 : Reset
        |    output out1 : Reset
        |    wire w : Reset
        |    w is invalid
        |    out0 <= w
        |    out1 <= w
        |    out0 <= in0
        |    out1 <= in1
        |""".stripMargin
      )
    }
  }
```
This also crashes on `v1.2.2`. It would be legal if last connect semantics were accepted, but due to backwards type inference introduced in this PR, `w` is not allowed to drive both a `UInt<1>` and an `AsyncReset`.

4. 

```scala
  // This tests for a bug unrelated to support or lackthereof for last connect in inference
  it should "take into account both internal and external constraints on Module port types" in {
    val result = compile(s"""
      |circuit top :
      |  module child :
      |    input i : AsyncReset
      |    output o : Reset
      |    o <= i
      |  module top :
      |    input in : AsyncReset
      |    output out : AsyncReset
      |    inst c of child
      |    c.o is invalid
      |    c.i <= in
      |    out <= c.o
      |""".stripMargin)
    result should containTree { case Port(_, "o", Output, AsyncResetType) => true }
  }
```
Unrelated bugfix that can lead to very strange behavior.

5.

```scala
  it should "not crash on combinational loops" in {
    an [CheckCombLoops.CombLoopException] shouldBe thrownBy {
      val result = compile(s"""
        |circuit top :
        |  module top :
        |    input in : AsyncReset
        |    output out : Reset
        |    wire w0 : Reset
        |    wire w1 : Reset
        |    w0 <= in
        |    w0 <= w1
        |    w1 <= w0
        |    out <= in
        |""".stripMargin,
        compiler = new LowFirrtlCompiler
      )
    }
  }
```
This crashes on `1.2.2` and `1.2.x` alike.

It could change the types of some registers, but only if the reset signal **is invalid**.

### Desired Merge Strategy

The history is ugly but I want it on the PR to keep some context. This should be squashed so that it doesn't show up in the actual git history.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/transforms/InferResets.scala
+++ b/src/main/scala/firrtl/transforms/InferResets.scala
@@ -36,7 +36,11 @@ object InferResets {
   private case class TypeDriver(tpe: Type, target: () => ReferenceTarget) extends ResetDriver {
     override def toString: String = s"TypeDriver(${tpe.serialize}, $target)"
   }
-
+  // When a [[ResetType]] is invalidated, we record the InvalidDrive
+  // If there are no types but invalid drivers, we default to BoolType
+  private case object InvalidDriver extends ResetDriver {
+    def defaultType: Type = Utils.BoolType
+  }
 
   // Type hierarchy representing the path to a leaf type in an aggregate type structure
   // Used by this [[InferResets]] to pinpoint instances of [[ResetType]] and their inferred type
@@ -136,6 +140,16 @@ class InferResets extends Transform {
             for ((i, j) <- points) {
               markResetDriver(locs(i), exps(j))
             }
+          case IsInvalid(_, lhs) =>
+            val exprs = Utils.create_exps(lhs)
+            for (expr <- exprs) {
+              // Ignore leaves that are not of type ResetType
+              // Unlike in markResetDriver, flow is irrelevant for invalidation
+              if (expr.tpe == ResetType) {
+                val target = makeTarget(expr)
+                map.getOrElseUpdate(target, mutable.ListBuffer()) += InvalidDriver
+              }
+            }
           case WDefInstance(_, inst, module, _) =>
             instMap += (inst -> module)
           case Conditionally(_, _, con, alt) =>
@@ -143,12 +157,10 @@ class InferResets extends Transform {
             val altMap = new DriverMap
             onStmt(conMap)(con)
             onStmt(altMap)(alt)
-            // Default to outerscope if not found in alt
-            val altLookup = altMap.orElse(map).lift
             for (key <- conMap.keys ++ altMap.keys) {
               val ds = map.getOrElseUpdate(key, mutable.ListBuffer())
               conMap.get(key).foreach(ds ++= _)
-              altLookup(key).foreach(ds ++= _)
+              altMap.get(key).foreach(ds ++= _)
             }
           case other => other.foreach(onStmt(map))
         }
@@ -167,17 +179,22 @@ class InferResets extends Transform {
     val res = mutable.Map[ReferenceTarget, Type]()
     val errors = new Errors
     def rec(target: ReferenceTarget): Type = {
-      val drivers = map(target)
+      val drivers = map.getOrElse(target, Nil)
       res.getOrElseUpdate(target, {
-        val tpes = drivers.map {
-          case TargetDriver(t) => TypeDriver(rec(t), () => t)
-          case td: TypeDriver => td
+        val tpes = drivers.flatMap {
+          case TargetDriver(t) => Some(TypeDriver(rec(t), () => t))
+          case td: TypeDriver  => Some(td)
+          case InvalidDriver   => None
         }.groupBy(_.tpe)
-        if (tpes.keys.size != 1) {
-          // Multiple types of driver!
-          errors.append(DifferingDriverTypesException(target, tpes.toSeq))
+        tpes.keys.size match {
+          // This can occur if something of type Reset has no driver
+          case 0 => InvalidDriver.defaultType
+          case 1 => tpes.keys.head
+          case _ =>
+            // Multiple types of driver!
+            errors.append(DifferingDriverTypesException(target, tpes.toSeq))
+            tpes.keys.head
         }
-        tpes.keys.head
       })
     }
     for ((target, _) <- map) {
--- a/src/main/scala/firrtl/transforms/InferResets.scala
+++ b/src/main/scala/firrtl/transforms/InferResets.scala
@@ -15,7 +15,9 @@ import scala.collection.mutable
 import scala.util.Try
 
 object InferResets {
+  @deprecated("This is no longer in use and will be removed", "1.3")
   final class DifferingDriverTypesException private (msg: String) extends PassException(msg)
+  @deprecated("This is no longer in use and will be removed", "1.3")
   object DifferingDriverTypesException {
     def apply(target: ReferenceTarget, tpes: Seq[(Type, Seq[TypeDriver])]): DifferingDriverTypesException = {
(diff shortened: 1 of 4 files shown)


src/test/scala/firrtlTests/InferResetsSpec.scala
```diff
@@ -410,5 +405,90 @@ class InferResetsSpec extends FirrtlFlatSpec {
...
+      |    output out : Reset
+      |    wire w : Reset
+      |    w is invalid
+      |    out <= w
+      |    out <= in
+      |""".stripMargin)
+    result should containTree { case DefWire(_, "w", AsyncResetType) => true }
+  }
+
+  it should "infer from connections, ignoring \"winning\" invalidation" in {
```
albert-magyar: Maybe reword this as "ignoring the fact that the invalidation wins"?

jackkoenig: yeah that's better

src/test/scala/firrtlTests/InferResetsSpec.scala
```diff
@@ -410,5 +405,90 @@ class InferResetsSpec extends FirrtlFlatSpec {
...
+      |    inst c of child
+      |    c.o is invalid
+      |    c.i <= in
+      |    out <= c.o
+      |""".stripMargin)
+    result should containTree { case Port(_, "o", Output, AsyncResetType) => true }
+  }
+
+  it should "not crash on combinational loops" in {
+    an [CheckCombLoops.CombLoopException] shouldBe thrownBy {
```
seldridge:
```suggestion
    a [CheckCombLoops.CombLoopException] shouldBe thrownBy {
```

src/test/scala/firrtlTests/InferResetsSpec.scala
```diff
@@ -410,5 +405,109 @@ class InferResetsSpec extends FirrtlFlatSpec {
...
+      |    output out : Reset
+      |    wire w : Reset
+      |    w is invalid
+      |    out <= w
+      |    out <= in
+      |""".stripMargin)
+    result should containTree { case DefWire(_, "w", AsyncResetType) => true }
+  }
+
+  it should "infer from connections, ignoring \"winning\" invalidation" in {
```
albert-magyar:
```suggestion
  it should "infer from connections, ignoring the fact that the invalidation wins" in {
```

src/main/scala/firrtl/transforms/InferResets.scala
```diff
@@ -156,54 +177,53 @@ class InferResets extends Transform {
...
+          graph.addPairWithEdge(v, u)
+          graph.addPairWithEdge(u, v)
+        case TypeDriver(tpe, _) =>
+          // Use nodes we already made, saves memory
+          val u = tpe match {
+            case AsyncResetType => asyncNode
+            case Utils.BoolType => syncNode
+            case other          => throwInternalError(s"Shouldn't have $other here")
+          }
+          graph.addPairWithEdge(u, v)
```
albert-magyar: This "fake" edge to the type seems like a clever way to share the work for the inference with the check _without_ interspersing check code into the connection edge creation.

src/main/scala/firrtl/transforms/InferResets.scala
```diff
@@ -156,54 +177,53 @@ class InferResets extends Transform {
...
+          graph.addPairWithEdge(v, u)
+        case InvalidDriver =>
+          graph.addVertex(v)   // Must be in the graph or won't be inferred
+      }
     }
+    val async = graph.reachableFrom(asyncNode)
+    val sync = graph.getVertices -- async
```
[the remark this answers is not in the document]
jackkoenig: I'll write a comment how it works because it may be non-obvious in a future fix/refactor.

src/main/scala/firrtl/transforms/InferResets.scala
```diff
@@ -156,54 +177,53 @@ class InferResets extends Transform {
...
     Try {
-      errors.trigger()
-      res.toMap
+      if (async.contains(syncNode)) throw InferResetsException(graph.path(asyncNode, syncNode))
+      ( sync.view.collect { case Var(t) => t -> syncNode.tpe } ++
+       async.view.collect { case Var(t) => t -> asyncNode.tpe }).toMap
```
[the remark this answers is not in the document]
jackkoenig: I like that better, I'll do that

jackkoenig:
Once this passes tests, I'm gonna :shipit: (but will squash and merge it myself to cleanup commit message)

PR #1399, opened 2020-02-18
Revert "Repl seq mem renaming"

jackkoenig:
Reverts freechipsproject/firrtl#1286

Unfortunately this causes issues in rocket-chip for the `SRAMAnnotation` which uses `Named` (https://github.com/chipsalliance/rocket-chip/blob/4f0cdea85c8a2b849fd582ccc8497892001d06b0/src/main/scala/util/Annotations.scala#L19).

While this fix is needed, it needs to be part of a larger fix to the whole Target and renaming system.


jackkoenig:
With this merged, we should be a go on `1.2.3` @ucbjrl, and we should be getting close on `1.3.0` along with @seldridge's changes.

PR #1407, opened 2020-02-21
sbt compatible publish for mill

sequencer:
### Contributor Checklist

### Type of Improvement
- bug fix

### API Impact
None
None
### Desired Merge Strategy

- [ ] Did you review?

Currently `mill` will only publish in `~/.ivy2/local/edu.berkeley.cs/firrtl-2.12.10_2.12/1.3-SNAPSHOT/`, which is non-compatible to sbt build.
This patch will overwrite `artifactName` from `firrtl-$crossVersion_$artifactScalaVersion` to `firrtl-$artifactScalaVersion`, giving the same directory with sbt.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -122,4 +122,6 @@ class firrtlCrossModule(crossVersion: String) extends ScalaModule with SbtModule
       Developer("jackbackrack", "Jonathan Bachrach", "https://eecs.berkeley.edu/~jrb/")
     )
   )
+  // make mill publish sbt compatible package
+  def artifactName = "firrtl"
 }


jackkoenig:
While this looks fine, I think it highlights a potentially larger issue. Why is mill publishing the artifact as `firrtl-2.12.10_2.12`? I get that it's a cross module but there is semantic meaning in that String, mainly that `firrtl_2.12` is FIRRTL publishes against Scala 2.12, the `2.12.10` is (to my understanding) nonstandard. Is this something that other `mill` users have run into and solved this way?

sequencer:
@jackbackrack I look back into mill souces:
`_2.12` is defined for `artifactSuffix`,which is based on `scalaVersion`, it can be changed to `_2.12.10` with `def crossFullScalaVersion = true`,
`firrtl-2.12.10` is from [`artifactName`](https://github.com/lihaoyi/mill/blob/c7873b6f2d977324f215209854a1f415e2882622/scalalib/src/JavaModule.scala#L559-L564), which is from `millOuterCtx.segments ++ Seq(millOuterCtx.segment)`. It means split user commands and join them with '-'. Thus it leads to `CrossModule` parameters exist in `artifactName`.
Maybe this behavior can help Cross in small version?(but open `crossFullScalaVersion` can also do this).
`utest` from lihaoyi also has this behavior:
https://github.com/lihaoyi/utest/blob/9067c938660831d2cbe7b73ef388256a728ba0f0/build.sc#L5

Emmm, I'm not sure what's the design purpose here.

jackkoenig:
Interesting, my guess is it's a wart but who knows. The fact that `utest` has the same thing convinces me that this is the right solution.

PR #1412, opened 2020-02-27
[RFC] Factor-out common test classes; package them

davidbiancolin:
We write a ton of FIRRTL passes in Golden Gate & FireSim. It would be awesome if we could test those passes using the same infrastucture as FIRRTL itself uses -- this would both reduce code duplication and provide a simpler path for upstreaming passes to FIRRTL. This PR is a first pass at factoring out some of the key classes to do this. 
(Because of the way we manage the firrtl dependency in Chipyard/Firesim/Rocket Chip, unfortunately i can't just add a managed dependency on firrtl's test sources.)

I've basically just moved FirrtlSpec.scala and PassTests.scala files into `src/test/scala/firrtl/testutills`, and named the new package firrtl.testutils
Otherwise i have made no attempt to do any class-level refactoring. 

Outstanding questions:
- I'm pretty sure we don't want to assemble these into a fat jar. Should we just collect the classes belonging to `firrtl.testutils` or also include some other depedencies in the jar.
- Should these be published too?
- What other classes should be factored out?

### Contributor Checklist

### Type of Improvement
- code refactoring

### API Impact

Breaking: changes location of common test classes.

None

### Desired Merge Strategy
Squash

- [x] Did you review?


albert-magyar:
There is currently waaaay too much stuff in the top level `firrtlTests` package, so I think anything to better organize that is great for that reason as well.

albert-magyar:
I can review the changes today.

albert-magyar:
I pulled in the upstream changes with `Dependency` stuff.

build.sbt
```diff
@@ -91,6 +91,21 @@ test in assembly := {} // Should there be tests?
...
+  case PathList("firrtlTests", xs @ _*) => MergeStrategy.discard
+  case x =>
+    val oldStrategy = (assemblyMergeStrategy in (Test, assembly)).value
+    oldStrategy(x)
+}
+
+assemblyJarName in (Test, assembly) := s"firrtl-test.jar"
+
+assemblyOutputPath in (Test, assembly) := file("./utils/bin/" + (Test / assembly / assemblyJarName).value)
+
```
jackkoenig: Just stripping the package `firrtlTests` still includes other tests: `https://github.com/freechipsproject/firrtl/tree/master/src/test/scala/firrtl`

I'm okay with this approach since the utility of this outweighs blocking on a perfect solution, but I hope we can eventually make this a multi-project build such that we can also publish the `testutils` package to maven.

ucbjrl:
Should we add the `Please Merge` label?

albert-magyar:
I think we're basically waiting on a milestone tag -- this breaks the API but only for test, which is not part of the old jar, so it's not part of our binary compatibility constraint.

albert-magyar:
@ucbjrl this is currently tagged for 1.3.0 -- do you think it should go in that release or should we just mark it 1.3.x, merge it, and only include it in 1.3.1?

ucbjrl:
Is it relatively stable? As far as I'm concerned, that would be the prime reason **not** to include it in 1.3.0. If it's going to be merged into master before 1.3.0 is released, I'd say it should be part of the 1.3.0 release.

albert-magyar:
I don't see any reason why this would change. The source changes (moving around things to the `testutils` package) won't change, and the new assembly in test JAR is just creating a JAR that a) didn't exist before and b) shouldn't get published, so if the `build.sbt` changes it won't affect anything release-facing.

@jackkoenig @davidbiancolin any thoughts?

jackkoenig:
I think it's fine for `1.3.0`.

PR #1413, opened 2020-03-02
Remove DiGraph.seededLinearize

### LINKED ISSUE #1247: Remove DiGraph.seededLinearize
tl;dr: My `DiGraph.seededLinearize` method doesn't make any sense and should probably be removed.

I'm filing this as a tracking issue in case anybody tries to use this method.

### Checklist

`DiGraph.seededLinearize` was a method I added to try and tease out one possible linearization from the space of all possible linearizations. I did this by adding a "seed" which would be used to see the initial DFS. However, seeding a DFS doesn't matter as nodes are being walked **in the order in which the edges are stored**. So, this method doesn't really have a point except it works for all test cases currently explored in the dependency API.

The expected behavior is that this would result in a specific order in which nodes are walked. This can be done inside the `seededLinearize` method or with an implicit ordering. However, it likely makes more sense to just remove this method.

If a user that wants a specific ordering can instead, produce a new `DiGraph` with everything in the order they want, e.g., [the approach taken in the Dependency API PR](https://github.com/freechipsproject/firrtl/pull/1123/commits/da7c4293e25281b0eaf837c64f4467f956b52d10#diff-4aa6423289d5fcb0bcd5aee390090c4bR206).

### Steps to Reproduce

(Intentionally poorly documented steps to reproduce because I'm not intending somebody to jump in and fix this. I'm only expecting for this method to be removed in 1.3.0.)

Running the following test case with the old logic will fail: 
 
 - https://github.com/freechipsproject/firrtl/pull/1123/commits/9b9ef2826e0a6b86835620333c93f39dd7c8f966

### Your environment

- `dependency-api-2` branch: 9b9ef2826e0a6b86835620333c93f39dd7c8f966 (this commit will probably disappear due to a rebase)

### External Information

None.

seldridge:
This removes the `seededLinearize` method added in #1079.

The original motivation here was to be able to seed the order in which vertices in a graph were explored during topological sorting. However, this method is really just providing an ordering for the _initial_ set of BFS vertices. This isn't enough as you really want to explore all vertices _and their edges_ in a specific order. The better way to do this is to sort the edges the way you want and then pass this into the existing topological sort method, `linearize`. ([This is how it works in the current version of the `DependencyManager`](https://github.com/freechipsproject/firrtl/blob/master/src/main/scala/firrtl/options/DependencyManager.scala#L192).)

Fixes #1247.

### Contributor Checklist

### Type of Improvement

- API Removal

### API Impact

This removes the `DiGraph.seededLinearize` method.

None.

### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/graph/DiGraph.scala
+++ b/src/main/scala/firrtl/graph/DiGraph.scala
@@ -64,52 +64,6 @@ class DiGraph[T] private[graph] (private[graph] val edges: LinkedHashMap[T, Link
     */
   def findSinks: Set[T] = reverse.findSources
 
-  /** Linearizes (topologically sorts) a DAG using a DFS. This can be seeded with an order to use for the DFS if the user
-    * wants to tease out a special ordering of the DAG.
-    * @param seed an optional sequence of vertices to use. This will default to the vertices ordering provided by getVertices.
-    * @throws CyclicException if the graph is cyclic
-    * @return a Map[T,T] from each visited node to its predecessor in the
-    * traversal
-    */
-  def seededLinearize(seed: Option[Seq[T]] = None): Seq[T] = {
-    // permanently marked nodes are implicitly held in order
-    val order = new mutable.ArrayBuffer[T]
-    // invariant: no intersection between unmarked and tempMarked
-    val unmarked = new mutable.LinkedHashSet[T]
-    val tempMarked = new mutable.LinkedHashSet[T]
-
-    case class LinearizeFrame[A](v: A, expanded: Boolean)
-    val callStack = mutable.Stack[LinearizeFrame[T]]()
-
-    unmarked ++= seed.getOrElse(getVertices)
-    while (unmarked.nonEmpty) {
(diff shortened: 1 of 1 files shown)


src/main/scala/firrtl/graph/DiGraph.scala
```diff
@@ -81,7 +104,7 @@ class DiGraph[T] private[graph] (private[graph] val edges: LinkedHashMap[T, Link
     case class LinearizeFrame[A](v: A, expanded: Boolean)
     val callStack = mutable.Stack[LinearizeFrame[T]]()
 
-    unmarked ++= seed.getOrElse(getVertices)
+    unmarked ++= getVertices
```
albert-magyar: Looks good now.

seldridge:
Yeah, the diff algorithm can't handle what's going on... (I was trying to clean this up in a rebase, but forgot to actually remove the use of the `seed`... Fixing in a rebase.)

PR #1420, opened 2020-03-03
Support Java API doc Scaladoc/Unidoc References

seldridge:
This enables automatic Scaladoc linking to Java API docs using the same strategy that Scala upstream uses.

Linking to Scala API documentation already works.

### Contributor Checklist

### Type of Improvement

- documentation

### API Impact

None.

None.

### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -136,6 +136,18 @@ lazy val publishSettings = Seq(
 lazy val docSettings = Seq(
   doc in Compile := (doc in ScalaUnidoc).value,
   autoAPIMappings := true,
+  apiMappings ++= {
+    Option(System.getProperty("sun.boot.class.path")).flatMap { classPath =>
+      classPath.split(java.io.File.pathSeparator).find(_.endsWith(java.io.File.separator + "rt.jar"))
+    }.map { jarPath =>
+      Map(
+        file(jarPath) -> url("https://docs.oracle.com/javase/8/docs/api")
+      )
+    }.getOrElse {
+      streams.value.log.warn("Failed to add bootstrap class path of Java to apiMappings")
+      Map.empty[File,URL]
+    }
+  },
   scalacOptions in Compile in doc ++= Seq(
     "-diagrams",
     "-diagrams-max-classes", "25",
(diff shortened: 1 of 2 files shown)


seldridge:
I have no earthly idea why this isn't building on Travis... :woman_shrugging: This works locally with a clean clone and a nuked `~/.ivy2`.

ucbjrl:
It looks like the plugin has been compiled for Java 8 and won't work with Java 11 or 13.

seldridge:
Ugh, thanks Jim. Scala does something different for linking to Java API docs. I can at least add that.

ucbjrl:
The incompatibility doesn't appear to be fundamental in nature. If you generate a full log, you get:
```
[error] scala.MatchError: 11 (of class java.lang.String)
[error] 	at com.thoughtworks.sbtApiMappings.BootstrapApiMappings$.$anonfun$globalSettings$1(BootstrapApiMappings.scala:28)
[error] 	at sbt.internal.util.Init$Value.$anonfun$apply$3(Settings.scala:823)
```
and the code at that location is:
```
  private[sbtApiMappings] def defaultBootstrapJavadocUrl = {
    val javaVersion = sys.props("java.specification.version") match {
->      case VersionNumber(Seq(1, javaVersion, _*), _, _) => javaVersion // 1.6-1.8
        case VersionNumber(Seq(javaVersion, _*),    _, _) => javaVersion // 9+
        case _ =>                                            8
    }
    if (javaVersion >= 11) {
      url(s"https://docs.oracle.com/en/java/javase/${javaVersion}/docs/api/index.html")
    } else {
      url(s"https://docs.oracle.com/javase/${javaVersion}/docs/api/index.html")
    }
  }
```
`java.specification.version` for post Java 8 differs between Oracle Java and OpenJDK. See https://stackoverflow.com/questions/5103121/how-to-find-the-jvm-version-from-a-program
We need more standards.

seldridge:
I switched this to not use the sbt-api-mappings plugin to avoid the issue that @ucbjrl found (also that plugin is sparsely updated making me nervous). Instead, this uses the same strategy that Scala uses to link against Java API documentation. The small example included works for unidoc.

PR #1421, opened 2020-03-03
Check sign of primop constants where appropriate

albert-magyar:
**Type of change:** better error messages / checks
**API impact:** none
**Code generation impact:** prevents generation of some broken code

Currently, negative constants arguments are banned for `Pad`, `Head`, `Tail`, `Bits`, `IncP`, and `DecP`, but this is never actually checked. This PR expands `CheckHighFormLike` to include these extra checks.

This also fixes a bug where a wrong-arity `Bits` could cause an `IndexOutOfBoundsException`.

**Before this commit, the following broken code makes it all the way through the compiler:**
```
circuit m:
  module m:
    input i1: UInt<1>
    input i2: UInt<1>
    output o: UInt<2>
    o <= bits(add(i1, i2), 0, -1)
```

**Surprising output:**
```Verilog
module m(
  input        i1,
  input        i2,
  output [1:0] o
);
  assign o = i1 + i2;
endmodule
```


albert-magyar:
@jackkoenig, I will need to manually modify the backport since this involves `IncP` et al.

src/main/scala/firrtl/passes/Checks.scala
```diff
@@ -83,24 +83,31 @@ trait CheckHighFormLike {
           errors.append(new IncorrectNumConstsException(info, mname, e.op.toString, nc))
       }
 
+      def nonNegativeConsts(): Unit = {
+        e.consts.map(_.toInt).filter(_ < 0).foreach {
```
jackkoenig: While this is huge improvement that I am happy to approve as is, it does leave a (less-common) bug open:
```
circuit m:
  module m:
    input i1: UInt<1>
    input i2: UInt<1>
    output o: UInt<2>
    o <= bits(add(i1, i2), 0, -11823845700)
```
You get
```
[error] Exception in thread "main" firrtl.passes.CheckHighFormLike$LsbLargerThanMsbException:  @[:test.fir@6.4]: [module m] Primop bits lsb 1061056188 > 0.
```

Unfortunately, `BigInt.toInt` just truncates into a 32-bit Int 👿

albert-magyar: True. I just fixed that and the other pre-existing Int-wrapping bug (for LSB > MSB check).

albert-magyar:
I updated the exceptions to use `BigInt`, which is technically an API change. I'll manually fix this for the backport @jackkoenig.

albert-magyar:
@jackkoenig let's watch to see if Mergify backports this.

If it does, I'll still have to manually fix it up since I changed the type of a field in one of the exceptions.

PR #1422, opened 2020-03-04
Revert inline nots

albert-magyar:
Based on the [discussion on chisel-users](https://groups.google.com/forum/#!topic/chisel-users/EOJPKvznbH8), I also added a formal check for a minimal, non-working FIRRTL input that generates the Verilog pattern Øyvind and @ekiwi were looking at. **I added it as a new formal check, but it is expected to fail for this PR since it is broken in master.**

```
circuit AddNot:
  module AddNot:
    input a: UInt<8>
    input b: UInt<8>
    output o: UInt<9>
    o <= add(a, not(b))
```

### Contributor Checklist

### Type of Improvement: 
 - bug fix

### API Impact
- removes `InlineNots`

- fixes the emission of incorrect Verilog discovered by @oharboe

### Desired Merge Strategy

- [ ] Did you review?


src/main/scala/firrtl/Emitter.scala
```diff
@@ -325,6 +322,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
      }
 
      def checkCatArgumentLegality(e: Expression): Unit = e match {
+       case _: UIntLiteral | _: SIntLiteral | _: WRef | _: WSubField =>
```
chick: This line appears as first statement in `checkArgumentLegality` why duplicate it here.
And just curious: is `WSubIndex` different from `WRef` and `WSubField` that it is not included here?

jackkoenig:
We should backport this right? It's also broken on 1.2.X

ucbjrl:
I think we should backport this and get a new release out ASAP.

jackkoenig:
I've added the `1.2.X` Milestone. @albert-magyar please remove if you disagree

jackkoenig:
Rather than a separate big LEC test, I think we should group these smaller ones in the regular Scala tests using `firrtlEquivalenceTest` like https://github.com/freechipsproject/firrtl/blob/b90589f5cd9d4048ada2a05d5225874791546170/src/test/scala/firrtlTests/transforms/CombineCatsSpec.scala#L34

I think we could write this test and put it with `ignore` for now so that Travis passes.

albert-magyar:
Yeah @jackkoenig this should definitely be 1.2.x, I just forgot. It will need manual fixup since removing the transform itself is obviously a binary-compatibility problem.

jackkoenig:
@albert-magyar we could also leave the transform, ignore the tests and just don't run it. But yeah removing and fixing up the backport is also fine.

src/main/scala/firrtl/Emitter.scala
albert-magyar: This is a no-op, good catch. I will delete it. `WSubIndex` is illegal in Low FIRRTL :)

(`WSubField` is legal in Low FIRRTL for memory ports)

albert-magyar:
I reverted the const-prop change since the value of the strength reduction is reduced in the absence of `InlineNots`.

We've also seen in #1240 that rolling strength reduction into `ConstantPropagation` can sometimes hurt QoR.

.travis.yml
```diff
@@ -82,6 +82,12 @@ jobs:
         - yosys -V
         - "travis_wait 30 sleep 1800 &"
         - ./.run_formal_checks.sh Ops
+    - stage: test
+      name: "Formal equivalence: AddNot"
+      script:
+        - yosys -V
+        - "travis_wait 30 sleep 1800 &"
+        - ./.run_formal_checks.sh AddNot
```
jackkoenig: Thoughts on putting this as a test in the ScalaTests rather than a full Travis stage?

albert-magyar: The DUT and reference both use the `VerilogEmitter` from the current version of the package, so how could we find emitter bugs with that?

Qui Verilogiet ipsos Veriloges?

albert-magyar: Good point about the wastefulness of the stage.

I put it into the same stage as `Ops` since they both run fast. I think this will cache the jars for your script?

This will also help cut Travis-related queueing delays.

jackkoenig: Ops is definitely better but I was thinking the style that `CombineCats` uses https://github.com/freechipsproject/firrtl/blob/b90589f5cd9d4048ada2a05d5225874791546170/src/test/scala/firrtlTests/transforms/CombineCatsSpec.scala#L34 where it checks against the `MinimumVerilogCompiler` that excludes this stuff

albert-magyar: `MinimumVerilogEmitter` still includes `InlineNots`. The only thing it filters is `DeadCodeElimination`.

albert-magyar: I think that fixing `MinimumVerilogEmitter` to be a more meaningful "safe baseline" would be good, but for this PR my tinfoil hat is firmly attached :)

jackkoenig: > MinimumVerilogEmitter still includes InlineNots. The only thing it filters is DeadCodeElimination

Ah well that's not good. We should make it truly Minimum. Anyway for this PR I think what you're doing is right then.

albert-magyar:
Merging this since @jackkoenig got it through SiFive regressions when cherry-picking into pre-inline-nots 1.2.x.

ucbjrl:
The above comment: `mergify bot added the Backported label 14 minutes ago` may be true, but the action it claims: `This PR has been backported to 1.2.X` appears not to be true, possibly because it failed checks. Should someone manually cherry-pick it into 1.2.x?

albert-magyar:
I'm manually fixing it up -- the `Backported` label maybe isn't as clear as it could be. Currently it just means that the backport PR has been opened.

jackkoenig:
Maybe there's a bug in Mergify

ucbjrl:
I couldn't find evidence of a backport PR.

PR #1429, opened 2020-03-06
Make mergify open backport PRs & signal on failed cherry-picks

albert-magyar:
@jackkoenig @ucbjrl, this is why we didn't see a backports for #1421 (and possibly others).

Mergify added a new option that now makes it optional to open backport PRs when there are cherry-pick conflicts. It is false by default, which means anything that doesn't cleanly cherry-pick gets dropped by default.

By setting this to `True`, we should get a nice PR that tells us it fails like this [example in a plain link that hopefully doesn't become a GitHub "mention" and troll the owners of that repo (but honestly, who knows what magic GitHub does with anything these days).](https://github.com/Mailu/Mailu/pull/932)

Relevant mergify-engine PR from the last couple weeks: Mergifyio/mergify-engine#741

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -29,6 +29,7 @@ pull_request_rules:
       backport:
         branches:
           - 1.2.x
+        ignore_conflicts: True
       label:
         add: [Backported]
 
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -14,6 +14,7 @@ pull_request_rules:
       - base=master
       - label="Please Merge"
       - label!="DO NOT MERGE"
+      - label!="bp-conflict"
     actions:
       merge:
         method: squash
@@ -30,6 +31,7 @@ pull_request_rules:
         branches:
           - 1.2.x
         ignore_conflicts: True
+        label_conflicts: "bp-conflict"
       label:
         add: [Backported]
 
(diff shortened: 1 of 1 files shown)


jackkoenig:
I remember seeing complaints about the old behavior (that you're reintroducing) in other repos because the PR would conflict but CI wouldn't catch it then it would be merged with the git conflict text in the merged files (lol). I still prefer this behavior since I think it's reasonable that your CI should be able to catch such issues. The one caveat is for pure documentation changes which we certainly don't have checks for.

jackkoenig:
Thanks for figuring this out @albert-magyar! Can you make a similar PR to chisel3?

albert-magyar:
We could filter auto-merge for PRs that have the boilerplate that would show up on a conflict?

jackkoenig:
> We could filter auto-merge for PRs that have the boilerplate that would show up on a conflict?

Good idea

albert-magyar:
@jackkoenig, they added a new ability to label the backport conflict PRs literally hours ago: Mergifyio/mergify-engine#771

I don't think it's out in the live build yet, so maybe we should wait?

jackkoenig:
Yeah let's wait on it, we haven't been burned yet so there isn't a rush.

jackkoenig:
Conflict label seems to be merged: https://github.com/Mergifyio/mergify-engine/pull/771

PR #1433, opened 2020-03-07
Provide an annotation mix-in that marks RTs as dontTouch

davidbiancolin:
In Golden Gate & FireSim we have a bunch of passes that run conditionally to do things like synthesize debug primitives or event counters. In chisel elaboration, we preemptively mark these nodes as DontTouch even though these synthesis passes may not be run, precluding potential optimizations. 

This PR adds a small trait hierarchy that lets an annotation mark its RTs as DontTouch, so that the Annotation can be easily removed by it's associated Transform later.  This presents new optimization opportunities to the downstream compiler, without having to reason about separate DontTouchAnnotations that may already be present on the same targets. 

### Contributor Checklist

### Type of Improvement
- new feature/API

### API Impact

- API expanding.

- No changes to existing verilog. 

### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/transforms/OptimizationAnnotations.scala
+++ b/src/main/scala/firrtl/transforms/OptimizationAnnotations.scala
@@ -10,7 +10,7 @@ case object NoDCEAnnotation extends NoTargetAnnotation
 
 /** Lets an annotation mark its ReferenceTarget members as DontTouch
   *
-  * This permits a transform to run and remove it's associated annotations,
+  * This permits a transform to run and remove its associated annotations,
   * thus making their ReferenceTargets new candidates for optimization. This
   * removes the need for the pass writer to reason about pre-existing
   * DontTouchAnnotations that may touch the same node.
--- a/src/main/scala/firrtl/transforms/OptimizationAnnotations.scala
+++ b/src/main/scala/firrtl/transforms/OptimizationAnnotations.scala
@@ -50,8 +50,6 @@ object DontTouchAnnotation {
     throw new DontTouchNotFoundException(module, component)
 }
 
-
-
 /** An [[firrtl.ir.ExtModule]] that can be optimized
   *
   * Firrtl does not know the semantics of an external module. This annotation provides some
--- a/src/main/scala/firrtl/transforms/OptimizationAnnotations.scala
+++ b/src/main/scala/firrtl/transforms/OptimizationAnnotations.scala
@@ -8,11 +8,33 @@ import firrtl.passes.PassException
 /** Indicate that DCE should not be run */
 case object NoDCEAnnotation extends NoTargetAnnotation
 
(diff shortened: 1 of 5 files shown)


src/main/scala/firrtl/transforms/OptimizationAnnotations.scala
```diff
@@ -8,11 +8,33 @@ import firrtl.passes.PassException
...
+trait HasDontTouches { self: Annotation =>
+  def dontTouches: Iterable[ReferenceTarget]
+}
+
+/**
+  * A globalized form of HasDontTouches which applies to all ReferenceTargets
+  * provided with the annotation
+  */
+trait DontTouchAllTargets extends HasDontTouches { self: Annotation =>
+  def dontTouches(): Iterable[ReferenceTarget] = getTargets.collect {
```
albert-magyar:
```suggestion
  def dontTouches: Iterable[ReferenceTarget] = getTargets.collect {
```

src/test/scala/firrtlTests/DCETests.scala
```diff
@@ -11,6 +11,13 @@ import firrtl.passes.memlib.SimpleTransform
 import java.io.File
 import java.nio.file.Paths
 
+case class AnnotationWithDontTouches(target: ReferenceTarget)
+    extends SingleTargetAnnotation[ReferenceTarget] with HasDontTouches {
+  def targets = Seq(target)
+  def duplicate(n: ReferenceTarget) = this.copy(n)
+  def dontTouches(): Seq[ReferenceTarget] = targets
```
albert-magyar:
```suggestion
  def dontTouches: Seq[ReferenceTarget] = targets
```

albert-magyar:
I would really like to use this. I frequently mix `dontTouch` onto the targets of LOTS of annotations, and this a) streamlines the process and b) allows you to "drop" the `dontTouch` restriction from your specified annotation without messing with existing `dontTouch`es.

Because this makes `DontTouchAnnotation` extend the new trait, it doesn't add the extra burden of making passes find two different kinds of `dontTouch`es. However, it would need to go into 1.3.0 since it will break binary compatibility.

src/main/scala/firrtl/transforms/OptimizationAnnotations.scala
```diff
@@ -8,11 +8,33 @@ import firrtl.passes.PassException
 /** Indicate that DCE should not be run */
 case object NoDCEAnnotation extends NoTargetAnnotation
 
+/** Lets an annotation mark its ReferenceTarget members as DontTouch
+  *
+  * This permits a transform to run and remove it's associated annotations,
```
albert-magyar:
```suggestion
  * This permits a transform to run and remove its associated annotations,
```

Issue #1436, opened 2020-03-10
Module name conflicts are silently ignored


albert-magyar:
Similar to #1096 but for actual module name conflicts.

### Checklist

- [x] Did you specify the current behavior?
- [x] Did you specify the expected behavior?
- [x] Did you provide a code example showing the problem?
- [ ] Did you describe your environment?
- [ ] Did you specify relevant external information?

### Steps to Reproduce
Use the following FIRRTL:
```
circuit main:
  module main:
    input i: UInt<1>
    output o: UInt<1>

    inst x of dup
    x.i <= i
    o <= x.o

  module dup:
    input i: UInt<1>
    output o: UInt<1>
    o <= i

  module dup:
    input i: UInt<1>
    output o: UInt<1>
    o <= not(i)
```

### What is the current behavior?
Compiles all the way to the following Verilog:
```Verilog
module main(
  input   i,
  output  o
);
  wire  x_i;
  wire  x_o;
  dup x (
    .i(x_i),
    .o(x_o)
  );
  assign o = x_o;
  assign x_i = i;
endmodule
module dup(
  input   i,
  output  o
);
  assign o = ~i;
endmodule
```

### What is the expected behavior?
Throw a `PassException`.

Issue #1096, opened 2019-06-14
ExtModule with defname may conflict with namespace


seldridge:
**Type of issue**: bug report

**If the current behavior is a bug, please provide the steps to reproduce the problem:**

Define an `ExtModule` with a `defName` that maps to a module that already exists in the circuit. This doesn't result in an error, but I think it probably should. A better approach would be that module names should be mangled to not conflict with any `ExtModule` `defName`s.

  * **What is the current behavior?**

You can get an `ExtModule` to resolve to a module in the circuit.

  * **What is the expected behavior?**

If something is an `ExtModule` it shouldn't resolve to anything contained in the circuit.

  * **Please tell us about your environment:**
    (examples)
    - version: `3.2-SNAPSHOT`

**Impact**: no functional change

**Development Phase**: request

PR #1437, opened 2020-03-10
Check for name collisions of Modules

albert-magyar:
Currently, there is no checking of whether module names collide (see #1436). There is also no checking of whether `ExtModule` `defname`s collide with pure-FIRRTL module names (see #1096).

**Type of improvement:** bugfix
**API impact:** none
**Backend code generation impact:** avoids generating some illegal code and some code that randomly chooses from among aliased modules
**Desired merge strategy:** merge commit

### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/src/test/scala/firrtlTests/CheckSpec.scala
+++ b/src/test/scala/firrtlTests/CheckSpec.scala
@@ -333,6 +333,23 @@ class CheckSpec extends FlatSpec with Matchers {
     }
   }
 
+  s"Defnames that conflict with pure-FIRRTL module names" should "throw an exception" in {
+    val input =
+      s"""|circuit bar :
+          |  module bar :
+          |    input i : UInt<8>
+          |    output o : UInt<8>
+          |    o <= i
+          |  extmodule dup :
+          |    input i : UInt<8>
+          |    output o : UInt<8>
+          |    defname = bar
+          |""".stripMargin
+    assertThrows[CheckHighForm.DefnameConflictException] {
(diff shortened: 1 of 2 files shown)


src/test/scala/firrtlTests/CheckSpec.scala
```diff
@@ -307,6 +307,49 @@ class CheckSpec extends FlatSpec with Matchers {
...
+          |    output o : UInt<8>
+          |    o <= not(i)
+          |""".stripMargin
+    assertThrows[CheckHighForm.ModuleNameNotUniqueException] {
+      try {
+        checkHighInput(input)
+      } catch {
+        case e: firrtl.passes.PassExceptions => throw e.exceptions.head
+      }
+    }
```
seldridge: Throwing the head of the cause is a clever way to unwrap this.

albert-magyar: Yeah, it's pretty annoying to make a test cause a single `PassException` in many cases, even when it's possible.

seldridge: Maybe we could write a custom method `assertThrowsOrContainsCause[A <: Throwable]` that will iterate over all throwables until it finds or fails to find type `A`.

albert-magyar: That's a good idea. I'll add that -- I know I'd use it elsewhere.

PR #1439, opened 2020-03-10
[mergify] Update match string for labeling backported PRs

albert-magyar:
More in the quest to make the backport flow work again. The body of backport PR messages got changed in Mergifyio/mergify-engine#768, so I updated the match string to work again.

This fixes the cause of #1438 not getting the "Backport" label.


albert-magyar:
If we merge this, the label action should trigger on #1438, which should trigger the no-review merge action.

Issue #1309, opened 2020-01-11
EliminateTargetPaths and Renaming/Duplicating Annotations


seldridge:
### Checklist

- [x] Did you specify the current behavior?
- [x] Did you specify the expected behavior?
- [x] Did you provide a code example showing the problem?
- [x] Did you describe your environment?
- [x] Did you specify relevant external information?

### What is the current behavior?

Consider the following FIRRTL:

```
circuit Foo:
  module Bar:
    node x = UInt<1>(0)
    skip
  module Foo:
    inst bar of Bar
```

And an annotation targeting `~Foo|Bar>x`.

When `EliminateTargetPaths` runs to duplicate `~Foo|Foo/bar:Bar` this results in the following (simplified) rename:

```
~Foo|Foo/bar:Bar => ~Foo|Foo/bar:Qux
```

However, the annotation associated with `~Foo|Bar>x` is just left around.

### What is the expected behavior?

*Something* needs to happen with the annotation. Specifically, one of two things:

1. If the duplicated instance is the only instance in the whole design, then the annotation should be renamed as `~Foo|Bar>x => ~Foo|Qux>x`.
1. If the duplicated instance is not the only instance, then the annotation should be copied with a rename of `~Foo|Bar>x => [~Foo|Bar>x, ~Foo|Qux>x]`

Additionally, the actual instance should be renamed.

### Steps to Reproduce

Run `EliminateTargetPaths` on the circuit above requesting to duplicate the `bar` instance. Note: if running this inside of `EliminateTargetPathsSpec`, you may need to patch to fix #1308. The existence of `DontTouchAnnotations` in many of the tests result in spurious `ResolvePaths` annotations being left around that can make this work in limited scenarios.

### Your environment

- current FIRRTL master 2dc005d500ffcccef5def5938dc7ead7c68644b3

### External Information

This is related to #611, but is a bit of the reverse. #611 talks about how to handle _deduplication_ of annotations. This issue is specifically concerned with a solution to __duplication__ of annotations. 

Additionally, a solution for both #611 and this issue are likely required for the solution proposed in #1301.

seldridge:
Fixed via #1370

PR #1442, opened 2020-03-11
Implement InstanceTarget Behavior for Dedup + EliminateTargetPaths

azidar:
### Summary of changes
- `RenameMap` Behavior
  - Prevent transitive renaming A -> B -> C (continueRenaming)
  - Added AST lookups for regular renames, so given rename (~A|A -> ~B|B), ~Top|A is renamed to ~Top|B.
  - Added transform to `RenameMap` initialization, and update all locations of `RenameMap()`
- Target
  - Add `ModuleTarget(module: String)` apply to create AST `ModuleTarget`
  - Override `toString` as `serialize` for `CompleteTarget`
  - Expansion of stripHierarchy to enable stripping `InstanceTarget`s to become `ModuleTarget`s
- Annotations
  - Bugfix in extractComponents where Products were not iterated over
  - SingleTargetAnnotation[Named] now converts to AST targets to preserve sticky behavior over dedup
- Eliminate Target Paths
  - Make DuplicationHelper use `LinkedHashMap`, as we iterate over its contents and convert to Seq in `def makePathless`
  - Add `DupedResult` to map original module to new module targets
  - Update renaming to record a map from all relative instance paths to original module, to new module target
  - Consumes `DedupedResult` to give better name to new duplicated module if it was originally deduplicated
  - Reorder modules in attempt to preserve original ordering, pre-deduplication
  - Move utility functions to object
- Dedup
  - Changed `NoDedupAnnotation` to contain `ModuleTarget`, rather than `ModuleName`
  - Added `DedupedResult` to map original module to the duplicate module
  - Consumes `DupedResult` to pick better name, if it existed
  - Updates renaming to chain the following: instancify deduped modules, remap differently named internal signals, then remap AST modules
  - Move utility functions to object
  - Remove annotations as part of determination of dedup correctness
- Dead Code Elimination
  - Add deletion of ASTModules
- Tests
  - Morphism Spec to ensure Dedup -> EliminateTargetPaths and EliminateTargetPaths -> Dedup patterns work properly
  - Update existing tests to make sure they work properly

### Contributor Checklist

### Type of Improvement

- bug fix
- documentation
- new feature/API

### API Impact

Biggest change is that DedupModules works between modules with different Target annotations. Another change is adding AST ModuleTargets, which are used when converting between Named and Target in SingleTargetAnnotation.

### Desired Merge Strategy

- [ ] Did you review?


mwachs5:
Is there a high-level description of what this PR is doing / how it's changing behavior?

seldridge:
@mwachs5: Sure, I can try to elaborate.

This PR is attempting to correct problems with how annotations are renamed by duplication (the `EliminateTargetPaths` transform) and deduplication. Currently this doesn't do the right thing. There is not intended to be any new behavior added by this PR. This PR is primarily intended to fix dup/dedup behavior and to then make the pattern of dup/transform/dedup tractable (more info in the `Background` section below).

Basically, this is trying to solve the following two problems:

1. When an instance is duplicated, what should happen to its annotations?
1. When an instance is deduplicated, what should happen to its annotations?

This attempts to always "do the right thing" meaning that annotations are not merged blindly or disallow dup/dedup. Instead, annotations are converted back and forth between `ModuleTarget`s and `InstanceTarget`s (with minimal paths). I did some early work on getting duplication better (#1309/#1370). This PR drives it all the way home.

### Background

There's a pattern for writing transforms where you'd like to "unfold" the AST to make the transform simple to write. Consider the following circuit (which has no candidates for deduplication):


```
circuit Foo:
  module Baz:
    node x
  module Bar:
    inst baz1 of Baz
    inst baz2 of Baz
  module Foo
    inst bar1 of Bar
    inst bar2 of Bar
```

Then assume that you have an annotation that is targeting `node x`, but only in one instance:

```
~Foo|Foo/bar1:Bar/baz1:Baz>x
```

If your transform is supposed to make modifications to `x`, you can't blindly modify the `Baz` module. (This may affect other instances.) So, the transform has to be painfully aware of the rest of the circuit, whether or not `Baz` is instantiated multiple times, whether or not every instance of `Baz` has the same annotation, etc. This gets very complicated, very quickly. However, it's something that almost every transform has to reason about.

To help with this, there's a pattern of **dup/transform/dedup**. What this means is, you conservatively duplicate everything you want to modify to make these separate modules, you then trivially transform these modules without worrying about any of the issues above, then you deduplicate.

The transform that does duplication is `EliminateTargetPaths` and you know about `DedupModules`.

However, in order for this pattern to work, duplication and deduplication need to "do the right thing" with all annotations. They need to not be blocked by annotations and you want annotations to properly morph based on how the underlying AST is folded/unfolded. Currently this is busted.

### Testing

I was originally trying to think about a good way to specify this in tests. Adam suggested thinking about identities that hold across transforms. The initial `MorphismSpec` was me trying to write this up in functional programming parlance. Basically, there should be transforms that have idempotent properties (`IdempotencyFixture`) and transforms that are inverses (`RightInverseFixture`). Adam then built this out more as a means of testing the behavior of dup/dedup.

There are stubs in the testing for a related pair of transforms, grouping and inlining/flattening. There are also probably a lot of situations where this type of testing makes sense that haven't been identified.

### Feedback

- Does this make sense?
- I assume some documentation about this pattern would be useful?

mwachs5:
I think what would help is clearly "do the right thing" means in the different cases? Is there a pointer to what the expected/desired behavior is for the annotations ?

seldridge:
High level, the "right thing" for deduplication is that you let modules merge, but their annotations become attached to minimal instances.

Consider the following circuit:
```
circuit Foo:
  module Baz:
    node y
  module Bar:
    node x
  module Foo:
    inst bar of Bar
    inst baz of Baz
```

With some annotation attached to a module `Bar`:
```
~Foo|Baz>y
```

When you deduplicate this to:
```
circuit Foo:
  module Bar:
    node x
  module Foo:
    inst bar of Bar
    inst baz of Bar
```

The annotation should become:
```
~Foo|Foo/baz:Bar>x
```

The right thing for duplication is then to invert this back to the original (with the exact same names of everything). There's a lot of tests in the `MorphismSpec` that are all written like this. For some annotations and some stuff to duplicate, make sure that running duplication followed by deduplication gets me back where I started.

albertchen-sifive:
I think you can replicate the behavior of the AST module using just the `update` method? If your `moduleTarget` gets renamed to a bunch of `InstanceTargets` you could just turn those `InstanceTarget`s back into `moduleTarget`s by using the `ofModule` fields. for relative instance targets, you could retain the same information if you use an extra target to keep track of the root module and then compute the relative paths in your transform.

Changing the meaning of the `circuit` field seems like a pretty big change to me. in this case, I'd rather use more `Target`s or a different `update` method to keep track of the things I'm targeting rather than complicating the `Target` semantics and the `RenameMap` logic. And I think `circuit` is still valuable as a namespacing tool.

jackkoenig:
I have a use case that I want to mention. Say I have some InstanceTarget:

```scala
 val top = CircuitTarget("Top")
 val a = top.module("Top").instOf("dut", "DUT").instOf("foo", "Foo")
```


I then want to delete module `Foo`:
```scala
  val rm = RenameMap()
  rm.delete(top.module("Foo")) // same as rm.record(top.module("Foo"), Seq())
```

If I then lookup the rename for my instance target, `master` (currently 1a03e63) just falls on its face:
```scala
  // Exception in thread "main" firrtl.RenameMap$IllegalRenameException: ofModule
  //   ModuleTarget(Top,Foo) cannot be renamed to Vector() - an ofModule can only be 
  //   renamed to a single IsModule with the same circuit  
  rm.get(a)
```

This branch does better, it doesn't throw an error, but the `get` returns `None`. This implies that there is no change to `a`, but that seems wrong since the Module to which it points was deleted.

src/main/scala/firrtl/RenameMap.scala
```diff
@@ -401,7 +409,19 @@ final class RenameMap private (val underlying: mutable.HashMap[CompleteTarget, S
...
+      //   Given Renames:
+      //     - (~Top|Top/a:A -> ~Top|A)
+      //     - (~Top|A -> ~Top|B)
+      //   Renaming (~Top|Top/a:A) should return (~Top|A), not (~Top|B)
+      // This is due to our splitting recursive renaming into three parts (top, mid, bot)
+      def continueRenaming[T <: CompleteTarget](original: Seq[CompleteTarget], ts: Seq[CompleteTarget]): Boolean =
+        ts match {
+          case Seq(_: T) if original != ts => false
+          case _ => true
+        }
```
albertchen-sifive, line 424: self-renames should stop renaming as well right? this looks like it would return `true` on self-rename.

seldridge:
@jackkoenig: I think is doing the right thing and the bug is in the transform that is recording the renames. If the transform also changed the instances, then it should be recording those renames.

To actually delete the module and all instances would require circuit-aware renaming. While I think circuit-aware renaming is reasonable and comes up often in specific transforms like [`EliminateTargetPaths.renameModules`](https://github.com/freechipsproject/firrtl/pull/1442/files#diff-240ff27fedb90ffe7f1d5cf3cf9b32c3R47), it may be easier to keep the methods for `record`, `delete`, etc. minimal, sharp, and not circuit-aware.

Sharp rename methods let you build up the rename map for weird renames like the following (delete `Foo`, but split it's instances into separate modules): consider a transform that is modifying instances as it goes and will delete unused modules at the end. You might build a rename map in the following way:

```
~Top|Top/dut:Dut/foo:Foo -> ~Top|Top/dut:Dut/foo:Bar
~Top|Top/dut:Dut/foo:Foo -> ~Top|Top/dut:Dut/foo2:Baz
~Top|Foo -> []
```

The final `delete` shouldn't affect the previous renames.

All that said, I'm all in favor of helper methods like `deleteAll(c: Circuit, m: ModuleTarget)`?

jackkoenig:
@seldridge I'm certainly open to alternative ways to express it, and a deleteAll helper would be fine. I guess it's not clear to me why the behavior I was assuming wouldn't support what you're saying. The semantics are already that the most specific thing matches. In your example, since there are also instance target renames, only the most precise instance target would match.

To be more precise, if you had the renames:
```
~Top|Top/dut:Dut/foo:Foo -> ~Top|Top/dut:Dut/foo:Bar
~Top|Top/dut:Dut/foo:Foo -> ~Top|Top/dut:Dut/foo2:Baz
~Top|Foo -> []
```
Then when you lookup `~Top|Top/dut:Dut/foo:Foo`, because there is a rename for that exact path, that one would match.

If `~Top|Foo -> []` **doesn't** delete instance paths that refer to instances of `Foo` (assuming no more precise rename applies), then what even is the point of it? Should `~Top|Foo -> ~Top|Bar` apply to instance targets?

seldridge:
Reading through this a couple times, I think the question is: **should renames be greedy or not?**

I had always thought of renames as not greedy and implemented as a single membership query of a set of targets. With greedy renames, you need multiple queries of the set or you need circuit information to record all renames at once.

I think with both greedy or non-greedy renames you can describe "weird" circuit modifications like deleting a module, but not deleting instance references.

@jackkoenig wrote:
> If `~Top|Foo -> []` doesn't delete instance paths that refer to instances of `Foo` (assuming no more precise rename applies), then what even is the point of it? 

I tend to think of this in terms of renames as diffs on the circuit. Or: does a transform exist that could create this rename? A pedantic example (that throws an internal error in `CInferTypes`...) would be something which removes a module, but doesn't remove instances:

Before:
```
circuit Top:
  module Foo:
    skip
  module Top:
    inst foo of Foo
```

After:
```
circuit Top:
  module Top:
    inst foo of Foo
```

@jackkoenig wrote:
> Should `~Top|Foo -> ~Top|Bar` apply to instance targets?

There's a history of it not working this way (see: #1309, #1370). I'd argue no.

From the circuit diff persepctive, the following represents `~Top|Foo -> ~Top|Bar`, but leaves instances intact.

Before:
```
circuit Top:
  module Foo:
    skip
  module Top:
    inst foo of Foo
```

After:
```
circuit Top:
  module Bar:
    skip
  module Top:
    inst foo of Foo
```

Note, that both of these could be represented with your proposed scheme and explicit self renames stating that instances _were not_ removed like `~Top|Top/foo:Foo -> ~Top|Top/foo:Foo`. This is a bit different from how things work now, but this could go either way.

jackkoenig:
> I had always thought of renames as not greedy and implemented as a single membership query of a set of targets. With greedy renames, you need multiple queries of the set or you need circuit information to record all renames at once (or circuit information and a different datastructure).

Note that it is already the case that there are multiple queries: https://github.com/freechipsproject/firrtl/blob/1a03e6356e451136d522d5a9acba374dd8972b24/src/main/scala/firrtl/RenameMap.scala#L398 

The original rename map definitely did only do 1 lookup which requires expressing all possible renames, as Uniquify does (using the old API): https://github.com/freechipsproject/firrtl/blob/1a03e6356e451136d522d5a9acba374dd8972b24/src/main/scala/firrtl/passes/Uniquify.scala#L281)

In my opinion, this is pretty problematic though, it requires a massive amount of work and object creation (effectively flattening all possible leaf names), when the actual information could be encoded very simply with a more sophisticated rename.

Notably, there are at least 2 levels of lookup, because we don't require you to do all possible `ReferenceTarget` renames of things _within_ a module when you rename the Module.

To me, it's about being able to express useful things efficiently. If I want to rename Module `F` to `Foo` for all instances, what do I have to do to make it work for all InstanceTargets?

eg. `~A|A/b:B/c:C/d:D/e:E/f:F`

`~A|F -> ~A|Foo` seems perfectly adequate to me 

I sure hope I don't have to do:
`~A|F -> ~A|Foo`  
`~A|E/f:F -> ~A|E/f:Foo`  
`~A|D/e:E/f:F -> ~A|D/e:E/f:Foo`  
`~A|C/d:D/e:E/f:F -> ~A|C/d:D/e:E/f:Foo`  
`~A|B/c:C/d:D/e:E/f:F -> ~A|B/c:C/d:D/e:E/f:Foo`  
`~A|A/b:B/c:C/d:D/e:E/f:F -> ~A|A/b:B/c:C/d:D/e:E/f:Foo`  
(rinse and repeat for every instance)  

I'm not really clear on what the utility of renaming Modules without renaming the instances is. If you want to rename a Module and then rename instances differently, *then* I think you should explicitly rename the InstanceTargets to the different targets.

Otherwise it seems like we're making this uncommon case easy, and the common case (of renaming all instances) hard. Even with a nicer API, it requires doing `InstanceGraph.findInstancesInHierarchy` every time you want to rename a Module.

Consider a simple but useful transform of prefixing all modules for namespacing reasons since Verilog doesn't have any packages or namespaces. Do we really want to have to do `InstanceGraph. fullHierarchy` which calls `DiGraph. pathsInDAG`, an algorithm that it notes is exponential time?

jackkoenig:
This discussion is interesting though, and it reminds me of an idea someone had (me? I can't remember) a long time ago of having multiple RenameMaps with different semantics (or the possibility of custom ones although that's sketchy).

Even if my idea of greedy renaming so that more complex things can be encoded in fewer renames, one transform still requires massive flattening to do renaming: LowerTypes

Consider `wire w : { a : UInt<1>, b : { c : UInt<1>, d : UInt<1> } }`
Renaming this is super inefficient, it requires:
```
w -> [w_a, w_b_c, w_b_c]
w.a -> [w_a]
w.b -> [w_b_c, w_c_d]
w.b.c -> [w_b_c]
w.b.d -> [w_b_d]
```
This is worse than mere flattening, because it requires every possible intermediate aggregate reference to duplicate all of the renaming of the children (see how `w.b` duplicates `w.b.c` and `w.b.d` and `w` duplicates all leaf references).

LowerTypes makes sense as a potential user of a more efficient/custom RenameMap. Rather than doing this superlinear flattening of all possibilities, it would be much more computationally efficient in practice to do the flattening only lazily, only do this for targets that are actually asked for rather than having to do it for all of them. Thus the "LowerTypesRenameMap" could record simply the types of references (eg. `w` -> { a : UInt<1>, b : { c : UInt<1>, d : UInt<1> } }`) then if it happens to get asked for a target referring to `w` or a subreference, it could generate the right rename.

This Greedy vs. non-Greedy thing seems kind of similar since they could just be alternative standard RenameMap options. Food for thought.

seldridge:
I think you're right. (Here's how I got to that point:)

Is the set of all full path `InstanceTarget`s pointing at the same module equal to one `ModuleTarget`?

I think yes. Then the latter representation is "minimal" or "folded" while the former is "bigger" or "unfolded".

Consider this circuit:
```
circuit Top:
  module Foo:
    skip
  module Top:
    inst foo1 of Foo
    inst foo2 of Foo
```

If you annotate all `Foo` with `InstanceTarget`s:
```
~Top|Top/foo1:Foo
~Top|Top/foo2:Foo
```

You can then fold these to a `ModuleTarget` without information loss:
```
~Top|Foo
```

If we then say that these two are equivalent, then renaming to delete a complete set of `InstanceTarget`s or a single `ModuleTarget` should produce the same result. How we actually implement that, I'm not sure...

I've thought about this more from the perspective of a cleanup transform _after_ `Dedup` runs where you combine annotations into some "minimal" form. I think for the equivalency of folded and unfolded representations, the behavior you're asking for, @jackkoenig, should work.

azidar:
Re: @albertchen-sifive 
> I think you can replicate the behavior of the AST module using just the update method? If your moduleTarget gets renamed to a bunch of InstanceTargets you could just turn those InstanceTargets back into moduleTargets by using the ofModule fields. for relative instance targets, you could retain the same information if you use an extra target to keep track of the root module and then compute the relative paths in your transform.

I completely agree that this is possible. It was one of my original points with our debate actually. The point remains that if you want to keep the circuit around because you care about the namespace, but you also want sticky behavior, then you can do as you say.

> Changing the meaning of the circuit field seems like a pretty big change to me. in this case, I'd rather use more Targets or a different update method to keep track of the things I'm targeting rather than complicating the Target semantics and the RenameMap logic. And I think circuit is still valuable as a namespacing tool.

The way I see it is its giving more meaning to the concept of 'namespacing', rather than getting rid of it as a namespace. One place where I see it as being useful is when a transform reports a rename, it makes sense to report it regardless of the circuit namespace. For example, say I have targets '~Top|Bar', and '~OtherTop|Bar' , and transform that deletes module `Bar`. We want a way to ensure we delete both of my targets, so how does the transform record the deletion? Without this PR, you would have to delete all possible circuit-target-prefixes, while with this PR you simply delete '~Bar|Bar'.

Re @jackkoenig & @seldridge 
>I have a use case that I want to mention. Say I have some InstanceTarget... I then want to delete module Foo... If I then lookup the rename for my instance target, master (currently 1a03e63) [and this PR] just falls on its face.

I read through all the comments, and I agree with @jackkoenig that this is a bug, given that our current semantics are to support a 'greedy' renaming model. I pushed a bugfix that @jackkoenig can let me know if it works :)

src/main/scala/firrtl/RenameMap.scala
azidar: I added some clarifying comments, but this code is used to stop the partial renaming from occurring. If a partial rename is a self-rename, then it shouldn't necessarily stop because sub-parts of it could be renamed. This is really only necessary for examples like the one in the comment, where you get a circular rename if you rename InstanceTarget -> ModuleTarget, then ModuleTarget -> ModuleTarget.

jackkoenig:
@azidar I believe you have fixed my use case:
```scala
  val top = CircuitTarget("Top")
  val a = top.module("Top").instOf("dut", "DUT").instOf("foo", "Foo")

  def func(delme: Target): Unit = {
    val rm = RenameMap()
    println(s"Deleting ${delme.serialize}")
    rm.delete(delme)
    println(s"${a.serialize} renames to ${rm.get(a)}")
  }
  func(top.module("Foo"))
  func(CircuitTarget("Foo").module("Foo"))
```
This is illustrating what happens if you delete `Foo` in circuit `Top` as well as `Foo` in circuit `Foo`:
```
[info] Deleting ~Top|Foo
[info] ~Top|Top/dut:DUT/foo:Foo renames to Some(List())

[info] Deleting ~Foo|Foo
[info] ~Top|Top/dut:DUT/foo:Foo renames to None
```
As desired, if you delete `Foo` in `Top`, the instance path is deleted.

I also agree that it makes sense to **not** delete the instance path if you delete `Foo` in circuit `Foo` since the instance path exists outside of `Foo`'s AST. That is to say, this makes sense to me!

azidar:
Actually @jackkoenig I disagree, I'm not convinced that deleting `~Foo|Foo` shouldn't delete `~Top|Top/dut:DUT/foo:Foo`. I think you can see `~Foo|Foo` as selecting all Foo's regardless of circuit, and if that is the case, then I think the current behavior is buggy.

jackkoenig:
I'm having an issue with renaming `ModuleName`

On master (eb52a5c)
```scala
import firrtl._
import firrtl.annotations._

case class MyAnno(target: ModuleName) extends SingleTargetAnnotation[ModuleName] {
  def duplicate(n: ModuleName): MyAnno = MyAnno(n)
}

object MyMain extends App {

  val cname = CircuitName("Top")

  val anno = MyAnno(ModuleName("Foo", cname))

  val rm = RenameMap(Map(ModuleName("Foo", cname) -> List(ModuleName("Bar", cname))))

  val res = anno.update(rm)

  println(s"Before: ${anno.serialize}")
  println(s"After : ${res.map(_.serialize)}")
}
```
Prints:
```
[info] Before: MyAnno(ModuleName(Foo,CircuitName(Top)))
[info] After : List(MyAnno(ModuleName(Bar,CircuitName(Top))))
```

With this branch (0ab2379), the same Scala doesn't compile, gets the error:
```
overloaded method value apply with alternatives:
[error]   (associatedTransform: firrtl.Transform)firrtl.RenameMap <and>
[error]   (map: scala.collection.Map[firrtl.annotations.Named,Seq[firrtl.annotations.Named]])firrtl.RenameMap
[error]  cannot be applied to (scala.collection.immutable.Map[firrtl.annotations.ModuleName,List[firrtl.annotations.ModuleName]])
[error]   val rm = RenameMap(Map(ModuleName("Foo", cname) -> List(ModuleName("Bar", cname))))
[error]
```
I'm not really sure why since you didn't remove any `RenameMap.apply` methods, perhaps an implicit conversion issue.

Anyway this doesn't work because `Map` is invariant in the key type parameter, so we can just cast:
```diff
14c14
<   val rm = RenameMap(Map(ModuleName("Foo", cname) -> List(ModuleName("Bar", cname))): Map[Named, Seq[Named]])
---
>   val rm = RenameMap(Map(ModuleName("Foo", cname) -> List(ModuleName("Bar", cname))))
```

With that change, it compiles and runs but prints the following:
```
[info] Before: MyAnno(ModuleName(Foo,CircuitName(Top)))
[info] After : List(MyAnno(ModuleName(Foo,CircuitName(Top))))
```
It didn't rename `Foo`.

Any thoughts on what's up? Is my code wrong?

src/main/scala/firrtl/transforms/Dedup.scala
```diff
@@ -66,35 +92,124 @@ class DedupModules extends Transform with PreservesAll[Transform] {
...
+          (args map (_.serialize))
+        builder ++= "printf(" + (strs mkString ", ") + ")" + info.serialize
+      case other: Statement =>
+        builder ++= ("  " * nindent)
+        builder ++= other.serialize
+    }
+    val builder = new mutable.StringBuilder()
+    serialize(builder, 0)(s)
+    builder.hashCode()
+  }
```
jackkoenig, line 212: Not as part of this PR, but we need to do something like this for normal serialization

src/main/scala/firrtl/annotations/Annotation.scala
```diff
@@ -59,29 +60,52 @@ trait SingleTargetAnnotation[T <: Named] extends Annotation {
...
+              case newTarget: T@unchecked =>
+                try {
+                  duplicate(newTarget)
+                }
+                catch {
+                  case _: java.lang.ClassCastException =>
+                    val msg = s"${this.getClass.getName} target ${target.getClass.getName} " +
+                      s"cannot be renamed to ${newTarget.getClass}"
+                    throw AnnotationException(msg)
+                }
```
jackkoenig, line 96: Can you write some comments describing this logic? And maybe a couple of helper functions would help just by giving names to parts of this larger function.

src/main/scala/firrtl/annotations/transforms/EliminateTargetPaths.scala
```diff
@@ -171,6 +257,47 @@ class EliminateTargetPaths extends Transform {
...
+    val newModule2Original = state.circuit.modules.map {
+      m => ct.module(m.name)
+    }.map {
+      t => (t, renameMap.get(ct.module(t.name)))
+    }.flatMap {
+      case (origMod, Some(Seq(newMod: IsModule))) =>
+        Some(Target.referringModule(newMod).module -> origMod.module)
+      case _ =>
+        None
+    }.toMap
```
jackkoenig, line 274:
```suggestion
    val newModule2Original = state.circuit.modules.flatMap { m =>
      val t = ct.module(m.name)
      renameMap.get(ct.module(t.name))  // isn't t a ModuleTarget? how can we call .name on it?
               .collect { case Seq(newMod: IsModule) =>
                 Target.referringModule(newMod).module -> t.module
               }
   }.toMap
```
Or if you prefer the original, at least do `state.circuit.modules.view.map` so that we don't create all intermediate collections.

src/main/scala/firrtl/annotations/transforms/EliminateTargetPaths.scala
```diff
@@ -171,6 +257,47 @@ class EliminateTargetPaths extends Transform {
...
+    // If previous instance target mapped to a single previously deduped module, return original name
+    // E.g. if previously ~Top|Top/foo:Foo was deduped to ~Top|Top/foo:Bar, then
+    //  Eliminate target paths on ~Top|Top/foo:Bar should rename to ~Top|Top/foo:Foo, not
+    //  ~Top|Top/foo:Bar___Top_foo
+    val newModuleNameMapping = newModule2Original ++ previouslyDeduped.flatMap {
+      case (current: IsModule, (orig: ModuleTarget, idx)) =>
+        renameMap.get(current) match {
+          case Some(Seq(ModuleTarget(_, m))) => Some(m -> orig.name)
+          case _ => None
+        }
```
jackkoenig, line 285:
```suggestion
        renameMap.get(current).collect { case Seq(ModuleTarget(_, m)) => m -> orig.name }
```

jackkoenig (commented):
I still need to review the logic a bit more, but here are some thoughts/questions.

src/main/scala/firrtl/annotations/Annotation.scala
```diff
@@ -59,29 +60,52 @@ trait SingleTargetAnnotation[T <: Named] extends Annotation {
...
 }
 
 /** [[MultiTargetAnnotation]] keeps the renamed targets grouped within a single annotation. */
 trait MultiTargetAnnotation extends Annotation {
-  /** Contains a sequence of [[Target]].
+  /** Contains a sequence of targets.
     * When created, [[targets]] should be assigned by `Seq(Seq(TargetA), Seq(TargetB), Seq(TargetC))`
-    * */
+    */
   val targets: Seq[Seq[Target]]
```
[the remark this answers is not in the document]
azidar: Its so that the targets can keep their one-to-many tracking. So if you are targeting Seq(Seq(foo), Seq(bar)), it can expand to Seq(Seq(foo_x, foo_y), Seq(bar_x, bar_y)). I think it could probably be a def.

src/main/scala/firrtl/RenameMap.scala
```diff
@@ -399,7 +414,22 @@ final class RenameMap private (val underlying: mutable.HashMap[CompleteTarget, S
...
+      //   Renaming (~Top|Top/a:A) should return (~Top|A), not (~Top|B)
+      // This is due to our splitting recursive renaming into three parts (top, mid, ast, bot)
+      // We continue renaming even if original == ts.head because if it returns the same target, there are
+      //   still additional renaming of different parts of the target which must occur. If the target actually
+      //   should not be renamed, it must go through all partial renaming parts to actually determine that fact.
+      def continueRenaming[T <: CompleteTarget](original: CompleteTarget, ts: Seq[CompleteTarget]): Boolean =
+        ts match {
+          case Seq(t: T @unchecked) if original != t => false
+          case _ => true
+        }
```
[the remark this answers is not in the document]
azidar: Forgot to add the classtag, good catch :)

PR #1445, opened 2020-03-12
fix transform Inline Nots

ingallsj:
### Contributor Checklist
- [X] Did you add Scaladoc to every public function/method?
- [X] Did you add at least one test demonstrating the PR?
- [X] Did you delete any extraneous printlns/debugging code?

### Type of Improvement
- bug fix
- backend code generation

### API Impact
none

Fixes https://github.com/freechipsproject/firrtl/pull/1422 by checking that the type (including width) is the same for both the LHS and any RHS terms inlined.
Re-applies https://github.com/freechipsproject/firrtl/pull/1270 to latest master head.

### Desired Merge Strategy
- Squash: The PR will be squashed and merged (choose this if you have no preference).

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/AddDescriptionNodes.scala
+++ b/src/main/scala/firrtl/AddDescriptionNodes.scala
@@ -79,6 +79,7 @@ class AddDescriptionNodes extends Transform with PreservesAll[Transform] {
          Dependency[firrtl.transforms.InlineBitExtractionsTransform],
          Dependency[firrtl.transforms.PropagatePresetAnnotations],
          Dependency[firrtl.transforms.InlineCastsTransform],
+         Dependency[firrtl.transforms.InlineNotsTransform],
          Dependency[firrtl.transforms.LegalizeClocksTransform],
          Dependency[firrtl.transforms.FlattenRegUpdate],
          Dependency(passes.VerilogModulusCleanup),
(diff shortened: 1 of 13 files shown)


johningalls-sifive:
@albert-magyar sorry that you had to deal with this in https://github.com/freechipsproject/firrtl/pull/1422 !
@jackkoenig thanks for the help with the solution.

johningalls-sifive:
@albert-magyar agree with everything you said about wanting to be safe and conservative composing these inlining transforms.

> This check can be expressed by simply not recursing into children whose type does not match that of their parent.

I put in a check that does this (I think) in commit e847505 commented with "do not recurse into children whose type does not match that of their parent".  Is that right?

albert-magyar:
> I put in a check that does this (I think) in commit [e847505](https://github.com/freechipsproject/firrtl/commit/e847505a01582e119ee04fc037ff8cc5b37e5296) commented with "do not recurse into children whose type does not match that of their parent". Is that right?

It looks like the right flavor of check in the case statement, but I suspect that it doesn't have the intended effect since the match happens after the recursive call and `rhs` would actually return the recursively-modified expression. I could actually be totally mistaken based on the specific low-level details of how you've written it, though, since I didn't look much beyond the diff of e847505.

One alternative would be to return `expr` instead of `rhs` in that case, which would have the effect of doing all the recursive modifications but then "throwing them away" and going back to the original expression tree if a width change is encountered on the way back up the tree. That might be less efficient than just not doing the recursion, but that is a negligible factor given how shallow the expression trees are after `SplitExpressions`, and it would be a very simple code change.

johningalls-sifive:
> One alternative would be to return `expr` instead of `rhs`.

Ah, gotcha, okay, I made this a little clearer / more overt in c837550.

ingallsj:
subsumed by https://github.com/freechipsproject/firrtl/pull/1817

PR #1461, opened 2020-03-18
Change annotation logging

jackkoenig:
* Change from log-level debug to trace
* Serialize as JSON rather than .serialize on each annotation

### Contributor Checklist

### Type of Improvement

- logging cleanup

### API Impact

It doesn't affect any Scala code, but changes the logging format. When annotations were new, it made sense to log them at the `debug` level because it was far less verbose than the circuit, these days they are extremely verbose. I also changed it to use JSON serialization instead of the relatively unused pretty printing `serialize` method. This makes it possible to take the output of logging and feed it back into the compiler which is very useful.

No impact

### Desired Merge Strategy

Don't Care

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/stage/transforms/UpdateAnnotations.scala
+++ b/src/main/scala/firrtl/stage/transforms/UpdateAnnotations.scala
@@ -3,7 +3,7 @@
 package firrtl.stage.transforms
 
 import firrtl.{AnnotationSeq, CircuitState, RenameMap, Transform, Utils}
-import firrtl.annotations.{Annotation, DeletedAnnotation}
+import firrtl.annotations.{Annotation, DeletedAnnotation, JsonProtocol}
 import firrtl.options.Translator
 
 import scala.collection.mutable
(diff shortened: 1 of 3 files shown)


src/main/scala/firrtl/stage/transforms/UpdateAnnotations.scala
```diff
@@ -22,8 +22,8 @@ class UpdateAnnotations(val underlying: Transform) extends Transform with Wrappe
 
     logger.info(s"Form: ${result.form}")
 
-    logger.debug(s"Annotations:")
-    remappedAnnotations.foreach( a => logger.debug(a.serialize) )
+    logger.trace(s"Annotations:")
+    logger.trace(JsonProtocol.serialize(remappedAnnotations))
```
jackkoenig: The code in this file is basically duplicated from `Compiler.scala`, @seldridge do you think it would be reasonable to move these into some common internal utility?

src/main/scala/firrtl/stage/transforms/UpdateAnnotations.scala
```diff
@@ -22,8 +22,8 @@ class UpdateAnnotations(val underlying: Transform) extends Transform with Wrappe
 
     logger.info(s"Form: ${result.form}")
```
chick: Seems like you might want to remove this blank line to make it exactly like the code in Compiler.scala. IntelliJ is catching more and more duplicated code, the more the same the more easily matched.
```suggestion
```

jackkoenig:
> Going forward, we need better logging that is actually useful. I agree that annotation logging is currently more of a last-ditch tracing effort.

Importantly, this vastly reduces the noise in log-level debug such that the current debug prints might actually be useful (like in Dedup and DCE)

PR #1470, opened 2020-03-24
Eliminate warnings on `sbt doc` and `sbt unidoc`

chick:
- removed toFirrtl reference on MultiTargetAnnotation
- lots of places where package path has to be added to comment references
- Change to use `/** text starts here` convention when wrong in comment with a doc fix.
  - Did not exhaustively change these
- Wrestled doc example in RenderDiGraph#renderNode, not sure if I won
- Cleaned up InferWidths & CatchExceptions imports
- Added missing license message to a couple of files.
- fixed a couple of stale parameter names in scaladoc
- Added @unchecked to stop erasure warning in Emitting where emission annotations are collected
- Change types to [_] on match in RenameMap#recordAll to fix erasure warning

### Contributor Checklist

- [ ] Did you review?


chick:
@jackkoenig Task generated by meeting yesterday. scaladoc is painfully finicky.

seldridge:
@chick: #1471 enforces no future warnings in Scaladoc via `-Xwarnings-fatal`. This requires turning on a few additional flags to remove feature warnings. Additionally, I'd like to get this into CI so that documentation always builds warning-free. This is, admittedly, draconian, but documentation is serious business.

src/main/scala/firrtl/transforms/InlineBitExtractions.scala
```diff
@@ -35,10 +37,10 @@ object InlineBitExtractionsTransform {
     case _ => expr // Not a candidate
   }
 
-  /** Mapping from references to the [[firrtl.ir.Expression Expression]]s that drive them */
+  /** Mapping from references to the [[firrtl.ir.Expression]]s that drive them */
```
jackkoenig: I actually prefer this one how it was, that's intentional syntax so that the text shows up as `Expression` with the correct link rather than as `firrtl.ir.Expression`, same with the ones below (a couple of which may have been broken because they went over multiple lines.

chick: I'll go back through and check, I think at one point there seemed to be situation where that form was causing a warning, but it could have been a mixed up with some other idiosyncrasy of scaladoc

chick: That seems ok, it does not work however in situations where this works
```
    * @throws scala.IllegalArgumentException if vprime is not a subset of V
```
but this gives a warning
```
    * @throws [[scala.IllegalArgumentException IllegalArgumentException]] if vprime is not a subset of V
```

chick: I updated all the files I changed to use your preferred form

seldridge: The `@throws ExceptionName` is the correct, undocumented syntax. The first argument is special and isn't supposed to be a link.

Unrelated, I usually write this as the underlying Java exception as opposed to the Scala type alias. That doesn't matter, though.

PR #1504, opened 2020-04-10
EliminateTargetPaths: don't duplicate modules with only one instance

albertchen-sifive:
### Contributor Checklist

This PR prevents `EliminateTargetPaths` from renaming modules with only a single instance. This is done by adding a prepocessing step that shortens target paths if they contain inst/ofMod pairs that are lone instances.

example:
```
circuit Foo:
  module Baz:
    skip
  module Bar:
    inst baz of Baz
    inst bazzz of Baz
  module Foo:
    inst bar of Bar
```
resolving `~Foo|Foo/bar:Bar/baz/Baz` and `~Foo|Foo/bar:Bar/baz/Bazzz` will create two `RenameMap`s. The first one renames
```
~Foo|Foo/bar:Bar/baz:Baz -> ~Foo|Bar/baz:Baz
~Foo|Foo/bar:Bar/bazzz:Baz -> ~Foo|Bar/bazzz:Baz
```
Then the usual duplication happens on `~Foo|Bar/baz:Baz` and `~Foo|Bar/bazzz:Baz`, which will only try to duplicate `Baz`. Finally, the two `RenameMap`s are chained together.

#### Type of Improvement

bug fix: this [unit test](https://github.com/freechipsproject/firrtl/blob/master/src/test/scala/firrtlTests/annotationTests/EliminateTargetPathsSpec.scala#L401-L418) for `EliminateTargetPaths` does not actually check anything. Running it shows that it the property it should be testing does not actually hold. `EliminateTargetPaths` turns this:
```
circuit Foo:
  module Baz:
    skip
  module Bar:
    inst baz of Baz
    skip
  module Foo:
    inst bar of Bar
```

into this:
```
circuit Foo :
  module Baz___Foo_bar_baz :
    skip
  module Bar___Foo_bar :
    inst baz of Baz___Foo_bar_baz
    skip
  module Foo :
    inst bar of Bar___Foo_bar
```

#### API Impact

Should not affect the current API, aside from changing module names.

It should prevent modules from being renamed unnecessarily to long module path names.

#### Desired Merge Strategy

Squash

#### Release Notes

don't duplicate lone instance modules in EliminateTargetPaths

- [x] Did you review?


mwachs5:
What does the fixed test do in this case now?

albertchen-sifive:
> What does the fixed test do in this case now?

it checks that the circuit remains unchanged https://github.com/albertchen-sifive/firrtl/blob/fix-EliminateTargetPaths-lone-instances/src/test/scala/firrtlTests/annotationTests/EliminateTargetPathsSpec.scala#L426 since all modules have only a single instance

seldridge:
That's a nice test I added... :woman_shrugging: 

What's the behavior with: https://github.com/freechipsproject/firrtl/pull/1442 ?

jackkoenig:
> resolving ~Foo|Foo/bar:Bar/baz/Baz and ~Foo|Foo/bar:Bar/baz/Bazzz will create two RenameMaps.

Does this mean that you get 1 RenameMap per instance? If I resolve all of the following: 
```
~Foo|Foo/a1:A
~Foo|Foo/a2:A
~Foo|Foo/b1:B
~Foo|Foo/b2:B
```
How many RenameMaps do I get? 2 or 4?

albertchen-sifive:
> How many RenameMaps do I get? 2 or 4?

sorry, wording is a bit weird. it will be at most 2 RenameMaps. first RenameMap has all the renames for single instance paths, second one is the duplication RenameMap. if there are no renames in the first pass, then just the duplication map will be used.

> What's the behavior with: #1442 ?

haven't read all of that PR, and I have to admit I still don't fully understand the AST module stuff. but I think this should still work fine with that PR? not sure

jackkoenig:
> sorry, wording is a bit weird. it will be at most 2 RenameMaps. first RenameMap has all the renames for single instance paths, second one is the duplication RenameMap. if there are no renames in the first pass, then just the duplication map will be used.

Got it, great.

This looks like a bug fix, is that correct? Is the only side effect of this PR that modules with only a single instance do not get duplicated in `EliminateTargetPaths`?

albertchen-sifive:
> This looks like a bug fix, is that correct? Is the only side effect of this PR that modules with only a single instance do not get duplicated in `EliminateTargetPaths`?

yes

PR #1528, opened 2020-04-14
Create new scopes for child statements of Conditionally

### LINKED ISSUE #1505: FIRRTL accepts invalid code
### Checklist

- [ ] Did you describe your environment? NA
- [ ] Did you specify relevant external information? NA

```                                                                                                                                                                                                                             
circuit MyModule : 
  module MyModule : 
    input clock : Clock
    input reset : UInt<1>
    output io : {flip en : UInt<1>, out : UInt<8>, wrap : UInt<1>}
    
    reg count : UInt<3>, clock with : (reset => (reset, UInt<3>("h00")))
    when io.en :
      node _T = eq(count, UInt<3>("h07"))
      node _T_1 = add(count, UInt<1>("h01"))
      node _T_2 = tail(_T_1, 1)
      count <= _T_2
      skip 
    node wrap = and(io.en, _T)
    io.out <= count
    io.wrap <= wrap 
```
Given the above, FIRRTL will generate Verilog just fine.

It should error, because `node wrap = and(io.en, _T)` refers to `_T` which is defined inside the `when` scope

### Steps to Reproduce

Yes, this comes from `chisel3.util.Counter` (lol whoops):
https://scastie.scala-lang.org/dtewI9H4R8mfdKRgxekx3A

albert-magyar:
**Type of change:** bug fix
**API impact:** changes what FIRRTL is accepted by the compiler
**Backend code-generation impact:** none for circuits that are still accepted
**Desired merge strategy:** merge
**Release notes:**
Unlike prior versions of FIRRTL, which did not implement new block scopes for the consequent and alternate child statements of a Conditionally (when) statement, declarations inside a child scope are not visible in the parent scope. This aligns the implementation with the specification.

* Fixes #1505 
* UPDATE: I added a behavior to support current CHIRRTL mem emission while being otherwise as strict as possible. It does rely on the Counter fix in freechipsproject/chisel3#1408, so it should still be on a major version.
* We can drop the adapter for 1.4.0 with the inferred mport re-implementation

~~I left in the failing tests as they illustrate the fact that CHIRRTL memory emission relies on the same buggy pattern as the Counter previously did. When an `mport` is declared inside a when, it is referenced outside the enclosing scope in several tests, which reflects a (now-illegal) idiom for unconditionally viewing the result of a read. However, I did manually fixup the `RightShiftExecutionTest` to reflect the fix for `Counter` in freechipsproject/chisel3#1408.~~

~~Since this would cause huge backwards incompatibilities at the CHIRRTL level, it probably makes sense to combine this with the upcoming CHIRRTL `mport` refactor and define a new pattern that doesn't rely on out-of-scope references.~~

I blended a little immutable management of the scope levels into the code to try to encapsulate things without having to refactor too much of `CheckHighForm`. It can benchmark this against the alternate approach of a flat mutable set with explicit deletion on scope exit, but I don't suspect either will be a bottleneck.

### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/passes/CheckHighForm.scala
+++ b/src/main/scala/firrtl/passes/CheckHighForm.scala
@@ -22,6 +22,10 @@ trait CheckHighFormLike { this: Pass =>
       moduleNS += name
       scopes.head += name
     }
+    def expandMPortVisibility(port: CDefMPort): Unit = {
+      // Legacy CHIRRTL ports are visible in any scope where their parent memory is visible
+      scopes.find(_.contains(port.mem)).getOrElse(scopes.head) += port.name
+    }
     def legalDecl(name: String): Boolean = !moduleNS.contains(name)
     def legalRef(name: String): Boolean = scopes.exists(_.contains(name))
     def childScope(): ScopeView = new ScopeView(moduleNS, new NameSet +: scopes)
@@ -243,7 +247,10 @@ trait CheckHighFormLike { this: Pass =>
         case sx: Connect => checkValidLoc(info, mname, sx.loc)
         case sx: PartialConnect => checkValidLoc(info, mname, sx.loc)
         case sx: Print => checkFstring(info, mname, sx.string, sx.args.length)
-        case _: CDefMemory | _: CDefMPort => errorOnChirrtl(info, mname, s).foreach { e => errors.append(e) }
+        case _: CDefMemory => errorOnChirrtl(info, mname, s).foreach { e => errors.append(e) }
+        case mport: CDefMPort =>
+          errorOnChirrtl(info, mname, s).foreach { e => errors.append(e) }
+          names.expandMPortVisibility(mport)
         case sx => // Do Nothing
       }
       s foreach checkHighFormT(info, mname)
--- a/src/main/scala/firrtl/passes/CheckHighForm.scala
+++ b/src/main/scala/firrtl/passes/CheckHighForm.scala
@@ -12,6 +12,21 @@ import firrtl.options.Dependency
 trait CheckHighFormLike { this: Pass =>
   type NameSet = collection.mutable.HashSet[String]
 
+  private object ScopeView {
+    def apply(): ScopeView = new ScopeView(new NameSet, List(new NameSet))
+  }
+
+  private class ScopeView private (moduleNS: NameSet, scopes: List[NameSet]) {
+    require(scopes.nonEmpty)
+    def declare(name: String): Unit = {
+      moduleNS += name
+      scopes.head += name
+    }
+    def legalDecl(name: String): Boolean = !moduleNS.contains(name)
+    def legalRef(name: String): Boolean = scopes.exists(_.contains(name))
+    def childScope(): ScopeView = new ScopeView(moduleNS, new NameSet +: scopes)
+  }
+
   // Custom Exceptions
   class NotUniqueException(info: Info, mname: String, name: String) extends PassException(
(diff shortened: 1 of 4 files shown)


albert-magyar:
As a compromise, I think it might be reasonable to include this change sooner, but with an exception in `CheckChirrtl` (but not `CheckHighForm`) that acts as a "compatibility mode," allowing the old behavior for **inferred mem ports only.**

We could remove that exception in 1.4.0 with the inferred mem port rework.

albert-magyar:
I added an implementation of that "scope usage adapter" compromise that is as strict as possible while accommodating current chisel3 code emission: an inferred `mport` is automatically made visible in the scope in which its parent memory is declared.

This adapter lets current chisel3 code (after the Counter change) work while making it a one-liner to drop the adapter, presumably with a major FIRRTL and Chisel release.

azidar:
Try running with rocket-chip?

ekiwi-sifive:
@albert-magyar Should one be able to shadow variables from the outerscope inside of a conditionally scope?
I.e. should this throw an error?

```
output a : UInt<1>
output b: UInt<1>

when a:
  node b = not(a)
  a <= b
```

albert-magyar:
> @albert-magyar Should one be able to shadow variables from the outerscope inside of a conditionally scope?
> I.e. should this throw an error?
> 
> ```
> output a : UInt<1>
> output b: UInt<1>
> 
> when a:
>   node b = not(a)
>   a <= b
> ```

This is one of the things that is not yet settled. The language of the spec loosely implies that shadowing is not allowed (i.e., shadowing would inherently prevent the following sentence from the spec from being entirely true).

> It is an error to refer to any component declared within a branch after the branch has ended.

ekiwi-sifive:
OK. The reason that I was asking is because I think that if we allowed shadowing, then dedup would be broken.

ekiwi-sifive:
> This is one of the things that is not yet settled. The language of the spec loosely implies that shadowing is not allowed (i.e., shadowing would inherently prevent the following sentence from the spec from being entirely true).

Should we add some wording about the different namespaces in firrtl? The ones I can think of are:
- a global namespace for module names
- a global namespace for width and bound variables
- a module level namespace for register, memories, wires, nodes, instances and ports (anything else?)
- a bundle level namespace for field names


I think we can distinguish namespace from scope, right? Like a namespace means that names have to be unique and a scope defines the name visibility.

jackkoenig:
> a global namespace for width and bound variables

I believe these are an implementation detail of width inference so I'm not sure this one is necessary

ekiwi-sifive:
> I believe these are an implementation detail of width inference so I'm not sure this one is necessary

Yes you are right, I guess the `VarWidth` nodes are serializable but they are not mentioned in the current firrtl spec.
I just recently was wondering what the correct namespace for these vars is and it took me a lot of time to find out.

jackkoenig:
> I just recently was wondering what the correct namespace for these vars is and it took me a lot of time to find out.

Yeah absolutely; and noting that namespace somewhere (as we did on the hashing PR) is very useful.

ekiwi-sifive:
Hi @albert-magyar, could you try if rebasing this change on master fixes the number of commits and changes that Github displays?

albert-magyar:
> Hi @albert-magyar, could you try if rebasing this change on master fixes the number of commits and changes that Github displays?

Github does some weird things sometimes :)

albert-magyar:
freechipsproject/chisel3#1518 now has elaboration time checks for this

ekiwi-sifive:
@azidar says its fine to merge!

Issue #1298, opened 2020-01-08
Module inlining runs after optimizations


jackkoenig:
This can prevent lots of optimizations. It's only problematic if you have multiple instances of the module being inlined which prevents cross-module Constant Propagation.

This should be trivial to fix with the Dependency API (see https://github.com/freechipsproject/firrtl/pull/1123) assuming there isn't any good reason for it running after optimizations.

### Checklist

- [x] Did you specify the current behavior?
- [x] Did you specify the expected behavior?
- [ ] Did you provide a code example showing the problem?
- [ ] Did you describe your environment?
- [x] Did you specify relevant external information?

### What is the current behavior?

Inlining runs after optimizations

### What is the ~~expected~~ desired behavior?

Inlining runs before optimizations

### Steps to Reproduce

Have a module instantiated multiple times marked for inlining that has constants driving its ports for only one instance.

### External Information

Nope

seldridge:
Note: this is property of *any* `inputForm=LowForm` transform. Slightly more accurately, low form transforms will run right before the emitter. Optimizations then only run if optimizations are requested. The behavior changes for low FIRRTL emitter, minimum verilog emitter, or verilog/system verilog emitter. :woman_shrugging: 

See https://github.com/freechipsproject/firrtl/issues/1165 for more info trying to document this.

Note that #1123 preserves this behavior if dependencies are specified using input/outputForm.

jackkoenig:
Ah yeah, thanks for the clarification @seldridge! I was just noting this and hopefully we can resolve this issue by leveraging your hard work 🙂

seldridge:
I think this is fixed on master. There's [no `optionalPrerequisite` defined for `InlineIntsances` now](https://github.com/freechipsproject/firrtl/blob/master/src/main/scala/firrtl/passes/Inline.scala#L29). Consequently, this runs before optimizations now:


```bash
./utils/bin/firrtl -i regress/RocketCore.fir --compiler verilog --target-dir test_run_dir --log-level info --inline RocketCore.IBuf
# Computed transform order in: 516.5 ms
# Determined Transform order that will be executed:
#   firrtl.stage.transforms.Compiler
#   ├── firrtl.passes.CheckChirrtl$
#   ├── firrtl.passes.CInferTypes$
#   ├── firrtl.passes.CInferMDir$
#   ├── firrtl.passes.RemoveCHIRRTL$
#   ├── firrtl.passes.ToWorkingIR$
[...]

Issue #1462, opened 2020-03-19
Parallel testing deadlock in environments with high degrees of parallelism


ucbjrl:
### Checklist

- [x] Did you specify the current behavior?
- [x] Did you specify the expected behavior?
- [ ] Did you provide a code example showing the problem?
- [x] Did you describe your environment?
- [x] Did you specify relevant external information?

### What is the current behavior?
```
$ sbt test
```
on the millennium machines will usually hang with multiple threads (according to `jdb`) looping in:
```
  [1] firrtl.transforms.DedupModules.<init> (Dedup.scala:46), pc = 21
  [2] firrtl.CompilerUtils$.getLoweringTransforms (Compiler.scala:409), pc = 142
```
Further investigation indicated that (according to `jdb`):
```
firrtl.stage.Forms$ = null
```
suggesting that `Forms` members were being accessed without the initialization of the enclosing object. The assumption is that although object initialization is protected against simultaneous execution by multiple threads, no such guarantee may be made about member references.

### What is the expected behavior?
```
$ sbt test
```
should eventually complete.

### Steps to Reproduce

Invoke `sbt test` on a system with a significant number of processors (a13.millennium.berkeley.edu has 24), and a [`build.sbt`](https://github.com/freechipsproject/firrtl/blob/960ea3bf75b8099fc1158b78c8554effb48e2bf5/build.sbt)configured for parallel test execution.

### Your environment

- Chisel Verions: N/A
- OS: Linux lxd-a13 4.15.0-47-generic #50-Ubuntu SMP Wed Mar 13 10:44:52 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
- Verilator version: Verilator 4.016 2019-06-16 rev UNKNOWN_REV

### External Information

This may be related to freechipsproject/chisel3#1053

ucbjrl:
Closed by #1463

PR #1534, opened 2020-04-20
Trait-base Dependency API Migration

seldridge:
This enables a smooth migration from `CircuitForm`-based dependency specification to Dependency API dependency specification. 

This adds a new trait `DependencyAPIMigration` that provides default implementations of Dependency API methods and `final override`s `inputForm`/`outputForm` to `UnknownForm`.

**Why is this PR touching so much stuff???**

Legacy infrastructure like `firrtl.Compiler`, `SeqTransform`, and `CoreTransform` (lowering compilers) and legacy testing infrastructure use legacy scheduling methods `getLoweringTransforms` and `mergeTransforms`. These methods cannot schedule transforms which have input/output forms of `UnknownForm`. Consequently, mixing in `DependencyAPIMigration` breaks **a ton** of stuff everywhere. To rectify this, this PR makes updates to legacy compiler infrastructure to have these use the Dependency API-based `firrtl.stage.transforms.Compiler` (a `TransformManager`) under the hood.

Closes https://github.com/freechipsproject/firrtl/pull/1533.

### Contributor Checklist

#### Type of Improvement

- code refactoring

- new feature/API

#### API Impact

This is both an expansion of the API and a modification of the API.

None.

#### Desired Merge Strategy

- [x] Did you review?

### DIFF
--- a/src/main/scala/firrtl/Compiler.scala
+++ b/src/main/scala/firrtl/Compiler.scala
@@ -13,6 +13,7 @@ import firrtl.ir.Circuit
 import firrtl.Utils.throwInternalError
 import firrtl.annotations.transforms.{EliminateTargetPaths, ResolvePaths}
 import firrtl.options.{DependencyAPI, Dependency, PreservesAll, StageUtils, TransformLike}
+import firrtl.stage.Forms
 import firrtl.stage.transforms.CatchCustomTransformExceptions
 
 /** Container of all annotations for a Firrtl compiler */
@@ -100,7 +101,9 @@ object CircuitState {
   * strictly supersets of the "lower" forms. Thus, that any transform that
   * operates on [[HighForm]] can also operate on [[MidForm]] or [[LowForm]]
   */
-@deprecated("CircuitForm will be removed in 1.3. Switch to Seq[TransformDependency] to specify dependencies.", "1.2")
+@deprecated(
+  "Mix-in the DependencyAPIMigration trait into your Transform and specify its Dependency API dependencies. See: https://bit.ly/2Voppre",
+  "FIRRTL 1.3")
 sealed abstract class CircuitForm(private val value: Int) extends Ordered[CircuitForm] {
   // Note that value is used only to allow comparisons
   def compare(that: CircuitForm): Int = this.value - that.value
@@ -119,7 +122,9 @@ sealed abstract class CircuitForm(private val value: Int) extends Ordered[Circui
   *
   * See [[CDefMemory]] and [[CDefMPort]]
   */
-@deprecated("Form-based dependencies will be removed in 1.3. Please migrate to the new Dependency API.", "1.2")
+@deprecated(
+  "Mix-in the DependencyAPIMigration trait into your Transform and specify its Dependency API dependencies. See: https://bit.ly/2Voppre",
+  "FIRRTL 1.3")
 final case object ChirrtlForm extends CircuitForm(value = 3) {
   val outputSuffix: String = ".fir"
 }
@@ -131,7 +136,9 @@ final case object ChirrtlForm extends CircuitForm(value = 3) {
   *
   * Also see [[firrtl.ir]]
   */
-@deprecated("Form-based dependencies will be removed in 1.3. Please migrate to the new Dependency API.", "1.2")
+@deprecated(
+  "Mix-in the DependencyAPIMigration trait into your Transform and specify its Dependency API dependencies. See: https://bit.ly/2Voppre",
+  "FIRRTL 1.3")
 final case object HighForm extends CircuitForm(2) {
   val outputSuffix: String = ".hi.fir"
 }
@@ -143,7 +150,9 @@ final case object HighForm extends CircuitForm(2) {
   *  - All whens must be removed
   *  - There can only be a single connection to any element
   */
-@deprecated("Form-based dependencies will be removed in 1.3. Please migrate to the new Dependency API.", "1.2")
+@deprecated(
+  "Mix-in the DependencyAPIMigration trait into your Transform and specify its Dependency API dependencies. See: https://bit.ly/2Voppre",
+  "FIRRTL 1.3")
 final case object MidForm extends CircuitForm(1) {
   val outputSuffix: String = ".mid.fir"
 }
@@ -154,7 +163,9 @@ final case object MidForm extends CircuitForm(1) {
   *  - All aggregate types (vector/bundle) must have been removed
   *  - All implicit truncations must be made explicit
   */
-@deprecated("Form-based dependencies will be removed in 1.3. Please migrate to the new Dependency API.", "1.2")
+@deprecated(
+  "Mix-in the DependencyAPIMigration trait into your Transform and specify its Dependency API dependencies. See: https://bit.ly/2Voppre",
+  "FIRRTL 1.3")
 final case object LowForm extends CircuitForm(0) {
   val outputSuffix: String = ".lo.fir"
 }
@@ -170,7 +181,9 @@ final case object LowForm extends CircuitForm(0) {
   * TODO(azidar): Replace with PreviousForm, which more explicitly encodes
   * this requirement.
   */
-@deprecated("Form-based dependencies will be removed in 1.3. Please migrate to the new Dependency API.", "1.2")
+@deprecated(
+  "Mix-in the DependencyAPIMigration trait into your Transform and specify its Dependency API dependencies. See: https://bit.ly/2Voppre",
+  "FIRRTL 1.3")
 final case object UnknownForm extends CircuitForm(-1) {
   override def compare(that: CircuitForm): Int = { sys.error("Illegal to compare UnknownForm"); 0 }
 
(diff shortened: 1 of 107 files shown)


src/main/scala/firrtl/Compiler.scala
```diff
@@ -100,7 +102,9 @@ object CircuitState {
   * strictly supersets of the "lower" forms. Thus, that any transform that
   * operates on [[HighForm]] can also operate on [[MidForm]] or [[LowForm]]
   */
-@deprecated("CircuitForm will be removed in 1.3. Switch to Seq[TransformDependency] to specify dependencies.", "1.2")
+@deprecated(
+  "Mix-in the DependencyAPIMigration trait into your Transform and specify it's Dependency API dependencies. See: https://bit.ly/2Voppre",
```
albert-magyar, line 106: s/it's/its here and other deprecation warnings

seldridge: Good catch. Thanks for checking the text. Incorrect grammar in the deprecation message isn't great. :grimacing: Fixed in the rebase.

seldridge:
@albert-magyar: The failure to migrate those three transforms was just missed on my part. Thanks for finding those. Rebased and updated.

The `Emitter`s have not been migrated, yet. I think it's reasonable to leave those as those really need a larger refactor. That can happen for 1.4.

Also, the notion of a `FirrtlCompilerAnnotation` needs to be rethought for 1.4. In terms of implementation, this is the closest thing we have to running a group of transforms.

src/main/scala/firrtl/passes/clocklist/ClockListTransform.scala
```diff
@@ -51,9 +51,11 @@ Usage:
...
 }
 
-class ClockListTransform extends Transform with RegisteredTransform {
-  def inputForm = LowForm
-  def outputForm = LowForm
+class ClockListTransform extends Transform with DependencyAPIMigration with RegisteredTransform {
+
+   override lazy val prerequisites = Forms.LowForm
+   override lazy val optionalPrerequisites = Seq.empty
+   override lazy val dependents = Forms.LowEmitters
```
jackkoenig, line 58: Is there a reason why sometimes you use `val` and sometimes `lazy val`? I actually think in all of these cases `def` would be _marginally_ better since the cost per call is very low. However, since we make so few transform objects I think the overhead of the `val` in the object is irrelevant anyway.

src/main/scala/firrtl/passes/wiring/WiringTransform.scala
```diff
@@ -36,9 +38,16 @@ case class SinkAnnotation(target: Named, pin: String) extends
...
+class WiringTransform extends Transform with DependencyAPIMigration {
+
+  override lazy val prerequisites = Forms.MidForm
+  override lazy val optionalPrerequisites = Seq.empty
+  override lazy val dependents = Forms.MidEmitters
+
+  override def invalidates(a: Transform): Boolean = {
+    val everything = new mutable.LinkedHashSet[Dependency[Transform]] ++ Forms.VerilogOptimized
+    (everything -- Forms.MinimalHighForm)(Dependency.fromTransform(a))
+  }
```
jackkoenig, line 50:
```suggestion
    val myInvalidates = Forms.VerilogOptimized.toSet -- Forms.MinimalHighForm
    myInvalidates.contains(Dependency.fromTransform(a))
  }
```
Can you describe what this is doing a little? Also is there any point to using a `LinkedHashSet` since we're ultimately just doing a lookup?

src/main/scala/firrtl/passes/clocklist/ClockListTransform.scala
albert-magyar: I think `val` has the advantage of being consistent with the restriction that it be a constant value without relying on implicit conventions surrounding idempotent `def`s that have never really been strictly adhered to. But since this isn't enforced at the API level, the benefit of that consistency is negligible.

src/main/scala/firrtl/passes/wiring/WiringTransform.scala
albert-magyar: To follow on to this, the set gets re-created one every `invalidates` call.

It might be worth providing a trait that lets users express their invalidates as a set of `Dependency[Transform]`, since that seems like it will be a common pattern.

src/main/scala/firrtl/passes/clocklist/ClockListTransform.scala
seldridge: Inconsistency on my part. I probably intended to always use `lazy val`.

Note: these fields are all shadowed by `private[options] lazy val`s. The `DependencyManager` is supposed to be using these to avoid a bunch of repeated conversions to sets. I looked and I'm actually not doing this.

So, what do you think? Use `def` everywhere? I'm fine with that.

ucbjrl: and don't forget the issues we've had with sharing data structures across multiple threads.

src/main/scala/firrtl/passes/wiring/WiringTransform.scala
jackkoenig: I was thinking the same thing and actually toyed around in Scastie a little, you could define `myInvalidates` as a private field in the class. I think it just depends on how often it gets called. Is the memory cost of the `Set` worth keeping around for the full compilation? I'm thinking no if this function only gets called 1-2 times per object, but if it's several times it may become worth it.

seldridge: What this is saying is that wiring invalidates everything in the locally observable universe except `ChirrtlToHighFirrtl`.

Because this is `inputForm=MidForm`/`outputForm=HighForm` this means that this, per the original algorithm in `mergeTransforms`, needs to rerun `IRToWorkingIR`, `ResolveAndCheck`, and `HighFirrtlToMiddleFirrtl`. 

Currently, `firrtl.stage.Forms` describes lower forms as supersets of higher forms (`Forms.LowForm` is `Forms.MidForm` with additional transforms), so this PR constructs a bigger set and removes what it doesn't invalidate. This PR is the same as [this case](https://github.com/freechipsproject/firrtl/blob/master/src/main/scala/firrtl/Compiler.scala#L258).

Would it be useful to document this on the method?

src/main/scala/firrtl/passes/clocklist/ClockListTransform.scala
jackkoenig: I *think* that's only an issue on `object`s right? Should be fine since this is a `class` right?

ucbjrl: My guess is any mutable field in a class (def or lazy) is an issue if instances of that class are shared between threads.

ucbjrl: Effectively, the thread that issues the first reference will end up defining the value. This may (or may not) be what you want.

src/main/scala/firrtl/passes/wiring/WiringTransform.scala
seldridge: @albert-magyar: a trait to specify invalidates this way makes sense. The main motivation for allowing for function-based sets with invalidates as it seemed like a representation of the universal set was required (a transform should be able to invalidate everything).

The `invalidates` method will get called inside the `DependencyManager` everything this transform runs. It seems reasonable to go from `O(n)` object creations to `O(1)` object creations. (A `DependencyManager` only ever constructs one instance of a transform.)

I'll update this to use a `private val`.

src/main/scala/firrtl/passes/clocklist/ClockListTransform.scala
albert-magyar: I think there are two separate issues we're conflating here:

- the issue with ordering initialization in multi-threaded contexts (#1462, causes _hangs_ but not inconsistencies). This is a [dead,live]lock issue with _all_ members, not just collections / data structures.

- thread-unsafe modifications to mutable data structures (see #1057). This is a data race that causes inconsistencies and is only an issue with mutable data structures. This is also simple to fix with Java threading modifiers.

src/main/scala/firrtl/passes/wiring/WiringTransform.scala
jackkoenig: Note that `Set[A]` extends `A => Boolean` so I think this `Dependency[Transform]` options should be `Dependency[Transform] => Boolean` and then could be implemented as a `Set`. The only problem is that I don't think you can implement a method as a value function, you have to have a separate private val and do something like this:
```scala
trait Foo { def func(x: Int): Boolean }
class Bar extends Foo {
  private val funcImpl = Set(...)
  def func(x: Int): Boolean = funcImpl(x)
}
```

src/main/scala/firrtl/passes/clocklist/ClockListTransform.scala
seldridge: Solution here: switching to `def` globally.

src/main/scala/firrtl/passes/wiring/WiringTransform.scala
seldridge: Additional commit (with @jackkoenig attribution) added to include the suggested fix. I'll relegate an improvement to the way to specify invalidaties to future work.

seldridge:
Note: I inadvertently changed the behavior of in-tree custom transforms with `inputForm=LowForm`. This PR causes the following transforms to now run before optimizations:

- `GetNamespace`
- `InlineInstances` (which actually fixes https://github.com/freechipsproject/firrtl/issues/1298)
- `ClockListTransform`
- `Flatten`
- `RenameModules`

I don't foresee this causing a problem, but I wanted to document the behavior change.

PR #1539, opened 2020-04-24
Implement InstanceTarget Behavior for Dedup + EliminateTargetPaths (without AST targets)

albertchen-sifive:
### Summary of Changes
This PR includes commits from https://github.com/freechipsproject/firrtl/pull/1442 and reverts the AST target changes so that @azidar's other changes can hopefully be merged in sooner. This PR summary is also copied from @azidar's PR with edits for the things that my commits change/delete/add. The main change is that the AST module changes were reverted and the RenameMap was changed so that no transitive renames occur even for self-renames.

Keeping it as a draft because I want to leave in my debugging helpers while I do some more testing. Changes should be pretty much done though.

* `RenameMap` Behavior
  
  * Prevent transitive renaming A -> B -> C (continueRenaming)
    * <b>change</b>: also prevent transitive renaming for self-renames
  * <s>Added AST lookups for regular renames, so given rename (~A|A -> ~B|B), ~Top|A is renamed to ~Top|B.</s>
  * Added transform to `RenameMap` initialization, and update all locations of `RenameMap()`
    * no change
* Target
  
  * <s>Add `ModuleTarget(module: String)` apply to create AST `ModuleTarget`</s>
  * Override `toString` as `serialize` for `CompleteTarget`
  * Expansion of stripHierarchy to enable stripping `InstanceTarget`s to become `ModuleTarget`s
* Annotations
  
  * Bugfix in extractComponents where Products were not iterated over
  * SingleTargetAnnotation[Named] now converts to AST targets to preserve sticky behavior over dedup
    * <b>change</b>: no longer converts to AST targets, instead converts renamed targets to local targets using `Target.referringModule` to preserve sticky behavior
* Eliminate Target Paths
  
  * Make DuplicationHelper use `LinkedHashMap`, as we iterate over its contents and convert to Seq in `def makePathless`
  * Add `DupedResult` to map original module to new module targets
  * Update renaming to record a map from all relative instance paths to original module, to new module target
  * Consumes `DedupedResult` to give better name to new duplicated module if it was originally deduplicated
  * Reorder modules in attempt to preserve original ordering, pre-deduplication
  * Move utility functions to object
  * <b>addition</b>: Bugfix: add self-renames to prevent `ofModule _ of target _ cannot be renamed to Vector(_, _, _, ...)` errors
* Dedup
  
  * Changed `NoDedupAnnotation` to contain `ModuleTarget`, rather than `ModuleName`
  * Added `DedupedResult` to map original module to the duplicate module
  * Consumes `DupedResult` to pick better name, if it existed
  * Updates renaming to chain the following: instancify deduped modules, remap differently named internal signals, then remap AST modules
  * Move utility functions to object
  * Remove annotations as part of determination of dedup correctness
  * <b>addition</b>: Bugfix: add instance renames so that deduped modules have their instances properly renamed
* Dead Code Elimination
  
  * <s>Add deletion of ASTModules</s>
* Tests
  
  * Morphism Spec to ensure Dedup -> EliminateTargetPaths and EliminateTargetPaths -> Dedup patterns work properly
  * Update existing tests to make sure they work properly
    * <b>addition</b> add Dedup tests to demonstrate instance renaming bug, EliminateTargetPaths for ofModule rename bug, and update RenameMap tests

### Contributor Checklist

#### Type of Improvement

- bug fix

- new feature/API

#### API Impact

Biggest change is that DedupModules works between modules with different Target annotations. <s>Another change is adding AST ModuleTargets, which are used when converting between Named and Target in SingleTargetAnnotation.</s>

#### Desired Merge Strategy

- [ ] Did you review?


src/main/scala/firrtl/RenameMap.scala
```diff
@@ -399,90 +496,114 @@ final class RenameMap private (val underlying: mutable.HashMap[CompleteTarget, S
...
     } else {
-      val getter = recursiveGet(errors)(_)
+
+      // Used to determine if renaming has ended early. This is to prevent the case
+      //   where a fully recursive renaming occurs, e.g.
+      //   Given Renames:
+      //     - (~Top|Top/a:A -> ~Top|A)
+      //     - (~Top|A -> ~Top|B)
+      //   Renaming (~Top|Top/a:A) should return (~Top|A), not (~Top|B)
+      // This is due to our splitting recursive renaming into three parts (top, mid, ast, bot)
```
azidar, line 506:
```suggestion
      // This is due to our splitting recursive renaming into three parts (top, mid, bot)
```

src/main/scala/firrtl/RenameMap.scala
```diff
@@ -312,10 +319,12 @@ final class RenameMap private (val underlying: mutable.HashMap[CompleteTarget, S
   }
 
   /** Checks for renames of only the path portion of an [[InstanceTarget]]
-    * Recursively checks parent [[IsModule]]s until a match is found
-    * First checks all parent paths from longest to shortest. Then
-    * recursively checks paths leading to the encapsulating module.
-    * Stops on the first match found.
+    * Recursively checks parent [[IsModule]]s until a match is found First
```
azidar, line 322:
```suggestion
    * Recursively checks parent [[IsModule]]s until a match is found. First
```

src/main/scala/firrtl/RenameMap.scala
```diff
@@ -399,90 +496,114 @@ final class RenameMap private (val underlying: mutable.HashMap[CompleteTarget, S
...
+      //   where a fully recursive renaming occurs, e.g.
+      //   Given Renames:
+      //     - (~Top|Top/a:A -> ~Top|A)
+      //     - (~Top|A -> ~Top|B)
+      //   Renaming (~Top|Top/a:A) should return (~Top|A), not (~Top|B)
+      // This is due to our splitting recursive renaming into three parts (top, mid, ast, bot)
+      // We continue renaming even if original == ts.head because if it returns the same target, there are
+      //   still additional renaming of different parts of the target which must occur. If the target actually
+      //   should not be renamed, it must go through all partial renaming parts to actually determine that fact.
+      def continueRenaming[T <: CompleteTarget](original: CompleteTarget, ts: Seq[CompleteTarget]): Boolean =
```
azidar, line 510: I think this is wrong actually, we need to capture the type with a classtag.

```suggestion
      def continueRenaming[T <: CompleteTarget : ClassTag](original: CompleteTarget, ts: Seq[CompleteTarget]): Boolean =
```

albertchen-sifive: I'm actually no longer using this function and forgot to delete it. I updated the {reference/instance/module}Get functions to return `Option` instead, so renaming stops when they return `Some`. since renaming can stop on self renames now.

src/test/scala/firrtlTests/annotationTests/MorphismSpec.scala
```diff
@@ -0,0 +1,570 @@
...
+    val annotations: AnnotationSeq = Seq.empty
+
+    val finalAnnotations: Option[AnnotationSeq] = None
+
+    lazy val state = CircuitState(Parser.parse(input), UnknownForm, annotations)
+  }
+
+  trait RightInverseFixture extends CircuitFixture {
+
+    /** An endomorphism */
```
albert-magyar, line 90: I think what this is getting at is that it is a function from the set of annotation sequences to the set of annotation sequences. Would it be possible to simplify some of the nomenclature to maybe "a function on an annotation sequence" to make it a bit more accessible?

src/test/scala/firrtlTests/annotationTests/MorphismSpec.scala
```diff
@@ -0,0 +1,570 @@
...
+    )
+
+    override val finalAnnotations: Option[AnnotationSeq] = Some(Seq(
+      AnAnnotation(CircuitTarget("Top").module("Top").instOf("baz", "Baz").instOf("foo", "Foo")),
+      AnAnnotation(CircuitTarget("Top").module("Top").instOf("qux", "Baz").instOf("foo", "Foo"))
+    ))
+    test()
+  }
+
+  it should "be idempotent with per-module annotations" in new IdempotencyDedupModulesFixture {
```
albert-magyar, line 532: These tests are excellent.

src/test/scala/firrtlTests/annotationTests/MorphismSpec.scala
```diff
@@ -0,0 +1,570 @@
...
+                           from: Option[AnAnnotation] = None,
+                           cause: Option[String] = None
+                         ) extends Annotation {
+    override def update(renames: RenameMap): Seq[AnAnnotation] = {
+      if (target.isDefined) {
+        renames.get(target.get) match {
+          case None => Seq(this)
+          case Some(Seq()) => Seq(AnAnnotation(None, Some(this), renames.associatedTransform.map(_.name)))
+          case Some(targets) =>
+            //TODO: Add cause of renaming, requires FIRRTL change to RenameMap
```
albert-magyar, line 28: Is this already accomplished with the associatedTransform name labeling?

src/test/scala/firrtlTests/annotationTests/MorphismSpec.scala
seldridge: This languages dates from my original `MorphismSpec`. As far as I know the language is correct, but inaccessible. I'd probably prefer to keep it, but explain it. I was trying to be pedantic about the FP ideas here so that whatever the test was doing had some basis in a mathematical property.

Yeah, it's just saying that this is a mapping from `CircuitState => CircuitState`.

albertchen-sifive:
split out `RenameMap` changes into a separate PR. https://github.com/freechipsproject/firrtl/pull/1591

azidar:
Can this PR also remove the `associatedTransform` from the RenameMap? I don't think its necessary, and @albert-magyar has some concerns about its introduction in general.

albertchen-sifive:
> Can this PR also remove the `associatedTransform` from the RenameMap? I don't think its necessary, and @albert-magyar has some concerns about its introduction in general.

I removed the `associatedTransform` and resolved merge conflicts.

src/main/scala/firrtl/transforms/Dedup.scala
```diff
@@ -65,40 +109,164 @@ class DedupModules extends Transform with DependencyAPIMigration with PreservesA
...
+                old -> newRelPath.instOf(newInstName, mod)
+              }
+            }
+          }
+
+          // Add all relative paths to referredModule to map to new instances
+          def addRecord(from: IsMember, to: IsMember): Unit = from match {
+            case x: ModuleTarget =>
+              instanceify.record(x, to)
+            case x: IsComponent =>
```
albert-magyar, line 219: The rename still doesn't seem to handle the effects of retyping on the `component` part of a `ReferenceTarget`. See #1625.

src/main/scala/firrtl/RenameMap.scala
```diff
@@ -36,7 +40,11 @@ object RenameMap {
   * @define noteDistinct @note Rename to/tos will be made distinct
   */
 // TODO This should probably be refactored into immutable and mutable versions
-final class RenameMap private (val underlying: mutable.HashMap[CompleteTarget, Seq[CompleteTarget]] = mutable.HashMap[CompleteTarget, Seq[CompleteTarget]](), val chained: Option[RenameMap] = None) {
+final class RenameMap private (
+    val underlying: mutable.HashMap[CompleteTarget, Seq[CompleteTarget]] = mutable.HashMap[CompleteTarget, Seq[CompleteTarget]](),
+    val chained: Option[RenameMap] = None,
+    val associatedTransform: Option[Transform] = None
```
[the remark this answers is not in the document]
albert-magyar: This was resolved by dropping the transform argument.

src/main/scala/firrtl/transforms/Dedup.scala
albertchen-sifive: this is fixed by https://github.com/freechipsproject/firrtl/pull/1624 right? and dedup no longer considers annotations when calculating tags, so I think the other issues are no longer a problem.

albert-magyar: Yeah, it does seem to be fixed in #1624. I had it backwards about which of the two places that consider retypes would be fixed in #1624, so this seems like it would actually compose properly with the fix in #1624, whereas `master` does not.

albert-magyar:
One more question: if all modules in a "dedup set" are annotated identically, it seems that this will still give `InstanceTargets`.

Are there any thoughts about a way to get these "consistent" annotations into `ModuleTarget`s without the heavy-handed step of `EliminateTargetPaths` and losing the dedup?

I actually think that this is great as-is. I think we should try to avoid making the responsibilities of any one transform too great, and it might make sense to add another transform that coalesces annotations that are across all instances of a module, or to extend `EliminateTargetPaths` for this case in a separate PR.

src/test/scala/firrtlTests/annotationTests/MorphismSpec.scala
```diff
@@ -0,0 +1,570 @@
...
+        CircuitTarget("Top").module("Top")
+      )
+  }
+
+
+  behavior of "EliminateTargetPaths"
+
+  trait RightInverseEliminateTargetsFixture extends RightInverseFixture with DefaultExample {
+    override val f: Seq[Transform] = Seq(new firrtl.transforms.DedupModules)
+    override val g: Seq[Transform] = Seq(new firrtl.annotations.transforms.EliminateTargetPaths)
```
albert-magyar, line 308: This is a silly nit, but `EliminateTargetPaths` is actually not the inverse of `DedupModules`, due to retyping.

seldridge: Not silly, but I think it should be clear that this is an inverse where equivalence is defined structurally.

albertchen-sifive:
> One more question: if all modules in a "dedup set" are annotated identically, it seems that this will still give `InstanceTargets`.
> 
> Are there any thoughts about a way to get these "consistent" annotations into `ModuleTarget`s without the heavy-handed step of `EliminateTargetPaths` and losing the dedup?
> 
> I actually think that this is great as-is. I think we should try to avoid making the responsibilities of any one transform too great, and it might make sense to add another transform that coalesces annotations that are across all instances of a module, or to extend `EliminateTargetPaths` for this case in a separate PR.

yes, it will rename to `InstanceTarget`s even if deduped modules are annotated identically. I like the idea of extending `EliminateTargetPaths` to account for this

src/main/scala/firrtl/annotations/transforms/EliminateTargetPaths.scala
```diff
@@ -67,7 +67,7 @@ object EliminateTargetPaths {
     }
     def onStmt(s: Statement): Statement = s map onStmt match {
       case w@WDefInstance(info, name, module, tpe) if toRename.contains(module) => w.copy(module = toRename(module))
-      case w@DefInstance(info, name, module) if toRename.contains(module) => w.copy(module = toRename(module))
+      case w@DefInstance(info, name, module, _) if toRename.contains(module) => w.copy(module = toRename(module))
```
albert-magyar, line 70: You can actually drop this line entirely (or the one above it) for the `1.4.0` version. The WDefInstance unapply will match the same stuff.

albert-magyar: I don't know if you're going to backport this PR (looks like it's marked 1.4.0), but make sure to put it back in if you do :)

albertchen-sifive: cool, glad that stuff is cleaned up now!

PR #1549, opened 2020-04-30
Uniquify mems as if they're vectors (emit Uniquified Verilog)

seldridge:
Change the behavior of `Uniquify` when generating unique names for memories. Previously, memories are treated like instances with ports. However, memories are never actually lowered like this. Consider that memories are lowered in one of two ways:

1. Using `ReplSeqMem` -- The memory is blackboxed. It's only potential source of namespace conflict via lowering is due to the name of the memory.
2. By the Verilog emitter -- if this happens, then the memory is emitted as `reg foo [0:31]`. Here, the contribution to the Verilog namespace is actually a **vector**.

This PR changes `Uniquify` to view the contribution of a memory to a module's namespace as situation (2).

This allows for the postcondition of `Uniquify` (that you can blindly lower types) to be true for FIRRTL-emitted Verilog that includes memories. Currently this is not true. *In effect, the `VerilogEmitter` invalidates `Uniquify` and this PR makes it not do that.*

One test case is included to show this behavior. Something like this is reproduced below. If you start with the following module `Foo` which contains memory `bar` and a wire `bar_0`:
```
circuit Foo:
  module Foo:
    mem bar:
      data-type => UInt<1>
      depth => 2
      read-latency => 0
      write-latency => 1
      read-under-write => undefined
    wire bar_0: UInt<1>
    bar_0 is invalid
```

You get the following Verilog currently (with `--no-dce` and trimmed to exclude randomization logic):
```verilog
module Foo(
);
  reg  bar [0:1];
  wire  bar_0;
  assign bar_0 = 1'h0;
endmodule
```

If you were to then blindly lower `bar` in Verilog you'd get a name conflict on `bar_0`. This PR changes it so that you instead get `reg bar_` *which does not conflict with `bar_0` if lowered*.

Contrast this with the behavior of `bar` being a wire:

```
circuit Foo:
  module Foo:
    wire bar: UInt<1>[2]
    wire bar_0: UInt<1>

    bar is invalid
    bar_0 is invalid
```

This will compile to:
```verilog
module Foo(
);
  wire  bar__0;
  wire  bar__1;
  wire  bar_0;
  assign bar__0 = 1'h0;
  assign bar__1 = 1'h0;
  assign bar_0 = 1'h0;
endmodule
```

This PR can be viewed as aligning these two naming behaviors.

### Contributor Checklist

#### API Impact

No impact.

Possible name changes in emitted Verilog for memories not blackboxed.

#### Desired Merge Strategy

#### Release Notes

- Prevent mems emitted as Verilog vectors from colliding with other signal names if lowered by a downstream tool

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/passes/Uniquify.scala
+++ b/src/main/scala/firrtl/passes/Uniquify.scala
@@ -92,7 +92,7 @@ object Uniquify extends Transform with DependencyAPIMigration {
       case sx: WDefInstance => Seq(Field(sx.name, Default, sx.tpe))
       case sx: DefMemory => sx.dataType match {
         case (_: UIntType | _: SIntType | _: FixedType) =>
-          Seq(Field(sx.name, Default, memType(sx)))
+          Seq(Field(sx.name, Default, VectorType(sx.dataType, sx.depth.toInt)))
         case tpe: BundleType =>
           val newFields = tpe.fields map ( f =>
             DefMemory(sx.info, f.name, f.tpe, sx.depth, sx.writeLatency,
--- a/src/test/scala/firrtlTests/UniquifySpec.scala
+++ b/src/test/scala/firrtlTests/UniquifySpec.scala
@@ -316,4 +316,24 @@ class UniquifySpec extends FirrtlFlatSpec {
     if (TestOptions.accurateTiming)
       renameMs shouldBe < (baseMs * threshold)
   }
+
+  it should "rename to avoid collisions with lowered bits" in {
+    val input =
+     """|circuit Foo:
+        |  module Foo:
+        |    mem _T :
+        |      data-type => UInt<1>
+        |      depth => 2
+        |      read-latency => 0
+        |      write-latency => 1
+        |      read-under-write => undefined
+        |    wire _T_1 : UInt<1>
+     """.stripMargin
+    val expected = Seq(
+      "mem _T_ :",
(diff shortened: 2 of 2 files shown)


seldridge:
This needs some thought. This will blow up the heap for large memories.

seldridge:
I'm closing this as I think the motivation to provide "uniquifed" emitted Verilog is pretty weak. I'm also not sure that "uniquified" is a useful attribute for downstream tools consuming Verilog.

That said, this PR brought up one critical issue: the internal datastructure used for uniquification and namespaces are extremely inefficient for vectors. Currently, namespaces are handled with a set. It would be more efficient to use something different for vectors instead of flattening the vectors into the namespace. Specifically, an interval tree should work for the two types of queries/operations that matter here:

  1. Does some number overlap with an existing interval? (e.g., does `foo_5: UInt<1>` overlap with `foo: UInt<1>[4]` ?)
  2. Insertion of new intervals into the interval tree

Due to the linear ordering of temporaries in the IR, e.g., `_T_1` is likely declared before `_T_2`, efficient implementation of (2) definitely needs a self-balancing tree (since the expected insertion pattern is guaranteed to be the worst case).

Jack has suggested switching to a trie for the namespace. I think that makes sense as a secondary optimization for the string portion of the namespace. However, I don't think that's currently as problematic as numbers are treated.

Issue #1466, opened 2020-03-22
Turn off coverage gathering for simulation-only initial blocks


mwachs5:
### Checklist

- [x] Did you write out a description of the feature you want to see?
- [x] Did you look around for any related features?
- [x] Did you specify relevant external information? -- *Not aware of any*

### Feature Description

During code emission, it would be nice to turn off (optionally) coverage for the random initialization code initial blocks. These constitute a large portion of lines of code and are difficult to cover since they are randomly initialized.
For certain tool flows it can be turned off by wrapping the  `initial block` with `// VCS coverage off` and `// VCS coverage on`

Similar to the addition of `ifndef SYNTHESIS`:
https://github.com/freechipsproject/firrtl/blob/5c0c0018d812d57270035a9d3bd82e2289acf4ec/src/main/scala/firrtl/Emitter.scala#L717

If done as requested, it would look like the following:
```
`ifndef SYNTHESIS
// VCS coverage off
initial begin
...

end // initial
// VCS coverage on
`endif // SYNTHESIS
```

Or, it could be done in a vendor-agnostic way with a define like:

```
`COVERAGE_ON

`COVERAGE_OFF
```

where users could define these macros as they see fit

### Related Features

Similar to the code for `ifndef SYNTHESIS added here: 
https://github.com/freechipsproject/firrtl/blob/5c0c0018d812d57270035a9d3bd82e2289acf4ec/src/main/scala/firrtl/Emitter.scala#L717

mwachs5:
perhaps this should be expanded for generic `BEFORE_INITIAL and `AFTER_INITIAL macros which can be populated with a variety of macros by the final verilog user

debs-sifive:
@mwachs5 can this be closed, since #1550 has been merged?

mwachs5:
Closed by #1550 🎉

PR #1550, opened 2020-04-30
before/after initial block macros

debs-sifive:
### Contributor Checklist

* I added some inline comments (not Scaladoc)
* https://github.com/freechipsproject/www.chisel-lang.org/pull/80

#### Type of Improvement

- new feature/API

#### API Impact

New macros have been added:
* `BEFORE_INITIAL`, which is called before the emitted (non-empty) initial block if it is defined
* `AFTER_INITIAL`, which is called after the emitted (non-empty) initial block if it is defined

`verilogToCpp` has been deprecated for `verilogtoCppWithExtraCmdLineArgs`, which optionally takes in a list of extra command line arguments.

These macros will now be emitted in the generated Verilog if there is a non-empty initial block. They are guarded by `ifdef` statements, for example:
```
`ifdef BEFORE_INITIAL
`BEFORE_INITIAL
`endif
initial begin
...
```

#### Desired Merge Strategy

- Squash: The PR will be squashed and merged (choose this if you have no preference.)

#### Release Notes

Users may define the following macros if they wish to specify behavior before/after initial blocks (for example, turning coverage on and off):
* `BEFORE_INITIAL`, which is called before the emitted (non-empty) initial block if it is defined
* `AFTER_INITIAL`, which is called after the emitted (non-empty) initial block if it is defined

`verilogToCpp` has been deprecated for `verilogtoCppWithExtraCmdLineArgs`, which optionally takes in a list of extra command line arguments.

- [ ] Did you review?


src/main/scala/firrtl/Emitter.scala
```diff
@@ -979,6 +979,10 @@ class VerilogEmitter extends SeqTransform with Emitter {
         }
         emit(Seq("`endif"))
         emit(Seq("`ifndef SYNTHESIS"))
+        // User-defined macro of code to run before an initial block
+        emit(Seq("`ifdef BEFORE_INITIAL"))
+        emit(Seq("`BEFORE_INITIAL"))
+        emit(Seq("`endif"))
```
jackkoenig, line 985: We didn't do this for the `RANDOMIZE` stuff, but should we prefix our macros with `FIR` or `FIRRTL` just in case FIRRTL-generated Verilog is being used along side other Verilog that may have it's own macros?

mwachs5:
closes #1466

src/main/scala/firrtl/Emitter.scala
debs-sifive: I also think this is a good idea, so I updated the macros specific to this PR and opened up issue #1554 to prompt discussion for the other macros.

mwachs5: oh i left my comment on #1554 before I read this discussion, as you were :)

src/test/scala/firrtl/testutils/FirrtlSpec.scala
```diff
@@ -254,12 +254,15 @@ object FirrtlCheckers extends FirrtlMatchers {
...
-  /** Checks that the emitted circuit has the expected lines in order, all lines will be normalized */
+  /** Checks that the emitted circuit contains the expected lines contiguously and in order;
+    * all lines will be normalized
+    */
   def containLines(expectedLines: String*) = new CircuitStateStringsMatcher(expectedLines)
 
   class CircuitStateStringsMatcher(expectedLines: Seq[String]) extends Matcher[CircuitState] {
     override def apply(state: CircuitState): MatchResult = {
       val emitted = state.getEmittedCircuit.value
+      println(expectedLines)
```
azidar, line 265: Remove println?

src/main/scala/firrtl/util/BackendCompilationUtilities.scala
```diff
@@ -107,7 +107,8 @@ trait BackendCompilationUtilities extends LazyLogging {
     vSources: Seq[File],
     cppHarness: File,
     suppressVcd: Boolean = false,
-    resourceFileName: String = firrtl.transforms.BlackBoxSourceHelper.defaultFileListName
+    resourceFileName: String = firrtl.transforms.BlackBoxSourceHelper.defaultFileListName,
+    extraCmdLineArgs: Seq[String] = Seq.empty
```
azidar, line 111: This is break binary compatiblity with the 1.3.X branch, which means it can't be backported. Instead, create a new function with the same name and the additional optional argument, call that function from this function, and deprecate this function.

jackkoenig: I'm a little late here, but I'm not a huge fan of deprecating `verilogToCpp` in favor of a very long name `verilogToCppWithExtraCmdLineArgs`, can we either:
1) Not deprecate `verilogToCpp`
2) Have `verilogToCppWithExtraCmdLineArgs` be named `verilogToCpp` instead and just overload?

debs-sifive: This was the initial approach, but it turns out you can't overload methods with default arguments :(

debs-sifive: i'm not really sure how deprecation works, but is it possible that when `verilogToCpp` is removed, to rename `verilogToCppWithExtraCmdLineArgs` to `verilogToCpp`?

jackkoenig: We could just not deprecate `verilogToCpp`, it's not that big of a deal to have two functions

debs-sifive: That sounds reasonable to me. Any opinions, @azidar ?

jackkoenig:
@Mergifyio backport 1.3.x

debs-sifive:
> Would this work better as an emission option like `firrtl --disable-coverage-in-initial-blocks`?

Would the emission option allow for the coverage dis/enabling to happen during simulation time? If so, I think your approach is fine. But I think verif engineers find it useful to be able to set a macro to toggle coverage on a simulation-by-simulation basis.

seldridge:
I was not thinking it would be configurable based on reasoning that verification engineers should not be computing coverage of initial randomization blocks. Two questions:

1. Do we make use of the run-time configuration when in a flow which computes coverage?
2. Is it ever meaningful for coverage to be computed for initial blocks, i.e., non-synthesizable code?

If it does need to be configurable, it would be cool for this to NOT be handled with a macro, but with some other mechanism.

PR #1556, opened 2020-05-01
Add LegalizeAndReductionsTransform

jackkoenig:
Workaround for https://github.com/verilator/verilator/issues/2300
present in Verilator versions v4.026 - v4.032. This transform turns AND
reductions for expressions > 64-bits into an equality check with all
ones. It is included in all Verilog emitters.

Would appreciate review from @seldridge on some of the Dependency API stuff. I copy-pasted `LegalizeClocksTransform` but didn't add it as a dependency to `AddDescriptionNodes`, `VerilogModulusCleanup`, `VerilogPrep`, `FlattenRegUpdate`, nor `RemoveKeywordCollisions`. I also didn't have `InlineCasts` invalidate it. Should I?

I'm also looking for feedback on if we should include this workaround. I think we should; it's a pretty annoying bug that is hard to find (only will come up during simulation) and emitting Verilog for maximal compatibility is our general ethos. I'm open to argument that it should be optional to turn on (or perhaps possible to turn off).

### Contributor Checklist

#### Type of Improvement

   - bug workaround

#### API Impact

No change to API

Stops emitting and reductions for > 64-bit expressions, instead emits equality check with all ones

#### Desired Merge Strategy

Don't care

#### Release Notes

Replace AND reductions of > 64-bit expressions with equality check against all ones.
Workaround for https://github.com/verilator/verilator/issues/2300: bug in Verilator versions v4.026 - v4.032.

- [x] Did you review?

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -8,13 +8,14 @@ import scala.collection.mutable
 
 import firrtl.ir._
 import firrtl.passes._
+import firrtl.transforms.LegalizeAndReductionsTransform
 import firrtl.annotations._
 import firrtl.traversals.Foreachers._
 import firrtl.PrimOps._
 import firrtl.WrappedExpression._
 import Utils._
 import MemPortUtils.{memPortField, memType}
-import firrtl.options.{HasShellOptions, ShellOption, StageUtils, PhaseException, Unserializable}
+import firrtl.options.{Dependency, HasShellOptions, ShellOption, StageUtils, PhaseException, Unserializable}
 import firrtl.stage.{RunFirrtlTransformAnnotation, TransformManager}
 // Datastructures
 import scala.collection.mutable.ArrayBuffer
@@ -180,7 +181,9 @@ class VerilogEmitter extends SeqTransform with Emitter {
   def inputForm = LowForm
   def outputForm = LowForm
 
-  override def prerequisites = firrtl.stage.Forms.LowFormOptimized
+  override def prerequisites =
+    Dependency[LegalizeAndReductionsTransform] +:
+    firrtl.stage.Forms.LowFormOptimized
 
   override def optionalPrerequisiteOf = Seq.empty
 
@@ -1160,7 +1163,9 @@ class VerilogEmitter extends SeqTransform with Emitter {
 
 class MinimumVerilogEmitter extends VerilogEmitter with Emitter {
 
-  override def prerequisites = firrtl.stage.Forms.LowFormMinimumOptimized
+  override def prerequisites =
+    Dependency[LegalizeAndReductionsTransform] +:
+    firrtl.stage.Forms.LowFormMinimumOptimized
 
   override def transforms = new TransformManager(firrtl.stage.Forms.VerilogMinimumOptimized, prerequisites)
     .flattenedTransformOrder
--- a/src/main/scala/firrtl/transforms/LegalizeReductions.scala
+++ b/src/main/scala/firrtl/transforms/LegalizeReductions.scala
@@ -1,0 +1,49 @@
+package firrtl
+package transforms
+
+import firrtl.ir._
+import firrtl.Mappers._
+import firrtl.options.{Dependency, PreservesAll}
+import firrtl.Utils.BoolType
+
+
+object LegalizeAndReductionsTransform {
+
+  private def allOnesOfType(tpe: Type): Literal = tpe match {
+    case UIntType(width @ IntWidth(x)) => UIntLiteral((BigInt(1) << x.toInt) - 1, width)
+    case SIntType(width) => SIntLiteral(-1, width)
+
+  }
+
+  def onExpr(expr: Expression): Expression = expr.map(onExpr) match {
+    case DoPrim(PrimOps.Andr, Seq(arg), _,_) if bitWidth(arg.tpe) > 64 =>
+      DoPrim(PrimOps.Eq, Seq(arg, allOnesOfType(arg.tpe)), Seq(), BoolType)
+    case other => other
+  }
+
+  def onStmt(stmt: Statement): Statement = stmt.map(onStmt).map(onExpr)
+
+  def onMod(mod: DefModule): DefModule = mod.map(onStmt)
+}
+
+/** Turns andr for expression > 64-bit into equality check with all ones
+  *
+  * Workaround a bug in Verilator v4.026 - v4.032 (inclusive).
+  * For context, see https://github.com/verilator/verilator/issues/2300
+  */
+class LegalizeAndReductionsTransform extends Transform with DependencyAPIMigration with PreservesAll[Transform] {
+
+  override def prerequisites =
+    firrtl.stage.Forms.WorkingIR ++
+    Seq( Dependency(passes.CheckTypes),
+         Dependency(passes.CheckWidths))
+
+  override def optionalPrerequisites = Nil
+
+  override def optionalPrerequisiteOf = Nil
+
+  def execute(state: CircuitState): CircuitState = {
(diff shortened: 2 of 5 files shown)


jackkoenig:
Rebased to fix `src/test/scala/firrtlTests/LoweringCompilersSpec.scala`, note change to expected transforms.

albert-magyar:
Also, it is not necessary to add it to the `VerilogMinimumOptimized` list to make it run. The form lists are just conveniences for legacy code -- the only thing that actually depends on this transform is the emitter. There is the separate question of whether it should be mandatory, which would reflect its status as either an optional or full prerequisite, and whether it is added as a target for default compilers if it is optional.

Right now, the use of the same `Forms` for both widely-used prerequisite lists and "Compiler target Seqs" is a convenience to make the existing transforms work, but it provides a high degree of over-specification of prerequisites that we should strive to avoid with new code.

seldridge:
The only `optionalPrerequisiteOf` to consider are the verilog emitters (`Forms.BackendEmitters`) if this transform is definitely supposed to run before specific emitters. (Or: check the result that you get as it is often a legal ordering to run an emitter before a custom transform.)

jackkoenig:
> Therefore, I would say that the ideal thing to do would be to make this transform have only HighForm (this is essentially universal to hide the "CHIRRTL hack"), CheckTypes, and InferWidths as prerequisites, and no optionalPrerequisites or optionalPrerequisiteOfs.

> No transform should invalidate this transform.

All this makes sense, it's good to start using the dependency API the right way! I've pushed a commit that reflects your suggestions, although I'll note that I made the prereqs as follows:
```scala
  override def prerequisites =
    firrtl.stage.Forms.WorkingIR ++
    Seq( Dependency(passes.CheckTypes),
         Dependency(passes.CheckWidths))
```

`Forms.HighForm` includes `Resolved` and `Deduped`, so to be more minimal I just based it on `Forms.WorkingIR`. Does it really need working IR? No but I think that serves as a good way to hide the "CHIRRTL hack" as you said.

I also removed it from `Forms.VerilogMinimumOptimized`, instead adding it directly as a prereq to each Emitter, but am open to making it not required.

jackkoenig:
> The only optionalPrerequisiteOf to consider are the verilog emitters (Forms.BackendEmitters) if this transform is definitely supposed to run before specific emitters.

If we make it non-mandatory, then yeah it would need all of the Verilog emitters as `optionalPrequisiteOf`.

albert-magyar:
> `Forms.HighForm` includes `Resolved` and `Deduped`, so to be more minimal I just based it on `Forms.WorkingIR`. Does it really need working IR? No but I think that serves as a good way to hide the "CHIRRTL hack" as you said.

That's a very good point. The inclusion of `HighForm` as a prerequisite on anything is kind of a wart, and I agree that `WorkingIR` seems like a better choice here. This will probably be revisited before 1.4.0.

> If we make it non-mandatory, then yeah it would need all of the Verilog emitters as optionalPrequisiteOf

Either that or alternatively the emitters could list it as an `optionalPrerequisite`. Since "reverse" edges are kind of an extra tool for dependency injection, it might be wise to make core transforms rely on `optionalPrerequisite` instead of `optionalPrerequisiteOf` for core transforms to avoid the "action at a distance" concern of having to look at both sides.

albert-magyar:
Also, are we not going to backport this to 1.2?

seldridge:
The failures on `LoweringCompilersSpec` and `CustomTransformSpec` are spurious. We should figure out what to do about these, though..

`LoweringCompilersSpec` is checking that the generated transform order matches the old pre-Dependency API transform order. It includes a small library for "patching" the old order with new transforms. This is good as it provides a record of how things work now vs. before. However, with `CircuitForm` deprecated this test is really just a gate on new transforms.

`CustomTransformSpec` is a similar thing that's looking for a specific placement of a custom `inputForm=LowForm` transforms (the custom transform should run between the last transform and the emitter). If the transform right before the emitter changes, then the test breaks. We can probably remove this test as well or just check that it runs before the emitter.

jackkoenig:
> Also, are we not going to backport this to 1.2?

Wasn't planning to because I deleted the automation for it (can only have 1 milestone on a PR) but we probably should

jackkoenig:
> One thought in favor of "yes": to support per-simulator specialization we would want to invest in the right infrastructure for specialized emitters. That may end up being very simple, but it might be worth resolving in a separate PR.

I agree that per-simulator specialization is beyond the scope of this PR. I don't think that formalized support os a prerequisite for a minor work around like this one, but I would agree with not mandating more dramatic workarounds, especially one that impact a lot of the generated code.

> One thought in favor of "no": the Verilator issue is a bug, not a broader limitation. It seems like the right solution for most users will be to just bump Verilator.

This is true and I share this opinion in most cases. I disagree on this particular issue because this is a very subtle bug that took several engineering hours to track down. I would definitely vote against if the bug manifested as Verilator crashing noisily so that it would be obvious. Most people do not update their tools that consistently so it is likely that many people will have version of Verilator with the bug for a while.

I don't know how often Ubuntu updates its packages (does it ever?), but Ubuntu 20.04 has Verilator 4.028 (a buggy version). I'll note that if Ubuntu doesn't update it's packages, then we may have impacted users for a very very long time which kind of goes against my argument since I don't think we should have this workaround for like 2 years, I was just thinking until 1.4.

My argument is that this workaround has little impact on the emitted Verilog (low cost) while removing at least one rare, but large source of user frustration. I think we should have it on by default for a little while such that it will be more likely most people have updated to a newer version.

albert-magyar:
That makes sense to me. I’m down with on by default.

sequencer:
maybe we can check our verilator version with `verilator -V` to decide if we need to execute this Transform or not? if in a buggy version, send a explicit warning to user, and execute this Transform.

aswaterman:
There’s no guarantee that the verilog will be generated on a system with
the same version of verilator that the system consuming the verilog has.

mwachs5:
> Also, are we not going to backport this to 1.2?
> 
> Wasn't planning to because I deleted the automation for it (can only have 1 milestone on a PR) but we probably should

I thought we didn't need to backport this to 1.2 since the `andR` was only introduced in 1.3? Or was that only that chisel 3.3 started using it...

albert-magyar:
That’s a good point, those are not emitted in Chisel 3.2. However, that
also means that there’s no real downside to enabling this pass in FIRRTL
1.2 for chisel-based designs, plus it has the benefit of supporting FIRRTL
designs from other flows that might use Andr.

seldridge:
Agree with @aswaterman. There's should be no check in FIRRTL itself about the Verilator version. 

However, whatever tool is invoking Verilator, e.g., testers or ChiselTest, does have the appropriate scope to perform such a check and should be turning on this transform if it knows it's using a specific Verilator version that has this bug.

By that logic we can either turn this on by default (I'm cool with it) or we leave it off and people who need it are expected to turn it on (Rocket Chip). 

We really need analog of `-march` that relates to the "host" we're targeting  that relates to what context of how the generated Verilog will be used. It's not clear to me how CPU bugs are handled in compilers, when the fixes are turned on by default, and when the fixes are turned off by default. Out of scope for this PR.

sequencer:
> There’s no guarantee that the verilog will be generated on a system with
the same version of verilator that the system consuming the verilog has.

Yes... I didn't realized that.

I just consult the upgrade strategy of Ubuntu to a DD(Debian Developer), he said a trivial patch to Ubuntu is required, if we ask bumping ventilator, or wait until next Ubuntu release and it will steal packages from Debian. So for Ubuntu users, it seems to be a very long time issue...
But as for Arch or Debian, this seems to be only a temporary issue which will be solved when 4.034 of verilator is released.

jackkoenig:
@seldridge I've fixed the tests, let me know what you think

jackkoenig:
@Mergifyio backport 1.2.x

PR #1475, opened 2020-03-25
Add basic transform benchmarking infrastructure and speed up Resolve Kinds

jackkoenig:
This is the first in a larger effort I'm working on to eliminate low hanging fruit performance issues. This is primarily a proof of concept and to show that even in seemingly simple passes we have pretty obvious room for performance improvement.

#### Resolve Kinds Benchmarking

I would appreciate someone checking my numbers, perhaps using BOOM or some large design from Chipyard.

These numbers were collected by running `sbt benchmark:assembly` then running using the fat jar.

```
$ java -version
openjdk version "1.8.0_232"
OpenJDK Runtime Environment (build 1.8.0_232-8u232-b09-0ubuntu1~16.04.1-b09)
OpenJDK 64-Bit Server VM (build 25.232-b09, mixed mode)
$ java -Xmx8G utils/bin/firrtl-benchmark.jar firrtl.benchmark.hot.ResolveKindsBenchmark <large_sifive_design>.pb 10 10
```

Times in milliseconds

| Commit | Desc | Mean | Median | Stddev |
| --- | --- | --- | --- | --- |
| e5ca9ac | no change | 571 | 571 | 49 |
| 567b271 | use HashMap | 342 | 341 | 20 |
| 5148c7f | eliminate traversal | 267 | 256 | 22 |

### Contributor Checklist

- [ ] Did you add Scaladoc to every public function/method?
- [ ] Did you add at least one test demonstrating the PR?
- [x] Did you delete any extraneous printlns/debugging code?
- [x] Did you specify the type of improvement?
- [x] Did you state the API impact?
- [x] Did you specify the code generation impact?
- [x] Did you request a desired merge strategy?

### Type of Improvement

 - performance improvement 

### API Impact

Technically this changes some APIs in `ResolveKinds`. This isn't backportable so maybe I should do this by copying the code and modifying the copy then deprecating the old `ResolveKinds`?

### Backend Code Generation Impact

None

### Desired Merge Strategy

 - Rebase: You will rebase the PR onto master and it will be merged with a merge commit.

### Reviewer Checklist (only modified by reviewer)
- [ ] Did you add the appropriate labels?
- [ ] Did you mark the proper milestone (1.2.x, 1.3.0, 1.4.0) ?
- [ ] Did you review?
- [ ] Did you mark as `Please Merge`?

PR #1582, opened 2020-05-09
Speed up CheckHighForm

jackkoenig:
This fixes a subtle performance regression from https://github.com/freechipsproject/firrtl/pull/870 and then because it highlighted some unnecessary error checking I improved performance further.

I have manually checked that this does not violate binary compatibility when backported to `1.3.x`

### Improvement 1

https://github.com/freechipsproject/firrtl/pull/870 made the error reporting in `CheckHighForm` (and `CheckChirrtl`) more detailed for Expressions when it's a type or width error. This includes eagerly serializing every single `Expression` twice. I made the functions accepting the serialization take their arguments lazily which prevents the serialization unless it's an error. 6-characters for a huge performance win.

### Improvement 2

In debugging (1), I noticed that we are checking the types of all Expressions in `CheckHighForm` (and `CheckChirrtl`). Because `Expressions` are recursive, this is `O(2^d*n)` where `d` is the depth of aggregate types and `n` is the number of types. 

There are no `Expressions` that have user-specified `Types`--all `Expression` types are derived from `Statement` declarations. The only errors found on `Types` here are negative Vec sizes and negative widths. Because such errors will be caught on the `Statement`s upstream, and errors for derived values in `Expressions` aren't useful anyway, I solve this issue by just turning off checking of `Types`. I'm pretty sure this is correct but would appreciate a second opinion

### Benchmark results

I have some benchmarking infrastructure I've been working on (see https://github.com/freechipsproject/firrtl/pull/1475 although I'm migrating to JMH). I haven't finished #1475 because it's really hard to measure the small improvements that those changes give, but using the same approach these improvements are obvious:

For a medium-sized design benchmarking `CheckHighForm`:
| commit | runtime (ms) | error (ms) |
| --- | --- | --- |
| master | 941 | 103 |
| Improvement 1 | 296 | 24 |
| Both improvements | 169 | 21 |

From a 4-7x speedup

**EDIT**: For a larger design, unsurprisingly improvements are larger:

| commit | runtime (ms) | error (ms) |
| --- | --- | --- |
| master | 18092 | 1804 |
| Improvement 1 | 14476 | 1090 |
| Both improvements | 1011 | 77 |

~15-20x speedup

### Contributor Checklist

#### Type of Improvement

   - performance improvement      

#### API Impact

No effect

No effect

#### Desired Merge Strategy

#### Release Notes
Fix performance regression in CheckChirrtl and CheckHighForm in `1.3.0` and improve performance further.

- [x] Did you review?

### DIFF
--- a/src/main/scala/firrtl/passes/CheckHighForm.scala
+++ b/src/main/scala/firrtl/passes/CheckHighForm.scala
@@ -191,7 +191,6 @@ trait CheckHighFormLike { this: Pass =>
         case ex => ex foreach validSubexp(info, mname)
       }
       e foreach checkHighFormW(info, mname + "/" + e.serialize)
-      e foreach checkHighFormT(info, mname + "/" + e.serialize)
       e foreach checkHighFormE(info, mname, names)
     }
 
--- a/src/main/scala/firrtl/passes/CheckHighForm.scala
+++ b/src/main/scala/firrtl/passes/CheckHighForm.scala
@@ -151,14 +151,14 @@ trait CheckHighFormLike { this: Pass =>
       case _ => // Do Nothing
     }
 
-    def checkHighFormW(info: Info, mname: String)(w: Width): Unit = {
+    def checkHighFormW(info: Info, mname: => String)(w: Width): Unit = {
       w match {
         case wx: IntWidth if wx.width < 0 => errors.append(new NegWidthException(info, mname))
         case wx => // Do nothing
       }
     }
 
-    def checkHighFormT(info: Info, mname: String)(t: Type): Unit = {
+    def checkHighFormT(info: Info, mname: => String)(t: Type): Unit = {
       t foreach checkHighFormT(info, mname)
       t match {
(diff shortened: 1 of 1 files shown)


src/main/scala/firrtl/passes/CheckHighForm.scala
```diff
@@ -191,7 +191,6 @@ trait CheckHighFormLike { this: Pass =>
         case ex => ex foreach validSubexp(info, mname)
       }
       e foreach checkHighFormW(info, mname + "/" + e.serialize)
```
albert-magyar, line 193:
```suggestion

```
Can you not drop this by the same logic? It doesn't do any meaningful work, but it would keep things kind of consistent.

jackkoenig: I originally thought the same thing, but literals actually do have user-defined types so we do need to check them.

albert-magyar:
Also, this might help reduce the massive stack footprint I see sometimes in `CheckHighForm` :)

PR #1592, opened 2020-05-11
Generalize keyword collision to name manipulation, Add {Lower,Upper}CaseNames Transforms

seldridge:
This generalizes the existing `RemoveKeywordCollisions` transform to a more generic `ManipulateNames` transform that will change all the names in a FIRRTL circuit that meet a condition using some function.

This PR then adds a `LowerCaseNames` and `UpperCaseNames` transform that will change all the names in a FIRRTL circuit to be either `lowercase` or `UPPERCASE`. The motivation here is to match (entirely arbitrary) internal company/institution naming conventions. One test is added for each transform.

The FIRRTL command line is extended with the `--change-name-case <lower|upper>` option.

### Contributor Checklist

#### API Impact

No impact. Pure API extension to add new transforms.

None. Users can optionally lower case or upper case all names. This is not turned on by default one way or the other.

#### Desired Merge Strategy

#### Release Notes

- Add {Lower,Upper}CaseNames transforms, --change-name-case <lower|upper>

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/features/LetterCaseTransform.scala
+++ b/src/main/scala/firrtl/features/LetterCaseTransform.scala
@@ -1,0 +1,38 @@
+// See LICENSE for license details.
+
+package firrtl.features
+
+import firrtl.{analyses, Namespace, passes, Transform}
+import firrtl.options.Dependency
+import firrtl.stage.Forms
+import firrtl.transforms.ManipulateNames
+
+/** Parent of transforms that do change the letter case of names in a FIRRTL circuit */
+abstract class LetterCaseTransform extends ManipulateNames {
+  override def prerequisites = Seq(Dependency(passes.LowerTypes))
+  override def optionalPrerequisites = Seq.empty
+  override def optionalPrerequisiteOf = Forms.LowEmitters
+  override def invalidates(a: Transform) = a match {
+    case _: analyses.GetNamespace => true
+    case _                        => false
+  }
+
+  protected def newName: String => String
+
+  final def condition = _ => true
+
+  final def manipulate = (a: String, ns: Namespace) => newName(a) match {
+    case `a` => a
+    case b   => ns.newName(b)
+  }
+}
+
+/** Convert all FIRRTL names to lowercase */
+final class LowerCaseNames extends LetterCaseTransform {
+  override protected def newName = (a: String) => a.toLowerCase
+}
+
+/** Convert all FIRRTL names to UPPERCASE */
+final class UpperCaseNames extends LetterCaseTransform {
+  override protected def newName = (a: String) => a.toUpperCase
+}
--- a/src/main/scala/firrtl/transforms/ManipulateNames.scala
+++ b/src/main/scala/firrtl/transforms/ManipulateNames.scala
@@ -1,0 +1,222 @@
+// See LICENSE for license details.
+
+package firrtl.transforms
+
+import firrtl._
+import firrtl.analyses.InstanceGraph
+import firrtl.annotations.{Named, CircuitName, ModuleName, ComponentName}
+import firrtl.Mappers._
+import firrtl.options.Dependency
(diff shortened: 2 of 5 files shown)


src/main/scala/firrtl/transforms/ManipulateNames.scala
```diff
@@ -0,0 +1,249 @@
...
+  }
+
+}
+
+/** Parent of transforms that do change the letter case of names in a FIRRTL circuit */
+abstract class LetterCaseTransform extends ManipulateNames {
+  override def prerequisites = Seq(Dependency(passes.LowerTypes))
+  override def optionalPrerequisites = Seq.empty
+  override def optionalPrerequisiteOf = Forms.LowEmitters
+  override def invalidates(a: Transform) = false
```
albert-magyar, line 229: This invalidates `analyses.GetNamespace` (which is not really used). It might be worth thinking about a "sensitive analysis" trait that things like this can invalidate.

src/main/scala/firrtl/transforms/ManipulateNames.scala
```diff
@@ -0,0 +1,249 @@
...
+  def execute(state: CircuitState): CircuitState = {
+    val renames = RenameMap()
+    renames.setCircuit(state.circuit.main)
+    state.copy(circuit = run(state.circuit, renames), renames = Some(renames))
+  }
+
+}
+
+/** Parent of transforms that do change the letter case of names in a FIRRTL circuit */
+abstract class LetterCaseTransform extends ManipulateNames {
```
albert-magyar, line 225: Is it worth thinking about a new package for new "feature" transforms? Moving all the core lowering transforms into a core lowering package will be something of a pain point, but it might be good to start something like `firrtl.features` or something.

albert-magyar: I think we can do the improvement later, the only necessary change for now would be reflecting the `GetNamespace` invalidation.

seldridge: Since this is 1.3.x, it's probably better to start the new package now as opposed to this getting into 1.3.x and then us needing to go through a deprecation cycle to move `LetterCaseTransform`. I'm good with `firrtl.features`. Any idea of a convention that compilers use for the separation of core vs. X?

src/main/scala/firrtl/transforms/ManipulateNames.scala
seldridge: Updated to invalidate this. Good catch.

src/main/scala/firrtl/transforms/ManipulateNames.scala
seldridge: Updated in the rebase to add `firrtl.features`.

src/main/scala/firrtl/features/LetterCaseTransform.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  protected def newName: String => String
+
+  final def condition = _ => true
+
+  final def manipulate = (a: String, ns: Namespace) => newName(a) match {
+    case `a` => a
+    case b   => ns.newName(b)
+  }
+}
```
jackkoenig, line 28: This is a method that returns a new function every time it's called, is that what you want? I grant there maybe CSE in Scala but I'm not sure for this kind of thing

jackkoenig: Actually nevermind, I'm sure it's just a single function, the method is returning a function literal.

albert-magyar: Yeah, it might be more clear that it doesn't have meaningful space overhead as a `val`, but a pure parameterless def with a constant return value should be fine.

PR #1594, opened 2020-05-11
Don't try deduping the main module of a circuit

albert-magyar:
**Type of improvement:** other enhancement
**API impact:** none

Currently, `DedupModules` attempts to dedup things against `circuit.main`.

This is not actually useful for two reasons:
- It will almost never accomplish anything in practice.
- Even in theory, it can only successfully dedup if the dupe of main is never instantiated, as is the case for the example in the test.

However, it can cause issues, since `buildRTLTags` will construct "temporary" annotations as part of the "dedup-ability" check. Since `circuit.main` isn't filtered out, this can lead to the construction of annotations with targets where the main module is renamed, but the target's circuit field is not. This can violate assumptions in the annotations, which restricts the ability of the annotations to express these annotations with `require` or `assert`.

One alternative would be to make the logic a bit more complicated in `buildRTLTags` for these cases, but since it's much easier to just prevent `circuit.main` from participating in a dedup group, and since deduping the top module is not generally useful, this change is a lot more straightforward.

### Contributor Checklist

- [] Did you add text to be included in the Release Notes for this change?

- [ ] Did you review?


albert-magyar:
> Makes sense to me. Duplicate hierarchies that would dedup will be handled by DCE, which if we migrate to mid or high form (a restricted version at least) can mitigate the performance impact of the additional modules.

There will be no performance penalty, since Chisel cannot generate code that will result in deduping against `circuit.main`. The only hypothetical example would be linking, and dedup across linking (and the `CircuitTarget` implications) is not something that is fully thought out.

albert-magyar:
@Mergifyio backport 1.3.x

PR #1603, opened 2020-05-13
[WIP] Implement CircuitGraph and IRLookup to firrtl.analyses

sequencer:
## Contributor Checklist

#### Type of Improvement
  - new feature/API     

#### API Impact
This PR implements `CircuitGraph` and `IRLookup` by rebasing #937 by @azidar into master branch.
status:
  - [x] rebase into master
  - [x] remove unrelated codes
  - [x] fix compiling
  - [x] fix testing

Generally this PR adds `CircuitGraph` and `ConnectionGraph` to construct connection information by Circuit IR, and adds `IRLookup` to lookup by `Target`.
Since DependencyAPI has been merged, I wondering if we should:
  - use `ConnectionGraphTransform`, `CircuitGraphTransform`, `InstanceGraphTransform` to store the these graphs. And add correspond analysis `Annotation`
  - when some circuit need to use a certain graph, it will add a prerequisite to that Transform and directly read that Annotation. 
  - when circuit is changed by other transforms, analysis transform will be invalided.

#### Desired Merge Strategy

##### Existing codes changelog: 
- add `def path: Seq[(Instance, OfModule)]` to `trait Target`
- implement `path` to `case class GenericTarget`
- add `encapsulatingModule: String`, `encapsulatingModuleTarget: ModuleTarget`, `leafModule: String` to `trait IsMember`
- remove `encapsulatingModule` in `trait IsComponent`, since it has been implemented in `IsMember`
- implement `path` to `CircuitTarget`
- implement `leafModule` to `ModuleTarget`
- add `def isClock: Boolean`, `def isInit: Boolean`, `def isReset: Boolean`, `def noComponents: ReferenceTarget`, `def leafSubTargets(tpe: firrtl.ir.Type): Seq[ReferenceTarget]`, `def allSubTargets(tpe: firrtl.ir.Type): Seq[ReferenceTarget]`, `def leafModule: String` to `ReferenceTarget`.
- add `def addReference(rt: ReferenceTarget): ReferenceTarget` , and implement `def leafModule: String` to `case class InstanceTarget`.

##### New features:
- Create `ConnectionGraph` based on `Circuit`, which is a `DiGraph[ReferenceTarget]` to record graph metadata for `ReferenceTarget` for IR.
- Create `CircuitGraph` based on `ConnectionGraph`, which use `ConnectionGraph` to expose more public API for circuit connection information.
- Create `IRLookup` based on `ConnectionGraph`, which let user to map `ReferenceTarget` to specific element in the circuit.

- [ ] Did you review?


src/main/scala/firrtl/graph/DiGraph.scala
```diff
@@ -34,7 +34,11 @@ object DiGraph {
 }
 
 /** Represents common behavior of all directed graphs */
-class DiGraph[T] private[graph] (private[graph] val edges: LinkedHashMap[T, LinkedHashSet[T]]) {
+trait DiGraphLike[T] {
+  private[graph] val edges: LinkedHashMap[T, LinkedHashSet[T]]
+
+  private[graph] val prev: LinkedHashMap[T, T]
```
albert-magyar, line 40: This isn't used in anything operating on a `DiGraphLike`, so maybe it should only be in a separate trait extended by the derived classes that use it? As far as I can tell, it looks like it could even be private to the derived classes.

sequencer: I made `DiGraph` to public and abstract to solve this issue, is this a good way to do so?

albert-magyar: I guess I mean why does `DiGraph` have `prev` if it isn't used?

sequencer:
rebase on to `always-wir` to fix #1621.

sequencer:
In 5a9f910 `Bundle` type cache by `IRLookup` is different from `Expression` bundle type. solution I can come up with is changing all `Seq` to `Vec` in `IR` `case class`, or update to 2.13. Any good suggestions?

jackkoenig:
> In 5a9f910 Bundle type cache by IRLookup is different from Expression bundle type.

Can you expand on this, what precisely is the issue?

The types are not equal but they both still implement structural equality:
```scala
Welcome to Scala 2.12.11 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_162).
Type in expressions for evaluation. Or try :help.

scala> import scala.collection.mutable.ArrayBuffer
import scala.collection.mutable.ArrayBuffer

scala> Seq(1, 2, 3) == ArrayBuffer(1, 2, 3)
res0: Boolean = true
```

> solution I can come up with is changing all Seq to Vec in IR case class, or update to 2.13

Both of these would be massively backwards incompatible. We cannot just switch to 2.13 without a long period of cross compiling to 2.12 (we haven't even dropped 2.11 yet, so we need at least a major release cycle of adding 2.13 and deprecating 2.11). I have thought about switching to a concrete type instead of `Seq` for performance reasons, but that too is a massive undertaking that shouldn't undertake lightly.

src/main/scala/firrtl/graph/DiGraph.scala
sequencer: I think there still needs some codes clean up after I fixes some tests.

sequencer:
Sorry @jackkoenig I just found the bug, is not related to 'ArrayBuffer' and `List`.
It was an internal logic issue from original codes.

albert-magyar:
@sequencer, that does appear to be a bug in `Utils.expandRef` -- at the very least, it doesn't match the description from the API spec in the Scaladoc for the method. Unfortunately, this is one of the pain points of using a method that doesn't get used in more than one place in the codebase.

I made the behavior of `expandRef` line up with the description in #1624 -- feel free to try that out and see if it fixes your issue.

sequencer:
Every tests passed! Thank you! @albert-magyar, after #1624 upstream, I will rebase to master.

src/main/scala/firrtl/graph/DiGraph.scala
```diff
@@ -48,7 +53,7 @@ class DiGraph[T] private[graph] (private[graph] val edges: LinkedHashMap[T, Link
     * @param v the specified node
     * @return a Set[T] of all vertices that v has edges to
     */
-  def getEdges(v: T): Set[T] = edges.getOrElse(v, Set.empty)
+  def getEdges(v: T, prevOpt: Option[collection.Map[T, T]] = None): Set[T] = edges.getOrElse(v, Set.empty)
```
albert-magyar, line 56: One of my main comments with the original PR was about the addition of this extra parameter to `getEdges`.

This parameter is only used in `BFS`, and it fundamentally changes the behavior of the underlying implementation of BFS. My thought is that it would make more sense to just rely on overloading `BFS` in the derived class or something like that.

src/main/scala/firrtl/graph/DiGraph.scala
```diff
@@ -29,12 +28,18 @@ object DiGraph {
...
-    new DiGraph(edgeDataCopy)
+    new DiGraph[T] {
+      val edges = edgeDataCopy
+    }
   }
 }
 
-/** Represents common behavior of all directed graphs */
-class DiGraph[T] private[graph] (private[graph] val edges: LinkedHashMap[T, LinkedHashSet[T]]) {
+abstract class DiGraph[T] {
```
albert-magyar, line 37: What is the incentive for making this abstract?

src/main/scala/firrtl/graph/DiGraph.scala
```diff
@@ -29,12 +28,18 @@ object DiGraph {
...
+
+  private[graph] val edges: LinkedHashMap[T, LinkedHashSet[T]]
+
+  private[graph] val prev = new mutable.LinkedHashMap[T, T]()
```
albert-magyar, line 41: This is not used.

sequencer: cleaned up

src/main/scala/firrtl/graph/DiGraph.scala
sequencer: reverted.

sequencer:
wait #1624 to be merged, codes cleaned up.

sequencer:
quite strange, in https://travis-ci.org/github/freechipsproject/firrtl/jobs/689972026:
 /home/travis/build/freechipsproject/firrtl/src/main/scala/firrtl/analyses/ConnectionGraph.scala:412:3: Could not find any member to link for "DiGraph".
I did import `DIGraph` with:
`import firrtl.graph.{CyclicException, DiGraph, MutableDiGraph}`

sequencer:
@azidar ready for review now, sorry for the delay.

src/main/scala/firrtl/analyses/CircuitGraph.scala
```diff
@@ -0,0 +1,137 @@
...
+    * @return
+    */
+  def apply(circuit: Circuit): CircuitGraph = new CircuitGraph(ConnectionGraph(circuit))
+
+  /** Return a nicely-formatted string of a path of [[firrtl.annotations.ReferenceTarget]]
+    * @param path
+    * @param tab
+    * @return
+    */
+  def prettyToString(path: Seq[ReferenceTarget], tab: String = ""): String = {
```
azidar, line 27: I wonder if this `path` is confusing, because we use the same word to mean either the instance path to a non-local reference, or a connection path through a graph. Perhaps change it to `connectionPath` or something like that?

src/main/scala/firrtl/analyses/CircuitGraph.scala
```diff
@@ -0,0 +1,137 @@
...
+      case (it, DefInstance(_, instance, ofModule, _)) => it.instOf(instance, ofModule)
+    }
+  }
+
+  /** Return the sequence of nodes from source to sink, inclusive
+    * @param source
+    * @param sink
+    * @return
+    */
+  def path(source: ReferenceTarget, sink: ReferenceTarget): Seq[ReferenceTarget] =
```
azidar, line 92: Similarly, I think we should rename this.

src/main/scala/firrtl/analyses/ConnectionGraph.scala
```diff
@@ -0,0 +1,684 @@
...
+    * Top>a.in            List()
+    * Top/a:A>in          List(Top>a.in)
+    * Top/a:A>b.in        List(Top>a.in)
+    * Top/a:A/b:B/in      List(Top/a:A>b.in, Top>a.in)
+    * Top/a:A/b:B/out     List(Top/a:A>b.in, Top>a.in)
+    * Top/a:A>b.out       List(Top>a.in)
+    * Top/a:A>out         List(Top>a.in)
+    * Top>a.out           List()
+    * Top>out             List()
+    * when we reach Top/a:A>in the stack is List
```
azidar, line 70: It seems like this comment ended abrupty? Are we missing something?

src/main/scala/firrtl/analyses/ConnectionGraph.scala
```diff
@@ -0,0 +1,684 @@
...
+    */
+  def shortCut(target: ReferenceTarget): Set[ReferenceTarget] = getShortCut(target).get
+
+  /** Returns a new, reversed connection graph where edges point from sinks to sources
+    *
+    * @return
+    */
+  def reverseConnectionGraph: ConnectionGraph = new ConnectionGraph(circuit, digraph.reverse, irLookup)
+
+  def getTag(node: ReferenceTarget,
```
azidar, line 130: Should we add scaladoc for this? It seems like all of the tagging stuff is not used in this PR. I believe it was necessary for the clock domain tracking stuff, which has been stripped away. Personally, I think this code is not particularly generic for the connectivity graph, and I remember trying to refactor it anyways in a different branch. How do you feel about removing it for now? This includes `getTag`, `setTag`, `tagPath`, etc.

src/main/scala/firrtl/analyses/IRLookup.scala
```diff
@@ -0,0 +1,343 @@
...
+import firrtl.annotations.TargetToken._
+import firrtl.annotations._
+import firrtl.ir._
+import firrtl.passes.MemPortUtils
+import firrtl.{DuplexFlow, ExpKind, SinkFlow, Flow, InstanceKind, Kind, SourceFlow, MemKind, PortKind, RegKind, UnknownFlow, Utils, WInvalid, WireKind}
+
+import scala.collection.mutable
+
+object IRLookup {
+  def apply(circuit: Circuit): IRLookup = ConnectionGraph(circuit).irLookup
```
azidar, line 14: When I wrote this, I couldn't think of a better way to construct the IRLookup independently from the ConnectionGraph, yet still guaranteed to be in sync with a connection graph if you did want it. I think in a future PR we could potentially improve this. No action needed, besides maybe a note?

src/main/scala/firrtl/analyses/IRLookup.scala
```diff
@@ -0,0 +1,343 @@
...
+}
+
+/** Handy lookup for obtaining AST information about a given Target
+  * @param declarations Maps references (not subreferences) to declarations
+  * @param modules Maps module targets to modules
+  */
+class IRLookup private[analyses] ( private val declarations: mutable.LinkedHashMap[ModuleTarget, mutable.LinkedHashMap[ReferenceTarget, FirrtlNode]],
```
azidar, line 21: I believe @jackkoenig ran into a performance issue where the hashmaps got too big. We should just keep an eye out for that, especially for the `flowCache`, `kindCache`, `tpeCache` and `exprCache`. We may want to guard the lookups with the `ModuleTarget` first.

src/main/scala/firrtl/analyses/ConnectionGraph.scala
sequencer: Removing it :)

src/main/scala/firrtl/analyses/IRLookup.scala
sequencer: Sure, in my next PRs, I'll find some other way to construct `IRLookup`.

sequencer: In my next PRs I think I should rewrite IRLookup construction.

src/main/scala/firrtl/analyses/IRLookup.scala
sequencer: In my next PRs I think I should rewrite IRLookup construction.

sequencer: Currently, `IRLookup` is constructed by BFS when constructing `ConnectionGraph`, I think:
1. we can use simple linear search to visit AST to construct `IRLookup`, parallel level will be much higher, but need additional search.
2. when visiting AST, we can directly generate flow, kind, tpe, expr etc Map, rather using cache to consult, but it will lead to huge memory usage. which is your prefer? the mutable maps in IRLookup are so scary.
3. performance issue you mentioned has been fixed by ModuleTarget guarding

ekiwi-sifive:
@sequencer can you rebase this on the latest `master` please?
This is quite a big PR and there might be some Scala 2.13 incompatible changes.

sequencer:
Sorry for the delay!
Work on it today!

sequencer:
> This is quite a big PR and there might be some Scala 2.13 incompatible changes.

I think it's OK now:)

src/main/scala/firrtl/analyses/CircuitGraph.scala
```diff
@@ -0,0 +1,137 @@
...
+  lazy val reverseConnectionGraph = connectionGraph.reverseConnectionGraph
+
+  // AST Circuit
+  val circuit = connectionGraph.circuit
+
+  // AST Information
+  val irLookup = connectionGraph.irLookup
+
+  // Module/Instance Hierarchy information
+  lazy val instanceGraph = new InstanceGraph(circuit)
```
ekiwi-sifive, line 51: This is going to race with this PR of mine: https://github.com/freechipsproject/firrtl/pull/1800

So depending on who gets in first, the other will have to fix some things.

src/main/scala/firrtl/analyses/CircuitGraph.scala
```diff
@@ -0,0 +1,137 @@
...
+  *
+  * @param connectionGraph Source-to-sink connectivity graph
+  */
+class CircuitGraph private[analyses] (val connectionGraph: ConnectionGraph) {
+
+  // Reverse (sink-to-source) connectivity graph
+  lazy val reverseConnectionGraph = connectionGraph.reverseConnectionGraph
+
+  // AST Circuit
+  val circuit = connectionGraph.circuit
```
ekiwi-sifive, line 45: Are you sure that you want to expose this as a `val`? Making it a method could facilitate making internal changes later.

src/main/scala/firrtl/analyses/CircuitGraph.scala
sequencer: Sure I'll use `InstanceKeyGraph`.

ekiwi-sifive: The current version of `InstanceKeyGraph` isn't feature complete. The feature complete version will only be available once #1800 is merged. We could merge #1603 first and then I would be happy to add the change to `InstanceKeyGraph` in #1800.

sequencer: Thank you! If the change takes time, you can just deprecate it and leave the upgrade patch to me.

ekiwi-sifive: It is already part of https://github.com/freechipsproject/firrtl/pull/1800. So you are good.

src/main/scala/firrtl/analyses/CircuitGraph.scala
ekiwi-sifive: Fixed in https://github.com/freechipsproject/firrtl/pull/1813

PR #1643, opened 2020-05-25
recore of Attributes

sequencer:
This pull request rebases great work #1172 by @grebe to master.

### Contributor Checklist

#### Type of Improvement
- new feature/API

#### API Impact
- break:
convert `DescriptionAnnotation` into `sealed trait`
- add:
`DocStringAnnotation`, `AttributeAnnotation`

None

#### Desired Merge Strategy

#### Release Notes
Verilog can have both attributes and doc string now.
 
- [x] Did you review?

### DIFF
--- a/src/main/scala/firrtl/AddDescriptionNodes.scala
+++ b/src/main/scala/firrtl/AddDescriptionNodes.scala
@@ -7,12 +7,22 @@ import firrtl.annotations._
 import firrtl.Mappers._
 import firrtl.options.{Dependency, PreservesAll}
 
-sealed trait DescriptionType
-case object DocStringDescription extends DescriptionType
-case object AttributeDescription extends DescriptionType
+sealed trait DescriptionAnnotation extends Annotation {
+  val named: Named
+  val description: String
 
-case class DescriptionAnnotation(named: Named, description: String, tpe: DescriptionType) extends Annotation {
-  def update(renames: RenameMap): Seq[DescriptionAnnotation] = {
+}
+
+case class DocStringAnnotation(named: Named, description: String) extends DescriptionAnnotation {
+  def update(renames: RenameMap): Seq[DocStringAnnotation] = {
+    renames.get(named) match {
+      case None => Seq(this)
+      case Some(seq) => seq.map(n => this.copy(named = n))
+    }
+  }
+}
+case class AttributeAnnotation(named: Named, description: String) extends DescriptionAnnotation {
+  def update(renames: RenameMap): Seq[AttributeAnnotation] = {
     renames.get(named) match {
       case None => Seq(this)
       case Some(seq) => seq.map(n => this.copy(named = n))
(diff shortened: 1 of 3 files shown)


sequencer:
@jackkoenig @azidar Any ideas to fix the unidoc issue?

kammoh:
@sequencer this is what I did, but I suppose there should be a cleaner way to do the partitioning as well.

```
diff --git a/src/main/scala/firrtl/AddDescriptionNodes.scala b/src/main/scala/firrtl/AddDescriptionNodes.scala
index eaf5d2df..5a6b3061 100644
--- a/src/main/scala/firrtl/AddDescriptionNodes.scala
+++ b/src/main/scala/firrtl/AddDescriptionNodes.scala
@@ -185,22 +185,22 @@ class AddDescriptionNodes extends Transform with DependencyAPIMigration with Pre
     * @return List of `Description`s with some descriptions merged
     */
   def mergeDescriptions(descs: Seq[Description]): Seq[Description] = {
-    val (docs: Seq[DocString], nodocs) = descs.partition {
+    val (docs, nodocs) = descs.partition {
       case _: DocString => true
       case _ => false
     }
-    val (attrs: Seq[Attribute], rest) = nodocs.partition {
+    val (attrs, rest) = nodocs.partition {
       case _: Attribute => true
       case _ => false
     }
 
     val doc = if (docs.nonEmpty) {
-      Seq(DocString(StringLit.unescape(docs.map(_.string.string).mkString("\n\n"))))
+      Seq(DocString(StringLit.unescape(docs.map(_.asInstanceOf[DocString].string.string).mkString("\n\n"))))
     } else {
       Seq()
     }
     val attr = if (attrs.nonEmpty) {
-      Seq(Attribute(StringLit.unescape(attrs.map(_.string.string).mkString(", "))))
+      Seq(Attribute(StringLit.unescape(attrs.map(_.asInstanceOf[Attribute].string.string).mkString(", "))))
     } else {
       Seq()
     }
```

sequencer:
quite strange, why would type annotation blocking unidoc?

albert-magyar:
The unidoc CI is set up to turn non-deprecation warnings into errors, but since it is subject to some degree of caching, PRs often reveal pre-existing warning problems when files re-compile for the first time in a while.

The "ideal" thing to do is to fix the underlying warnings, which here would involve turning, unchecked matches against parameterized types (like `Seq[DocString]`) into matches not attempting to spuriously match the parameter by using a wildcard (like `Seq[_]`) or using the `@unchecked` Scala compiler annotation.

Usually we just title those commits with the names of the issues they fix; here, something like "Fix unchecked type parameter matches in AddDescriptionNodes" would be ideal.

I can also help with this and add one extra commit to this branch to fix those up.

sequencer:
Sorry for the delay, I was on the road last week .
I think this PR is ready for review now. What do your think? @albert-magyar @grebe

azidar:
@grebe do you care about reviewing this? Otherwise I don't mind :)

src/main/scala/firrtl/AddDescriptionNodes.scala
```diff
@@ -7,31 +7,79 @@ import firrtl.annotations._
...
 import firrtl.options.{Dependency, PreservesAll}
 
-case class DescriptionAnnotation(named: Named, description: String) extends Annotation {
-  def update(renames: RenameMap): Seq[DescriptionAnnotation] = {
+/**
+  * A base trait for `Annotation`s that describe a `FirrtlNode`.
+  * Usually, we would like to emit these descriptions in some way.
+  */
+sealed trait DescriptionAnnotation extends Annotation {
+  def named: Named
```
azidar, line 15: These should be Target, not Named.

src/main/scala/firrtl/AddDescriptionNodes.scala
```diff
@@ -7,31 +7,79 @@ import firrtl.annotations._
...
+  def description: String
+}
+
+/**
+  * A docstring description (a comment).
+  * @param named the object being described
+  * @param description the docstring describing the object
+  */
+case class DocStringAnnotation(named: Named, description: String) extends DescriptionAnnotation {
```
azidar, line 24: Similarly, this should be Target.

src/main/scala/firrtl/AddDescriptionNodes.scala
```diff
@@ -7,31 +7,79 @@ import firrtl.annotations._
...
+    }
+  }
+}
+
+/**
+  * An Verilog-style attribute.
+  * @param named the object being given an attribute
+  * @param description the attribute
+  */
+case class AttributeAnnotation(named: Named, description: String) extends DescriptionAnnotation {
```
azidar, line 38: Same, this should be Target.

sequencer:
I personally prefer sealed since `AttributeAnnotation` and  `DocStringAnnotation` are extending from it, both of which are used in the `Emitter`, user should not provide their own implementation to `DocStringAnnotation`, since they cannot modify `Emitter` without changing firrtl codes.

PR #1645, opened 2020-05-27
Support Memory Initialization for Simulation and FPGA Flows

ekiwi:
### Contributor Checklist

  - it is an annotation, and thus outside of the firrtl spec

#### Type of Improvement
- new feature

#### API Impact

Add new `InitMemoryAnnotation` as well as a sealed `MemoryInitValue` trait.

Changes the emitter to generate an `init` block with initial value assignments if a memory is annotated with a `MemoryScalarInit` or `MemoryArrayInit`. If the new `InitMemoryAnnotation` is not used, this change should not affect the code generation.

The generated `init` code should work with `verilator` (which is checked in the unit tests) as well as yosys (not checked in the unit test, but according to their documentation, this is the way to initialize memories).

#### Desired Merge Strategy

Squash

#### Release Notes
* A new `InitMemoryAnnotation` makes it possible to deterministically initialize memories for simulation and FPGA flows.

- [x] Did you review?


src/main/scala/firrtl/Emitter.scala
```diff
@@ -15,8 +14,9 @@ import firrtl.PrimOps._
 import firrtl.WrappedExpression._
 import Utils._
 import MemPortUtils.{memPortField, memType}
-import firrtl.options.{Dependency, HasShellOptions, ShellOption, StageUtils, PhaseException, Unserializable}
+import firrtl.options.{Dependency, HasShellOptions, PhaseException, ShellOption, StageUtils, Unserializable}
```
ekiwi, line 17: I will clean this up.

ekiwi:
This needs two follow up changes:
1) add support for `InitMemoryAnnotation` to `treadle`
2) add Chisel frontend support (this will have to deal with the question of non-ground-type memories)

src/main/scala/firrtl/passes/LowerTypes.scala
```diff
@@ -280,15 +285,17 @@ object LowerTypes extends Transform with DependencyAPIMigration {
...
         m copy (ports = portsx)
       case m: Module =>
-        m copy (ports = portsx) map lowerTypesStmt(memDataTypeMap, m.info, m.name, renames)
+        m copy (ports = portsx) map lowerTypesStmt(memDataTypeMap, m.info, m.name, renames, initializedMems)
     }
   }
 
   def execute(state: CircuitState): CircuitState = {
+    // TODO: how do we match memories that are not in the top module
+    val initializedMems = state.annotations.collect{case m : MemoryInitAnnotation if !m.isRandomInit => m.target.ref}.toSet
```
ekiwi, line 294: This needs to be fixed. Maybe someone like @seldridge or @azidar could point me to an example on how to correctly match all annotated memories no matter which module they are in.

albert-magyar: This does in fact match all annotated memories. However, it is associating them with their declared name only, which is only meaningful in the scope of an individual module. A better approach would be:
```suggestion
    val initializedMems = state.annotations.collect {
      case m : MemoryInitAnnotation if !m.isRandomInit => (m.target.module, m.target.ref)
    }.toSet
```

src/main/scala/firrtl/passes/LowerTypes.scala
```diff
@@ -210,6 +211,10 @@ object LowerTypes extends Transform with DependencyAPIMigration {
         sx.dataType match {
           case _: GroundType => sx
           case _ =>
+            // right now only ground type memories can be initialized
+            if(initializedMems.contains(sx.name)) {
```
albert-magyar, line 215:
```suggestion
            if(initializedMems.contains((mname, sx.name))) {
```

albert-magyar: Again, this is just an example of how to fix the issue.

albert-magyar:
Another issue is that these annotations will not be handled correctly when applied to a `ReferenceTarget` pointing to a memory in a particular _instance_ of a module. This can arise either from a user directly attaching the annotation to an instance, or as a result of `DedupModules`.

However, this has nothing to do with how you've implemented the PR -- it appears to be a pre-existing flaw that would also affect the other emission options. I imagine it was not diagnosed since this "dedup annotated modules" pattern was under-utilized, as it is still waiting on #1539 for best functionality. The solution will involve having the `Emitter` extend `ResolvedAnnotationPaths`, and could just be a separate PR after #1359.

Note that this is not an issue for your changes to `LowerTypes`, since an annotation to an aggregate-typed memory in a single instance of a module is still an error that must be detected by analyzing that module.

src/main/scala/firrtl/passes/LowerTypes.scala
seldridge: Albert's approach works. I'd probably write this as a `Set[ReferenceTarget]` or `Set[Target]` and then do checking inside each module. This pushes towards a pattern of walking the AST while tracking the target. This assumes, as [Albert mentions](https://github.com/freechipsproject/firrtl/pull/1645#issuecomment-634874207) that the circuit has been properly duplicated such that all memory targets have no instances in their path.

albert-magyar: @seldridge, an alternative to duplicating before `LowerTypes` is just to consider that "an error in one instance is still an error in the module."

Note the use of encapsulatingModule, which is the actual module containing the reference. The module field of the ReferenceTarget is the root of its cross-module reference path.

I.e, if I wish to target mem m in module t, but only in the instance defined by the hierarchical path q.r.s in module q, the module would be q, while the encapsulatingModule (that actually contains the memory) would be t.

ekiwi:
So how to we proceed? I was hoping to get this into Firrtl 1.4 and then add Chisel and Treadle support once 1.4 is released.

Are there any style issues you would like me to address?
Should the `MemoryInitValue` sealed trait be moved into the same file as the annotation or is it ok if it lives with the emission options?

albert-magyar:
I only looked at this to the extent to try to suggest the fix to matching the memories. It seems you've fixed that now.

ekiwi-sifive:
As we discussed in the dev meeting today: @albert-magyar / @jackkoenig : please let me know which init block the memory initialization should go into and which defines should guard it (if any).

albert-magyar:
Have you looked at #1050? It would be good to see if there are opportunities to leverage shared code and make the APIs as homogeneous as possible.

ekiwi-sifive:
> Have you looked at #1050? It would be good to see if there are opportunities to leverage shared code and make the APIs as homogeneous as possible.

I re-used the `EmissionOption` infrastructure in the Emitter code that was introduced by PR#1050. However, while #1050 used the `reset` field of the `DefRegister` construct to carry the initial value, I needed to put the initial value in the Annotation since there is not way in the Firrtl IR to describe an init value for a memory.

albert-magyar:
Thanks, that sounds good. I also like that you do not introduce new ifdefs but do support the old ifdef if there are no memories initialized with the new API.

ekiwi-sifive:
@seldridge the last feature missing here is to ensure that the annotation can be serialized. Can you give me a hint on what is necessary to ensure that?

After that is done, this is ready for the final review.

src/main/scala/firrtl/EmissionOption.scala
```diff
@@ -9,6 +9,18 @@ package firrtl
   */
 trait EmissionOption
 
+/** Emission customization options for memories */
+trait MemoryEmissionOption extends EmissionOption {
+  def initValue: MemoryInitValue = MemoryRandomInit()
+}
+
+sealed trait MemoryInitValue
+case class MemoryRandomInit() extends MemoryInitValue
```
albert-magyar, line 18: This should be a case object since it's a nullary product type.
```suggestion
case object MemoryRandomInit() extends MemoryInitValue
```

seldridge:
For custom serializers, I think you need to implement a `CustomSerializer` typeclass and add it here: https://github.com/freechipsproject/firrtl/blob/632930723adc2f78d4d6445acf5f8bcc250a6c0c/src/main/scala/firrtl/annotations/JsonProtocol.scala#L23.

albert-magyar:
> @seldridge the last feature missing here is to ensure that the annotation can be serialized. Can you give me a hint on what is necessary to ensure that?

It looks like everything you use is a nesting of products and/or collections of basic, intrinsically serializable types. You will not need to do anything to make it serializable.

I tried the following in an annotation file (for a mem of depth 4), and it worked just fine:
```
[
    {
        "class": "firrtl.annotations.MemoryInitAnnotation",
        "target": "~MemTest|MemTest>m",
        "value": { "class": "firrtl.MemoryArrayInit", "values": [0, 1, 2, 3] }
    }
]
```
Unfortunately, if you take my suggestion of using a case object for the nullary `MemoryRandomInit`, **this will no longer be the case**. Perhaps that's why you made it a nullary case class. This is something that could potentially be addressed upstream.

src/main/scala/firrtl/EmissionOption.scala
albert-magyar: See my other comment

ekiwi-sifive:
> 
> Unfortunately, if you take my suggestion of using a case object for the nullary `MemoryRandomInit`, **this will no longer be the case**. Perhaps that's why you made it a nullary case class. This is something that could potentially be addressed upstream.

I just did that for uniformity reasons. But thanks for trying this out!

Should I just keep things the way they are right now?

albert-magyar:
> Should I just keep things the way they are right now?

Since the JSON protocol support for singletons doesn't have anything to do with this PR, I guess the two options are writing a custom serializer for your annotation or just leaving it the way it is. Since it would be easy to fix this later, it seems like this is maybe outside the scope of this PR, but it would be good if @davidbiancolin could chime in.

albert-magyar:
I have a new idea after thinking about this some more. The serialization protocol for annotations is designed to support custom annotation classes that ideally are product types with as simple a type signature as possible. While nesting a `MemoryInitValue` inside a `MemoryInitAnnotation` will currently deserialize without any additional code, this is something of an implementation-dependent behavior of the current JSON protocol.

A more robust way to get the same functionality (and sidestep the desire to use singletons to encode "flavor" of initialization) is to make a different annotation for each type of initialization. By extending a common trait, this would require only minor, non-structural changes to the other code.

```
sealed trait MemoryInitAnnotation
  extends SingleTargetAnnotation[ReferenceTarget]
  with MemoryEmissionOption

case class MemoryRandomInitAnnotation(target: ReferenceTarget)
  extends MemoryInitAnnotation {
  ...
}

case class MemoryScalarInitAnnotation(target: ReferenceTarget, value: BigInt)
  extends MemoryInitAnnotation {
  ...
}

case class MemoryArrayInitAnnotation(target: ReferenceTarget, values: Seq[BigInt])
  extends MemoryInitAnnotation {
  ...
}
```

What do you think about this?

ekiwi:
> What do you think about this?

I like it. See last commit.

ekiwi:
@albert-magyar : now that `MemoryInitValue` isn't part of the annotation anymore, should I make it packet private?

albert-magyar:
> @albert-magyar : now that `MemoryInitValue` isn't part of the annotation anymore, should I make it packet private?

Only if you want, it's sealed and very clear what each of them is, so I think there's no risk in leaving them public.

ekiwi-sifive:
> Only if you want, it's sealed and very clear what each of them is, so I think there's no risk in leaving them public.

OK. Then let's leave it as it is.

PR #1646, opened 2020-05-27
Limit recursion (and reduce total work) in CheckWidths

albert-magyar:
**Type of change:** performance improvement
**API impact:** none

I was experiencing stack overflows (and insane runtime when increasing the stack size) in `CheckWidths` for some very very large designs with deep expression trees. I tried changing the implementation to limit recursion depth, and it fixed my problem.

Here is some extra rationale for the extra code complexity:
* I tried moving to an iterative approach, but the overhead of materializing the order slowed down the common case, where an expression tree has 1-3 nodes.
* The recursion limit and switch to DFS at a moderate depth seems to add minimal overhead.
* By observing that `WSubField`, `WRef`, `WSubIndex`, and `WSubAccess` _cannot_ contain subtrees that cause the errors checked by this transform, I was able to reduce the total amount of work done in expression tree checking.

Taking all of these together, this change seems to be neutral for small designs and the regression designs but very helpful for pathological designs.

### Contributor Checklist

- [N/A] Did you update the FIRRTL spec to include every new feature/behavior?
- [N/A] Did you add at least one test demonstrating the PR?

- [x] Did you review?


jackkoenig:
Results are noisy but looks like a slight improvement across the board for our standard designs. For a large, crazy design with deeper Expressions it speed things up a solid 18% in CheckWidths:
```
Before: 1209.3 ± 181.7 ms
After:  987.4  ± 105.1 ms
```

benchmark/src/main/scala/firrtl/benchmark/hot/PassBenchmark.scala
```diff
@@ -0,0 +1,32 @@
...
+  hot.util.benchmark(warmup, runs)(passFactory().run(preState.circuit))
+}
+
+object ResolveKindsBenchmark extends PassBenchmark(() => ResolveKinds)
+
+object CheckHighFormBenchmark extends PassBenchmark(() => CheckHighForm)
+
+object CheckWidthsBenchmark extends PassBenchmark(() => CheckWidths)
+
+object InferBinaryPointsBenchmark extends PassBenchmark(() => new InferBinaryPoints)
```
seldridge, line 32: This starts to motivate a `PassBenchmark` that takes a string argument off the command line.

src/main/scala/firrtl/passes/CheckWidths.scala
```diff
@@ -109,46 +109,65 @@ object CheckWidths extends Pass with PreservesAll[Transform] {
...
-            case (IntervalType(Closed(la), Closed(ua), _), IntervalType(Closed(lb), Closed(ub), _)) if (ua < lb) || (ub < la) =>
-              errors.append(new DisjointSqueeze(info, target.serialize, sqz))
-            case other =>
+    def check_width_e_inner(info: Info, target: Target, expr: Expression): Unit = {
+      expr match {
+        case e @ UIntLiteral(v, w: IntWidth) if math.max(1, v.bitLength) > w.width =>
+          errors.append(new WidthTooSmall(info, target.serialize, v))
+        case e @ SIntLiteral(v, w: IntWidth) if v.bitLength + 1 > w.width =>
+          errors.append(new WidthTooSmall(info, target.serialize, v))
+        case e @ DoPrim(op, Seq(a, b), _, tpe) =>
```
seldridge, line 118: `DoPrim` code cleanup looks good.

src/main/scala/firrtl/passes/CheckWidths.scala
```diff
@@ -109,46 +109,65 @@ object CheckWidths extends Pass with PreservesAll[Transform] {
...
+            case _ =>
+          }
+        case _ =>
+      }
+    }
+
+    def check_width_e(info: Info, target: Target, recDepth: Integer)(expr: Expression): Unit = {
+      check_width_e_inner(info, target, expr)
+      expr match {
+        case _: Mux | _: ValidIf | _: DoPrim =>
```
seldridge, line 147: Can you include a comment here on why the recursion is limited? (As you state: ref/subfield/subaccess/subindex aren't going to be checked by `check_width_e_inner`, so there's no point in traversing them.)

src/main/scala/firrtl/passes/CheckWidths.scala
```diff
@@ -109,46 +109,65 @@ object CheckWidths extends Pass with PreservesAll[Transform] {
...
+              errors.append(new HeadWidthException(info, target.serialize, n, bitWidth(a.tpe)))
+            case (Tail, Seq(n)) if (hasWidth(a.tpe) && bitWidth(a.tpe) < n) =>
+              errors.append(new TailWidthException(info, target.serialize, n, bitWidth(a.tpe)))
+            case (AsClock, _) if (bitWidth(a.tpe) != 1) =>
+              errors.append(new MultiBitAsClock(info, target.serialize))
+            case (AsAsyncReset, _) if (bitWidth(a.tpe) != 1) =>
+              errors.append(new MultiBitAsAsyncReset(info, target.serialize))
+            case _ =>
+          }
+        case _ =>
```
seldridge, line 140: Related to comment on `check_width_e`, can you include a comment for this case on why the expressions are not recursively traversed (no point as others won't be checked).

albert-magyar:
Good suggestion, I added some comments to make it clearer what each piece is doing and why children of `WRef` et al. are ignored.

Issue #567, opened 2017-05-10
Improve verilog backend for formal tools


azidar:
Emit assert and coverage points from verilog backend.

ekiwi-sifive:
Finally fixed by @tdb-alcorn !

(can someone with the right permissions close this issue please?)

seldridge:
Closed via #1653. :tada:

PR #1653, opened 2020-06-02
Basic model checking API

tdb-alcorn:
This is an effort to address #567, which has been open for 3 years. The goal is to make it possible to add just enough model checking apparatus to FIRRTL to allow Chisel users to produce Verilog with `assume, assert, cover` statements such that the result can be checked by [Symbiyosys](https://symbiyosys.readthedocs.io/en/latest/). Symbiyosys is the explicit backend for testing this functionality because it's free and because it recognizes only a minimal kernel of System Verilog's model checking features, so anything that works in Symbiyosys should work with more advanced tools.

I propose adding the following three statements to FIRRTL:

- `assert(clk, cond, en, msg)`
- `assume(clk, cond, en, msg)`
- `cover(clk, cond, en, msg)`

In all three statements, `clk`, `cond`, and `en` must ultimately have concrete type `UInt<1>` (i.e. `Bool`). They are the clock signal, the condition being checked, and an enable signal, respectively. `msg` must be a string literal, and it provides the model checker with an error message to be produced if the check fails. 

To use the new statements, you need to use the System Verilog emitter. The Verilog emitter does not emit the new statements, and it depends on a transform that removes them and issues a warning. In the future, the Verilog emitter may be patched to produce simulation constructs for verification statements (e.g. `printf` for failing asserts).

#### TODOs:

- [x] implement assert
- [x] implement assume
- [x] implement cover
- [x] handle changing submodule assumes to asserts
- [x] fix expand whens to `and` block condition with formal statement enable
- [x] implement to/from proto (`convert`)
- [x] check types pass
- [x] dead code elimination (delete if not enabled)
- [x] legalize clocks

### Contributor Checklist

#### Type of Improvement

- new feature/API

#### API Impact

Three new statements, no changes to other API features.

New Verilog is emitted in a completely separate block within the file. This should have no effect on any existing Verilog emitted.

#### Desired Merge Strategy

- Squash

#### Release Notes

This release includes a new model checking API. It includes three new statements for model checking:

- `assert(clk, cond, en, msg)`
- `assume(clk, cond, en, msg)`
- `cover(clk, cond, en, msg)`

Each new statement maps directly to its System Verilog equivalent. In all three statements, `clk`, `cond`, and `en` must ultimately have concrete type `UInt<1>` (i.e. `Bool`). They are the clock signal, the condition being checked, and an enable signal, respectively. `msg` must be a string literal, and it provides the model checker with an error message to be produced if the check fails.

When emitting model checking statements, the Verilog emitter wraps them in an `\`ifdef FORMAL` block, excluding them from synthesis flows but making them available to model checkers, which define the `FORMAL` flag (such as Symbiyosys). This new API should have no effect on synthesis results.

- [ ] Did you review?


tdb-alcorn:
@azidar Would you or another FIRRTL dev be willing to help me understand how to approach this change?

edwardcwang:
1) What's the difference between assert and assume in this context?

2) Should new/experimental FIRRTL AST structures be marked in some way so that users are aware that it is experimental?

3) How much of this belongs in the AST vs as part of annotations or other metadata?

mwachs5:
should the `check`, `assume`, and `assert` include additional information other than the boolean? For example a string error message?

tdb-alcorn:
@edwardcwang to answer your questions:

>     1. What's the difference between assert and assume in this context?

assert tells the model checker to verify that the expression holds, whereas assume tells the model checker not to bother exploring any situations where the expression doesn't hold.

>     2. Should new/experimental FIRRTL AST structures be marked in some way so that users are aware that it is experimental?

I think the idea of marking new FIRRTL structures as experimental is interesting, but do you think it would actually help anyone? My understanding is that right now most people interact with FIRRTL through Chisel anyway, so perhaps just marking it as something different at the Chisel level would be enough.
 
>     3. How much of this belongs in the AST vs as part of annotations or other metadata?

I did consider using annotations to generate special `assert_foo` and `assume_foo` wires (etc.), but it seemed to me that depending on naming conventions instead of language constructs is fraught with danger. Furthermore, it was unclear to me how we could go from annotations/metadata to emitting the desired Verilog statements, except through bending the FIRRTL->Verilog emitter in awkward ways.

IMO, attacking the problem head on by creating a minimal, clean API will create fewer problems down the road. An added benefit of implementing this functionality as a first-class language feature is that downstream tools can standardize on it more reliably. For example, I hope to eventually create a bounded model checker that operates directly on FIRRTL to complement Treadle. To make it easier for other tools to consume, it's important that this API contain only the bare essentials, which (I hope) is what I'm proposing. Other fancy SV model checking constructs can be implemented at the Chisel level and eventually boil down to this small handful of basic FIRRTL statements.

tdb-alcorn:
@mwachs5:

> should the `check`, `assume`, and `assert` include additional information other than the boolean? For example a string error message?

Good idea! I think it makes sense for `check` to have a message string explaining what you're intending to check. The same goes for `cover`, so that you can explain why it is necessary to be able to reach the specified state (liveness). For `assume`, I'm not so sure: the point of assume is to speed up the model checker by giving it hints about which states it doesn't need to consider. The only time assume is violated is when the state you're assuming isn't reachable on all timesteps (unsatisfiable), which usually isn't a problem with the DUT per-se, but rather a problem with how you've constructed the model checking logic. In other words, `assume` violations happen while you're writing the tests, whereas `check` and `cover` violations happen while you're working on the DUT. But I guess it can't hurt to have a string message as an option.

ekiwi:
> should the `check`, `assume`, and `assert` include additional information other than the boolean? For example a string error message?

I would recommend putting any debug information like error messages into an annotation. This might require giving each assertion/assumption/cover a name so that you can actually target it through an annotation. It would help use keep the actual IR as minimal as possible and since you can always add more annotation, this could enable people to come up with more specific libraries that might add additional information for different kinds of assertions.

ekiwi:
> The new expression is `past(expr, n)` where expr can be of any type and n is a literal integer. The intended meaning of `past` is the value of `expr` `n` cycles ago. It maps to SV's `$past(expr, n)` expression.

In the interest of keeping things simple, I would hold off on adding a `past` construct for now. While it might result in a better formula for the model checker, you can always just add a chain of registers if you want to emulate it and this way we do not need to define what the value of `past` is in the first cycle of the model checking run.

tdb-alcorn:
> I would recommend putting any debug information like error messages into an annotation. 

Makes sense to me.

> ```
> assert a0 = eq(a, 0)
> ```

Do you think it would be sensible to make `check`, `cover`, `assume` expressions instead of statements? Then you could write `node a0 = assert(eq(a,0))` and

seldridge:
One comment: the `VerilogEmitter` is technically Verilog 2001 which doesn't have `assert` in my understanding. This is a great candidate for going in the `SystemVerilogEmitter`, however. 

I think the ideal approach will be to get chisel3 to emit these new IR nodes to replace the current implementation of `chisel3.assert`. The mechanism for how these nodes are lowered/emitted would then depend on the backend. The `VerilogEmitter` should emit a mux and a `$fatal`. The `SystemVerilogEmitter` should emit `assert`.

@tdb-alcorn: The `VerilogEmitter` isn't terrifically extensible and there's been talk of building a new `SystemVerilogEmitter` that is more extensible. One idea here is to support custom IR nodes that define how they should be emitted. I think that `assert` may be fundamental enough to warrant inclusion in the IR. However, if there's push back here then a custom node may be the way to go.

ekiwi:
> Do you think it would be sensible to make `check`, `cover`, `assume` expressions instead of statements? Then you could write `node a0 = assert(eq(a,0))`.

In my understanding, `assert` has side effect in that it can fail the simulation. Thus it should be a statement.
The problem with making it an expression is that it could get constant-prop'd e.g. if used like this `or(UInt<1>(1), assert(...))`

Do you have a use-case for using the result of an assert?

edwardcwang:
Hey Tom @tdb-alcorn ,

> I think the idea of marking new FIRRTL structures as experimental is interesting, but do you think it would actually help anyone? My understanding is that right now most people interact with FIRRTL through Chisel anyway, so perhaps just marking it as something different at the Chisel level would be enough.

One line of thought is inspired from how the RISC-V ISA is 'modularized'. For example, if one is working on a library or tool that uses FIRRTL, it would be great to be able to, say, support the main FIRRTL AST independently from extensions - e.g. we could imagine that there is an (experimental) model-checking extension. We could then also say that a circuit contains just the "bare minimum" needed for synthesis if it only uses the core AST, while there could be circuits that contain additional information/metadata such as model checking info. Of course it would also be interesting to consider how annotations fit into this schema.

albert-magyar:
> * `check(bool_expr)` which will map to SV's `assert`
> * `assume(bool_expr)` which will map to SV's `assume`
> * `cover(bool_expr)` which will map to SV's `cover`

It may be desirable to restrict the emission of these constructs to clocked Verilog procedures. Since the general idea with FIRRTL is to attach clocks to statements rather than to the contexts in which they appear, it might make sense to add a clock argument to some or all of these. This would help document the semantic intent of the assertion, and would avoid the possibility of race-condition false positives that may appear with unclocked immediate assertions (though there may be properties of emitted FIRRTL-generated Verilog that already mitigate this issue).

To this end, the output of `CheckSpec` does not appear to be legal Verilog, since the immediate assertion `assert(areEqual)` isn't part of a procedure.

As others mentioned, there is the question of deciding on the final set of supported constructs, and one alternative is to use annotations to encode assertions / properties / etc, which allows arbitrarily rich sets of target language constructs to be added without modifying the FIRRTL compiler. However, I am **very much** in favor of including basic immediate checks (cover / assert / assume) into the IR, since I think it's better to put such universal constructs into the IR without relying on annotations.

As it turns out, the issue with your assert statements not being emitted seems to be related to the `Emitter` code relying on mutation. I put some inline comments in the code; unfortunately, using the `simulate` side-effect-having function would solve your emission problems, but it requires a clock, since it appears to have been written to support the intrinsically clocked `print` and `stop` statements.

src/main/scala/firrtl/Emitter.scala
```diff
@@ -879,6 +885,10 @@ class VerilogEmitter extends SeqTransform with Emitter {
           simulate(sx.clk, sx.en, stop(sx.ret), Some("STOP_COND"), sx.info)
         case sx: Print =>
           simulate(sx.clk, sx.en, printf(sx.string, sx.args), Some("PRINTF_COND"), sx.info)
+        // formal
+        case sx: Check =>
+          check(sx.arg)
```
albert-magyar, line 890: The enclosing match is used as a way to call side-effect-having functions -- the evaluated value of the match is not used.

The `Print` and `Stop` cases call the impure `simulate` to cause procedures containing the appropriate Verilog displays or finishes to be emitted. If `Check` had a clock, you could make some call resembling:
`simulate(sx.clk, Utils.one, check(sx.arg), ...`

Unfortunately, there doesn't seem to be a utility like `simulate` for emitting wildcard-sensitivity procedures, so you'd have to extend it to make the clock argument optional.

Your intuition to just return a value is certainly understandable given that there seems to be a lot of action-at-a-distance going on in this function :)

albert-magyar:
As an aside, I would vote for calling the thing mapping to a SystemVerilog immediate assertion `assert`, rather than `check`. We already have to deal with a lot of constructs that have names that are meaningful in the implementation language (print, add, head, tail, type, etc.), so I think that avoiding terms that are overloaded in that way is not worth using less idiomatic names.

ekiwi-sifive:
@albert-magyar I agree with your point about treating an `assert` as a clocked (and I guess also gated) statement similar to a `printf`. However, do you have a good idea on how to name the statement? We need a way to annotate an `assert` in order to attach debug information that will be used in case the assertion fails.

src/main/scala/firrtl/Emitter.scala
tdb-alcorn: Thanks Albert! I'm adding a `formals` array buffer to hold the formal checks, to be emitted at the end of the file within an `\`ifdef FORMAL` block.

tdb-alcorn:
Thanks for your feedback so far everyone! Summarizing the discussion a little here for anyone catching up:

- `check` should just be `assert`
- `assume`, `assert`, `cover` should all be clocked
- model checking statements need a way to be identified so that annotations can provide metadata like error messages. One idea is to do something like `assert a0 = exp`, similar to the `node` statement
- `past` is unnecessary because it can be implemented with shift registers. It could be a high-level convenience in Chisel but need not be part of FIRRTL
- none of these affordances make sense in Verilog 2001. The SystemVerilog emitter needs to behave differently from the Verilog emitter when dealing with model checking statements
- question raised by @edwardcwang: is it possible and greatly desirable to make the model checking affordances into a separate FIRRTL extension to keep synthesis and verification stuff separate? My follow-up question: what changes to FIRRTL would this require?

I've been thinking some more about how we'd actually get this done and there are a few issues that stick out:

1. Handling model checking statements that occur within `when` blocks. Since the conditional statement is disallowed in LoFIRRTL, we need a way to move such statements to the top level of the module. One idea I've had is to prepend the condition to the expression checked by the statement. E.g. in Chisel
```
when (cond) {
  assert(foo)
}
```
would become LoFIRRTL
```
assert(or(not(cond), foo))
```
Are there any drawbacks to this approach?
2. Handling submodules correctly. When instantiating a submodule, `assume`s should become `assert`s. If we don't do this then there are bugs that model checking will be unable to detect. Take this example (in Chisel for clarity):
```
class Inner extends Module {
  val io = IO(new Bundle {
    val in = Input(Bool())
    val out = Output(Bool())
  })
  io.out := false.B  // bug, should equal io.in

  // inner module will pass model checking (incorrectly, because of this buggy assumption)
  assume(io.in === false.B)
  assert(io.out === io.in)
}

class Outer extends Module {
  val io = IO(new Bundle {
    val in = Input(Bool())
    val out = Output(Bool())
  })

  val inner = Module(new Inner)  // on instantiating a submodule, internal assumes are converted to asserts
  inner.io.in := io.in
  io.out := inner.io.out

  // outer module will fail model checking (correctly) because the buggy assumption fails as an assertion
  assert(io.out === io.in)
}
```
should generate Verilog
```

module Inner(
  input clock,
  input reset,
  input in,
  output out
);
  assign io_out = 1'b0;

`ifdef FORMAL
  always @(posedge clock) begin
    assert(io_in == 1'b0);  // converted from assume to assert because Inner is a submodule
    assert(io_out == 1'b0);
  end
`endif
endmodule

module Outer(
  input clock,
  input reset,
  input in,
  output out
);
  wire  inner_io_in;
  wire  inner_io_out;
  Inner inner (
    .io_in(inner_io_in),
    .io_out(inner_io_out)
  );
  assign io_out = inner_io_out;
  assign inner_io_in = io_in;

`ifdef FORMAL
  always @(posedge clock) begin
    assert(io_out == io_in);
  end
`endif
endmodule
```
I'm not at all sure of how to handle this type of thing from within Firrtl. It's probably going to require a bit of extra machinery to pass some context around about which module is the top.
3. Enabling liveness checks through an `eventually` statement. Symbiyosys supports the `s_eventually` keyword which helps you verify that a design will reach a desired state at some point. You use it like this in System Verilog: `assert(s_eventually io_valid)`. It is very useful (see footnote). IMO a liveness affordance merits inclusion in this basic FIRRTL model checking toolkit because liveness properties are very easy to understand and natural to reach for when first starting to add verification. Is there a way to implement this using existing FIRRTL constructs? If not, is adding an `eventually` statement a good idea?

(footnote): While coming up with examples for this PR I tried my hand at verifying the GCD example from the the [Chisel3 template project](https://github.com/freechipsproject/chisel-template/blob/release/src/main/scala/gcd/GCD.scala) and actually found a bug using the liveness check `assert(s_eventually y == 0)`. If you emit Verilog for this module, add this liveness check and check it in Symbiyosys' `live` mode you'll find that it fails. The bug is that if the `io_value1 = 0` is initially loaded then the GCD module never finishes because it loops indefinitely subtracting 0 from `y`. I was unable to think of a reasonable way to find this bug without the liveness checker, and therefore the `s_eventually` affordance.

ekiwi-sifive:
Thanks for the great summary @tdb-alcorn !


>     * model checking statements need a way to be identified so that annotations can provide metadata like error messages. One idea is to do something like `assert a0 = exp`, similar to the `node` statement

While I initially suggested that, I forgot about the need for the statement to be clocked. It might make more sense to model it after a `printf` statement. To solve the annotation issue it might make sense to allow a label with any kind of `printf`, `stop`, `assume`, etc. like:

```
printf(clk, cond, "bla bla") : my_print
```



>     1. Handling model checking statements that occur within `when` blocks.

Your solution should work! Alternatively I wonder if it would be possible to reuse the lowering passes that operate on `printf` and `stop` by adding an enabled `condition` to the formal statements similar to how the current statements work.



> I'm not at all sure of how to handle this type of thing from within Firrtl. It's probably going to require a bit of extra machinery to pass some context around about which module is the top.

The top module should be known to any compiler pass. The problem might be more of a UX thing since it might seem a bit magical that changing the toplevel module changes assumes to asserts (it makes sense to me, however it might not be immediately clear to users).



> 3. Enabling liveness checks through an `eventually` statement.

This would be cool. We should maybe try and find out how these map to the `justice` properties in the `btor2` format: http://fmv.jku.at/papers/NiemetzPreinerWolfBiere-CAV18.pdf

seldridge:
The existing lowering for `stop` is mostly doing what you describe. Your approach looks fine. Just anding assert condition (or your DeMorgan'd version) should be fine. The following Scastie snippet shows the lowering of `stop`: https://scastie.scala-lang.org/seldridge/844HVJucRIGlebtDBN2kww/271. Note that this includes a check that you're out of reset.

> It's probably going to require a bit of extra machinery to pass some context around about which module is the top.

Context is available since the `Circuit` IR node includes a `main: String` parameter.

tdb-alcorn:
Great, the only remaining question for now is how to handle an `eventually` affordance. I think it warrants inclusion for the reasons outlined above (it's really useful), and because it's difficult to see how you would implement `eventually` semantics using only existing FIRRTL constructs.

I propose adding a 5th argument to `assert` and `assume` to be an eventually flag. The syntax is `assert(clk, cond, en, eventually, msg)` where `clk` is a clock, `cond` is the expression you’re asserting, `en` is an enable signal (nothing will be asserted if `en` is false), and `msg` is some kind of error message that will be displayed if it fails. `eventually` is an integer literal, either 0 or 1 by convention, 1 means treat the assert as a liveness check (`assert(s_eventually ...)` in SV), and 0 or any other value means treat the assert normally. Same syntax for `assume`.

One alternative would be to make an entirely separate statement `assertLive(...)` and `assumeLive(...)`, but I'm inclined to introduce fewer keywords and just use arguments to differentiate the two related affordances. Any other ideas or comments?

ekiwi-sifive:
> Any other ideas or comments?

I would hold off and focus on safety properties for now. Once that is merge we can think about how to deal with liveness properties. This would make the PR easier to review.

tdb-alcorn:
FYI I've updated the top comment in the conversation to reflect the current state of the PR. After I've finished the TODOs and other tasks I'll change this PR from draft to "ready for review".

albert-magyar:
@tdb-alcorn, we were talking about a few things related to this PR, and an idea came up about separating the atomic proposition and clocking from the assert/assume/cover directive.

The idea here is that we would introduce a `Property` node that takes a boolean value and a clock. This property is a declaration, so it can have a name.

Downstream of this, `assert`/`cover`/`assume`/`print`/`stop` could be handled in a more consistent way, as each would just be a statement that references a property and potentially contains a string message.

Again, this is just an idea. I think the main difference is in that it tries to avoid code duplication in a slightly different way from the "Formal" grouping. Regardless, it would be awesome if you would be able to have a Zoom meeting with some of the other FIRRTL developers to talk about more ideas; I think @azidar was hoping to schedule a meeting about this some time. One other thing to discuss is the degree to which it makes sense to tie the `assert` etc. nodes to a specific use case of formal verification, since the ability to use them in dynamic verification could also be a feature.

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -501,6 +501,71 @@ case class Print(
   def foreachString(f: String => Unit): Unit = Unit
   def foreachInfo(f: Info => Unit): Unit = f(info)
 }
+
+// formal
+
+abstract class Formal(
```
tdb-alcorn, line 507: Make enum, pattern match enumeration value

tdb-alcorn: Name bikeshedding:
 - Property
 - VerificationStatement
 - Model
 - Spec

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -501,6 +501,71 @@ case class Print(
...
+  val name: String,
```
tdb-alcorn, line 508: op?

src/main/scala/firrtl/passes/CheckTypes.scala
```diff
@@ -308,6 +310,11 @@ object CheckTypes extends Pass with PreservesAll[Transform] {
             errors.append(new PrintfArgNotGround(info, mname))
           if (wt(sx.clk.tpe) != wt(ClockType)) errors.append(new ReqClk(info, mname))
           if (wt(sx.en.tpe) != wt(ut)) errors.append(new EnNotUInt(info, mname))
+        case sx: Formal =>
+          if (wt(sx.clk.tpe) != wt(ClockType)) errors.append(new ReqClk(info, mname))
+          if (wt(sx.cond.tpe) != wt(ut) && wt(sx.cond.tpe) != wt(st))
```
tdb-alcorn, line 315: should only be uint

src/main/scala/firrtl/passes/formal/AssertSubmoduleAssumptions.scala
```diff
@@ -0,0 +1,33 @@
...
+
+
+object AssertSubmoduleAssumptions extends Pass {
+  override def prerequisites: Seq[TransformDependency] =
+    firrtl.stage.Forms.Deduped
+
+  override def invalidates(a: Transform): Boolean = a match {
+    case _ => false
+  }
+
```
seldridge, line 18: This should probably include the following to make sure it runs before the middle emitter. I also tend to be explicit with all dependencies as opposed to letting them inherit from `Pass` or `Transform`.

```scala
def optionalPrerequisites = Seq.empty
def optionalPrerequisiteOf = firrtl.stage.Forms.MidEmitters
```

src/main/scala/firrtl/passes/formal/AssertSubmoduleAssumptions.scala
```diff
@@ -0,0 +1,33 @@
...
+package firrtl.passes.formal
+
+import firrtl.passes.{Pass, PullMuxes, ReplaceAccesses, ResolveFlows}
+import firrtl.ir.{Assert, Assume, Circuit, Statement}
+import firrtl.options.Dependency
+import firrtl.stage.TransformManager.TransformDependency
+import firrtl.{CircuitState, FirrtlUserException, Transform}
+
+
+object AssertSubmoduleAssumptions extends Pass {
```
tdb-alcorn, line 11: Annotation to enable/disable this pass
See NoDCE annotation

What should the default be? Solicit feedback from formal people

Non-blocking

src/main/scala/firrtl/Emitter.scala
```diff
@@ -1078,6 +1110,16 @@ class VerilogEmitter extends SeqTransform with Emitter {
         emit(Seq("`endif // SYNTHESIS"))
       }
 
+      if (formals.keys.nonEmpty) {
+        emit(Seq("`ifdef FORMAL"))
```
tdb-alcorn, line 1114: do we need this block? do all tools that accept SV accept assert, assume, cover

tdb-alcorn: if we keep this block, what should the name be?

tdb-alcorn: FIR_VERIFICATION
FIR_FORMAL

ekiwi-sifive:
👍 `Verification`

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -501,6 +501,71 @@ case class Print(
...
+  val info: Info,
+  val clk: Expression,
+  val cond: Expression,
```
tdb-alcorn, line 511: pred

src/main/scala/firrtl/passes/CheckTypes.scala
ekiwi-sifive: `UInt<1>` to be exact

src/main/scala/firrtl/Emitter.scala
jchang0: Agree with @tbd-alcorn about not enclosing it in ifdef FORMAL.   assert, assume, and cover are not restricted to formal, and all simulation tools would benefit from it.

src/main/scala/firrtl/passes/formal/AssertSubmoduleAssumptions.scala
jchang0: Simulation tools automatically treat assume as asserts.  So this pass may not be necessary.

jchang0:
> I propose adding a 5th argument to `assert` and `assume` to be an eventually flag. The syntax is `assert(clk, cond, en, eventually, msg)` where `clk` is a clock, `cond` is the expression you’re asserting, `en` is an enable signal (nothing will be asserted if `en` is false), and `msg` is some kind of error message that will be displayed if it fails. `eventually` is an integer literal, either 0 or 1 by convention, 1 means treat the assert as a liveness check (`assert(s_eventually ...)` in SV), and 0 or any other value means treat the assert normally. Same syntax for `assume`.


wouldn't clk just be inherited from the default clock of the module, just like everything else in Chisel?

I rather not call out eventually on this level.  There are a whole class of temporal operators that we would want to throw if we allow things such as eventually.

> 
> One alternative would be to make an entirely separate statement `assertLive(...)` and `assumeLive(...)`, but I'm inclined to introduce fewer keywords and just use arguments to differentiate the two related affordances. Any other ideas or comments?

Probably better alternative.

jchang0:
Would also like to add the argument "label" to these properties.

e.g. assume(cond, en, label, msg)

  It will generate SVA labels for the properties so that it'll be easier index when using tools that reads in verilog.

jackkoenig:
> wouldn't clk just be inherited from the default clock of the module, just like everything else in Chisel?

> I rather not call out eventually on this level. 

Note that this is the IR API, not the Chisel API. In Chisel you would just use the implicit clock.

tdb-alcorn:
@azidar I think this is in a good state for review now. I've addressed the main critiques & ideas from yesterday. Apologies if you're the wrong person to at-mention, I wasn't sure who this will be assigned to.

src/main/scala/firrtl/stage/Forms.scala
```diff
@@ -57,7 +57,8 @@ object Forms {
          Dependency[passes.ExpandWhensAndCheck],
          Dependency[passes.RemoveIntervals],
          Dependency(passes.ConvertFixedToSInt),
-         Dependency(passes.ZeroWidth) )
+         Dependency(passes.ZeroWidth),
+         Dependency(passes.formal.AssertSubmoduleAssumptions))
```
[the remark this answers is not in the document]
tdb-alcorn: This transform (like DCE) is optional and can be disabled, so I don't think it makes sense to make it part of the spec for mid form. I couldn't find a better way to inject it into the default compiler flow other than to just add it to the dependencies for mid (I tried to find a way to add it as a "default" custom transform but couldn't make it work). Is there a better way?

src/main/scala/firrtl/Emitter.scala
```diff
@@ -1078,6 +1092,16 @@ class VerilogEmitter extends SeqTransform with Emitter {
...
```
albert-magyar, line 1096: We should not ifdef these unless there is a specific reason we must.

src/main/scala/firrtl/Emitter.scala
```diff
@@ -1078,6 +1092,16 @@ class VerilogEmitter extends SeqTransform with Emitter {
...
+        for ((clk, content) <- formals if content.nonEmpty) {
+          emit(Seq(tab, "always @(posedge ", clk, ") begin"))
+          for (line <- content) emit(Seq(tab, tab, line))
```
albert-magyar, line 1099: This is SystemVerilog, so there should be some way to prevent emitting this for the "Verilog" flow.

ekiwi-sifive: Or just to print a warning? Which tool would actually reject these constructs because of the `.v` file ending?

albert-magyar: It's not about the file extension -- if we have a "Verilog" backend, it should produce Verilog (95, 2001, or 2005).

ekiwi-sifive: OK. So the solution would be to move the code that Tom added to the `VerilogEmitter` class at the end of the `Emitter.scala` file and remove the `StageUtils.dramaticWarning`?
What do you suggest?

tdb-alcorn: I've fixed this in the latest changes, please take a look!

tdb-alcorn:
It looks like the CI tests that are failing are the LoweringCompilersSpec ones where the set of transforms to be run is checked against some older version. When I add a new transform (e.g. `RemoveVerificationStatements`) and add it to the Verilog emitter this shows up in that set and causes the test to fail. What's the best way to address this? Currently I just manually find out the index of transform I added and slice it out of the list of expected transforms in each test, but this feels very hacky and a bit dangerous.

tdb-alcorn:
> What are the "X", "Y", and "Z" in the example message?

In "X equals Y when Z is valid", I intended for X and Y to be two signals in the module, `X==Y` to be the predicate being tested and Z to be the enable for the statement.

seldridge:
`LoweringCompilersSpec` was a crutch that I added to check the old hard-coded transform order against the new Dependency API-derived transform order. This was insanely useful to get the Dependency API to be load bearing. However, as we move more and more away this point, the test isn't testing anything and is just a PR nuissance.

For 1.4 (or in 1.4.1) we should rip this out.

To your point: if the PR adds a transform, then this should use an `Add` patch and if it removes a transform it should use a `Del` patch. It's janky, I realize... I cooked this up after some iterations on how to lock the old transform order and then keep a record of exactly how we differ from the that.

ekiwi-sifive:
@edwardcwang I read a little more about the cover statement. For formal verification it seems to be treated almost like an assert: The formal tool tries to find a trace that makes the predicate in the cover statement true or tries to prove that it can never be covered. I think the main difference is that the result's interpretation is flipped: while an assertion failure is bad and proving that it can never fail is good, hitting a cover point is good and proving that it can never be covered is bad.

@tdb-alcorn : is this how you have been using `cover` with `yosys`?

albert-magyar:
> I think the main difference is that the result's interpretation is flipped: while an assertion failure is bad and proving that it can never fail is good, hitting a cover point is good and proving that it can never be covered is bad.

The nice thing is that the existential qualifier makes it easier to map to checking whether a TS satisfies a safety property: it's possible to check if the negation of the covered property is ever "violated."

src/main/scala/firrtl/Emitter.scala
albert-magyar: Thanks, will do!

ekiwi-sifive:
I think the user should be alerted if their verification statements are removed.
It could be a single warning (not one per statement) that suggest that the use turn on the SystemVerilog emission if they want their statement to be preserved.
Silently removing them in the default setting seems like a user-hostile approach.

ekiwi-sifive:
We also need a transform that goes from `assert(...)` to the `when ...: printf(...) ; stop(...)` construct that we are currently emitting for the Chisel `assert` construct.
However, I think this does not need to be in this PR, but can be added in a follow up PR before the `1.4` release.

tdb-alcorn:
> is this how you have been using `cover` with `yosys`?

Yes, and also in the way that @albert-magyar suggested. The two core ways that I use cover is:
1. check that ready/valid can be both high and low on any decoupled ports (e.g. `cover(clk, io_in_ready == 1, ...)` and `cover(clk, io_in_ready == 0, ...)`.
2. check that the enable is reachable for any assert statements (e.g. if I `assert(clk, foo, en, ...)` then I should also `cover(clk, en, 1, ...)` to avoid a silent bug in which the enable is never reachable so the assert never fails, regardless of the behaviour of `foo`.

spec/spec.tex
```diff
@@ -1024,6 +1024,45 @@ \subsubsection{Format Strings}
 \item \verb|\'| : Single quote
 \end{itemize}
 
+\subsection{Verification}
+
+To facilitate simulation, model checking and formal methods, there are three non-synthesizable verification statements available: assert, assume and cover. Backends are free to generate the corresponding model checking constructs in the target language, but this is not required by the FIRRTL specification. For example, the System Verilog emitter produces System Verilog assert, assume and cover statements. However, backends should not generate any new synthesizable hardware when handling to verification statements. Backends that do not obey this rule should issue a warning.
```
ekiwi-sifive, line 1029: > However, backends should not generate any new synthesizable hardware when handling to verification statements. Backends that do not obey this rule should issue a warning.

I am not sure I understand this line. Do we need this?
Something like @albert-magyar 's GoldenGate compiler might actually generate HW in order to check assertions on the FPGA. I don't think that he needs to emit a warning for that.

On the other hand, we might want to require that backends that do not support these statements should issue a warning and then ignore them.

spec/spec.tex
```diff
@@ -1024,6 +1024,45 @@ \subsubsection{Format Strings}
...
+
+Each type of verification statement requires a clock signal, a predicate signal, an enable signal and an explanatory message string literal. The predicate and enable signals must have single bit unsigned integer type. When an assert is violated the explanatory message may be issued as guidance. The explanatory message may be phrased as if prefixed by the words ``Verifies that...''.
+
+\subsubsection{Assert}
+
+The assert statement verifies that the predicate is true on the rising edge of any clock cycle when the enable is true. In other words, it verifies that enable implies predicate.
+
+\begin{lstlisting}
+wire clk:Clock
+wire pred:UInt<1>
```
ekiwi-sifive, line 1039: how about we change that to `node pred = eq(X, Y)`?

spec/spec.tex
```diff
@@ -1024,6 +1024,45 @@ \subsubsection{Format Strings}
...
+wire en:UInt<1>
```
ekiwi-sifive, line 1040: maybe `node en = z_valid`

ekiwi-sifive: this might address @edwardcwang question as to the meaning of the message

src/main/scala/firrtl/Emitter.scala
```diff
@@ -15,8 +14,9 @@ import firrtl.PrimOps._
 import firrtl.WrappedExpression._
 import Utils._
 import MemPortUtils.{memPortField, memType}
-import firrtl.options.{Dependency, HasShellOptions, ShellOption, StageUtils, PhaseException, Unserializable}
+import firrtl.options.{Dependency, HasShellOptions, PhaseException, ShellOption, StageUtils, Unserializable}
```
ekiwi-sifive, line 17: this is annoying thing that IntelliJ does. However, while unrelated to this PR, I think we should just keep it this way so that future edits to the Emitter don't have to manually undo IntelliJ's import sorting.

src/main/scala/firrtl/Emitter.scala
```diff
@@ -781,6 +790,14 @@ class VerilogEmitter extends SeqTransform with Emitter {
       lines += Seq("`endif // SYNTHESIS")
     }
 
+    def addFormal(clk: Expression, en: Expression, stmt: Seq[Any], info: Info) = {
```
ekiwi-sifive, line 793: can we move this an `formalStatement` to the SystemVerilog emitter?

src/main/scala/firrtl/Visitor.scala
```diff
@@ -328,6 +328,17 @@ class Visitor(infoMode: InfoMode) extends AbstractParseTreeVisitor[FirrtlNode] w
         case "attach" => Attach(info, ctx_exp map visitExp)
         case "printf(" => Print(info, visitStringLit(ctx.StringLit), ctx_exp.drop(2).map(visitExp),
           visitExp(ctx_exp(0)), visitExp(ctx_exp(1)))
+        // formal
+        case "assert" => Verification(Formal.Assert, info, visitExp(ctx_exp(0)),
+          visitExp(ctx_exp(1)), visitExp(ctx_exp(2)),
+          visitStringLit(ctx.StringLit))
+        case "assume" => Verification(Formal.Assume, info, visitExp(ctx_exp(0)),
+      visitExp(ctx_exp(1)), visitExp(ctx_exp(2)),
+      visitStringLit(ctx.StringLit))
```
ekiwi-sifive, line 337: what's up with the indentation here?

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -501,6 +501,38 @@ case class Print(
...
+  pred: Expression,
+  en: Expression,
+  msg: StringLit
+) extends Statement with HasInfo {
+  def serialize: String = op + "(" + Seq(clk, pred, en).map(_.serialize)
+    .mkString(", ") + ", \"" + msg.serialize + "\")" + info.serialize
+  def mapStmt(f: Statement => Statement): Statement = this
+  def mapExpr(f: Expression => Expression): Statement =
+    copy(clk = f(clk), pred = f(pred), en = f(en))
+  def mapType(f: Type => Type): Statement = this
```
ekiwi-sifive, line 525: @azidar should this map over the types of the `clk`, `pred` and `en` Expressions?
I personally have never used `mapType`.

spec/spec.tex
tdb-alcorn: Do you think it's important to explicitly call out that the type of pred is `UInt<1>`?

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -501,6 +501,38 @@ case class Print(
...
+  def mapString(f: String => String): Statement = this
```
ekiwi-sifive, line 526: @azidar should this map over the `msg` String?

src/main/scala/firrtl/proto/FromProto.scala
```diff
@@ -8,7 +8,7 @@ import java.io.{File, FileInputStream, InputStream}
 import collection.JavaConverters._
 import FirrtlProtos._
 import com.google.protobuf.CodedInputStream
-import Firrtl.Statement.ReadUnderWrite
+import Firrtl.Statement.{ReadUnderWrite, Formal}
```
ekiwi-sifive, line 11: `Formal` should be `Verification`

src/main/scala/firrtl/transforms/formal/RemoveVerificationStatements.scala
```diff
@@ -0,0 +1,41 @@
...
+  with PreservesAll[Transform] {
+
+  override def prerequisites: Seq[TransformDependency] = Seq.empty
+  override def optionalPrerequisites: Seq[TransformDependency] = Seq.empty
+  override def optionalPrerequisiteOf: Seq[TransformDependency] =
+    Seq( Dependency[VerilogEmitter],
+      Dependency[MinimumVerilogEmitter])
+
+  def removeVerification(s: Statement): Statement = s match {
+    case _: Verification => EmptyStmt
```
ekiwi-sifive, line 28: This should emit a warning (does `logger.warn(....)` work?) the first time a statement is replaced.
The warning should point the user to the SystemVerilog emitter and explain how to enable it.

src/main/scala/firrtl/Emitter.scala
tdb-alcorn: Not unless we make a separate SystemVerilogRenderer too, which seems like over-kill. Plus, by leaving it in, if/when we want to have the Verilog emitter emit muxed `printf`s for asserts it will be easy to implement by simply updating the addFormalStatement method.

src/main/scala/firrtl/proto/FromProto.scala
tdb-alcorn: ? There is already a `Firrtl.Statement.Verification`. Verification is used to denote the statement and formal is used to denote the "operator" (assert, assume, cover)

spec/spec.tex
ekiwi-sifive: That is the thing that gets lost in my suggestion. Yeah, maybe we should have a `wire` and a `connect` which would combine both points.

```
wire pred: UInt<1>
pred <= eq(X, Y)
```

spec/spec.tex
ekiwi-sifive: Same as above: adding a `en <= Z_valid` is better than converting to a `node`

src/main/scala/firrtl/Emitter.scala
ekiwi-sifive: OK. Then this is good enough for now.

src/main/scala/firrtl/proto/FromProto.scala
ekiwi-sifive: I see. The name `Formal` is a little confusing, but I remember us talking about this on the Zoom call and we only came up with thing we did not want to call it like `Kind` and `Type`.

albert-magyar: The only directly analogous situation is DoPrim, where the field for the “variety” of operator is just called `op`

ekiwi-sifive: So `VerificationOp` ?

tdb-alcorn:
@seldridge I'm seeing some indeterminacy in the LoweringCompilersSpec. When I run it anywhere from 0-5 tests might fail. It seems that the ordering produced by the TransformManager may not be fully deterministic: have you seen this before?

seldridge:
I think this was fixed in https://github.com/freechipsproject/firrtl/pull/1686. Can you try a rebase/merge?

src/main/scala/firrtl/proto/FromProto.scala
tdb-alcorn: > So `VerificationOp` ?

Eh, it's not evocative but I guess I could live with it. But think of how this Formal enum is actually used in the code, like:
```
case Verification(Formal.Assert, ...) =>
```
IMO it reads fine and it's obvious what is meant. Writing that as `case Verification(VerificationOp.Assert, ...)` feels pretty bloodless.

spec/spec.tex
tdb-alcorn: I think that we should establish a contract that formal statements should have no effect on synthesis, in order to avoid [Heisenbugs](https://en.wikipedia.org/wiki/Heisenbug). Any backend that needs/wants to violate that contract should be loud and explicit about it to prevent surprises.

The idea to have backends that don't support them issue a warning is good: I've now added that for the Verilog emitter, I'll add it to the spec too.

ekiwi-sifive: When would a system ever add undesired synthesizable hardware?
How would that possibly cause a heisenbug assuming that the tool is correct?
What about dead code elimination that might be inhibited by adding an `assert` statement? Does that count as "generating new synthesizable hardware" ?

src/main/scala/firrtl/proto/FromProto.scala
ekiwi-sifive: Not a blocker. I don't care that much.

src/main/scala/firrtl/ir/IR.scala
ekiwi-sifive: Let's leave this in its current form as it is consistent with `Print`.

src/main/scala/firrtl/ir/IR.scala
ekiwi-sifive: Let's leave this in its current form as it is consistent with `Print`.

src/main/scala/firrtl/transforms/formal/RemoveVerificationStatements.scala
```diff
@@ -0,0 +1,53 @@
...
+    })
+  }
+
+  def execute(state: CircuitState): CircuitState = {
+    val newState = state.copy(circuit = run(state.circuit))
+    if (removedCounter > 0) {
+      StageUtils.dramaticWarning(s"$removedCounter verification statements " +
+        "were removed when compiling to Verilog because the basic Verilog " +
+        "standard does not support them. If this was not intended, compile " +
+        "to System Verilog instead using the `-X sverilog` compiler flag.")
```
ekiwi-sifive, line 49: Awesome! 🎉

ekiwi-sifive:
Once the tests pass we should have someone (I volunteer) try the formal statements out with yosys and SymbiYosys on their local computer and then I would think we are ready to merge.

spec/spec.tex
```diff
@@ -1024,6 +1024,51 @@ \subsubsection{Format Strings}
 \item \verb|\'| : Single quote
 \end{itemize}
 
+\subsection{Verification}
+
+To facilitate simulation, model checking and formal methods, there are three non-synthesizable verification statements available: assert, assume and cover. Backends are free to generate the corresponding model checking constructs in the target language, but this is not required by the FIRRTL specification. For example, the System Verilog emitter produces System Verilog assert, assume and cover statements. However, backends should not generate any new synthesizable hardware when handling to verification statements. Backends that do not obey this rule should issue a warning.
```
albert-magyar, line 1029:
```suggestion
To facilitate simulation, model checking and formal methods, there are three non-synthesizable verification statements available: assert, assume and cover. Backends are free to generate the corresponding model checking constructs in the target language, but this is not required by the FIRRTL specification. For example, the SystemVerilog emitter produces SystemVerilog assert, assume and cover statements. However, backends should not generate any new synthesizable hardware when handling to verification statements. Backends that do not obey this rule should issue a warning.
```

src/test/scala/firrtlTests/formal/VerificationSpec.scala
```diff
@@ -0,0 +1,55 @@
+package firrtlTests.formal
+
+import firrtl.{SystemVerilogCompiler}
+import firrtl.testutils.FirrtlFlatSpec
+import logger.{LogLevel, Logger}
+
+class VerificationSpec extends FirrtlFlatSpec {
+  behavior of "Formal"
+
+  it should "generate Verilog statements within FORMAL block" in {
```
albert-magyar, line 10: This text is outdated now. Thanks for making the changes to the emission :)

spec/spec.tex
ekiwi-sifive: This changes `System Verilog` to `SystemVerilog` (👍)  in case anyone else is wondering...

albert-magyar: Hah, yes, the diff is not that clear.

src/test/scala/firrtlTests/formal/VerificationSpec.scala
albert-magyar, line 10:
```suggestion
  it should "generate immediate SVA statements within a procedure" in {
```

tdb-alcorn:
> So the only blocker left for me is the language around "not generating new synthesizable hardware" which I personally find confusing and unnecessary.

Fair. I'm happy to leave it out of the spec for now, but I think we should talk about later with a broader group (next dev meeting perhaps?) because I think it will help new backend implementers in the future. By giving them a safe convention to stick to, it sends the message that deviating from that convention should be weighed carefully because there are potential pitfalls.

tdb-alcorn:
@ekiwi-sifive Tests are passing, feel free to try it out with Symbiyosys! It's working for me.

ekiwi-sifive:
So I tried out a simple example with the firrtl version from Tom's branch:

**Safety.fir**
```.fir
; this example follows the one from the SymbiYosys introduction
; https://symbiyosys.readthedocs.io/en/latest/quickstart.html
; but uses a synchronous reset instead of a init preset in order
; to try out assumptions
circuit Safety:
  module Safety:
    input clk: Clock
    input preset: AsyncReset
    input reset: UInt<1>
    output counter: UInt<5>
    reg counter_r: UInt<5>, clk with: (reset => (reset, UInt<5>(0)))
    when eq(counter_r, UInt<5>(15)):
      counter_r <= UInt<5>(0)
    else:
      counter_r <= add(counter_r, UInt<5>(1))
    counter <= counter_r

    ; logic to track whether this is the first cycle
    reg _first: UInt<1>, clk with: (reset => (preset, UInt<1>(1)))
    _first <= UInt<1>(0)
    ; assume design is reset in the first cycle
    assume(clk, eq(reset, UInt<1>(1)), _first, "Reset should be asserted in the first cycle")

    ; make sure that when we are not in reset the invairance holds
    assert(clk, lt(counter, UInt<5>(31)), not(reset), "Counter should be less than 31")
```

**Safety.anno.json**
```.json
[
    {
      "class":"firrtl.annotations.PresetAnnotation",
      "target":"~Safety|Safety>preset"
    }
]
```

_The PresetAnnotation is needed in order to track the first cycle thus allowing us to properly reset the design_

Now you can run firrtl:
```.sh
./utils/bin/firrtl -i test/integration/Safety.fir -faf test/integration/Safety.anno.json
```

Which presents you with a nice, informative warning message:
```
------------------------------------------------------------------------------
Warning: 2 verification statements were removed when compiling to Verilog because the basic Verilog standard does not support them. If this was not intended, compile to System Verilog instead using the `-X sverilog` compiler flag.
------------------------------------------------------------------------------
```

Thus we switch to the SystemVerilog emitter:
```.sh
./utils/bin/firrtl -i test/integration/Safety.fir -faf test/integration/Safety.anno.json -X sverilog
```
which results in a `Safety.sv` file with the following contents (a removed the randomization logic to make this more readable):
```.sv
module Safety(
  input        clk,
  input        reset,
  output [4:0] counter
);
  reg [4:0] counter_r;
  wire [5:0] _GEN_0 = counter_r == 5'hf ? 6'h0 : counter_r + 5'h1;
  reg  _first = 1'h1;
  assign counter = counter_r;
  always @(posedge clk) begin
    if (reset) begin
      counter_r <= 5'h0;
    end else begin
      counter_r <= _GEN_0[4:0];
    end
    _first <= 1'h0;
  end
// [...] (removed randomization)
  always @(posedge clk) begin
    if (_first) begin
      assume(reset);
    end
    if (~reset) begin
      assert(counter < 5'h1f);
    end
  end
endmodule
```

Now we can create a SymbiYosys file as described in their [Quick Start Guide](https://symbiyosys.readthedocs.io/en/latest/quickstart.html):

**Safety.sby**
```.ini
[options]
mode bmc
depth 100

[engines]
smtbmc

[script]
read -formal Safety.sv
prep -top Safety

[files]
Safety.sv
```

Then we launch the tool:
```.sh
sby Safety.sby
```

Which does not find any errors 100 cycles into the design, using `yices2` as the solver.

src/test/scala/firrtlTests/formal/VerificationSpec.scala
tdb-alcorn: Fixed in 22876578ddfc1246f9341279f78b242397b79a06

tdb-alcorn: Fixed in 22876578ddfc1246f9341279f78b242397b79a06

spec/spec.tex
tdb-alcorn: Fixed in 22876578ddfc1246f9341279f78b242397b79a06

src/main/scala/firrtl/Emitter.scala
tdb-alcorn: Fixed

spec/spec.tex
tdb-alcorn: Fixed

PR #1654, opened 2020-06-02
Fix performance regression in DedupModules

jackkoenig:
Fix a performance bug in DedupModules introduced in #1539. Stop
recalculating the same expensive datastructures for each module,
potentially multiple times.

The correctness is easy to see. I hoisted a couple of datastructures out of the returned function.

### Benchmarking results

Times in seconds. This change is still benchmarking but results for small were promising.

|  Design | Before 1539 | 1539 | This PR |
| --- | --- | --- | --- |
| small | 39.4 ± 1.2 | 44.1 ± 0.8 | 37.7 ± 1.9 |
| medium | 86.9 ± 1.1 | 148.2 ± 2.0 | 90.7 ± 3.3 |

### Contributor Checklist

- [N/A] Did you add Scaladoc to every public function/method?
- [N/A] Did you update the FIRRTL spec to include every new feature/behavior?
- [N/A] Did you add at least one test demonstrating the PR?

#### Type of Improvement

 - performance improvement 

#### API Impact

None

None

#### Desired Merge Strategy

#### Release Notes
Fix performance regression in DedupModules

- [x] Did you review?

### DIFF
--- a/src/main/scala/firrtl/transforms/Dedup.scala
+++ b/src/main/scala/firrtl/transforms/Dedup.scala
@@ -165,7 +165,7 @@ class DedupModules extends Transform with DependencyAPIMigration with PreservesA
     }.toMap
 
     // get the ordered set of instances a module, includes new Deduped modules
-    val getChildrenInstances = (mod: String) => {
+    val getChildrenInstances = {
       val childrenMap = instanceGraph.getChildrenInstances
       val newModsMap: Map[String, mutable.LinkedHashSet[WDefInstance]] = dedupMap.map {
         case (name, m: Module) =>
@@ -175,7 +175,7 @@ class DedupModules extends Transform with DependencyAPIMigration with PreservesA
         case (name, m: DefModule) =>
           m.name -> mutable.LinkedHashSet.empty[WDefInstance]
       }.toMap
-      childrenMap.get(mod).getOrElse(newModsMap(mod))
+      (mod: String) => childrenMap.get(mod).getOrElse(newModsMap(mod))
     }
 
     val instanceNameMap: Map[OfModule, Map[Instance, Instance]] = {


src/main/scala/firrtl/transforms/Dedup.scala
```diff
@@ -165,7 +165,7 @@ class DedupModules extends Transform with DependencyAPIMigration with PreservesA
     }.toMap
 
     // get the ordered set of instances a module, includes new Deduped modules
-    val getChildrenInstances = (mod: String) => {
+    val getChildrenInstances = {
       val childrenMap = instanceGraph.getChildrenInstances
       val newModsMap: Map[String, mutable.LinkedHashSet[WDefInstance]] = dedupMap.map {
```
albert-magyar, line 170: I see that your way prevents re-computation, but since these are constant across the calls to `getChildrenInstances` there any reason not to just pull these data structures out of this anonymous scope, make them `val`s scoped to `run` and then make `getChildrenInstances` itself a one-liner def?

jackkoenig: I could do that. I did it this way partially to minimize the diff, although also partially that I kind of like the pattern of using anonymous scopes to contain vals that are only used inside of it, I find it self-documenting that these aren't used later. I'm happy to change it if you think it's more canonical or clear though.

albert-magyar: The return value is a function returning a mutable set, which means that it actually requires some detailed reading to know that the fact that data structures scoped to this anonymous closure are not later mutated.

I certainly don't object to just merging it as-is, but I think that the "anonymous closure to hide random local variables" is maybe a little less accessible. The subtle difference of an immediately-evaluated closure in a pure expression position vs. a traditional `val` function closure was what obscured the repetitive evaluation.

This is all totally a silly preference thing. I assumed you did it just to minimized the diff, but if you like it you should keep it :)

PR #1669, opened 2020-06-07
Build ArrayBuffers in Block.mapStmt

jackkoenig:
Micro-optimize `Block.mapStmt`, to construct `ArrayBuffer`s and flatten out nested `Block`s and `EmptyStmt`s. This will enable cleaning up awkward concatenations we often do at the end of complex transforms to prevent too much nesting ([for example](https://github.com/freechipsproject/firrtl/blob/14099e1d6b7937206cfc5309cbb0e044afc358d0/src/main/scala/firrtl/passes/ExpandWhens.scala#L48)). Furthermore, it should improve FlameGraphs in a few places. It has the added benefit of just cleaner intermediate FIRRTL circuits since `EmptyStmts` are always removed (although only on the next `mapStmt`).

This will need tweaking when we add 2.13 support, with the main candidate being `VectorBuilder`, but this is the best approach that I could find.

I'll note the benefits are small and should be revisited, but this does seem to speed things up slightly while cleaning up the process of iterating on Blocks.

### Benchmark Results

This change has a very small impact (Scala does a pretty good job of `map` on its datastructures, shocker), so I checked several implementations:

1. The current implementation
2. ArrayBuffer (the winner)
3. ArrayBuffer2 (tried to improve with some specialized code, no real benefit for awkward code)
4. VectorBuilder
5. ListBuffer

I also tried both Java8 and GraalVM Java11. The data appears to show that `ArrayBuffer` helps a little, `ArrayBuffer2` is maybe slightly better but the code was super awkward so probably not worth it. VectorBuilder and ListBuffer were worse (but much more so on OpenJDK8 than on GraalVM).

#### GraalVM Java 11

  | master | ArrayBuffer | ArrayBuffer2 | VectorBuilder | ListBuffer
-- | -- | -- | -- | -- | --
Icache.fir | 3.4 ± 0.2 | 3.4 ± 0 | 3.4 ± 0.1 | 3.4 ± 0.1 | 3.4 ± 0.1
small | 45.1 ± 1.9 | 43.2 ± 1.1 | 43.7 ± 0.8 | 41.6 ± 0.3 | 44.3 ± 2.1
medium | 153.9 ± 2.9 | 152.3 ± 8.2 | 144.8 ± 4.5 | 177.4 ± 9.7 | 141.1 ± 2.2
large | 424.6 ± 3.9 | 404.5 ± 2.1 | 434.1 ± 29.5 | 440.7 ± 19.5 | 432.8 ± 20.1

#### OpenJDK 8

  | master | ArrayBuffer | ArrayBuffer2 | VectorBuilder | ListBuffer
-- | -- | -- | -- | -- | --
Icache.fir | 2.7 ± 0.1 | 2.6 ± 0 | 2.7 ± 0 | 2.7 ± 0 | 2.7 ± 0
small | 48.2 ± 1.3 | 51.1 ± 1 | 50.7 ± 0.7 | 53.9 ± 0.7 | 53.9 ± 3.9
medium | 194 ± 3.4 | 187.1 ± 4.5 | 186.3 ± 5.5 | 233.6 ± 6.5 | 201.4 ± 8.4
large | 540 ± 13 | 535.7 ± 6.3 | 530.2 ± 5.6 | 649.8 ± 0.8 | 591.6 ± 6

#### Sanity Check for Massive Designs (GraalVM Java 11)

Because very large arrays can have issues, I wanted to sanity check that `ArrayBuffer` is actually decent on a couple bigger designs, it does just fine there:

  | master | ArrayBuffer
-- | -- | --
huge | 1275.4 ± 18.8 | 1188.8 ± 29.7
large multi-core | 1995.9 ± 92.6 | 1994.4 ± 72.8

### Contributor Checklist

- [NA] Did you add at least one test demonstrating the PR?

#### Type of Improvement

   - performance improvement  
  - code refactoring              

No impact on Verilog, intermediate emitted FIRRTL should no longer have skips (or at least have far fewer, log-level trace will still have skips from the previous transform).

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/ir/IR.scala
+++ b/src/main/scala/firrtl/ir/IR.scala
@@ -422,7 +422,21 @@ object Block {
 
 case class Block(stmts: Seq[Statement]) extends Statement {
   def serialize: String = stmts map (_.serialize) mkString "\n"
-  def mapStmt(f: Statement => Statement): Statement = Block(stmts map f)
+  def mapStmt(f: Statement => Statement): Statement = {
+    val res = new scala.collection.mutable.ArrayBuffer[Statement]()
+    var it = stmts.iterator
+    while (it.hasNext) {
+      it.next() match {
+        case EmptyStmt => // flatten out
+        case b: Block =>
+          val prev = it
+          it = b.stmts.iterator ++ prev
+        case other =>
+          res.append(f(other))
+      }
+    }
+    Block(res)
+  }
   def mapExpr(f: Expression => Expression): Statement = this
   def mapType(f: Type => Type): Statement = this
   def mapString(f: String => String): Statement = this


jackkoenig:
Tests failed because some tests are expecting the skips 🙄

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -421,8 +421,29 @@ object Block {
...
+  def mapStmt(f: Statement => Statement): Statement = {
+    val res = new scala.collection.mutable.ArrayBuffer[Statement]()
+    var its = stmts.iterator :: Nil
+    while (its.nonEmpty) {
+      val it = its.head
+      if (it.hasNext) {
+        it.next() match {
+          case EmptyStmt => // flatten out
+          case b: Block =>
+            its = b.stmts.iterator :: its
```
seldridge, line 437: I'm playing around with my transform checking work and I noticed that this behavior may be unintuitive. 

Specifically, it seems counterintuitive that doing a `map` with an identity function can actually (superficially) modify the circuit, e.g.:

```scala
/* The following: */
Block(Seq(EmptyStmt, 
          Block(DefWire(NoInfo, "foo", UIntType(IntWidth(1)))))).mapStmt(identity)

/* Returns: */
Block(ArrayBuffer(DefWire(NoInfo, "foo", UIntType(IntWidth(1)))))
```

This has the effect of making transforms which don't have meaningful modifications to the circuit, superficially modify the circuit. Namely, I'm seeing a false positive invalidation in one of my tests where I think `ExpandConnects` is adding a `Block` and then `CInferTypes` is "changing the circuit" by stripping it, even though `CInferTypes` does no real work. Clearly, the issue is `ExpandConnects` adding a block (and my test treating this as a false positive), but the behavior of `mapStmt(identity)` still smells weird to me.

jackkoenig: > Specifically, it seems counterintuitive that doing a map with an identity function can actually (superficially) modify the circuit

Oh yeah this is totally wrong. Thinking about `Statement` as a `Functor`, this is a violation of the first functor law: Functors must preserve identity morphisms.

I'd also note that there's a pretty bad inconsistency with `foreach`, which while not a violation of a law to my knowledge, has the potential for confusion and nasty bugs.

How should we resolve this?
* We can revert and reproduce the flattening functionality in a different, more explicit way.
* Alternatively we could define `==` (and `hashCode`) on `Block` to match these semantics

seldridge: Explicit functionality via a new transform, `FlattenStatements`, which (1) removes `EmptyStmt` and (2) flattens/joins `Block`s is the most intuitive solution to me. This would be part of one of the very early forms (`Forms.MinimalHighForm` or `Forms.Resolved`?) and I can deal with adding the appropriate invalidations as they're discovered (I'm pretty sure `ExpandConnects` is adding `Block`s...).

jackkoenig: I'm of two minds.

When I attempted this kind of thing in the past, I just tactically ran the flattening where necessary. The issue was that it actually slowed things down a little bit (like 5%), whereas this approach of overlaying the flattening with traversal has no overhead/perhaps even a slight win. It could be possible to find a to do both: make the flattening opt-in but also overlay with the next run, but that could also be tricky.

ekiwi-sifive: Can we just "fix" `ExpandConnects` to not emit that unnecessary block?

seldridge: Yes, we should fix that.

Nonetheless, the behavior here is weird. The old behavior was a `map` the new behavior is something like ` flatMap`, but not quite.

You can play around with this at the following Scastie. You can see the old behavior if you switch to Chisel 3.3.1.

https://scastie.scala-lang.org/vlk2fHeUSXmnXsefl71CrA

PR #1670, opened 2020-06-09
Handle splitting of compound expressions in Mux

sahandKashani:
SplitExpressions technically handles Muxes in the call to split(), but
split() is only called on a DoPrim and Muxes are not DoPrims. We therefore
end up with cases where muxes still contain nested expressions after calling
this pass.

This commit fixes the issue.

### Contributor Checklist

#### Type of Improvement

- bug fix

#### API Impact

No API changes.

Generated verilog should now contain an intermediate signal that breaks up compound expressions in Muxes, similarly to how all other nested non-mux expressions are broken up.

#### Desired Merge Strategy

Squash: The PR will be squashed and merged (choose this if you have no preference. -->

#### Release Notes

Fix bug where compound expressions in Muxes were not being split by the `SplitExpressions` pass.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/passes/SplitExpressions.scala
+++ b/src/main/scala/firrtl/passes/SplitExpressions.scala
@@ -55,6 +55,7 @@ object SplitExpressions extends Pass {
         def onExp(e: Expression): Expression =
           e map onExp match {
             case ex: DoPrim => ex map split
+            case ex: Mux => ex map split
             case ex => ex
          }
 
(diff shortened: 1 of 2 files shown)


sahandKashani:
Hi folks.

I see that a regression test failed with my modification, but I didn't get this specific error when running `sbt test`. Are they run through that command? If not, could anyone point me to how to run these tests locally so I can debug what went wrong?

Thanks!

sahandKashani:
Alright, I found the test. I'll close this PR while I investigate the regression failure.

Issue #1680, opened 2020-06-10
Compiler.internalTransform: timing information should not be logged as an error


ekiwi-sifive:
### The Problem

In `Compiler.internalTransform`, the time it took to compute the transform order as well as the compile time are logged as `errors`:

```.scala
logger.error(f"Computed transform order in: $timeResolveDependencies%.1f ms")
/* Show the determined transform order */
logger.info("Determined Transform order that will be executed:\n" + tm.prettyPrint("  "))
/* Run all determined transforms tracking how long everything takes to run */
val (timeExecute, annotationsOut) = firrtl.Utils.time { tm.transform(c.stateIn) }
logger.error(f"Total FIRRTL Compile Time: $timeExecute%.1f ms")
```

### The Solution

I would like to propose to change the `logger.error` to `logger.info` since the time taken to compute things obviously is not an error.

Changing this is the second puzzle piece (after https://github.com/freechipsproject/chisel3/issues/1473) in my attempt to get a silent Chisel/Firrtl compiler.

### Potential Problems with the Proposed Solution

I guess there might be backwards compatibility concerns, i.e. users complaining that they do not see the compile time anymore. Maybe this could be solved by setting the log level to `info` by default. However, currently there is no option to turn off these messages and I think that needs to be fixed.

PR #1685, opened 2020-06-10
Compiler: demote compile time to info instead of error

ekiwi-sifive:
fixes https://github.com/freechipsproject/firrtl/issues/1680

### DIFF
--- a/src/main/scala/firrtl/stage/phases/Compiler.scala
+++ b/src/main/scala/firrtl/stage/phases/Compiler.scala
@@ -100,12 +100,12 @@ class Compiler extends Phase with Translator[AnnotationSeq, Seq[CompilerRun]] wi
       val tm = new firrtl.stage.transforms.Compiler(targets)
       /* Transform order is lazily evaluated. Force it here to remove its resolution time from actual compilation. */
       val (timeResolveDependencies, _) = firrtl.Utils.time { tm.flattenedTransformOrder }
-      logger.error(f"Computed transform order in: $timeResolveDependencies%.1f ms")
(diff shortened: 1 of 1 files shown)


seldridge:
How far back should we backport this? 

This can be backported to 1.2.x, but the patch is totally different:

```diff
diff --git a/src/main/scala/firrtl/Compiler.scala b/src/main/scala/firrtl/Compiler.scala
index 4f17b4f0..744d9f92 100644
--- a/src/main/scala/firrtl/Compiler.scala
+++ b/src/main/scala/firrtl/Compiler.scala
@@ -487,7 +487,7 @@ trait Compiler extends LazyLogging {
       }
     }
 
-    logger.error(f"Total FIRRTL Compile Time: $timeMillis%.1f ms")
+    logger.info(f"Total FIRRTL Compile Time: $timeMillis%.1f ms")
 
     finalState
   }
```

ekiwi-sifive:
> How far back should we backport this?

Should this be backported at all? Seems like it would alter the user experience.

seldridge:
Re: backporting: I think the policy at this point has been to greedily backport. I.e., backport a PR as far back as it will reasonably go without an API breakage.

Or, inductively: if this PR provides a better user experience in 1.4, then it's a better 1.3.x experience, then it's a better 1.2.x experience...

Thoughts @jackkoenig.

jackkoenig:
I wouldn't backport this, at least not to `1.2.x`. I think the argument can be made for `1.3.x` since it's kind of the "current stable release branch". In general I see these kinds of things as changing on major boundaries but I don't feel that strongly about it.

PR #1689, opened 2020-06-12
RemoveIntervals: fixup kinds of refs to replaced nodes

albertchen-sifive:
This updates `RemoveIntervals` so that it fixes up `WRef`s to the nodes that it replaced with wires. alternatively we could make `RemoveIntervals` invalidate `ResolveKinds`, but it's simple enough to fix it in `RemoveIntervals` since it needs to traverse through the expressions anyway and I didn't want to mess up `LoweringCompilersSpec`.

### Contributor Checklist

#### Type of Improvement

- bug fix

#### API Impact

none

none

#### Desired Merge Strategy

- Squash

#### Release Notes

fix bug in RemoveIntervals where refs to replaced nodes did not have the correct kind

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/passes/RemoveIntervals.scala
+++ b/src/main/scala/firrtl/passes/RemoveIntervals.scala
@@ -50,7 +50,7 @@ class RemoveIntervals extends Pass with PreservesAll[Transform] {
   def run(c: Circuit): Circuit = {
     val alignedCircuit = c
     val errors = new Errors()
-    val replacedNodes = mutable.HashMap.empty[OfModule, mutable.HashSet[String]]
+    val replacedNodes = mutable.HashMap.empty[OfModule, mutable.Set[String]]
     val wiredCircuit = alignedCircuit map makeWireModule(replacedNodes)
     val replacedCircuit = wiredCircuit map replaceModuleInterval(errors, replacedNodes)
     errors.trigger()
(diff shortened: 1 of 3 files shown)


src/main/scala/firrtl/passes/RemoveIntervals.scala
```diff
@@ -48,29 +50,32 @@ class RemoveIntervals extends Pass with PreservesAll[Transform] {
...
-    s map replaceTypeInterval map replaceStmtInterval(errors, mname) map replaceExprInterval(errors, info, mname)
+    s map replaceTypeInterval map replaceStmtInterval(errors, mname, replacedNodes) map replaceExprInterval(errors, info, mname, replacedNodes)
 
   }
 
-  private def replaceExprInterval(errors: Errors, info: Info, mname: String)(e: Expression): Expression = e match {
+  private def replaceExprInterval(errors: Errors, info: Info, mname: String, replacedNodes: Set[String])(e: Expression): Expression = e match {
+    case ref: WRef if replacedNodes(ref.name) =>
+      ref.copy(kind = WireKind)
     case _: WRef | _: WSubIndex | _: WSubField => e
```
jackkoenig, line 76: I see you're fixing the Kind, are the types correct? I see there is `replaceTypeInterval`, but as far as I can tell, this is not applied to `Expressions`.

albertchen-sifive: yeah It's used for ports, but it doesn't look like the types fixedup are afterwards

jackkoenig: I'm wondering if instead we should just invalidate types and kinds. Doing the kinds here is faster than running `ResolveKinds`, but if we have to run `InferTypes` anyway 🤷

albertchen-sifive: oh wait my bad. it does run `InferTypes` afterwards https://github.com/freechipsproject/firrtl/blob/master/src/main/scala/firrtl/passes/RemoveIntervals.scala#L54. so I guess just delete that line and invalidate `InferTypes` instead?

jackkoenig: Well unfortunately this mixes in `PreservesAll[Transform]` so we can't change that behavior in a backportable way. Perhaps the best approach is to do what you originally did, leave the `InferTypes` call alone (change it in a non-backportable 1.4 version of this PR) and just fixup the Kinds as you did.

jackkoenig: @seldridge thoughts?

seldridge: There should be two PRs or backport this and change it in 1.3.x:

1) 1.4: Remove `PreservesAll[Transform]` and add the appropriate invalidations
2) 1.3.x: Manually run the invalidated transforms at the finish of this transform

The lack of backportability of `PreservesAll` was my bad. This is intended to be a "helper" mix-in, but it's just busted. Sorry.

PR #1690, opened 2020-06-15
ir: store FileInfo string in escaped format

ekiwi-sifive:
This should speed up the common case as the compiler never operates on the unescaped string. The new escape function also fixes a bug where ']' was not escaped even though it is the delimiting character for FileInfo. In order to maintain backwards compatibility for the ProtoBuf format, this patch adds escape/unescape calls when going from/to protobuf format.
For better performance we should consider changing the protobuf format.

### Contributor Checklist

#### Type of Improvement

- performance improvement
- bug fix

#### API Impact

- deprecates constructing `FileInfo` with a `StringLit` (`StringLit` is still used for e.g. `Print` statements which have different escaping requirements)

- should not change the generated verilog

#### Desired Merge Strategy
- Squash

#### Release Notes
n/a

- [ ] Did you review?


src/main/scala/firrtl/ir/IR.scala
```diff
@@ -22,20 +22,46 @@ case object NoInfo extends Info {
...
+
+/** Stores the string of a file info annotation in its escaped form. */
+class FileInfo(val info: String) extends Info {
+  override def toString: String = " @[" + info + "]"
   //scalastyle:off method.name
   def ++(that: Info): Info = if (that == NoInfo) this else MultiInfo(Seq(this, that))
 }
+
+object FileInfo {
+  @deprecated("The use of FileInfo(...) is deprecated. Use FileInfo.fromUnEscaped(...) instead", "1.4")
```
seldridge, line 34: Two nits about writing deprecation warnings:

1. This will auto-generate the message about "FileInfo is deprecated" so the PR can leave that off.
2. The PR should include the library name in the "since" parameter.

The above two are somewhat canonical with how Scala does it. In Chisel and FIRRTL, we've tended to adopt also adding a "This will be removed in FIRRTL 1.5" message to be explicit with how long somebody has to get their code updated.

Consider:

```suggestion
  @deprecated("Use FileInfo.fromUnEscaped instead. FileInfo.apply will be removed in FIRRTL 1.5.", "FIRRTL 1.4")
```

I believe that will then generate the following message:

```
method apply in object FileInfo is deprecated (since FIRRTL 1.4): Use FileInfo.fromUnEscaped instead. FileInfo.apply will be removed in FIRRTL 1.5.
```

ekiwi-sifive: Thanks!

ekiwi-sifive:
@jackkoenig Please let me know your thoughts on how to break the API to the Chisel compiler.

jackkoenig:
I've been thinking about a bunch of course locator stuff, rather than changing `FileInfo`, why don't we introduce a new construct that becomes the default:
```scala
case class FileLineCol(file: String, line: Int, col: Int)
```
This construct could then store the filename in escaped format.

We can then think about either eliminating the `FileInfo` construct altogether since they're a leaky abstraction that have been abused, or also migrate them.

ekiwi-sifive:
> I've been thinking about a bunch of course locator stuff, rather than changing `FileInfo`, why don't we introduce a new construct that becomes the default:

I think splitting this info up into the different fields is good, once we actually operate on it in the compiler. Right now we never modify or even look at the content of a `FileInfo` node and thus it does not make sense to me to parse it into the separate fields and then put them back together once we serialize. Essentially, I think we should keep it as simple as possible until we actually need the extended functionality.

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -22,20 +22,45 @@ case object NoInfo extends Info {
...
+      .replace("\t", "\\\t")
+      .replace("]", "\\]")
+  }
+  /** removes the `\` in front of `\`, `\n`, `\t` and `]` */
+  def unescape(s: String): String = {
+    s.replace("\\]", "]")
+      .replace("\\\t", "\t")
+      .replace("\\\n", "\n")
+      .replace("\\\\", "\\")
+  }
```
jackkoenig, line 52: I think this escaping logic is sufficient (and arguably better) for in and out of FIRRTL (although we should check that with the Parser), but I'm a bit worried about the Verilog emission. If you have a unicode character in a source locator, currently we emit a unicode escape, eg.
```
y <= x @[file.scala\t123:11 😎]
```
Emits as
```
assign y = x; // @[file.scala\t123:11 \uD83D\uDE0E]
```

This PR changes it to emit as:
```
assign y = x; // @[file.scala\t123:11 😎]
```

This may be legal, but if we're going to change behavior there should be tests illustrating this and we need to be sure that this doesn't violate the Verilog 1995 spec.

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -22,20 +22,45 @@ case object NoInfo extends Info {
...
-case class FileInfo(info: StringLit) extends Info {
-  override def toString: String = " @[" + info.serialize + "]"
+
+/** Stores the string of a file info annotation in its escaped form. */
+case class FileInfo(escaped: String) extends Info {
+  override def toString: String = " @[" + escaped + "]"
   //scalastyle:off method.name
   def ++(that: Info): Info = if (that == NoInfo) this else MultiInfo(Seq(this, that))
+  def unescaped: String = FileInfo.unescape(escaped)
+}
```
jackkoenig, line 32: To be backwards compatible, this needs a constructor that accepts `info: StringLit` and an accessor for `info`: `def info: StringLit`.

Also we should think about `unapply`, anyone `unapplying` and using the result will have their code break from this change. We could manually implement the `unapply` and deprecate it; then we won't have an unapply but we can add it after removing the old one.

src/test/scala/firrtlTests/InfoSpec.scala
```diff
@@ -172,4 +172,27 @@ class InfoSpec extends FirrtlFlatSpec with FirrtlMatchers {
...
+    parseAnno("test\\\ntest").unescaped should be ("test\ntest")
+    parseAnno("test\\\ttest").escaped should be ("test\\\ttest")
+    parseAnno("test\\\ttest").unescaped should be ("test\ttest")
+    parseAnno("test\\\\test").escaped should be ("test\\\\test")
+    parseAnno("test\\\\test").unescaped should be ("test\\test")
+    parseAnno("test\\]test").escaped should be ("test\\]test")
+    parseAnno("test\\]test").unescaped should be ("test]test")
+    parseAnno("test[\\][\\]test").escaped should be ("test[\\][\\]test")
+    parseAnno("test[\\][\\]test").unescaped should be ("test[][]test")
+  }
```
jackkoenig, line 196: 1. `parseAnno` is the wrong name here (I'm guessing it's from copy-paste), but just like `parse` would be better or `parseInfo`.

2. I would like to see tests that use the actual ANTLR Parser (because this changes the emitted FIRRTL) as well as in and out of protobuf.

src/main/scala/firrtl/ir/IR.scala
ekiwi-sifive: I honestly completely forgot about the Verilog emission and need to fix that.
Thanks a lot for pointing that out!

The way I would do it is to call the `unescape` function and then the old `encodeJava` function in the Verilog emitter. I will add a commit shortly.

src/main/scala/firrtl/ir/IR.scala
ekiwi-sifive: > To be backwards compatible, this needs a constructor that accepts ``

I added a backwards compatible apply method. That should hopefully be enough. Do you really want to be backwards compatible for people who call the constructor directly (i.e. with `new`)?

I could try to add an unapply, but it would be pretty inefficient to use. Essentially you would have to call `unescape` for it.

src/test/scala/firrtlTests/InfoSpec.scala
ekiwi-sifive: 1. I am going to change it to `parseInfo`. I didn't consider that `anno` was already used for something else.

2. Isn't `firrtl.Parser.parse` invoking the antlr parser? I can add some protobuf tests.

ekiwi-sifive:
Thinking a little more about how `Info` should be serialized in the Emitter I stumbled over this current behavior: https://github.com/freechipsproject/firrtl/issues/1695

src/test/scala/firrtlTests/InfoSpec.scala
```diff
@@ -172,4 +172,27 @@ class InfoSpec extends FirrtlFlatSpec with FirrtlMatchers {
...
+    def input(anno: String): String =
+      s"""circuit m: @[$anno]
+        |  module m:
+        |    skip
+        |""".stripMargin
+    def parseAnno(anno: String): FileInfo = {
+      firrtl.Parser.parse(input(anno)).info.asInstanceOf[FileInfo]
+    }
+
+    parseAnno("test\\\ntest").escaped should be ("test\\\ntest")
```
ekiwi-sifive, line 186: This test is actually wrong: should be `test\\n` not `test\\\n`

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -22,20 +22,45 @@ case object NoInfo extends Info {
...
+  @deprecated("Use FileInfo.fromUnEscaped instead. FileInfo.apply will be removed in FIRRTL 1.5.", "FIRRTL 1.4")
+  def apply(info: StringLit): FileInfo = new FileInfo(escape(info.string))
+  def fromEscaped(s: String): FileInfo = new FileInfo(s)
+  def fromUnescaped(s: String): FileInfo = new FileInfo(escape(s))
+  /** prepends a `\` to: `\`, `\n`, `\t` and `]` */
+  def escape(s: String): String = {
+    s.replace("\\", "\\\\")
+      .replace("\n", "\\\n")
+      .replace("\t", "\\\t")
+      .replace("]", "\\]")
```
ekiwi-sifive, line 44: This is broken, should replace `\n` with `\\n`!
And then we have a problem when trying to escape something like `"test\\\\t"` which should be escaped to `"test\\t"` not `"test\t"`

ekiwi-sifive: TODO: replace with a custom Translator based on the apache commons: https://github.com/apache/commons-lang/blob/master/src/main/java/org/apache/commons/lang3/StringEscapeUtils.java#L57

ekiwi-sifive: The main difference should be that we do not escape unicode characters (or should we? at some point this should be documented in the Firrtl spec!) and `"` and that we do escape `]` because that is the delimiter used for the parser.

jackkoenig: If we can define that FIRRTL files are UTF-8 as you suggested (need to check that this is a good idea), then I think we should not escape unicode in emitted FIRRTL. But we certainly need it escaped in emitted Verilog.

ekiwi-sifive: > But we certainly need it escaped in emitted Verilog.
Yes. I am working on moving the Verilog focused escaping into the Emitter (other parts of the compiler should not have to know about Verilog).

ekiwi-sifive: Once I take more precise control over the escaping I can probably just add the unicode escaping on top of the Verilog stuff and unescape the `]` in one go.

ekiwi-sifive: This is the newer version using `commons-text` instead of `commons-lang`: https://github.com/apache/commons-text/blob/master/src/main/java/org/apache/commons/text/StringEscapeUtils.java#L62

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -22,24 +24,73 @@ case object NoInfo extends Info {
...
+  private val EscapedToVerilog = new AggregateTranslator(
+    new LookupTranslator(Seq(
+      // ] is the one character that firrtl needs to be escaped that does not need to be escaped in
+      "\\]" -> "]" : CharMap,
+      "\"" -> "\\\"",
+      // \n and \t are already escaped
+      "\b" -> "\\b",
+      "\f" -> "\\f",
+      "\r" -> "\\r",
+  ).toMap.asJava),
```
ekiwi-sifive, line 74: @jackkoenig : any idea why this wouldn't compile with Scala `2.11` but work fine with `2.12`?
```
[error] /Volumes/develop/firrtl/src/main/scala/firrtl/ir/IR.scala:74:3: illegal start of simple expression
[error]   ).toMap.asJava),
[error]   ^
[error] /Volumes/develop/firrtl/src/main/scala/firrtl/ir/IR.scala:78:1: ')' expected but '}' found.
[error] }
[error] ^
[error] two errors found
```

seldridge: Trailing comma on the previous line?

ekiwi-sifive: I think I may have found what it was upset about: the trailing `,`

src/main/scala/firrtl/ir/IR.scala
ekiwi-sifive: @jackkoenig : turns out at least `Scala 2.11` does not like having two `unapply` methods:
```
[error] /Volumes/develop/firrtl/src/main/scala/firrtl/ir/IR.scala:29:12: method unapply is defined twice
[error]   conflicting symbols both originated in file '/Volumes/develop/firrtl/src/main/scala/firrtl/ir/IR.scala'
```

So it seems like this would require us to make `FileInfo` not a case class.

> Also we should think about `unapply`, anyone `unapplying` and using the result will have their code break from this change. We could manually implement the `unapply` and deprecate it; then we won't have an unapply but we can add it after removing the old one.

seldridge:
I think this PR is pushing towards two separate, admirable issues:

1) What is the FIRRTL compiler in-memory representation of a path?
2) How is this emitted in FIRRTL IR and how is this emitted in Verilog?

We need to get a handle on this and make sure that it works for Linux paths, Windows paths, with and without Unicode, and potentially filesystems reporting things incorrectly. Things are clearly broken right now, so any improvement is good. I don't think this PR has to solve all of them.

I do wonder if there's a Java/Scala library that has this already figured out?

There's an old blog post from the Beets project where they're dealing with similar issues of multi-OS support for reading paths into a database: https://beets.io/blog/paths.html. This seems topical.

ekiwi-sifive:
> Things are clearly broken right now, so any improvement is good. I don't think this PR has to solve all of them.

Yes. So this PR is focused on preserving outside visible behavior while making some compilation cases faster. The one behavior change is that `]` should now be properly escaped, however, I would treat this as a bug fix rather than a new features.

ekiwi-sifive:
While this does speed up serialization, the results I got when trying to cold benchmark end-to-end are non-conclusive:

  |   | from ProtoBuf |   | from Firrtl |  
-- | -- | -- | -- | -- | --
revision | design | runtime (s) | SD | runtime (s) | SD
7d299b28 (PR) | small | 52.83 | 0.98 | 50.44 | 0.25
  | medium | 122.13 | 5.03 | 132.09 | 0.67
b750754c (old) | small | 45.06 | 3.45 | 51.64 | 0.61
  | medium | 137.78 | 39.96 | 122.22 | 8.05

ekiwi-sifive:
Some Hot Benchmarking (10 warumup, 10 runs):

Hot Benchmark (10,10):

JVM 14
File-info-escaped + medium: 887ms +/- 39ms
File-info-escaped + small: 434ms +/- 4ms
Old + medium: 1282ms +/- 18ms
Old + small: 638ms +/- 12ms

JVM 1.8
File-info-escaped + medium: 1728ms +/- 9ms
File-info-escaped + small: 851ms +/- 9ms
Old + medium: 2273ms +/- 182ms
old + small: 1091ms +/- 7ms

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -22,24 +24,73 @@ case object NoInfo extends Info {
...
+  def escape(s: String): String = EscapeFirrtl.translate(s)
+  /** removes the `\` in front of `\`, `\n`, `\t` and `]` */
+  def unescape(s: String): String = UnescapeFirrtl.translate(s)
+  /** take an already escaped String and do the additional escaping needed for Verilog comment */
+  def escapedToVerilog(s: String) = EscapedToVerilog.translate(s)
+
+  // custom `CharSequenceTranslator` for FIRRTL Info String escaping
+  type CharMap = (CharSequence, CharSequence)
+  private val EscapeFirrtl = new LookupTranslator(Seq(
+    "\\" -> "\\\\" : CharMap,
```
jackkoenig, line 53: Is this ensuring the whole `Seq` is a `Seq[CharMap]`? I think it'd be more clear to just do:
```scala
  private val EscapeFirrtl = new LookupTranslator(Seq[CharMap](
    "\\" -> "\\\\", ...
```
I've never seen this style of setting the type of the whole `Seq`

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -22,24 +24,73 @@ case object NoInfo extends Info {
   override def toString: String = ""
   def ++(that: Info): Info = that
 }
-case class FileInfo(info: StringLit) extends Info {
-  override def toString: String = " @[" + info.serialize + "]"
+
+/** Stores the string of a file info annotation in its escaped form. */
+case class FileInfo(escaped: String) extends Info {
```
jackkoenig, line 29: Unfortunately this is a backwards breaking change and there's no way to provide the old `unapply` and deprecate it. You're _supposed_ to be able to, but there's a bug in Scala scala/bug 11252 (not linking because I don't want to put them on blast).

I think this is worth the improvement and the best we can do we I'll accept it but I won't like it 🙅 .

src/main/scala/firrtl/proto/FromProto.scala
```diff
@@ -45,9 +45,9 @@ object FromProto {
       case Firrtl.SourceInfo.POSITION_FIELD_NUMBER =>
         val pos = info.getPosition
         val str = s"${pos.getFilename} ${pos.getLine}:${pos.getColumn}"
-        ir.FileInfo(ir.StringLit(str))
+        ir.FileInfo.fromUnescaped(str)
       case Firrtl.SourceInfo.TEXT_FIELD_NUMBER =>
-        ir.FileInfo(ir.StringLit(info.getText))
+        ir.FileInfo.fromUnescaped(info.getText)
```
jackkoenig, line 50: Ugh, we should stop escaping it in the proto, beyond the scope of this PR though because we obviously want to support old protos

src/main/scala/firrtl/ir/IR.scala
ekiwi-sifive: Thanks! This is definitely the right way to do it (TM). I was just trying until the type inference did what I wanted.

src/main/scala/firrtl/proto/FromProto.scala
ekiwi-sifive: Yeah, I added this to make sure that the proto format stays the same.

src/main/scala/firrtl/ir/IR.scala
ekiwi-sifive: > I think this is worth the improvement and the best we can do we I'll accept it but I won't like it

Yeah. I think at this point it is either breaking backwards compatibility or we decide to abandon the PR.

src/main/scala/firrtl/ir/IR.scala
ekiwi-sifive: Should be fixed in latest commit.

ekiwi-sifive:
## My Manual Regression Test

425354a4 (~master) vs 4ffdf9e (this PR)

**test.fir**
```.fir
circuit test:
  module test:
    input t1: UInt<1> @[test \ntest]
    input t2: UInt<1> @[test \ttest]
    input t3: UInt<1> @[test \]test]
    input t4: UInt<1> @[test 😎test]
    input t5: UInt<1> @[file.scala\t123:11 😎]
```

### Verilog

**test.v** produced by `java -cp firrtl.425354a4.jar firrtl.stage.FirrtlMain -i test.fir -X verilog`
```.v
module test(
  input   t1, // @[test \ntest]
  input   t2, // @[test \ttest]
  input   t3, // @[test ]test]
  input   t4, // @[test \uD83D\uDE0Etest]
  input   t5 // @[file.scala\t123:11 \uD83D\uDE0E]
);
endmodule
```

**test.v** produced by `java -cp firrtl.4ffdf9ec.jar firrtl.stage.FirrtlMain -i test.fir -X verilog`
```.v
module test(
  input   t1, // @[test \ntest]
  input   t2, // @[test \ttest]
  input   t3, // @[test ]test]
  input   t4, // @[test \uD83D\uDE0Etest]
  input   t5 // @[file.scala\t123:11 \uD83D\uDE0E]
);
endmodule
```

### LoFirrtl

**test.lo.fir** produced by `java -cp firrtl.425354a4.jar firrtl.stage.FirrtlMain -i test.fir -X low`
```.fir
circuit test :
  module test :
    input t1 : UInt<1> @[test \ntest]
    input t2 : UInt<1> @[test \ttest]
    input t3 : UInt<1> @[test ]test]
    input t4 : UInt<1> @[test \uD83D\uDE0Etest]
    input t5 : UInt<1> @[file.scala\t123:11 \uD83D\uDE0E]

    skip
```

_Note_: here the `]` isn't properly escaped and will fail parsing. The `😎` is escaped and will be unescaped when reparsing the firrtl.

**test.lo.fir** produced by `java -cp firrtl.4ffdf9ec.jar firrtl.stage.FirrtlMain -i test.fir -X low`
```.fir
circuit test :
  module test :
    input t1 : UInt<1> @[test \ntest]
    input t2 : UInt<1> @[test \ttest]
    input t3 : UInt<1> @[test \]test]
    input t4 : UInt<1> @[test 😎test]
    input t5 : UInt<1> @[file.scala\t123:11 😎]

    skip
```

_Note_: the `]` is properly escaped. We no longer escape Unicode characters when serializing to firrtl.

### Protobuf

TODO...

ekiwi-sifive:
## Chisel3 to Protobuf

I created a file called `Smile😎Spec.scala`:
```.scala
/*...*/

class Smile extends Module {
  val io = IO(new Bundle { val in = Input(Bool()) ; val out = Output(Bool())})
  io.in <= io.out
}
class SmileSpec extends FlatSpec {
  val c = Driver.elaborate(() => new Smile())
  Driver.dumpProto(c, Some(new File("smile.pb")))
}
```

With the current chisel3 `master` branch, `cat smile.pb`:
```

�
��
Smile
clock"
reset
)
io!*
in

out
J2+
_T%R:



ioin:


ioout�Smile😎Spec.scala 13:9
Smile⏎
```

After publishing this branch locally and running the same code, the resulting `simile.pb` file is exactly the same.

ekiwi-sifive:
## Protobuf to Verilog

This is after fixing the source of `Smile` to correctly connect `io.out := io.in`.

**Smile.v** produced by `java -cp firrtl.425354a4.jar firrtl.stage.FirrtlMain -i smile.pb -X verilog`
```.v
module Smile(
  input   clock,
  input   reset,
  input   io_in,
  output  io_out
);
  assign io_out = io_in; // @[Smile\uD83D\uDE0ESpec.scala 13:10]
endmodule
```

**Smile.v** produced by `java -cp firrtl.4ffdf9ec.jar firrtl.stage.FirrtlMain -i smile.pb -X verilog`
```.v
module Smile(
  input   clock,
  input   reset,
  input   io_in,
  output  io_out
);
  assign io_out = io_in; // @[Smile\uD83D\uDE0ESpec.scala 13:10]
endmodule

ekiwi-sifive:
@jackkoenig Do you understand why the travis check is still outstanding but there are no Details available?

jackkoenig:
>  Do you understand why the travis check is still outstanding but there are no Details available?

Nope but there doesn't appear to be a corresponding build here: https://travis-ci.com/github/freechipsproject/firrtl/pull_requests. Try pushing another commit

ekiwi-sifive:
TODO: rebase on latest master to fix conflicts.

ekiwi-sifive:
Now rebased on the latest main branch. Let's see if the tests still pass...

PR #1711, opened 2020-06-23
Emitter: disable some scalastyle warnings

ekiwi-sifive:
Adds some `scalastyle:off` comments to fix the most annoying `scalastyle` warnings without any code changes.

### Contributor Checklist

#### Type of Improvement
- code cleanup

#### API Impact
- none

- any change in code generation is a bug

#### Desired Merge Strategy
- squash

#### Release Notes

n/a

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -190,6 +190,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
 
   val outputSuffix = ".v"
   val tab = "  "
+  //scalastyle:off method.name
   def AND(e1: WrappedExpression, e2: WrappedExpression): Expression = {
     if (e1 == e2) e1.e1
     else if ((e1 == we(zero)) | (e2 == we(zero))) zero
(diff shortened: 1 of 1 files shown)


src/main/scala/firrtl/Emitter.scala
```diff
@@ -190,6 +190,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
 
   val outputSuffix = ".v"
   val tab = "  "
+  //scalastyle:off method.name
```
seldridge, line 193: I think if the PR uses `scalastyle:off`, then it would need to turn it back on at the end of the method?

There is an inline `scalastyle:ignore` that the PR could use on a per-line basis that might work better for the method name ones.

ekiwi-sifive: > I think if the PR uses `scalastyle:off`, then it would need to turn it back on at the end of the method?

Thanks for catching that. The style in which I was using it to disable things for individual classes or methods is used in some other Chisel related code base, but as you pointed out, it actually stays off for the rest of the file.

I think in the end the thing to do is to close these PRs and add a reminder on the developer's meeting schedule to run `scalafmt` before releasing `1.4`.

Issue #1679, opened 2020-06-10
Syntax for Labled Statements


ekiwi-sifive:
### Feature Description

We need to be able to label statements with an id in order to annotate them.

Statements that already have a name because they are declarations:
- Wires
- Registers
- Nodes
- Memories (incl. ports)
- Instances

Statements that I consider important to be able to annotate:
- Conditionals
- Stops
- Formatted Prints
- Verification (https://github.com/freechipsproject/firrtl/pull/1653)

Other Statements:
- Connects
- Partial Connects
- Invalidates
- Attaches
- Empty

### Type of Feature
- new firrtl feature

### Related Features
This would be nice for the new `Verfication` statements: https://github.com/freechipsproject/firrtl/pull/1653

### Design Decisions

Is this needed?
- I think we all agree that this is the case.

What should the label be?
- A valid, unique, firrtl identifier.

Should the label be optional?
- I would say yes, because this would automatically give us the ability to load older firrtl files.

Which statements should be annotated?
- I believe at the minimum: `when`, `else`, `stop`, `printf` + the new Verification statements
- Do people feel strongly about annotating any other statements?

What should the syntax be?
- I like `stmt(...) [ : lbl]` but while this would work for `stop` etc, it does become very ugly for `when` statements (`when c2 : : lbl`)
- Please comment bellow with any syntax suggestions that you have and let's see if we can get a consensus.

ekiwi-sifive:
We talked about this in the dev meeting and the consensus was that people should label signals going into statements instead of trying to label statements themselves.

PR #1712, opened 2020-06-23
verification: emit mesage as Verilog comment

ekiwi-sifive:
### Contributor Checklist

#### Type of Improvement

- backend comment generation

#### API Impact
- no API impact

- this adds the message included with a verification statement as a Verilog comment to the Verilog code

#### Desired Merge Strategy

- squash

#### Release Notes
-

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -458,7 +458,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
 
   def addFormalStatement(formals: mutable.Map[Expression, ArrayBuffer[Seq[Any]]],
                                  clk: Expression, en: Expression,
-                                 stmt: Seq[Any], info: Info): Unit = {
+                                 stmt: Seq[Any], info: Info, msg: StringLit): Unit = {
     throw EmitterException("Cannot emit verification statements in Verilog" +
       "(2001). Use the SystemVerilog emitter instead.")
   }
@@ -827,8 +827,8 @@ class VerilogEmitter extends SeqTransform with Emitter {
       lines += Seq("`endif // SYNTHESIS")
     }
 
-    def addFormal(clk: Expression, en: Expression, stmt: Seq[Any], info: Info) = {
-      addFormalStatement(formals, clk, en, stmt, info)
+    def addFormal(clk: Expression, en: Expression, stmt: Seq[Any], info: Info, msg: StringLit): Unit = {
+      addFormalStatement(formals, clk, en, stmt, info, msg)
     }
 
     def formalStatement(op: Formal.Value, cond: Expression): Seq[Any] = {
@@ -939,7 +939,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
         case sx: Print =>
           simulate(sx.clk, sx.en, printf(sx.string, sx.args), Some("PRINTF_COND"), sx.info)
         case sx: Verification =>
-          addFormal(sx.clk, sx.en, formalStatement(sx.op, sx.pred), sx.info)
+          addFormal(sx.clk, sx.en, formalStatement(sx.op, sx.pred), sx.info, sx.msg)
         // If we are emitting an Attach, it must not have been removable in VerilogPrep
         case sx: Attach =>
           // For Synthesis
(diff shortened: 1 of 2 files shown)


ekiwi-sifive:
There are cooler things we could try that would actually display the message once an assertion fails, but this simple change at least gives people a better idea of what the assertion is supposed to do if they look at the generated Verilog.

src/main/scala/firrtl/Emitter.scala
```diff
@@ -1258,8 +1258,9 @@ class SystemVerilogEmitter extends VerilogEmitter {
 
   override def addFormalStatement(formals: mutable.Map[Expression, ArrayBuffer[Seq[Any]]],
                                   clk: Expression, en: Expression,
-                                  stmt: Seq[Any], info: Info): Unit = {
+                                  stmt: Seq[Any], info: Info, msg: StringLit): Unit = {
     val lines = formals.getOrElseUpdate(clk, ArrayBuffer[Seq[Any]]())
+    lines += Seq("// ", msg.serialize)
     lines += Seq("if (", en, ") begin")
```
jackkoenig, line 1264: I'm about to add source locator information to always blocks so I'm wondering how the combination should look:

#### This PR

```verilog
always @(posedge clock) begin
  // assume input is 0xAA
  if (1'h1) begin
    assume(inputEquals0xAA);
  end
end
```

#### My idea for my PR

```verilog
always @(posedge clock) begin
  if (1'h1) begin // @[MyFile.scala 20:14]
    assume(inputEquals0xAA); // @[MyFile.scala 21:16]
  end
end
```

#### Combo?

```verilog
always @(posedge clock) begin
  // assume input is 0xAA
  if (1'h1) begin // @[MyFile.scala 20:14]
    assume(inputEquals0xAA); // @[MyFile.scala 21:16]
  end
end
```

jackkoenig: Also should the message comment maybe go inside the if?

ekiwi-sifive: > I'm about to add source locator information to always blocks so I'm wondering how the combination should look:

I think your suggested combo looks good.



> Also should the message comment maybe go inside the if?

I don't think so since the `en` signal which feeds into the `if` is part of the assertion.

jackkoenig: Oh fair, then what should I do with the source locator? I was assuming that was an `if` from a `when` in my example. I think it is more obvious to the user to put it on the `assume` itself but I don't feel that strongly

ekiwi-sifive: > Oh fair, then what should I do with the source locator?

Is it excessive to just add it to both lines? This way people won't complain about missing source locators.

seldridge:
Two questions: 

1) Any reason to not go with `$display` or `$error` output with this PR?
2) The comment mechanism is pretty close to a `DocStringAnnotation`. Can this be unified with that eventually?

I don't know much about SVA, but the [Dolous website shows `assert` being written this way](https://www.doulos.com/knowhow/sysverilog/tutorial/assertions/). It seems pretty straightforward to emit a `else $error(Assertion with message '<message>' was false)` or something. Usage of `$display` I think is wrong and would print everytime the assertion doesn't fail?

(2) is obviously blocked by a solution to target verification statements. I'm cool with getting a solution in before that, but it would be good to have some thought about if we want to unify this and a plan for how to unify it. Generally I'm in favor of a single mechanism to add comments.

ekiwi-sifive:
>     1. Any reason to not go with `$display` or `$error` output with this PR?

This was supposed to be the quick & easy PR that improves things a bit with no generated code changes. As I mentioned in my first comment, there are better things we can do, but they needs more considerations and testing with different tools.

ekiwi-sifive:
Also: I don't think `$display` and `$error` are going to work with the formal tool.

seldridge:
Good point. It's definitely way more informative than the current status.

We do have runway before 1.4.0. Do you think there's time to get something like https://github.com/freechipsproject/firrtl/pull/1725 in place for the `SystemVerilogEmitter` to provide some parity before then?

ekiwi-sifive:
> We do have runway before 1.4.0. Do you think there's time to get something like #1725 in place for the `SystemVerilogEmitter` to provide some parity before then?

I don't think that display will work for any formal tools and we will probably have to take some tool specific approaches:
* for open source `yosys` I think putting the assertion message at least into a Verilog comment is as good as it gets
* for commercial tools we can switch to emitting a named property and a print on assertion failure once we are able to name statements (which is blocked on https://github.com/freechipsproject/firrtl/issues/1679)

For simulation tools you can probably just run the pass from #1725 and it should work as expected.

PR #1624, opened 2020-05-15
Make Utils.expandRef actually return intermediate expressions per API docs

albert-magyar:
**Type of improvement:** bug fix
**API impact:** changes behavior of `Utils.expandRef` to match description
**Backend code-generation impact:** Will cause `DedupModules` to rename intermediate targets for composite types
**Release notes:**
The `expandRef` method in `Utils` now correctly returns possible intermediate references to subfields or subindices of composite types.

### Contributor Checklist
- [x] Did you add Scaladoc to every public function/method?
- [x] Did you update the FIRRTL spec to include every new feature/behavior?
- [x] Did you add at least one test demonstrating the PR?
- [x] Did you delete any extraneous printlns/debugging code?
- [x] Did you specify the type of improvement?
- [x] Did you state the API impact?
- [x] Did you specify the code generation impact?
- [] Did you request a desired merge strategy?
- [x] Did you add text to be included in the Release Notes for this change?

### Reviewer Checklist (only modified by reviewer)
- [ ] Did you add the appropriate labels?
- [ ] Did you mark the proper milestone (1.2.x, 1.3.0, 1.4.0) ?
- [ ] Did you review?
- [ ] Did you check whether all relevant Contributor checkboxes have been checked?
- [ ] Did you mark as `Please Merge`?

PR #1713, opened 2020-06-24
Don't Dedup modules if it would change semantics

### LINKED ISSUE #1703: Interaction between Chisel._, DedupModules, and ExpandConnects can result in dropped connections
### Checklist

When connecting things with Chisel._, it's possible that DedupModules and ExpandConnects interact badly with eachother to result in dropped wires (wires tied off to 0 instead of being connected as specified by the user).

The Chisel user's intended connectivity should not be changed by DedupModules and wires should end up connected properly in the final verilog.

OR an error should be thrown by Chisel/FIRRTL alerting user that this is not going to work in this case.

### Steps to Reproduce
This was originally discovered in implementing. https://github.com/chipsalliance/rocket-chip/pull/2528

Thanks to @jackkoenig for getting this cut-down example coded up:

https://scastie.scala-lang.org/yJUyW088TVOPhQpLlSX1Ng

Code duplicated below:

```scala
/* FIRRTL Dedup bug

This illustrates a problem with Dedup where Modules using <- (partial connect) from
compatibility mode can have their semantics changed by Dedup. Because partial connect
uses by-name connections, name-agnostic deduplication can change the names of the ports
thus changing the semantics of connections to those ports.

You can see this bug by changing the import between Chisel._ and chisel3._
Under chisel3._ things work correctly (bulk connections are flattened before FIRRTL)
But if you use Chisel._, the connection semantics change and the connections to bar
get dropped (defaulted to 0 because Chisel._)

*/
import Chisel._
//import chisel3._
import chisel3.stage.ChiselStage

class Foo extends Bundle {
  val foo = Input(Bool())
  val fuzz = Output(Bool())
}

class Bar extends Bundle {
  val bar = Input(Bool())
  val buzz = Output(Bool())
}

class FooModule extends Module {
  val io = IO(new Foo)
  io.fuzz := io.foo
}

class BarModule extends Module {
  val io = IO(new Bar)
  io.buzz := io.bar
}

class FooAndBarModule extends Module {
  
  val io = IO(new Bundle {
    val foo = new Foo
    val bar = new Bar
  })
  val foo = Module(new FooModule)
  val bar = Module(new BarModule)
  io.foo <> foo.io
  io.bar <> bar.io

}

println((new ChiselStage).emitVerilog(new FooAndBarModule()))
```

Here is the bad result:
```
module FooModule(
  input   io_foo,
  output  io_fuzz
);
  assign io_fuzz = io_foo; // @[main.scala 32:11]
endmodule
module FooAndBarModule(
  input   clock,
  input   reset,
  input   io_foo_foo,
  output  io_foo_fuzz,
  input   io_bar_bar,
  output  io_bar_buzz
);
  wire  foo_io_foo; // @[main.scala 46:19]
  wire  foo_io_fuzz; // @[main.scala 46:19]
  wire  bar_io_foo; // @[main.scala 47:19]
  wire  bar_io_fuzz; // @[main.scala 47:19]
  FooModule foo ( // @[main.scala 46:19]
    .io_foo(foo_io_foo),
    .io_fuzz(foo_io_fuzz)
  );
  FooModule bar ( // @[main.scala 47:19]
    .io_foo(bar_io_foo),
    .io_fuzz(bar_io_fuzz)
  );
  assign io_foo_fuzz = foo_io_fuzz; // @[main.scala 48:10]
  assign io_bar_buzz = 1'h0;
  assign foo_io_foo = io_foo_foo; // @[main.scala 48:10]
  assign bar_io_foo = 1'h0;
endmodule
```

### Your environment

- Chisel Versions: ` *   21ea734 - (9 weeks ago) Merge branch 'master' into 3.3.x - Jim Lawson`
- firrtl Version: `| * e599dba - (10 days ago) delete usages of toSet for determinism (#1686) (#1688) - mergify[bot]`

### External Information

https://github.com/chipsalliance/rocket-chip/pull/2528, https://github.com/chipsalliance/rocket-chip/pull/2487

jackkoenig:
If a module has ports of type Bundle that are used in aggregate
connections in parent modules, Dedup cannot change the names of the
fields of the Bundle or it would change the semantics of the connection.
Dedup now detects this case and refrains from agnostifying the ports of
such modules to prevent this issue.

Fixes #1703

@chick @ucbjrl I am very much hoping this can make it into the `1.3.2` release. It's fine that it missed the `1.4-SNAPSHOT` and `1.2.x` latest.

### Contributor Checklist

#### Type of Improvement

  - bug fix

#### API Impact

This can prevent deduplication of modules where FIRRTL was formerly (incorrectly) deduping them, but due to a second bug it would coincidentally undo the bug. If you consider the main test case but using `<=` instead of `<-`:
```
circuit FooAndBarModule :
  module FooModule :
    output io : {flip foo : UInt<1>, fuzz : UInt<1>}
    io.fuzz <= io.foo

  module BarModule :
    output io : {flip bar : UInt<1>, buzz : UInt<1>}
    io.buzz <= io.bar

  module FooAndBarModule :
    output io : {foo : {flip foo : UInt<1>, fuzz : UInt<1>}, bar : {flip bar : UInt<1>, buzz : UInt<1>}}

    inst foo of FooModule
    inst bar of BarModule
    io.foo <= foo.io       ; change from <-
    io.bar <= bar.io       ; changed from <-
```
It is incorrect for Dedup to Dedup `FooModule` and `BarModule` because afterwards these bulk connections are illegal. However, nothing checks this in normal compilation flows and then `ExpandConnects` assumes everything is correct and expands the connection in a way that is coincidentally correct.

Because Chisel does not emit aggregate `<=`, it is unlikely anyone will notice this.

No impact

#### Desired Merge Strategy

#### Release Notes

Fix #1703 

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/transforms/Dedup.scala
+++ b/src/main/scala/firrtl/transforms/Dedup.scala
@@ -5,14 +5,17 @@ package transforms
 
 import firrtl.ir._
 import firrtl.Mappers._
+import firrtl.traversals.Foreachers._
 import firrtl.analyses.InstanceGraph
 import firrtl.annotations._
 import firrtl.passes.{InferTypes, MemPortUtils}
-import firrtl.Utils.throwInternalError
+import firrtl.Utils.{kind, splitRef, throwInternalError}
 import firrtl.annotations.transforms.DupedResult
 import firrtl.annotations.TargetToken.{OfModule, Instance}
 import firrtl.options.{HasShellOptions, ShellOption}
 import logger.LazyLogging
+
+import scala.annotation.tailrec
 
 // Datastructures
 import scala.collection.mutable
@@ -446,6 +449,48 @@ object DedupModules extends LazyLogging {
     changeInternals({n => n}, retype, {i => i}, renameOfModule)(module)
   }
 
+  @tailrec
+  private def hasBundleType(tpe: Type): Boolean = tpe match {
+    case _: BundleType => true
+    case _: GroundType => false
+    case VectorType(t, _) => hasBundleType(t)
+  }
+
+  // Find modules that should not have their ports agnostified to avoid bug in
+  // https://github.com/freechipsproject/firrtl/issues/1703
+  // Marks modules that have a port of BundleType that are connected via an aggregate connect or
+  // partial connect in an instantiating parent
+  // Order of modules does not matter
+  private def modsToNotAgnostifyPorts(modules: Seq[DefModule]): Set[String] = {
+    val dontDedup = mutable.HashSet.empty[String]
+    def onModule(mod: DefModule): Unit = {
+      val instToModule = mutable.HashMap.empty[String, String]
+      def markAggregatePorts(expr: Expression): Unit = {
+        if (kind(expr) == InstanceKind && hasBundleType(expr.tpe)) {
+          val (WRef(inst, _, _, _), _) = splitRef(expr)
+          dontDedup += instToModule(inst)
+        }
+      }
+      def onStmt(stmt: Statement): Unit = {
+        stmt.foreach(onStmt)
+        stmt match {
+          case inst: DefInstance =>
+            instToModule(inst.name) = inst.module
+          case Connect(_, lhs, rhs) =>
+            markAggregatePorts(lhs)
+            markAggregatePorts(rhs)
+          case PartialConnect(_, lhs, rhs) =>
+            markAggregatePorts(lhs)
+            markAggregatePorts(rhs)
+          case _ =>
+        }
+      }
+      mod.foreach(onStmt)
+    }
+    modules.foreach(onModule)
+    dontDedup.toSet
+  }
+
   //scalastyle:off
   /** Returns
     *  1) map of tag to all matching module names,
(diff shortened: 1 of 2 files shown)


jackkoenig:
@albert-magyar I unrolled the bulk connects in https://github.com/freechipsproject/firrtl/pull/1624 (which is going to make backporting a pain), but wanted you to sanity check that I didn't break the thing your test is testing.

jackkoenig:
Backporting note, need to rename `DefInstance` => `WDefInstance` in `modsToNotAgnostifyPorts`

albert-magyar:
Seems like a great catch. The retyping can definitely do some subtle things. I can look at this in more detail tomorrow, but would this maybe be a time to consider whether retyping is worth it in general?

jackkoenig:
> would this maybe be a time to consider whether retyping is worth it in general?

I absolutely agree. I see this as a bugfix that does not change behavior, but for 1.4.0 I think we should reconsider the rules for Dedup. The current rules are to support the `GroupAndDedup` feature which takes advantage of the agnostification (including retyping), but I'm now thinking that should be an opt-in addition to `Dedup` and not the default behavior of the transform.

albert-magyar:
@jackkoenig, based on my quick look, this also appears to work for the analogous case where you do a `wire_same_type_as_inst <= inst` bulk connect.

PR #1724, opened 2020-06-25
Specify execution order of side-effect-having statements

albert-magyar:
**Type of improvement:** spec
**API impact:** none
**Release notes:**
The execution ordering rules maintained by the current FIRRTL compiler for side-effect-having statements (print, stop, verification statements) have been codified in the spec.

### Contributor Checklist

- [ ] Did you review?


ekiwi:
Are you sure we should not enforce that `printf` statements are executed before all others in order to ensure that all prints happen before the simulation exits?

albert-magyar:
> Are you sure we should not enforce that `printf` statements are executed before all others in order to ensure that all prints happen before the simulation exits?

That is not what is currently done in the compiler. "All prints happen first" would presumably apply across module boundaries, which would be very difficult to enforce in Verilog output.

spec/spec.tex
```diff
@@ -977,11 +977,17 @@ \subsection{Instances}\label{instances}
...
 The stop statement is used to halt simulations of the circuit. Backends are free to generate hardware to stop a running circuit for the purpose of debugging, but this is not required by the FIRRTL specification.
 
 A stop statement requires a clock signal, a halt condition signal that has a single bit unsigned integer type, and an integer exit code.
 
+For clocked statements that have side effects in the environment (stop, print, and verification
+statements), the order of execution of any such statements that are triggered on the same clock edge
+is determined by their syntactic order in the enclosing module. The order of execution of clocked,
+side-effect-having statements in different modules or with different clocks that trigger
+concurrently is undefined.
+
```
jackkoenig, line 990: This makes sense to me, but I'd like @azidar (on vacation so we'll have to wait) and @aswaterman to take a look.

I believe we do preserve the order of all simulation statements in the implementation, but we need to make sure we're doing this.

albert-magyar: The rationale is that the current (Chisel 3.3 + FIRRTL 1.3) implementation of asserts _already assumes_ that at least a partial ordering of simulation statements is maintained. Since there is no actual coupling of those two statements (especially after ExpandWhens), this feels like the weakest simple thing that will guarantee that. The idea here is that it's better have a violation of an explicit rule than an implicit assumption.

One could interpret the following sentence from section 5.3.1 in the spec as already implying the ordering from this addition, but the fact that it's listed with connection order makes it less clear than it could be.
> Ordering of statements is significant in a statement group. Intuitively, during elaboration, statements execute in order, and the effects of later statements take precedence over earlier ones.

albert-magyar: But I am certainly in no hurry to merge this, so whoever can take a look at it, that would be great.

ekiwi-sifive: @albert-magyar Do I understand this correctly in that if you have a print, a stop and another print that are all triggered together, we should see the output of the first print, but never the output of the second print since the simulation has been stopped?

edwardcwang:
Taking a step back... none of the "side-effect-having" statements described in this PR should be (hardware-)synthesizable, correct? In other words, it should not affect what logic gates are generated.

ekiwi-sifive:
> Taking a step back... none of the "side-effect-having" statements described in this PR should be (hardware-)synthesizable, correct? In other words, it should not affect what logic gates are generated.

Yes. This is about well defined simulation behavior.

spec/spec.tex
albert-magyar: If they appear in that order, yes.

PR #1725, opened 2020-06-25
Enable flow from assert to print + stop

albert-magyar:
**Type of improvement:** enhancement
**API impact:** none
**Backend code-generation impact:** allows code using the new `assert` statement from #1653 to map to the same conditional print-and-stop. This allows us to maintain the flow of sending Chisel assertions to `$display`/`$finish` blocks that will work in IEEE 1364 Verilog.
**Desired merge strategy:** rebase
**Release notes:**
The assert statement added in #1653 can be mapped to conditional prints and stops when targeting Verilog to aid compatibility with strict IEEE 1364 tools.

There is another question remaining: currently, nonzero-argument `Stop`s turn into `$fatal`, which is not part of IEEE 1364 Verilog. I can extend the Verilog emitter to use `$finish` for Verilog and `$fatal` for SystemVerilog to be stricter, if we feel that's warranted.

This depends on the ordering specified in #1724 (as does the current Chisel 3.3 + FIRRTL 1.3 flow).

The print / stop conditions leave in an unsplit expression, which is guaranteed to always be fine because it exclusively uses 1-bit boolean ops. I may add a comment to note that this doesn't invalidate `SplitExpressions` and specifically note that that case does not violate the postcondition of `SplitExpressions` as a comment in `SplitExpressions.scala`.

### Contributor Checklist

- [x] Did you review?

### DIFF
--- a/src/main/scala/firrtl/transforms/formal/ConvertAsserts.scala
+++ b/src/main/scala/firrtl/transforms/formal/ConvertAsserts.scala
@@ -1,0 +1,39 @@
+// See LICENSE for license details.
+
+package firrtl.transforms.formal
+
+import firrtl._
+import firrtl.ir._
+import firrtl.options.Dependency
+
+/** Convert Asserts
+  *
+  * Replaces all Assert nodes with a gated print-and-stop. This effectively
+  * emulates the assert for IEEE 1364 Verilog.
+  */
+object ConvertAsserts extends Transform with DependencyAPIMigration {
+  override def prerequisites = Nil
+  override def optionalPrerequisites = Nil
+  override def optionalPrerequisiteOf = Seq(
+    Dependency[VerilogEmitter],
+    Dependency[MinimumVerilogEmitter],
+    Dependency[RemoveVerificationStatements])
+
+  override def invalidates(a: Transform): Boolean = false
+
+  def convertAsserts(stmt: Statement): Statement = stmt match {
+    case Verification(Formal.Assert, i, clk, pred, en, msg) =>
+      val nPred = DoPrim(PrimOps.Not, Seq(pred), Nil, pred.tpe)
+      val gatedNPred = DoPrim(PrimOps.And, Seq(nPred, en), Nil, pred.tpe)
+      val stop = Stop(i, 1, clk, gatedNPred)
+      msg match {
+        case StringLit("") => stop
+        case _ => Block(Print(i, msg, Nil, clk, gatedNPred), stop)
+      }
+    case s => s.mapStmt(convertAsserts)
+  }
+
+  def execute(state: CircuitState): CircuitState = {
+    state.copy(circuit = state.circuit.mapModule(m => m.mapStmt(convertAsserts)))
+  }
+}
--- a/src/main/scala/firrtl/transforms/formal/RemoveVerificationStatements.scala
+++ b/src/main/scala/firrtl/transforms/formal/RemoveVerificationStatements.scala
@@ -19,7 +19,7 @@ class RemoveVerificationStatements extends Transform
   with PreservesAll[Transform] {
 
   override def prerequisites: Seq[TransformDependency] = Seq.empty
-  override def optionalPrerequisites: Seq[TransformDependency] = Seq.empty
+  override def optionalPrerequisites: Seq[TransformDependency] = Seq(Dependency(ConvertAsserts))
   override def optionalPrerequisiteOf: Seq[TransformDependency] =
     Seq( Dependency[VerilogEmitter],
       Dependency[MinimumVerilogEmitter])
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -16,7 +16,7 @@ import Utils._
 import MemPortUtils.{memPortField, memType}
 import firrtl.options.{Dependency, HasShellOptions, PhaseException, ShellOption, Unserializable}
 import firrtl.stage.{RunFirrtlTransformAnnotation, TransformManager}
-import firrtl.transforms.formal.RemoveVerificationStatements
+import firrtl.transforms.formal.{RemoveVerificationStatements, ConvertAsserts}
 // Datastructures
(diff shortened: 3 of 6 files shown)


src/main/scala/firrtl/transforms/formal/ConvertAsserts.scala
```diff
@@ -0,0 +1,36 @@
...
+    Dependency[MinimumVerilogEmitter],
+    Dependency[RemoveVerificationStatements])
+
+  override def invalidates(a: Transform): Boolean = false
+
+  def convertAsserts(stmt: Statement): Statement = stmt match {
+    case Verification(Formal.Assert, i, clk, pred, en, msg) =>
+      val nPred = DoPrim(PrimOps.Not, Seq(pred), Nil, pred.tpe)
+      val gatedNPred = DoPrim(PrimOps.And, Seq(nPred, en), Nil, pred.tpe)
+      Block(Print(i, msg, Nil, clk, gatedNPred), Stop(i, 1, clk, gatedNPred))
```
ekiwi, line 29: Should the `Print` be omitted when the assert message is empty?

albert-magyar: The message is mandatory, but I could certainly add a special case for the empty string. That seems like a sensible thing to add, so I'll put that in unless anyone objects.

albert-magyar: Added

ekiwi: Thanks! This way we can fall back to emitting a printf and an assert with an empty message in the Chisel frontend in order to implement a backwards compatible assert(x, "%d", y)

src/main/scala/firrtl/transforms/formal/ConvertAsserts.scala
```diff
@@ -0,0 +1,40 @@
+// See LICENSE for license details.
+
+package firrtl.transforms.formal
+
+import firrtl._
+import firrtl.ir._
+import firrtl.options.Dependency
+
+/**
+  * Convert Asserts
```
seldridge, line 10: Nit:
```suggestion
/** Convert Asserts
```

src/main/scala/firrtl/transforms/formal/RemoveVerificationStatements.scala
```diff
@@ -19,7 +19,7 @@ class RemoveVerificationStatements extends Transform
   with PreservesAll[Transform] {
 
   override def prerequisites: Seq[TransformDependency] = Seq.empty
-  override def optionalPrerequisites: Seq[TransformDependency] = Seq.empty
+  override def optionalPrerequisites: Seq[TransformDependency] = Seq(Dependency(ConvertAsserts))
```
seldridge, line 22: What's your thinking here for specifying both `optionalPrerequisiteOf` in `ConvertAsserts` and an `optionalPrerequistes` here?

I'm not opposed to it, just wondering if you've identified some benefit/good pattern as I would think one or the other is sufficient.

src/main/scala/firrtl/transforms/formal/ConvertAsserts.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  def convertAsserts(stmt: Statement): Statement = stmt match {
+    case Verification(Formal.Assert, i, clk, pred, en, msg) =>
+      val nPred = DoPrim(PrimOps.Not, Seq(pred), Nil, pred.tpe)
+      val gatedNPred = DoPrim(PrimOps.And, Seq(nPred, en), Nil, pred.tpe)
+      val stop = Stop(i, 1, clk, gatedNPred)
+      msg match {
+        case StringLit("") => stop
+        case _ => Block(Print(i, msg, Nil, clk, gatedNPred), stop)
+      }
```
seldridge, line 33: Should the message be modified to include source locator information similar to Chisel's current output of:

```scala
s"""|Assertion failed: $msg
    |    at $line""".stripMargin
```

Alternatively, maybe it's better to do this in the `chisel3` side?

(And this looks good with @ekiwi's suggestion to ignore the message if it doesn't exist.)

seldridge:
> There is another question remaining: currently, nonzero-argument Stops turn into $fatal, which is not part of IEEE 1364 Verilog. I can extend the Verilog emitter to use $finish for Verilog and $fatal for SystemVerilog to be stricter, if we feel that's warranted.

Great idea. I feel it's warranted. If we have multiple emitters, it seems sound to choose a standard and stick to it. At some point I thought we were using 1364-2001, though you may know better? Separate PR in terms of logistics.

src/main/scala/firrtl/transforms/formal/RemoveVerificationStatements.scala
albert-magyar: I actually am generally not in favor of this pattern (I would prefer to just use `optionalPrerequisites`), but the "two-sided" listing seems to have been adopted for other similar recent additions, so I was doing it for what I thought was maybe consistency.

src/main/scala/firrtl/transforms/formal/ConvertAsserts.scala
albert-magyar: I think it might be better to do it on the chisel side, otherwise the messages would be more informative in Verilog than in SystemVerilog

seldridge: That sounds good to me!

src/main/scala/firrtl/transforms/formal/RemoveVerificationStatements.scala
seldridge: I agree. I'm in favor of just an `optionalPrerequisites` specification. No harm done and it's consistent with existing work. Fine the way it is, then.

Auditing the transforms to remove dependency over-specification can be done at a later point. Part of that might be good to come up with some "style guide" for transform writers. Future work...

PR #1727, opened 2020-06-27
Avoid repeated inlining in FlattenRegUpdate

jackkoenig:
When-else structure can lead to the same complex mux structure being
the default on several branches in register update logic. When these are
inlined, it can lead to artifical unreachable branches that show up as
coverage holes in coverage of the emitted Verilog. This commit changes
the inlining logic to prevent inlining any reference expression that
shows up multiple times because this is a common indicator of the
problematic case.

### Contributor Checklist

#### Type of Improvement

  - backend code generation  

#### API Impact

No change to API

This changes code gen by no longer inlining muxes that would appear multiple times in register update logic. See detailed example in follow on comment.

#### Desired Merge Strategy

#### Release Notes

Improved Verilog emission logic to no longer inline the same `if-else` block in register update logic if it would appear more than once. This fixes some cases of pathological Verilog emission and has the added benefit of improving standard Verilog coverage metrics on the generated code.

- [ ] Did you review?


jackkoenig:
Example:

```
circuit Foo :
  module Foo :
    input clock : Clock
    input reset : UInt<1>
    output io : { flip in : UInt<8>, flip a : UInt<1>, flip b : UInt<1>, flip c : UInt<1>, flip d : UInt<1>, out : UInt<8>}

    reg reg : UInt<8>, clock with :
      reset => (reset, UInt<8>("h0"))
    when io.a :
      reg <= io.in
      node _T = eq(io.b, UInt<1>("h0"))
      when _T :
        reg <= UInt<2>("h2")
    when io.b :
      when io.c :
        when io.d :
          reg <= UInt<2>("h2")
    io.out <= reg
```

#### Without this change

```verilog
  reg [7:0] reg_;
  wire  _T = ~io_b;
  assign io_out = reg_;
  always @(posedge clock) begin
    if (reset) begin
      reg_ <= 8'h0;
    end else if (io_b) begin
      if (io_c) begin
        if (io_d) begin
          reg_ <= 8'h2;
        end else if (io_a) begin
          if (_T) begin // Impossible because _T = ~io_b
            reg_ <= 8'h2;
          end else begin
            reg_ <= io_in;
          end
        end
      end else if (io_a) begin
        if (_T) begin // Impossible because _T = ~io_b
          reg_ <= 8'h2;
        end else begin
          reg_ <= io_in;
        end
      end
    end else if (io_a) begin
      if (_T) begin // Redundant because _T = ~io_b
        reg_ <= 8'h2;
      end else begin
        reg_ <= io_in;
      end
    end
  end
```

#### With this change
```verilog
  reg [7:0] reg_;
  wire  _T = ~io_b;
  wire [7:0] _GEN_0 = _T ? 8'h2 : io_in;
  wire [7:0] _GEN_1 = io_a ? _GEN_0 : reg_;
  assign io_out = reg_;
  always @(posedge clock) begin
    if (reset) begin
      reg_ <= 8'h0;
    end else if (io_b) begin
      if (io_c) begin
        if (io_d) begin
          reg_ <= 8'h2;
        end else begin
          reg_ <= _GEN_1;
        end
      end else begin
        reg_ <= _GEN_1;
      end
    end else begin
      reg_ <= _GEN_1;
    end
  end
```

jackkoenig:
The failure is ICache equivalence check running out of memory. This change is *probably* correct.

jackkoenig:
Having run some analysis, this vastly improves standard coverage metrics of FIRRTL-generated Verilog:

Coverage improvement for a complex, multi-core design without and with this change

firrtl  |  Line  |  Toggle | Branch
---- | ---- | ---- | ----
clean |   88.70% | 88.78% | 90.08%
this PR | 97.54% | 88.76% |  95.18%

Line and branch coverage improvements illustrate how many of those unreachable branches are gone. This is a massive improvement for use of the Verilog in industry-standard verification flows.

jackkoenig:
Unfortunately, the source locators in register update logic PR conflicts heavily with this so I'll need to do some intense rebasing. But I now believe we should get this in to 1.4.

src/test/scala/firrtlTests/RegisterUpdateSpec.scala
```diff
@@ -0,0 +1,66 @@
...
+      |    r <= UInt(2)
+      |io.out <= r""".stripMargin
+    )
+    val verilog = result.getEmittedCircuit.value
+    result  shouldNot containLine ("r <= io_in;")
+    verilog shouldNot include     ("if (io_a) begin")
+    result  should    containLine ("r <= _GEN_0;")
+  }
+
+  it should "not let duplicate subtrees on one register affect another" in {
```
ekiwi-sifive, line 39: Do you think you could add a test that contains the expected firrtl after running the `FlattenRegUpdate` pass?
This would help by serving as a visual example of what `FlattenRegUpdate` does.

Similar to this: https://github.com/freechipsproject/firrtl/blob/master/src/test/scala/firrtlTests/WiringTests.scala#L71

src/main/scala/firrtl/transforms/FlattenRegUpdate.scala
```diff
@@ -58,38 +58,61 @@ object FlattenRegUpdate {
...
-    // The threshold is empirical but ample.
-    val flattenThreshold = 4
-    val numTimesFlattened = mutable.HashMap[Mux, Int]()
-    def canFlatten(m: Mux): Boolean = {
-      val n = numTimesFlattened.getOrElse(m, 0)
-      numTimesFlattened(m) = n + 1
-      n < flattenThreshold
-    }
+    // We want to flatten Mux trees for reg updates into if-trees for improved
+    // QoR for conditional updates. However, when-otherwise structure often
```
ekiwi-sifive, line 62: It is not clear to me what you mean by `when-otherwise structure`. Maybe you could rephrase this comment to require less background information on how register updates are emitted.

jackkoenig: I'll try to rephrase, it's not about how their emitted, it's about what last-connect semantics mean, given:
```scala
when (io.a) {
  reg  := foo
  when (!io.b)
    reg := bar
  }
}
// Semantically, at this point the "fallback" value for reg is:
//   mux(io.a, mux(!io.b, bar, foo), reg)
when (io.b) {
  when (io.c) {
    when (io.d) {
      reg := fizz
    }
  }
}
```
The problem being solved is that the 2nd when structure has 3 "holes" where the complex fallback currently (before this PR) gets inlined. The improvement in this PR is detecting this case and not inlining it.

src/test/scala/firrtlTests/RegisterUpdateSpec.scala
jackkoenig: I think I can, it's a little more annoying because the output of `FlattenRegUpdate` doesn't escape the VerilogEmitter, but I should be able to use a TransformManager to get the exactly output of it.

ekiwi-sifive:
Seems like `yosys` is running out of memory for the `ICache` equivalence check.

jackkoenig:
> Seems like yosys is running out of memory for the ICache equivalence check.

Yeah that tends to happen with any large change to the Verilog, I'll run locally and assuming it passes, waive the formal checks

src/test/scala/firrtlTests/RegisterUpdateSpec.scala
jackkoenig: Pushing tests in a sec, for posterity, I tried to use the nice tree matchers but it's a bit too verbose:
```scala
    fstate should containTree {
      case Connect(_,
             Reference("r2",_,_,_),
             Mux(Reference("io_b",_,_,_),
                 UIntLiteral(x, _),
                 Mux(Reference("io_a",_,_,_),
                     Reference("io_in",_,_,_),
                     Reference("r2",_,_,_),
                     _
                 ),
                 _
             )
            ) if x == 3 => true
    }
```
vs.
```scala
fstate should containLine ("""r2 <= mux(io_b, UInt<8>("h3"), mux(io_a, io_in, r2))""")
```

src/main/scala/firrtl/transforms/FlattenRegUpdate.scala
jackkoenig: Updated with an example, let me know what you think

src/test/scala/firrtlTests/RegisterUpdateSpec.scala
jackkoenig: Done.

jackkoenig:
This still ought to fail CI in ICache equivalence check, I'll push another commit with the ci skip once I run it manually.

src/main/scala/firrtl/transforms/FlattenRegUpdate.scala
```diff
@@ -58,38 +58,80 @@ object FlattenRegUpdate {
...
+    //
+    // _GEN_0 = mux(a, foo, r)
+    // _GEN_1 = mux(a, bar, _GEN_0)
+    // r <= mux(b, _GEN_1, _GEN_0)
+    //
+    // Inlining _GEN_0 into _GEN_1 would result in unreachable lines in the Verilog. While we could
+    // do some optimizations here, this is *not* really a problem, it's just that Verilog metrics
+    // are based on the assumption of human-written code and as such it results in unreachable
+    // lines. Simply not inlining avoids this issue and leaves the optimizations up to synthesis
+    // tools which do a great job here.
```
ekiwi-sifive, line 85: I like this explanation!

Is this related to the changes in https://github.com/freechipsproject/firrtl/pull/804 ?

jackkoenig:
The diff for the ICache illustrates the purpose of this PR really well I think:
```verilog
$ diff ICache.3113233.v ICache.7928ab2.v
304a305,307
>   wire [1:0] _GEN_93 = io_mem_0_a_ready ? 2'h2 : _GEN_91; // @[ICache.scala 169:29 ICache.scala 169:37]
>   wire [1:0] _GEN_94 = io_s2_kill ? 2'h0 : _GEN_93; // @[ICache.scala 170:25 ICache.scala 170:33]
>   wire [1:0] _GEN_95 = _T_850 ? _GEN_94 : _GEN_91; // @[Conditional.scala 29:59]
305a309,310
>   wire [1:0] _GEN_96 = io_mem_0_d_valid ? 2'h3 : _GEN_95; // @[ICache.scala 173:29 ICache.scala 173:37]
>   wire [1:0] _GEN_97 = _T_851 ? _GEN_96 : _GEN_95; // @[Conditional.scala 29:59]
440,491d444
<       end else if (_T_851) begin // @[Conditional.scala 29:59]
<         if (io_mem_0_d_valid) begin // @[ICache.scala 173:29]
<           state <= 2'h3; // @[ICache.scala 173:37]
<         end else if (_T_850) begin // @[Conditional.scala 29:59]
<           if (io_s2_kill) begin // @[ICache.scala 170:25]
<             state <= 2'h0; // @[ICache.scala 170:33]
<           end else if (io_mem_0_a_ready) begin // @[ICache.scala 169:29]
<             state <= 2'h2; // @[ICache.scala 169:37]
<           end else if (_T_848) begin // @[Conditional.scala 29:59]
<             if (s1_miss) begin // @[ICache.scala 165:22]
<               state <= 2'h1; // @[ICache.scala 165:30]
<             end
<           end
<         end else if (_T_848) begin // @[Conditional.scala 29:59]
<           if (s1_miss) begin // @[ICache.scala 165:22]
<             state <= 2'h1; // @[ICache.scala 165:30]
<           end
<         end
<       end else if (_T_850) begin // @[Conditional.scala 29:59]
<         if (io_s2_kill) begin // @[ICache.scala 170:25]
<           state <= 2'h0; // @[ICache.scala 170:33]
<         end else if (io_mem_0_a_ready) begin // @[ICache.scala 169:29]
<           state <= 2'h2; // @[ICache.scala 169:37]
<         end else if (_T_848) begin // @[Conditional.scala 29:59]
<           if (s1_miss) begin // @[ICache.scala 165:22]
<             state <= 2'h1; // @[ICache.scala 165:30]
<           end
<         end
<       end else if (_T_848) begin // @[Conditional.scala 29:59]
<         if (s1_miss) begin // @[ICache.scala 165:22]
<           state <= 2'h1; // @[ICache.scala 165:30]
<         end
<       end
<     end else if (_T_851) begin // @[Conditional.scala 29:59]
<       if (io_mem_0_d_valid) begin // @[ICache.scala 173:29]
<         state <= 2'h3; // @[ICache.scala 173:37]
<       end else if (_T_850) begin // @[Conditional.scala 29:59]
<         if (io_s2_kill) begin // @[ICache.scala 170:25]
<           state <= 2'h0; // @[ICache.scala 170:33]
<         end else if (io_mem_0_a_ready) begin // @[ICache.scala 169:29]
<           state <= 2'h2; // @[ICache.scala 169:37]
<         end else begin
<           state <= _GEN_91;
<         end
<       end else begin
<         state <= _GEN_91;
<       end
<     end else if (_T_850) begin // @[Conditional.scala 29:59]
<       if (io_s2_kill) begin // @[ICache.scala 170:25]
<         state <= 2'h0; // @[ICache.scala 170:33]
<       end else if (io_mem_0_a_ready) begin // @[ICache.scala 169:29]
<         state <= 2'h2; // @[ICache.scala 169:37]
493c446
<         state <= _GEN_91;
---
>         state <= _GEN_97;
496c449
<       state <= _GEN_91;
---
>       state <= _GEN_97;
```

src/main/scala/firrtl/transforms/FlattenRegUpdate.scala
jackkoenig: They are indeed related. #804 was an earlier attempt to solve the same problem. I believe that this approach solves the actual issue better (unreachable branches), but #804 is useful for other reasons. It's a more compiler-y optimization attempt whereas this solution is more recognizing that the existing inlining behavior is actually a de-optimization (ie. undoing Common SubExpression Elimination). The fact that this approach works so well is more of a property of last connect semantics than something fundamental, but it's still as useful observation and heuristic.

I think #804 is still worth investigating, since the Verilog from the original example is still kind of ugly:
```verilog
  always @(posedge clock) begin
    if (reset) begin
      reg_ <= 8'h0;
    end else if (io_b) begin
      if (io_c) begin
        if (io_d) begin
          reg_ <= 8'h2;
        end else begin
          reg_ <= _GEN_1;
        end
      end else begin
        reg_ <= _GEN_1;
      end
    end else begin
      reg_ <= _GEN_1;
    end
  end
```
#804 would turn this into
```verilog
  always @(posedge clock) begin
    if (reset) begin
      reg_ <= 8'h0;
    end else if (io_b && io_c && io_d) begin
      reg_ <= 8'h2;
    end else begin
      reg_ <= _GEN_1;
    end
  end
```

jackkoenig:
It took 10GB for Yosys to prove the ICache equivalent. I suspect this is due to the memory structures in it and perhaps Yosys 0.8 doesn't/can't represent them very efficiently. Perhaps it's not the best test since it has to be waived so often...

PR #1728, opened 2020-06-27
Upd build flow. Bump vers

tampler:
Preparations for 2.13 support


tampler:
Not actual since #1729

PR #1729, opened 2020-06-27
Add Support for Scala 2.13

tampler:


tampler:
Code seems okay. Need to fix UniDoc and add 2.13 target on CI

tampler:
Hi Jack

Thanks for reviewing this. 

1. As for several versions of code, that's the standard way to handle sources, when API differs. I did that for a good reason, after getting compile errors just switching the compiler version back and forth. Scala 2.11 and 2.12 are equal, 2.13 is different for those files 

That's a standard practice in Scala and is done in many popular FP libs, like [ZIO](https://github.com/zio/zio/tree/master/core/shared/src/main)
So I suggest to leave this for now get back to this issue in another PR

2. Okay, I'll setup CI here for this project  and will see how to fix the issue with UniDoc

jackkoenig:
> 1. As for several versions of code, that's the standard way to handle sources, when API differs. I did that for a good reason, after getting compile errors just switching the compiler version back and forth. Scala 2.11 and 2.12 are equal, 2.13 is different for those files

> That's a standard practice in Scala and is done in many popular FP libs, like ZIO
So I suggest to leave this for now get back to this issue in another PR

I don't disagree with the basic principle, I'm only taking issue with the full duplication of files rather than the factoring out of code that needs to differ. For comparison, using the following commands for this analysis:
```bash
# To find Scala-version specific, non-test code
find . -name 'scala-2.1*' -exec find {} -name '*.scala' \; | grep -v test | xargs wc -l

# To find non-test Scala code
find . -name '*.scala' | grep -v test | xargs wc -l
```
| Project | Scala-version specific | All non-test Scala |
| --- | --- | --- |
| ZIO | 1666 | 47103 |
| This PR | 8355 | 36249 |

We are a somewhat smaller project with 5x the amount of Scala-version specific code.

From the perspective of version control, I can't see in this diff what changes were necessary in these duplicated files for `2.13`. I would much rather do more focused changes, and I'm happy to help with them. I also don't want to do it in follow on PRs because it makes the git history much harder to track; this PR creates a break in the commit history for duplicated files like the `Emitter` and that complicates tracing the history of things.

> 2. Okay, I'll setup CI here for this project and will see how to fix the issue with UniDoc

Thanks for looking at that. Just as a tip, I don't think the issue has anything to do with `UniDoc` itself. I think there is a bad interaction between the ANTLR and Protobuf SBT plugins. In the most recent commit, a different test failed: https://travis-ci.com/github/freechipsproject/firrtl/jobs/356083778

tampler:
Closing this to clean the commit history. Follow up is in #1737

PR #1732, opened 2020-07-01
Faster dedup instance graph

ekiwi-sifive:
This PR is based on top of https://github.com/freechipsproject/firrtl/pull/1731

By not hashing a full `ir.DefInstance` node which can contain costly `BundleType` child nodes, we instead hash a custom `case class` which contains the only two fields that are ever used with `InstanceGraph`: `name` and `module`.
In addition we cache the graph vertices in order to speed up multiple calls to `findInstancesInHierarchy`.

### Contributor Checklist

#### Type of Improvement
 - performance improvement

#### API Impact
- add `FastInstanceGraph` to `firrtl.analyses`
- eventually `InstanceGraph` should be deprecated

- none

#### Desired Merge Strategy
- squash

#### Release Notes

- New `InstanceKeyGraph` that improves API and performance of `InstanceGraph` in the `Dedup` pass. Planned to eventually supersede `InstanceGraph` for all use-cases.

- [x] Did you review?


seldridge:
Out of curiosity, did you try something like caching the `hashCode` of a `WDefInstance`? I only ask because this has been a trivial (and mandatory) requirement for large annotations and improving runtime. E.g., https://github.com/freechipsproject/chisel3/pull/1485.

This may not be applicable here, though.

ekiwi-sifive:
> Out of curiosity, did you try something like memoizing the `hashCode` of a `WDefInstance`?

No, but there is no reason to key on the type. It doesn't make sense from an API perspective and even when using WDefInstance as return value, the type is never used.

albert-magyar:
It would be good to just add new signatures to `InstanceGraph` where necessary and switch the underlying implementation to not hash `WDefInstances` -- this is also a functionality improvement.

ekiwi-sifive:
> It would be good to just add new signatures to `InstanceGraph` where necessary and switch the underlying implementation to not hash `WDefInstances` -- this is also a functionality improvement.

But I would still have to re-implement the old methods which expose `WDefInstance` in the signature on top of the new internal representation, right? Since we cannot just remove public methods without deprecating them first.

albert-magyar:
> > It would be good to just add new signatures to `InstanceGraph` where necessary and switch the underlying implementation to not hash `WDefInstances` -- this is also a functionality improvement.
> 
> But I would still have to re-implement the old methods which expose `WDefInstance` in the signature on top of the new internal representation, right? Since we cannot just remove public methods without deprecating them first.

Yes.

It would be fine to deprecate them, and the difference in behavior arising from using `(name, module)` alone would make no change in the current "intended" use cases and would actually just be strictly better.

Given that some are parameterless and therefore not amenable to overloading, and given that none of the new public methods here rely on _user-supplied_ `WDefInstances` as hashed keys, an alternative would be to use your better underlying implementation strategies and just put `UnknownType` into the `WDefInstances` that appear as values in the resulting data structures. This would basically give all the speedup.

ekiwi-sifive:
> Yes.
> 
> It would be fine to deprecate them, and the difference in behavior arising from using `(name, module)` alone would make no change in the current "intended" use cases and would actually just be strictly better.

I thought about that and decided it was much easier to just make a new class and slowly deprecate the old one. Re-implementing old methods on top of a new representation could introduce more bugs and make the code less readable.



> Given that some are parameterless and therefore not amenable to overloading, and given that none of the new public methods here rely on _user-supplied_ `WDefInstances` as hashed keys, an alternative would be to use your better underlying implementation strategies and just put `UnknownType` into the `WDefInstances` that appear as values in the resulting data structures. This would basically give all the speedup.

Yeah, if I was allowed to just create new `WDefInstance` nodes with an `UnknownType` this would simplify the additional code. However, it would still make the code a lot more complicated because right now most methods are a very thin wrapper around the internal data structure.

albert-magyar:
The `graph` in `FastInstanceGraph` also has a very different meaning from the one in `InstanceGraph` since it is a tree. This would result in repeated work for some transforms that rely on traversals of the `InstanceGraph` to schedule work.

Maybe a good name for this would be `FastInstanceTree`?

ekiwi-sifive:
> The `graph` in `FastInstanceGraph` also has a very different meaning from the one in `InstanceGraph` since it is a tree. This would result in repeated work for some transforms that rely on traversals of the `InstanceGraph` to schedule work.
> 
> Maybe a good name for this would be `FastInstanceTree`?

I mistakenly thought that we need to know where the `DefInstance` occurred, however it turns out this is not the case: https://github.com/freechipsproject/firrtl/blob/5a8dc6ab72122193e8451ec06841e830ea805cb1/src/test/scala/firrtlTests/analyses/InstanceGraphTests.scala#L272

I am going to add [the commit removing `parent`](https://github.com/freechipsproject/firrtl/pull/1733/commits/13cee41cb2992bd3c0432933dad48955f13546d0) to this PR.

jackkoenig:
🚨  bikeshedding alert 🚨 
I don't think "Fast" is the right name here because I think you'll basically the same performance by just overriding `hashCode` in `DefInstance`. I'm down with cleaning up the API and looking at `InstanceGraph`, I agree it is probably just easiest to create a new one with different types for all of the methods. Not sure of the best name though, maybe just `InstGraph` so people would rather type it anyway?

A side note while I have all of your attention - returning mutable types like `LinkedHashMap` is a huge wart (I know we're all aware). Whenever we add 2.13 support I think we should make the API there be `SeqMap`, ie. never publish a 2.13 artifact with the `LinkedHashMap` return type.

ekiwi-sifive:
> A side note while I have all of your attention - returning mutable types like `LinkedHashMap` is a huge wart (I know we're all aware). Whenever we add 2.13 support I think we should make the API there be `SeqMap`, ie. never publish a 2.13 artifact with the `LinkedHashMap` return type.

In this PR, `FastInstanceGraph` does not expose any `mutable.LinkedHashMap` in the public API. I think there are some other methods in the follow up PR that do, because I did not know if changing this to Map would destroy the order.

jackkoenig:
> changing this to Map would destroy the order.

It would. In Scala 2.12 land we return `mutable.LinkedHashMap`, or we copy-paste `VectorMap` from Scala 2.13 and return that, or we create a custom type* but I think just keeping it the same as the current `InstanceGraph` is fine as long as we fix it in the firrtl_2.13 artifact.

ekiwi-sifive:
Just out of curiosity: what is the order that `InstanceGraph` is trying to maintain and why? The use cases that I have seen so far do not seem to depend on the order.

jackkoenig:
I'd have to do some archeology to find the specific cases but there were places where we iterated on the Map structure to create resulting hardware which needs to be deterministic (thus ordered). This may have been more of a `DiGraph` issue than an `InstanceGraph` issue, I'm not sure.

jackkoenig:
The fear is it's really easy to screw this up, so protect users from themselves by ensuring the iteration order is defined.

I've actually just had an idea here though. If we're creating a new type, we could have a different return value that gives the functionality but hides the mutability:
```scala
def orderedMap[A, B](xs: (A, B)*): Function1[A, B] with Iterable[(A, B)] =
  collection.mutable.LinkedHashMap(xs:_*)
```
You lose some normal "Map-like" functions but it gets the point across

ekiwi-sifive:
@albert-magyar: I tried out your suggested solution to change the graph used by `InstanceGraph` to use a different `Key`. See: https://github.com/ekiwi-sifive/firrtl/commit/8756515e28db34741b2bc87388227021c7b195d1

However, this revealed that changing the type in `WDefInstance` to `UnknownType` changes the order of the `owners` map in `WiringUtils.sinksToSources` and that changes the firrtl produced by the `Wiring` transform. (It should still be correct, but some connect statements are switched).

@jackkoenig and I decided to go back to the initial approach of creating a new class and eventually deprecating `InstanceGraph`. I am happy for any suggestions on what to name this new class.

seldridge:
Hmm... does switching `WiringUtils.sinksToSources` from using `mutable.HashMap` to `mutable.LinkedHashMap` fix it?

ekiwi-sifive:
> Hmm... does switching `WiringUtils.sinksToSources` from using `mutable.HashMap` to `mutable.LinkedHashMap` fix it?

It is not going to restore the old behavior. But it should make things more stable in the future.

ekiwi-sifive:
Just ran a cold compile benchmark on my local machine with OpenJDK 11 and three repetitions:


  | design | max heap (MiB) | SD | runtime (s) | SD
-- | -- | -- | -- | -- | --
master | small | 2819 | 155.81 | 48.35 | 1.61
FastInstanceGraph | small | 2728 | 207.16 | 48.36 | 2.39
master | medium | 3893 | 103.02 | 111.02 | 3.35
FastInstanceGraph | medium | 3497 | 162.3 | 104.9 | 3.26

src/main/scala/firrtl/analyses/FastInstanceGraph.scala
```diff
@@ -0,0 +1,117 @@
...
+  }
+}
+
+
+object FastInstanceGraph {
+  /** We want to only use this untyped version as key because hashing bundle types is expensive
+    * @param name the name of the instance
+    * @param module the name of the module that is instantiated
+    */
+  case class Key(name: String, module: String) {
```
azidar, line 67: Can we rename this to `InstanceKey`? Otherwise just `Key` seems ambiguous.

src/main/scala/firrtl/analyses/FastInstanceGraph.scala
```diff
@@ -0,0 +1,117 @@
...
+import firrtl.ir
+
+import scala.collection.mutable
+
+/** A class representing the instance hierarchy of firrtl Circuit
+  * This is a faster version of the old `InstanceGraph` which only uses
+  * pairs of InstanceName and Module name as vertex keys instead of using WDefInstance
+  * which will hash the instance type causing some performance issues.
+  */
+class FastInstanceGraph(c: ir.Circuit) {
```
azidar, line 17: Maybe we could call it `InstanceKeyGraph`? I'm also down for `InstGraph`.

azidar:
Actually in doing my due diligence, we should probably add a test case to demonstrate this new capability. Can you please add some tests? Thanks!

ekiwi-sifive:
> Actually in doing my due diligence, we should probably add a test case to demonstrate this new capability. Can you please add some tests?

Good catch! I actually found a non-determinism bug!

PR #1734, opened 2020-07-02
Error on ExtModules w/ same defname, diff. ports

seldridge:
Adds a high form check to ensure that external modules that have the
same defname and same parameters also have exactly the same ports.

Note that this is a simple check that the ports match exactly. If ports are reordered, then this will still fail. Also, this if the same defnames have *different* parameters, then this check can rightly be a false negative. The underlying Verilog may be parameterizing the width of ports based on the parameters meaning that any check here is going to be insufficient unless the Verilog is examined.

This came up as part of a discussion here: https://github.com/freechipsproject/firrtl/pull/1731#issuecomment-652771527.

### Contributor Checklist

- [nope] Did you add Scaladoc to every public function/method?

#### Type of Improvement

- new feature/API                   

#### API Impact

Circuits containing extmodules with the same defname and different ports will result in the circuit being rejected.

None.

#### Desired Merge Strategy

- [ ] Did you review?


src/main/scala/firrtl/passes/CheckHighForm.scala
```diff
@@ -81,9 +83,24 @@ trait CheckHighFormLike { this: Pass =>
       m => errors.append(new ModuleNameNotUniqueException(m.info, m.name))
     }
 
+    val extmoduleCollidingPorts = c.modules.collect {
+      case a: ExtModule => a
+    }.groupBy(a => (a.defname, a.params)).mapValues {
```
albert-magyar, line 88: Why look at the params here?

seldridge: Since the parameters might be used to parameterize port widths, then it's only safe to check that extmodules with the same defname and params must have exactly the same ports.

An additional, complementary check would be that all defnames have the same port names, but widths are ignored.

Do you think it makes sense to add that?

albert-magyar: Makes sense to me; I guess I was assuming the second check (based on the idea that normal Verilog-2005 params can't add/remove ports), but I was mostly wondering if this could get expensive. I guess there aren't normally a ton of blackboxes, so the "hashing types" problem isn't too bad.

seldridge:
Rebased. I can update this with @albert-magyar's request for the second check, though.

ekiwi-sifive:
> Rebased. I can update this with @albert-magyar's request for the second check, though.

Sounds good. Maybe you could add some comments as to why the port width has to match if the parameters are the same and why it is ok to diverge if this isn't the case.

Also have you thought about the edge case where one `ExtModule` declares a port as `Reset`  and the other as `UInt<1>` (same for `Clock` etc.). Should that be a warning or an error?

seldridge:
Sure, I can add those comments. It's not super clear what the check is as written.

Good thought on the edge cases. What do you think about the following?

1. Comparing `AsyncReset` and `UInt<1>` is an error
2. Comparing `[AsyncReset, UInt<1>]` and `Reset` is not an error
3. `InferResets` invalidates `CheckHighForm` because it may enable now checks of type (2) are checks of type (1)
4. Comparing `Clock` with anything other than `Clock` is an error
5. Comparing `UInt<n>` vs. `UInt<m>` for an extmodule that has no parameters is an error
6. Comparing `UInt<n>` vs. `UInt<m>` for an extmodule that has parameters is not an error (could be a warning, but this is a common case and we don't have a way to turn off the warning)

seldridge:
@ekiwi-sifive: Updated with the logic above. 

Possibly worth an @albert-magyar look, too.

seldridge:
The checks I added conflict with the test introduced in https://github.com/freechipsproject/firrtl/pull/1087. 

Specifically, the test is something like two modules with the same defname, same ports, different widths, and different types for ports (one is a `UInt` and one is an `SInt`). This PR changes `CheckHighForm` to reject this. 

Example circuit that is now rejected:

```
circuit Foo:
  extmodule Bar:
    output out: UInt<1>
    defname = Bar
    parameter A = 0

  extmodule Qux :
    output out: SInt<1>
    defname = Bar
    parameter A = 1

  module Foo:
    inst bar of Bar
    inst qux of Qux
```

@jackkoenig: Do you have any thoughts on if these are equivalent or different? 

It seems problematic to have them the same as connecting to one vs. connecting to the other will be different for a `UInt` in `Foo` vs. an `SInt` in `Foo`.

jackkoenig:
@seldridge I think your check is correct and the test from #1087 is wrong. The test only passes at the Verilog-level because we emit SInts as unsigned in Verilog. While changing that behavior would be tricky, I think it's right to error here.

seldridge:
@jackkoenig: Sounds good. I added a commit to the head of this patch series that updates the test.

seldridge:
I think this is a 1.2.x backport. @ekiwi-sifive: do you have any objection to that?

ekiwi-sifive:
> I think this is a 1.2.x backport. @ekiwi-sifive: do you have any objection to that?

I think it might break some not-strictly-correct but currently working firrtl which is why I think this might be a breaking change the should only go into 1.4. Feel free to convince me otherwise though.

jackkoenig:
> I think it might break some not-strictly-correct but currently working firrtl which is why I think this might be a breaking change that should only go into 1.4. Feel free to convince me otherwise though.

I agree with @ekiwi-sifive, we could try to backport a version that only catches cases that would already be broken in Verilog but dunno if it's worth it.

seldridge:
It is changing the API of what was previously accepted FIRRTL and it does seem odd to break that on a minor release. 1.4 it is. Can always backport later if a good reason presents itself. Thanks for the review and feedback.

PR #1737, opened 2020-07-04
Add scala 2.13 support

tampler:
This is the follow up of #1729 with a cleaned commit history


tampler:
Will finish this PR after #1758

PR #1741, opened 2020-07-07
Add Expression Fuzzer

albertchen-sifive:
This adds JQF fuzz/repro CLIs and build.sbt tasks to run them. Code for these main methods are in the new `jqf` scala project. `JQFFuzz.scala` and `JQFRepro.scala` are scala translations of the [Fuzz and Repro goals](https://github.com/rohanpadhye/JQF/tree/master/maven-plugin/src/main/java/edu/berkeley/cs/jqf/plugin) of the maven plugin. The `fuzzer/JQFFuzz` and `fuzzer/JQFRepro` sbt tasks are thin wrappers around these main methods.

```
sbt> fuzzer/jqfFuzz <testClassName> <testMethodName> <otherArgs>...
```
```
sbt> fuzzer/jqfRepro <testClassName> <testMethodName> <otherArgs>...
```

This PR also adds a `fuzzer` project that contains random generator combinators. Generators use the `GenMonad` typeclass to abstract away the random generator implementation. This allows AST generators build with these combinators to use both JQF and ScalaCheck as the underlying random number generator. To use the generator combinators with a different random number generator, define an instance of the `GenMonad[T]` typeclass. `GenMonad` instances for `SourceOfRandomnessGen` (requires an implicit `SourceOfRandomness`) and `org.scalacheck.Gen` are defined.

The `StateGen` case class wraps a function `State => G[(State, A)]` that represents a random generator function that keeps track of some state, e.g. the namespace of references generated. Expression generators are `StateGen` functions that take a `State` type for which there is an `ExprState[State]` typeclass instance. The `ExprState` typeclass provides methods for adding references to the global namespace, getting generator parameters, and generating sub-expressions.

Expression generators are collected into `ExprGen` objects. The `ExprGen` trait has abstract `Option[StateGen]` members for `boolUIntGen`, `uintGen`, `boolSIntGen`, and `sintGen`. Generators for 1-width types are special cased because some types cannot be generated without generating zero-width wires. Generators for other types can later be added by adding more methods to the `ExprGen` trait with a default implementation of `None`. The `ExprGen` companion object has `ExprGen` objects for all the `PrimOp`s and for literals and references.

The `FirrtlCompileTests` class just tests that generated circuits compile without crashing. The `FirrtlEquivalenceTests` class tests that the minimum verilog compiler and optimized verilog compiler produce equivalent results for generated circuits.

The `ExprGenParams` provides an instance of the `ExprState` typeclass. `ExprGenParams` has three parameters: `maxDepth: Int`, `maxWidth: Int`, and `generators: Seq[(Int, ExprGen)]`. The `maxDepth` parameter sets the maximum number of levels of nested sub-expressoins, the `maxWidth` parameter sets the maximum width of any generated expression, and the `generators` param is a list of frequency -> expression generator pairs. For example, the following generator this generator can be used to generate circuits with a lot of `Not`s to test something like a inline-nots transform.
```scala
val params = ExprGenParams(
  maxDepth = 50,
  maxWidth = 31,
  generators = Seq(
    30 -> NotDoPrimGen,
    1 -> AddDoPrimGen,
    1 -> SubDoPrimGen,
    ...
  )
)
params.generateSingleExprCircuit[SourceOfRandomnessGen]()
```

### Contributor Checklist

#### Type of Improvement

- new feature/API

#### API Impact

add

none

#### Desired Merge Strategy

- Squash

- [ ] Did you review?


jackkoenig:
My [potentially incorrect] understanding is that JQF generates a "corpus" of inputs it finds as useful. Is this true and if so is it something we can save (as a seed for future runs) and perhaps using your fuzzer generate `.fir` test cases that exercise interesting code paths?

Also is it possible to see the coverage from JQF? If we can get the `.fir` we could run it through normal scoverage, but I'm curious to see how their metrics match or differ from scoverage if we can.

albertchen-sifive:
> My [potentially incorrect] understanding is that JQF generates a "corpus" of inputs it finds as useful. Is this true and if so is it something we can save (as a seed for future runs) and perhaps using your fuzzer generate `.fir` test cases that exercise interesting code paths?

yeah, it saves a new input whenever a new code path is exercised (I think) or you can tell it to just save all inputs. yes you can use these to seed a future run of JQF

> Also is it possible to see the coverage from JQF? If we can get the `.fir` we could run it through normal scoverage, but I'm curious to see how their metrics match or differ from scoverage if we can.

yeah the JQFRepro command has an option to log coverage for an input file. I have no idea how to use the output file though

coverage log snippet:
```
(939642881) firrtl/passes/PadWidths$#width():34 --> firrtl/bitWidth$#apply(Lfirrtl/ir/Type;)Lscala/math/BigInt;
(939646977) firrtl/passes/PadWidths$#width():35 --> firrtl/ir/SIntLiteral#tpe()Lfirrtl/ir/Type;
(939646978) firrtl/passes/PadWidths$#width():35 --> firrtl/passes/PadWidths$#width(Lfirrtl/ir/Type;)I
(939651073) firrtl/passes/PadWidths$#fixup():43 --> firrtl/passes/PadWidths$#width(Lfirrtl/ir/Expression;)I
(939651074) firrtl/passes/PadWidths$#fixup():43 [-1]
(939651075) firrtl/passes/PadWidths$#fixup():44 [0]
(939651075) firrtl/passes/PadWidths$#fixup():44 [1]
(939651081) firrtl/passes/PadWidths$#fixup():44 --> firrtl/passes/PadWidths$#tx$1(Lfirrtl/ir/Expression;I)Lfirrtl/ir/GroundType;
(939651082) firrtl/passes/PadWidths$#fixup():44 --> firrtl/ir/DoPrim#<init>(Lfirrtl/ir/PrimOp;Lscala/collection/Seq;Lscala/collection/Seq;Lfirrtl/ir/Type;)V
(939651083) firrtl/passes/PadWidths$#fixup():45 [1]
(939655169) firrtl/passes/PadWidths$#onExp():57 --> firrtl/Mappers$#ExprMap(Lfirrtl/ir/Expression;)Lfirrtl/ir/Expression;
(939655170) firrtl/passes/PadWidths$#onExp():57 --> firrtl/Mappers$ExprMap$#map$extension(Lfirrtl/ir/Expression;Lscala/Function1;Lscala/Function1;)Lfirrtl/ir/Expression;
(939655171) firrtl/passes/PadWidths$#onExp():58 [1]
```

fuzzer/src/main/scala/firrtl/ExprGen.scala
```diff
@@ -681,6 +681,28 @@ object ExprGen {
     }
   }
 
+  object MuxGen extends ExprGen[Mux] {
+    def name = "mux"
+    private def imp[S: ExprState, G[_]: GenMonad](tpe: Type): StateGen[S, G, Mux] = {
+      for {
+        cond <- ExprState[S].exprGen(Utils.BoolType)
+        expr1 <- ExprState[S].exprGen(tpe)
+        expr2 <- ExprState[S].exprGen(tpe)
```
ekiwi-sifive, line 690: The spec seems to say that "The types of the two input expressions are equivalent". However in practice it seems like we happily except something like this:
```.fir
circuit m:
  module m:
    input a : UInt<1>
    input b : UInt<2>
    input sel: UInt<1>
    output c: UInt
    c <= mux(sel, a, b)
```

Thus it would probably make sense to relax the type requirements here.

build.sbt
```diff
@@ -193,3 +193,68 @@ lazy val benchmark = (project in file("benchmark"))
...
+      "edu.berkeley.cs.jqf" % "jqf-instrument" % JQF_VERSION,
+      "com.github.scopt" %% "scopt" % "3.7.1",
+    ),
+    // Compile / run / fork := true,
+    // outputStrategy := Some(StdoutOutput),
+  )
+
+
+lazy val jqfFuzz = sbt.inputKey[Unit]("")
+lazy val jqfRepro = sbt.inputKey[Unit]("")
```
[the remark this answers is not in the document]
albertchen-sifive: oops, the string should have a description of the task

jackkoenig:
The Unidoc test still failed although I reflexively hit rerun. Anyway not a great sign since I thought this issue was fixed...

albertchen-sifive:
Seems like the unidoc was failing because it wasn't using the correct compiler options to compile the `fuzzer` project. unidoc wasn't using the `better-monadic-for` compiler plugin that `fuzzer` depended on. And for some reason when I ran `fuzzer/compile` I didn't need `import scala.language.higherKinds`, but the unidoc task requires it. I ended up only using `better-monadic-for` plugin in two places, so I got rid of it and used a workaround. and I added the `scala.language.higherKinds` imports.

fuzzer/src/main/scala/firrtl/ExprGen.scala
```diff
@@ -0,0 +1,728 @@
...
+    def boolUIntGen[S: ExprState, G[_]: GenMonad]: Option[StateGen[S, G, Mux]] = uintGen.map(_(1))
+    def uintGen[S: ExprState, G[_]: GenMonad]: Option[Width => StateGen[S, G, Mux]] = {
+      Some { imp(isUInt = true) }
+    }
+
+    def boolSIntGen[S: ExprState, G[_]: GenMonad]: Option[StateGen[S, G, Mux]] = sintGen.map(_(1))
+    def sintGen[S: ExprState, G[_]: GenMonad]: Option[Width => StateGen[S, G, Mux]] = {
+      Some { imp(isUInt = false) }
+    }
+  }
```
jackkoenig, line 727: One thing that has never really made sense to me is why there are function definitions for each `boolUInt`, `uint`, `boolSInt`, and `sint`. How is this going to work when we add aggregates? 

I highlighted this particular block fo code because the private `imp` taking `isUInt` as a `Boolean` argument kind of highlights the point. What will this code look like with Bundles and Vecs?

albertchen-sifive: There are different definitions for each because some primops like `lt` `ge`, etc. can only generate `boolUInt` and some cannot generate single bit ground types without generating zero-width wires. For aggregates I don't think there will be these edge-cases so we can just add two more methods for bundles and vecs.

In this case the code for Bundles and Vecs would look like
```
private def imp[S: ExprState, G[_]: GenMonad](tpe: Type): StateGen[S, G, Mux] = {
  for {
    cond <- ExprState[S].exprGen(Utils.BoolType)
    expr1 <- ExprState[S].exprGen(tpe)
    expr2 <- ExprState[S].exprGen(tpe)
  } yield {
    Mux(cond, expr1, expr2, tpe)
  }
}
```
This could be used for all the types, but since we allow uint/sint types with differing widths the generator is a bit different here.

jackkoenig: I see, alright I think this can work.

jackkoenig:
I think I'm going to manually squash and merge this because otherwise the commit message is going to have a lot of messages from WIP commits.

Issue #1229, opened 2019-11-07
Source locators on sequential statements


SandeepRajendran:
### Checklist

- [x] Did you write out a description of the feature you want to see?
- [ ] Did you look around for any related features?
- [ ] Did you specify relevant external information?

### Feature Description

Combinational statements in the emitted verilog have source information linking it to the chisel line that it originated from. For eg:

`assign io_bpwatch_3_action = io_bp_3_control_action; // @[Breakpoint.scala 87:16]`

It would be helpful if there was similar information for sequential statements inside always blocks. Right now sequential statements look like:

```
  always @(posedge clock) begin
    if (reset) begin
      reg_debug <= 1'h0;
    end else begin
      if (insn_ret) begin
        if (_T_4609) begin
          if (exception) begin
            if (trapToDebug) begin
              reg_debug <= _GEN_42
            end
          end
    ...
 ```

 It would be useful for debugging purposes if there was information that looked like:
 ```
  always @(posedge clock) begin
    if (reset) begin
      reg_debug <= 1'h0;                // @[CSR.scala 300:20]
    end else begin
      if (insn_ret) begin
        if (_T_4609) begin
          if (exception) begin
            if (trapToDebug) begin
              reg_debug <= _GEN_42;  // @[CSR.scala 671:25]
            end
          end
    ...
 ```
### Type of Feature

- backend code generation

jackkoenig:
Related to https://github.com/freechipsproject/firrtl/issues/679

Issue #1695, opened 2020-06-16
Should a MultiInfo be recovered when serializing and parsing?


ekiwi-sifive:
Currently a `MultiInfo` gets serialized in such a way that it is impossible to recover:

```.scala
val info = MultiInfo(Seq(MultiInfo(Seq(FileInfo("a"))), FileInfo("b"), FileInfo("c")))
```

serializes to ` @[a b c]` which is then parsed as `FileInfo("a b c")`.

Is this the expected behavior?

ekiwi-sifive:
Going one step further: If this is the desired behavior, why don't we get rid of the `MultiInfo` node and change the `FileInfo.++` method to append to the string?

```.scala
// current implementation
def ++(that: Info): Info = if (that == NoInfo) this else MultiInfo(Seq(this, that))
// suggested implementation:
def ++(that: Info): Info = that match {
    case NoInfo => this
    case FileInfo(escapedThat) => FileInfo(escaped + " " + escapedThat)
}
```

ekiwi-sifive:
> Going one step further: If this is the desired behavior, why don't we get rid of the `MultiInfo` node and change the `FileInfo.++` method to append to the string?

As an alternative, if we want to keep the Strings separate until right before serialization, we could change `FileInfo` to have a `Seq[String]`.

ekiwi-sifive:
As discussed in the dev meeting, this issue is on hold until @jackkoenig has time to revamp the whole Info system.

jackkoenig:
I think the answer to your question "Should a MultiInfo be recovered when serializing and parsing?" is "yes", it's just a question of how exactly to do it. I don't think we should concatenate into a single FileInfo because of the tendency of multiple FIRRTL nodes with the same source locator being collapsed. I'm not sure of the right way to represent Multi-Info in the serialized form but it's definitely a problem worth solving. Perhaps having the proposed `FileLineCol` type would work because then `@[file.scala 1:23 other.scala 3:45]` could be parsed into a `MultiInfo(Seq(FileLineCol(...), FileLineCol(...)))` and anything that doesn't match would parse into a `FileInfo` as an arbitrary string.

ekiwi-sifive:
`file.scala 1:23 other.scala 3:45` might be a valid file name on Linux.
If we do want to keep them separate it seems much easier to just allow for multiple `@[ ... ]` blocks:
```
@[file.scala 1:23] @[other.scala 3:45]
```

seldridge:
[...]

PR #1743, opened 2020-07-09
Propagate source locators to register update always blocks

jackkoenig:
### Contributor Checklist

#### Type of Improvement

- performance improvement (incidental, likely worse over all)
- backend code generation    

#### API Impact

As currently implemented, this changes some internal, yet public APIs in ExpandWhens. If we wish to backport this change I would like to have it reviewed and approved with the change and then add back the APIs for compatibility.

The purpose of this change is to include source locators in register update in emitted Verilog. Given some input FIRRTL:
```
circuit Foo :
  module Foo :
    input clock : Clock
    input reset : UInt<1>
    output io : { flip in : UInt<8>, flip a : UInt<1>, out : UInt<8> }

    reg reg : UInt<8>, clock with : (reset => (reset, UInt<8>("h0"))) @[main.scala 15:20]
    reg <= UInt<2>("h2") @[main.scala 16:11]
    when io.a : @[main.scala 17:15]
      reg <= io.in @[main.scala 18:9]
    io.out <= reg @[main.scala 19:10]
```
On current master you get:
```verilog
  always @(posedge clock) begin
    if (reset) begin
      reg_ <= 8'h0;
    end else if (io_a) begin
      reg_ <= io_in;
    end else begin
      reg_ <= 8'h2;
    end
  end

```

With this change we get the correct source locators!
```verilog
  always @(posedge clock) begin
    if (reset) begin // @[main.scala 15:20]
      reg_ <= 8'h0; // @[main.scala 15:20]
    end else if (io_a) begin // @[main.scala 17:15]
      reg_ <= io_in; // @[main.scala 18:9]
    end else begin
      reg_ <= 8'h2; // @[main.scala 16:11]
    end
  end
```

#### Desired Merge Strategy

- Squash

#### Release Notes
Emit source locators in register update always blocks

#### Additional Discussion

The approach I took here is to create a parallel datastructure in the source locators to represent multiple source locators for muxes. Essentially, you can think about the source locator for an ExpandWhens-generated mux as a Triple of the source locators for the condition, the true value, and the false value. This approach also intentionally uses NoInfo in these structures to keep the Triple structure. I think this is best explained with an example showing the serialized circuit during compilation, with modifications to info serialization to make MultiInfos and NoInfos apparent. Given the same input as above:

After ExpandWhens:
```
circuit Foo : ()
  module Foo : ()
    input clock : Clock ()
    input reset : UInt<1> ()
    output io : { flip in : UInt<8>, flip a : UInt<1>, out : UInt<8>} ()

    reg reg : UInt<8>, clock with :
      reset => (reset, UInt<8>("h0")) @[main.scala 15:20]
    node _GEN_0 = mux(io.a, io.in, UInt<2>("h2")) @[ @[main.scala 17:15],  @[main.scala 18:9],  @[main.scala 16:11]]
    io.out <= reg @[main.scala 19:10]
    reg <= _GEN_0 ()
```
You'll notice the triple on `_GEN_0` and how these source locators correspond to the `when` and connections above

After RemoveReset:
```
circuit Foo : ()
  module Foo : ()
    input clock : Clock ()
    input reset : UInt<1> ()
    input io_in : UInt<8> ()
    input io_a : UInt<1> ()
    output io_out : UInt<8> ()

    reg reg : UInt<8>, clock with :
      reset => (UInt<1>("h0"), reg) @[main.scala 15:20]
    node _GEN_0 = mux(io_a, io_in, UInt<2>("h2")) @[ @[main.scala 17:15],  @[main.scala 18:9],  @[main.scala 16:11]]
    io_out <= reg @[main.scala 19:10]
    reg <= mux(reset, UInt<8>("h0"), _GEN_0) @[ @[main.scala 15:20],  @[main.scala 15:20],  ()]
```
You'll see that the connection to `reg` has source locators from the reg itself for reset, and a NoInfo hole for `_GEN_0` which doesn't really have a source locator

After FlattenRegUpdate (internal to VerilogEmitter):
```
circuit Foo : ()
  module Foo : ()
    input clock : Clock ()
    input reset : UInt<1> ()
    input io_in : UInt<8> ()
    input io_a : UInt<1> ()
    output io_out : UInt<8> ()

    reg reg : UInt<8>, clock with :
      reset => (UInt<1>("h0"), reg) @[main.scala 15:20]
    node _GEN_0 = mux(io_a, io_in, UInt<8>("h2")) @[ @[main.scala 17:15],  @[main.scala 18:9],  @[main.scala 16:11]]
    io_out <= reg @[main.scala 19:10]
    skip
    reg <= mux(reset, UInt<8>("h0"), mux(io_a, io_in, UInt<8>("h2"))) @[ @[main.scala 15:20],  @[main.scala 15:20],  @[ @[main.scala 17:15],  @[main.scala 18:9],  @[main.scala 16:11]]]
```
Here's the interesting bit, it keeps a nesting of `MultiInfo` to reflect the nesting of muxes. This parallel datastructure is recursively emitted while the muxes are recursively emitted by the VerilogEmitter.

##### Alternative Approaches

This approach works pretty well, but it obviously is a bit brittle because of the parallel book keeping. An alternative approach is to add source locators to Expressions. The approach I took in this PR is *much* less intrusive than that; and way less likely to break existing transforms, but we could consider that instead.

##### `.fir` Serialization and Parsing

This nested structure is important for emitting the best Verilog. This is not currently maintained in FIRRTL emission nor supported by parsing. @ekiwi brought up this issue in https://github.com/freechipsproject/firrtl/issues/1695 so if we accept the approach in this PR (which I think we should), I think we also need to address that issue.

He proposed supporting multiple infos, but this PR technically would necessitate arbitrary nesting of infos. We could do this by adopting some delineator for multiple infos within a `@[...]` (eg. `,`) and requiring it to be escaped if used in filenames. A

Alternatively, we could just support parsing a single flat sequence of infos, and then support the nesting this PR requires by doing SplitExpressions before emission. It's not elegant but it could work. It doesn't support `--info-mode append` very well (which is still the default... I think we should change it to `--info-mode use`), but that could be worked around by having the two infos from `append` be merged into a single FileInfo. Inelegant but 🤷 

I kind of ran out of steam while writing this and the previous section so please feel free to ask for clarification/examples.

#### Potential Performance Improvement

The incidental performance improvement (I haven't measured it) is due to changing how source locators are propagated in `ExpandWhens`. It previously would just collect all of the source locators by serializing Connects and concatenating if there's a collision. I can't remember exactly how it could happen (maybe removing LHS subaccesses?) but rewriting this code makes that particular issue go away. Note that other aspects of this PR may make typical performance worse.

- [ ] Did you review?


jackkoenig:
Note that while this approach is brittle, I ran it through the SiFive flow and it seems to work pretty well in practice. Try it out!

src/test/scala/firrtlTests/InfoSpec.scala
```diff
@@ -172,4 +192,51 @@ class InfoSpec extends FirrtlFlatSpec with FirrtlMatchers {
...
   }
+
+  "source locators for basic register updates" should "be propagated to Verilog" in {
+    val result = compileBody(s"""
+      |input clock : Clock
+      |input reset : UInt<1>
+      |output io : { flip in : UInt<8>, out : UInt<8>}
+      |reg r : UInt<8>, clock
+      |r <= io.in $Info1
+      |io.out <= r $Info2
```
seldridge, line 203: This test doesn't make use of `Info2` since it's only testing that the register update is propagated. Would it be clearer to remove it?

src/test/scala/firrtlTests/InfoSpec.scala
```diff
@@ -172,4 +192,51 @@ class InfoSpec extends FirrtlFlatSpec with FirrtlMatchers {
...
+  }
+
+  "source locators for register reset" should "be propagated to Verilog" in {
+    val result = compileBody(s"""
+      |input clock : Clock
+      |input reset : UInt<1>
+      |output io : { flip in : UInt<8>, out : UInt<8>}
+      |reg r : UInt<8>, clock with : (reset => (reset, UInt<8>("h0"))) $Info3
+      |r <= io.in $Info1
+      |io.out <= r $Info2
```
seldridge, line 216: `Info2` isn't checked in the test. Should it be removed?

src/main/scala/firrtl/Emitter.scala
```diff
@@ -545,17 +545,19 @@ class VerilogEmitter extends SeqTransform with Emitter {
       this(Seq(), Map.empty, m, moduleMap, "", new EmissionOptions(Seq.empty))(writer)
     }
 
-    val netlist = mutable.LinkedHashMap[WrappedExpression, Expression]()
+    val netlist = mutable.LinkedHashMap[WrappedExpression, (Expression, Info)]()
```
seldridge, line 548: I'm perfectly fine with tuples, but the choice of a tuple here is different from the modified transforms that use `InfoExpr`. Can you elaborate on your reasoning for using a tuple here and a case class elsewhere?

src/main/scala/firrtl/transforms/ConstantPropagation.scala
```diff
@@ -577,27 +595,30 @@ class ConstantPropagation extends Transform with DependencyAPIMigration with Res
...
       }
-      nodeMap(lname) = value
+      nodeMap(lname) = InfoExpr.wrap(info, value)
     }
 
     def constPropStmt(s: Statement): Statement = {
-      val stmtx = s map constPropStmt map constPropExpression(nodeMap, instMap, constSubOutputs)
+      val s0 = s.map(constPropStmt) // Statement recurse
+      val s1 = propagateDirectConnectionWithInfo(nodeMap, dontTouches)(s0) // hacky source locator propagation
+      val stmtx = s1.map(constPropExpression(nodeMap, instMap, constSubOutputs)) // propagate sub-Expressions
```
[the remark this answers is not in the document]
jackkoenig: I totally agree, I originally wrote this as:
```scala
      val stmtx = s.map(constPropStmt)
                   .pipe(propagateDirectConnectionWithInfo(nodeMap, dontTouches))
                   .map(constPropExpression(nodeMap, instMap, constSubOutputs))
```
But `pipe` is only in 2.13 and I thought it a bit self-indulgent to port it as a utility.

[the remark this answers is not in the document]
seldridge: Oh neat! Didn't know about `pipe`. (Or about `tap`.) I wanted this without knowing it for so long...

src/main/scala/firrtl/Emitter.scala
jackkoenig: It predates `InfoExpr`, might as well use `InfoExpr`, I'll change it.

src/main/scala/firrtl/WIR.scala
```diff
@@ -169,6 +169,34 @@ case object Dshlw extends PrimOp {
   }
 }
 
+// Used privately to include Infos in the Netlist
+private[firrtl] case class InfoExpr(info: Info, expr: Expression) extends Expression {
```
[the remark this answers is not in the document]
jackkoenig: That is correct that this should never leak from any of the transforms in which it is used. I'll update the comment

src/main/scala/firrtl/transforms/ConstantPropagation.scala
```diff
@@ -676,8 +698,9 @@ class ConstantPropagation extends Transform with DependencyAPIMigration with Res
       // Actually transform some statements
       stmtx match {
         // Propagate connections to references
-        case Connect(info, lhs, rref @ WRef(rname, _, NodeKind, _)) if !dontTouches.contains(rname) =>
-          Connect(info, lhs, nodeMap(rname))
+        case Connect(info0, lhs, rref @ WRef(rname, _, NodeKind, _)) if !dontTouches.contains(rname) =>
+          val (info1, value) = unwrap(nodeMap(rname))
+          Connect(orElse(info0, info1), lhs, value)
```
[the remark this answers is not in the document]
jackkoenig: This is the opposite because I forgot to update it, nice catch. I should also have a comment pointing to that function since these kind of should be in sync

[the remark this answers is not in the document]
jackkoenig: Honestly, this code looks like it should be unreachable (the node reference should be propagated during the `constPropExpression` above), but I ran scoverage and it does seem to be hit somehow. I might mark this as a TODO check if still necessary.

src/test/scala/firrtlTests/InfoSpec.scala
jackkoenig: I think it should be there because if you reversed the `InfoExpr.orElse` logic in `ConstProp` then deleting `Info2` would have the same behavior as this, but leaving `Info2` and reversing that logic would cause this test to fail.

src/test/scala/firrtlTests/InfoSpec.scala
jackkoenig: Same as above, put differently, it's checking that `Info1` is prioritized over `Info2`, I should add a comment mentioning this.

src/main/scala/firrtl/passes/ExpandWhens.scala
```diff
@@ -78,37 +74,29 @@ object ExpandWhens extends Pass {
...
+    }
+
+
     // Memoizes the node that holds a particular expression, if any
     val nodes = new NodeLookup
 
     // Seq of attaches in order
     lazy val attaches = mutable.ArrayBuffer.empty[Attach]
 
-    val infoMap: InfoMap = new InfoMap
```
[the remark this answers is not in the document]
jackkoenig: I'll remove it and then look at readding it (along with the original `def expandWhens`) in a follow on commit

src/main/scala/firrtl/transforms/ConstantPropagation.scala
```diff
@@ -14,6 +14,7 @@ import firrtl.graph.DiGraph
 import firrtl.analyses.InstanceGraph
 import firrtl.annotations.TargetToken.Ref
 import firrtl.options.Dependency
+import firrtl.InfoExpr.{unwrap, orElse}
```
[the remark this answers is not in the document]
jackkoenig: I'll prefix them all with the object.

src/main/scala/firrtl/transforms/FlattenRegUpdate.scala
```diff
@@ -7,11 +7,19 @@ import firrtl.ir._
...
 
 import scala.collection.mutable
 
 object FlattenRegUpdate {
 
+  // Combination function for dealing with inlining of muxes and the handling of Triples of infos
+  private def combineInfos(muxInfo: Info, tinfo: Info, finfo: Info): Info = {
+    val (eninfo, tinfoAlt, finfoAlt) = MultiInfo.demux(muxInfo)
+    // Use MultiInfo constructor to preserve NoInfos
+    new MultiInfo(List(eninfo, orElse(tinfo, tinfoAlt), orElse(finfo, finfoAlt)))
```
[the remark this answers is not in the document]
jackkoenig: That is an interesting way to express it. I do feel like there is some other way to express this kind of concept but it smells like it might be a Haskell/ML thing that doesn't translate well to Scala. In any case, it's partially why I put it in a function since at least the composition is separated from the business logic of dealing with the muxes.

src/test/scala/firrtlTests/InfoSpec.scala
jackkoenig: with another look I realized that this is different from the above, I'll remove it here.

src/test/scala/firrtlTests/InfoSpec.scala
jackkoenig: With another look I realized that what I said is not what this is checking (I am currently working on tests that do do this so it was on my mind). Will remove `Info2`.

jackkoenig:
@seldridge I have addressed your comments and also pushed a fix to the ConstantPropagation changes to make them work right. They previously did what they were supposed to but broke name swapping (when a temporary name directly drives a better name, we swap the names to keep the better name). I renamed the utility function `propagateDirectConnectionInfoOnly` because it doesn't do any const prop any more (that's what broke name swapping), it *just* does source locator propagation. It is a hack as the comment said, but a useful one in the meantime until I can make everything in ConstProp propagate source locators.

jackkoenig:
I still need to rebase on `master` though...

src/main/scala/firrtl/transforms/ConstantPropagation.scala
```diff
@@ -458,9 +459,12 @@ class ConstantPropagation extends Transform with DependencyAPIMigration with Res
...
   }
 
-  private def constPropNodeRef(r: WRef, e: Expression) = e match {
-    case _: UIntLiteral | _: SIntLiteral | _: WRef => e
-    case _ => r
+  private def constPropNodeRef(r: WRef, e: Expression): Expression = {
+    def doit(ex: Expression) = ex match {
+      case _: UIntLiteral | _: SIntLiteral | _: WRef => ex
+      case _ => r
+    }
```
seldridge, line 465: starsky-and-hutch

jackkoenig:
I resolved conflicts just in the Github GUI. I think this PR is done. Future work:

* Further improve source locator propagation
  * ConstProp
  * RemoveWires
  * SplitExpressions
* Formalize concept of "structured info"
  * How should it be [de]serialized?
  * What about non-mux Expressions that could be nested?
  * What kinds of structure are there?
    * Info that corresponds to sub-expressions (like what this PR does for muxes)
    * Multiple infos for the same thing (original purpose of `MultiInfo`, eg. multiple expressions collapsing to 1)
    * Caller and Callee infos or more generalized full call stacks

src/main/scala/firrtl/transforms/ConstantPropagation.scala
jackkoenig: Hahaha exactly. I wasn't sure what to call the new inner function since it's not a `loop`, nor is it `rec`ursive. I've seen `doit` used before as a throwaway function name so 🤷

azidar:
Fixes #1229

PR #1745, opened 2020-07-09
mask bits when propagating bitwise ops

albertchen-sifive:
Ran into this bug on one of the circuits generated by the fuzzer. Here is a minimized example.
```
circuit Foo:
  module Foo:
    output out1: UInt<2>
    output out2: UInt<2>
    out1 <= xor(SInt<2>(-1), SInt<2>(1))
    out2 <= or(SInt<2>(-1), SInt<2>(1))
```

gets constant propped to this
```
circuit Foo:
  module Foo:
    output out1: UInt<2>
    output out2: UInt<2>
    out1 <= UInt<2>(-2)
    out2 <= UInt<2>(-1)
```

which results in illegal verilog
```verilog
module Foo(
  output [1:0] out1,
  output [1:0] out2
);
  assign out1 = 2'h-2;
  assign out2 = 2'h-1;
endmodule
```

This PR updates `ConstantPropagation` to use a bit mask when propagating these bitwise operations, so that the value of the resulting `UIntLiteral` is not negative.

### Contributor Checklist

#### Type of Improvement

- bug fix

#### API Impact

no impact

adds `UIntLiteral.masked` helper method

fixes broken verilog, see example above

#### Desired Merge Strategy

- Squash

- [ ] Did you review?


jackkoenig:
Should we consider adding a requirement to the `UIntLiteral` definition that the value be positive? https://github.com/freechipsproject/firrtl/blob/3fee5f87ec598e5ac870c9d2cd606047106421bb/src/main/scala/firrtl/ir/IR.scala#L225

It might catch other similar bugs although manifesting them may be tough.

src/main/scala/firrtl/transforms/ConstantPropagation.scala
```diff
@@ -207,7 +207,10 @@ class ConstantPropagation extends Transform with DependencyAPIMigration with Res
   }
 
   object FoldAND extends FoldCommutativeOp {
-    def fold(c1: Literal, c2: Literal) = UIntLiteral(c1.value & c2.value, c1.width max c2.width)
+    def fold(c1: Literal, c2: Literal) = {
+      val width = (c1.width max c2.width).asInstanceOf[IntWidth]
+      UIntLiteral((c1.value & c2.value) & ((BigInt(1) << width.width.toInt) - 1), width)
+    }
```
jackkoenig, line 213: This logic looks good, since it's written 3 times, I think we should probably factor this logic into a common utility function. Arguably it makes sense to override the apply method with this, but as an alternative:
```scala
// In ir/IR.scala
object UIntLiteral {
  /** Utility to construct UIntLiterals masked by the width
    *
    * This supports truncating negative values as well as values that are too wide for the width
    */
 def  masked(value: BigInt, width: IntWidth): UIntLiteral = {
    val mask = (BigInt(1) << width.width.toInt) - 1
   UIntLiteral(value & mask, width)
  }
}
```

albertchen-sifive:
@Mergifyio backport 1.2.X 1.3.x

albertchen-sifive:
@Mergifyio backport 1.2.x

PR #1753, opened 2020-07-10
Remove overlapping inputForm=LowForm tests

seldridge:
Remove tests from LoweringCompilerSpec testing the placement of
inputForm=LowForm legacy, custom transforms. This behavior is already
tested in the CustomTransformSpec.
   
Refactor the test used in the CustomTransformSpec to assert that
inputForm=LowForm legacy transforms run right before the emitter (see
note below!). The new test looks only for a list of (customTransform,
emitter) in a sliding, size-2 window of the flattened transform order.
Previously, this was looking for a match before and after the custom
transform. The old implementation necessitate busywork updates of the
test when new transforms are added that changed the transform running
before the custom transform.
    
Note: this test, as written is intentionally wrong. When verification
statements were added, the test was changed to not do what it's
supposed to do. Namely, the test is supposed to ensure that an
inputForm=LowForm transform runs immediately before its emitter.
However, the test is actually checking that the custom transform runs
before transforms that convert and remove verification statements. I'm
intentionally leaving the test broken, but doing the refactor in order
to make this easier to manually backport to the 1.3.x branch.

This should backport to 1.3.x.

### Contributor Checklist

- [n/a] Did you add Scaladoc to every public function/method?
- [n/a] Did you update the FIRRTL spec to include every new feature/behavior?
- [n/a] Did you add at least one test demonstrating the PR?
- [x] Did you delete any extraneous printlns/debugging code?
- [x] Did you specify the type of improvement?
- [x] Did you state the API impact?
- [x] Did you specify the code generation impact?
- [x] Did you request a desired merge strategy?
- [x] Did you add text to be included in the Release Notes for this change?

#### Type of Improvement

- code cleanup

#### API Impact

None.

#### Backend Code Generation Impact

None.

#### Desired Merge Strategy

- Rebase: You will rebase the PR onto master and it will be merged with a merge commit.

#### Release Notes

None.

### Reviewer Checklist (only modified by reviewer)
- [ ] Did you add the appropriate labels?
- [ ] Did you mark the proper milestone (1.2.x, 1.3.0, 1.4.0) ?
- [ ] Did you review?
- [ ] Did you check whether all relevant Contributor checkboxes have been checked?
- [ ] Did you mark as `Please Merge`?

PR #1754, opened 2020-07-10
Integrate new transforms with firrtl.stage.Forms

seldridge:
Move new transforms, recently added, into existing or new sets of
transforms (defined in firrtl.stage.Forms).

One transform is a mandatory low FIRRTL optimization:

  - firrtl.transforms.LegalizeAndReductionsTransform

Previously, this was included as a prerequisite of all Verilog
emitters (minimum, normal, and SystemVerilog).

Two transforms associated with converting and removing the new
verification statements are moved into a new set of transforms,
AssertsRemoved:

  - firrtl.transforms.formal.ConvertAsserts
  - firrtl.transforms.formal.RemoveVerificationStatements

Previously, these transforms were directly added as prerequisites to
the minimum Verilog and normal Verilog emitter, but not the
SystemVerilog emitter.

The designation of inputForm=LowForm for legacy, custom transforms is
updated to include assertion removal transforms as part of their
optionalPrerequisites. This has the effect of continuing to cause
inputForm=LowForm transforms to run as late as possible (right before
the low FIRRTL, minimum Verilog, Verilog, or SystemVeriog emitter).

Tests are updated to reflect the new order in both CustomTransformSpec
and LoweringCompilersSpec. This commit also simplifies the test used
in the CustomTransformSpec to assert that inputForm=LowForm legacy
transforms run right before the emitter. The new test looks only for a
list of (customTransform, emitter) in a sliding, size-2 window of the
flattened transform order. Previously, this was looking for a match
before and after the custom transform. The old implementation
necessitate busywork updates of the test when new transforms are added
that changed the transform running before the custom transform.

Depends on #1753.

This should not be backported due to references to 1.4 verification statements.

### Contributor Checklist

- x ] Did you add text to be included in the Release Notes for this change?

#### Type of Improvement

- code refactoring

#### API Impact

This adds `firrtl.stage.Forms.AssertRemoved` to the API.

None.

#### Desired Merge Strategy

#### Release Notes

None

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/Compiler.scala
+++ b/src/main/scala/firrtl/Compiler.scala
@@ -313,7 +313,7 @@ trait Transform extends TransformLike[CircuitState] with DependencyAPI[Transform
   }
 
   override def optionalPrerequisites: Seq[Dependency[Transform]] = inputForm match {
-    case L => Forms.LowFormOptimized
+    case L => Forms.LowFormOptimized ++ Forms.AssertsRemoved
     case _ => Seq.empty
   }
 
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -7,16 +7,15 @@ import java.io.Writer
 import scala.collection.mutable
 import firrtl.ir._
 import firrtl.passes._
-import firrtl.transforms.{FixAddingNegativeLiterals, LegalizeAndReductionsTransform}
+import firrtl.transforms.FixAddingNegativeLiterals
 import firrtl.annotations._
 import firrtl.traversals.Foreachers._
 import firrtl.PrimOps._
 import firrtl.WrappedExpression._
 import Utils._
 import MemPortUtils.{memPortField, memType}
-import firrtl.options.{Dependency, HasShellOptions, PhaseException, ShellOption, Unserializable}
+import firrtl.options.{HasShellOptions, PhaseException, ShellOption, Unserializable}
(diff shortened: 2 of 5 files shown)


sequencer:
> One transform is a mandatory low FIRRTL optimization:
> `firrtl.transforms.LegalizeAndReductionsTransform`

I have a question to this.
i notice [debian-sid](https://packages.debian.org/sid/electronics/verilator), [homebrew](https://formulae.brew.sh/formula/verilator) and [archlinux](https://www.archlinux.org/packages/community/x86_64/verilator/) has updated to 4.036 or even 4.038, which won't contains bug of verilator/verilator#2300

My question is possible to remove this after some a milestone like fedora get this update?

I think distributions like Ubuntu or CentOS should not affect our development since they contains a very bad package maintaining policy.

sequencer:
Generally look good to me:)
Should we only merge this, or merge this after #1753 gets merged?

seldridge:
@sequencer: Can you take a look at this again? This is just rebased onto `master`, but I think you actually have review powers now (and thereby this has the potential to get merged).

seldridge:
I missed the earlier comment about `LegalizeAndReductionsTransform`. That's probably a @jackkoenig question on how long to leave it in. Ideally, if the transform comes out, it would be replaced with a change to the Verilog emitter that emits Verilog code that includes a compile-time check of the Verilator version. I'm not sure if this is possible, however.

PR #1758, opened 2020-07-11
WIP : Travis CI with a prebuilt Docker image

tampler:
1. Added `sbt-extras` for improved bin compatibility for different sbt versions

2. Added binary `launchers/1.3.13/sbt-launch.jar` due to a permanent error in sbt-extras. TODO: Find a better workaround

3. Added a pre-built Docker Image with the latest Yosys, Verilator and JVM 11 with Shenandoah GC


tampler:
Setting as a WIP since we need to add other checks and approve the flow

tampler:
Updated PR with the following:

1. Moved docker cmd to a separate script
2. Added docker image caching 

Now Travis fails with the following : `cp: cannot stat 'utils/bin/firrtl.jar': No such file or directory`
@jackkoenig Should we somehow change the approach and avoid using firrtl.jar ?

jackkoenig:
Looking at the changes this seems like a lot of churn just to fix a [hopefully] minor CI issue. I was thinking about what could be causing the failures and PR-ed a simple change that _might_ fix it: https://github.com/freechipsproject/firrtl/pull/1762

tampler:
Hi Jack

The idea is not to merely fix that issue with Unidoc for scala 2.13. The bigger ideas are
1. Improve CI integrity with a conteinerization strategy
2. Improve CI performance with fewer tasks and Docker image caching 

If you want, I may remote [sbt-extras](https://github.com/paulp/sbt-extras) dependency and leave only a fixed 1.3.13 version of SBT

PR #1776, opened 2020-07-17
RenameMapSpec: try rename instance and port

ekiwi-sifive:
I mostly want to know what the expected behavior is when trying to rename port as well as instance name.
Both things can occur together in a Uniquify-like pass.

Recording the field rename and the instance rename is apparent not enough; the third test fails.
What is the correct way to do this?

### DIFF
--- a/src/test/scala/firrtlTests/RenameMapSpec.scala
+++ b/src/test/scala/firrtlTests/RenameMapSpec.scala
@@ -838,11 +838,16 @@ class RenameMapSpec extends FirrtlFlatSpec {
     val top = CircuitTarget("top").module("top")
     val child = CircuitTarget("top").module("child")
 
-    val r = RenameMap()
-    r.record(child.ref("a"), Seq(child.ref("a_0"), child.ref("a_1")))
+    val portRenames = RenameMap()
+    portRenames.record(child.ref("a"), Seq(child.ref("a_0"), child.ref("a_1")))
+
+    val instanceRenames = RenameMap()
     val i = top.instOf("i", "child")
     val i_ = top.instOf("i_", "child")
-    r.record(i, i_)
+    instanceRenames.record(i, i_)
+
+    // The port and instance renames must be *explicitly* chained!
+    val r = portRenames.andThen(instanceRenames)
     r.get(i.ref("a")) should be (Some(Seq(i_.ref("a_0"), i_.ref("a_1"))))
   }
 }
--- a/src/test/scala/firrtlTests/RenameMapSpec.scala
+++ b/src/test/scala/firrtlTests/RenameMapSpec.scala
@@ -809,4 +809,40 @@ class RenameMapSpec extends FirrtlFlatSpec {
     r.delete(Mod)
     r.get(foo) should be (Some(Nil))
   }
+
+  it should "rename an instance if it has been renamed" in {
+    // this could happen if the instance name needed to be uniquified to avoid clashes in LowerTypes
+    val top = CircuitTarget("top").module("top")
+
+    val r = RenameMap()
+    val i = top.instOf("i", "child")
+    val i_ = top.instOf("i_", "child")
+    r.record(i, i_)
+    r.get(i) should be (Some(Seq(i_)))
+    r.get(i.ref("a")) should be (Some(Seq(i_.ref("a"))))
+  }
+
+  it should "rename references to an instance's ports if the ports of the module have been renamed" in {
+    // this could happen if the port name needed to be uniquified to avoid clashes in LowerTypes
+    val top = CircuitTarget("top").module("top")
+    val child = CircuitTarget("top").module("child")
+
+    val r = RenameMap()
+    r.record(child.ref("a"), Seq(child.ref("a_0"), child.ref("a_1")))
+    val i = top.instOf("i", "child")
+    r.get(i.ref("a")) should be (Some(Seq(i.ref("a_0"), i.ref("a_1"))))
+  }
+
+  it should "rename references to renamed instance's ports if the ports of the module have been renamed" in {
+    // this could happen if the instance as well as the port name needed to be uniquified to avoid clashes in LowerTypes
+    val top = CircuitTarget("top").module("top")
+    val child = CircuitTarget("top").module("child")
+
+    val r = RenameMap()
+    r.record(child.ref("a"), Seq(child.ref("a_0"), child.ref("a_1")))
+    val i = top.instOf("i", "child")
+    val i_ = top.instOf("i_", "child")
+    r.record(i, i_)
+    r.get(i.ref("a")) should be (Some(Seq(i_.ref("a_0"), i_.ref("a_1"))))
+  }
 }


seldridge:
tl;dr: This looks like a bug---the rename map is returning a target that doesn't exist in the circuit. Your recorded renames make sense to me. `RenameMap` should do the right thing here.

It appears that you can get the right behavior, but you have to record the full path rename (:grimacing:):

```
~top|top/i:child>a -> [~top|top/i_:child>a_0, ~top|top/i_:child>a_1]
```

However, there's no way that a transform should be required to record all possible renames for all possible (exponential) paths to a given reference in the circuit. This is way too expensive and runs counter to traversing modules in a circuit. It should be sufficient to record the module port rename and the instance name rename. 

Your renames are exactly what `ManipulateNames` produces in a similar situation and it consequently suffers from the same bug. E.g., if you try to lowercase all names in the following circuit:

```
circuit foo:
  module bar:
    input IN: UInt<1>
    output OUT: UInt<1>
    OUT <= IN
  module foo:
    input in: UInt<1>
    output out: UInt<1>
    inst BAR of bar
    BAR.IN <= in
    out <= BAR.OUT
```

You get a rename map of:

```
~foo|bar>OUT=>~foo|bar>out
~foo|bar>IN=>~foo|bar>in
~foo=>~foo
~foo|foo/BAR:bar=>~foo|foo/bar:bar
```

This looks fine, but does the invalid rename: `~foo|foo/BAR:bar>IN => ~foo|foo/BAR:bar>in`.

I'd propose fixing rename map to "do the right thing". Also, this probably motivates an optional wrapper transform that checks the legality of renames---e.g., when a target is renamed, it must map to a target that exists.

albertchen-sifive:
this PR https://github.com/freechipsproject/firrtl/pull/1539 changed the `RenameMap` behavior to no longer implicitly chain renames on different "parts" of a single target. i.e. it will stop renaming on the first rename it finds for any given target, so the last test returns
```
~top|top/i:child>a -> [~top|top/i_:child>a_0, ~top|top/i_:child>a_1]
```
because it applies the rename `~top|child>a -> [~top|child>a_0, ~top|child>a_1]` and stops.

It is now required to explicitly chain `RenameMap`s using the `.andThen` method. So to implement the rename for the last case:
```diff
     val top = CircuitTarget("top").module("top")
     val child = CircuitTarget("top").module("child")

-    val r = RenameMap()
-    r.record(child.ref("a"), Seq(child.ref("a_0"), child.ref("a_1")))
+    val r1 = RenameMap()
+    r1.record(child.ref("a"), Seq(child.ref("a_0"), child.ref("a_1")))
+    val r2 = RenameMap()
     val i = top.instOf("i", "child")
     val i_ = top.instOf("i_", "child")
-    r.record(i, i_)
-    r.get(i.ref("a")) should be (Some(Seq(i_.ref("a_0"), i_.ref("a_1"))))
+    r2.record(i, i_)
+    r1.andThen(r2).get(i.ref("a")) should be (Some(Seq(i_.ref("a_0"), i_.ref("a_1"))))
   }
 }

```

ekiwi-sifive:
Thanks @albertchen-sifive !

So if I have a transform that renames ports and instances it should be enough to create one rename map for all port renames and one for all other renames and then chain them before returning from the transform, correct?

albertchen-sifive:
> Thanks @albertchen-sifive !
> 
> So if I have a transform that renames ports and instances it should be enough to create one rename map for all port renames and one for all other renames and then chain them before returning from the transform, correct?

yes

seldridge:
Thanks for the clarification @albertchen-sifive.

I think one port rename and one module rename only works for a circuit with depth 2, though. If you start adding levels of depth, then you need to chain more. Consider the situation of adding a grandchild to the example where you're renaming `~top|top/i:child/j:granchild>a`.

A safe, naive approach would be to have one rename map for every module in the design chained in topologically sorted order (and yield an `O(m)` renaming cost for a circuit with `m` modules?).

I think you can knock this down to `O(lgm)`, but that would actually combining rename maps for modules which are at the same level in the topological sort.

Does this make sense or am I missing something?

ekiwi-sifive:
Is this problem slightly different? I.e. the original problem was the two step rename of port and instance.
The problem that you bring up, @seldridge , is that the path of an instance might change, right?
My pass does not use instance paths since all renames are for all instances of the module. Would I still be affected?

seldridge:
> Is this problem slightly different? I.e. the original problem was the two step rename of port and instance.

I think I went more general to the problem of generating a rename map for transforms that rename references and instances.

You're strictly asking about the behavior of a rename map with a port rename and an instance rename, yes?

> The problem that you bring up, @seldridge , is that the path of an instance might change, right?

Yeah.

If a transform, like `Uniquify`, renames instances then there may exist an annotation with a target using the longest possible path in the circuit. The rename map returned by that transform needs to correctly rename that path and that appears to necessitate chaining.

> My pass does not use instance paths since all renames are for all instances of the module. Would I still be affected?

I'm missing some context on what your transform is doing, but I was assuming that it had the capability to rename instances and ports. I think that any transform that can rename instances, e.g., `Uniquify` or `ManipulateNames`/`VerilogRename`, will generate incorrect rename maps unless it's doing this chaining. (`ManipulateNames` doesn't do this correctly; `Uniquify` appears to be using one rename map, so I think it's incorrect, but I did not test it.)

ekiwi-sifive:
> I'm missing some context on what your transform is doing, but I was assuming that it had the capability to rename instances and ports.

What I was trying to get at is that all renames are only at depth 2, i.e. module.instOf(....) -> module.instOf(...), but never deeper, i.e. no top.instOf(...).instOf(...).
I have no idea how the resolution is done, but I was hoping that something like `~top|top/i:child/j:granchild>a` would search for renames of `top.i` and then for renames of `child.j`, but my understanding here is probably wrong.

albertchen-sifive:
> A safe, naive approach would be to have one rename map for every module in the design chained in topologically sorted order (and yield an `O(m)` renaming cost for a circuit with `m` modules?).
> 
> I think you can knock this down to `O(lgm)`, but that would require combining rename maps for modules which are at the same level in the topological sort.

yeah this makes sense to me. The `InlineInstances` transform does what you described with the topological sort renaming.

ekiwi-sifive:
If we do the one RenameMap per module it would be interesting to explore if this has any impact on the performance of the common case in which instances are not renamed.

jackkoenig:
> If we do the one RenameMap per module it would be interesting to explore if this has any impact on the performance of the common case in which instances are not renamed.

I suspect this is going to be a pretty big performance problem for designs with large numbers of annotations and large numbers of modules.

Worse though, the user API here for instance path renaming is pretty bad. How is someone writing an inlining transform supposed to realize they need to create and chain `n` or at least `lg(n)` `RenameMaps` together? I can't remember exactly where but I remember being grumpy recently (like 2-3 months ago) that I had to enumerate (and rename) all instance paths in the design to do a simple module rename which is a different way of solving this same kind of problem.

I don't know how to solve this, but we really need to provide a simpler API for renaming modules and instances.

ekiwi-sifive:
Is this a valid approach to doing the renaming if port and instance names can be renamed? (https://github.com/freechipsproject/firrtl/pull/1784/commits/2dda0e691c384fa630fb9f29ca03de08425f16e1)

```.scala
    val moduleRenames: Map[String, RenameMap] = resultAndRenames.map{ case(m,r) => m.name -> r }.toMap
    val moduleOrderBottomUp = new InstanceKeyGraph(result).moduleOrder.reverseIterator
    val renames = moduleOrderBottomUp.map(m => moduleRenames(m.name)).reduce((a,b) => a.andThen(b))
```

PR #1780, opened 2020-07-20
Uniquify: should uniquify memory ports

ekiwi-sifive:
Is this a bug?

### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/src/test/scala/firrtlTests/UniquifySpec.scala
+++ b/src/test/scala/firrtlTests/UniquifySpec.scala
@@ -258,6 +258,36 @@ class UniquifySpec extends FirrtlFlatSpec {
     executeTest(input, expected)
   }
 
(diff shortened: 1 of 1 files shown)


ekiwi-sifive:
When I change the type of the memory to be a GroundType like `UInt<8>`, infer binary points seems to crash....

ekiwi-sifive:
Uniquifying memory ports does not actually seem to be a requirement. Neither Jack nor I could come up with a situtation in which it would be necessary. This is the case because all children of all memory ports are the same and the memory ports are never directly addressed in Verilog, only their fields. So if we have `r` and `r_data`, `r.data` will become `r_data` at the same time as `r_data.data` is flattened to `r_data_data` and everything is good.

PR #1783, opened 2020-07-22
[WIP] combined Uniquify and LowerTypes pass (old)

ekiwi-sifive:
### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/SymbolTable.scala
+++ b/src/main/scala/firrtl/SymbolTable.scala
@@ -72,5 +72,6 @@ private class StandardGlobalSymbolTable extends GlobalSymbolTable {
 
 object GlobalSymbolTable {
   def scanModuleTypes[T <: GlobalSymbolTable](c: Circuit, t: => T): T = scanModuleTypes(c.modules, t)
+  def scanModuleTypes(c: Circuit): GlobalSymbolTable = scanModuleTypes(c.modules, new StandardGlobalSymbolTable)
   def scanModuleTypes[T <: GlobalSymbolTable](m: Iterable[DefModule], t: => T): T = { m.foreach(t.declare) ; t }
 }
--- a/src/main/scala/firrtl/SymbolTable.scala
+++ b/src/main/scala/firrtl/SymbolTable.scala
@@ -39,8 +39,8 @@ private[firrtl] class StandardSymbolTable(global: GlobalSymbolTable)  extends Sy
 }
 
 object SymbolTable {
-  def scanModule(t: SymbolTable, m: DefModule): SymbolTable = {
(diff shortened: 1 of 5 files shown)


seldridge:
Since you're taking a hard look at `Uniquify` (thanks!) I've had two recent thoughts about this that may be worth considering as part of a refactor. These can obviously be future work, though:

1) Have you considered how uniquification should work for memories? If `ReplSeqMem` runs, then the memory is blackboxed and there's no way for the readers/writers/readwriters to enter the namespace. If `ReplSeqMem` doesn't run, then the memory is lowered in the Verilog emitter to an unpacked array which can produce Verilog that doesn't preserve the `Uniquify` invariance. This can be trivially fixed by adding all addresses in a memory to the namespace, but this has drastic performance implications: https://github.com/freechipsproject/firrtl/pull/1549, However, this has enormous performance implications, which leads to point question (2).

2) Have you considered using a data structure other than a set to store the namespace? Uniquify already has big problems with large vecs, but this is exacerbated if memories enumerate their addresses. E.g., `s"""output out: UInt<1>[${BigInt(2).pow(22)}]"""` takes 2 minutes to uniquify. It seemed like this could be remedied by switching to a tree of prefixes instead of a flat `Set[String]`. The numeric prefix would need to be something like an interval tree (so that you can quickly query if some index or range overlaps with the existing namespace). I got as far as realizing this would need a self-balancing tree (as numeric signals appear sequentially in a circuit, e.g., `_T_1`, `_T_2`, ...) and stopped.

ekiwi-sifive:
>     1. Have you considered how uniquification should work for memories?

I did add uniquification for memory ports with this PR, but after thinking about this a bit more and talking to Jack, I think I am going to remove it in the final version of this PR. In general I have come to believe that all that the new `LowerTypes` pass needs to do is to rename things in such a way that it does not create any conflicts.

As a secondary concern I am trying to rename things exactly the way that Uniquify does for now in order to make regression testing much easier.

I have to admit that I never quite understood your concerns in #1549. Since we do emit memories as arrays, there is no need to treat them like bundles where you could have conflicts with the suffixes `_0`, `_1` ...

I have not tried changing the `Namespace` data structure used for the renaming, but changing that is almost orthogonal to the changes in this PR. For large vectors as in your example we might want to consider emitting them as Verilog arrays at some point, thus completely removing the need to rename them. Verilog arrays have to problem that they are only supported inside a module, however: 1) I think many large Vectors will not be part of a Module's port 2) we can come up with a standardized way to lower them once they are part of a module's ports, i.e., we would unpack and repack them when going through module boundaries.

ekiwi-sifive:
superseded by #1784

Issue #1165, opened 2019-08-23
Transform CircuitForms are Ambiguous


seldridge:
**tl;dr: when you specify `LowForm` or `HighForm` as your `inputForm`/`outputForm` in a custom transform, you may either be inadvertently writing a transform that relies on `ResolveAndCheck` or your transform may not be portable between `-X low` and `-X verilog`.**

This weirdness is preserved in #1123.

## High Form Ambiguity

For a custom transform that outputs `HighForm`, the re-lowering (or lack thereof) is a function of the `inputForm` due to `getLoweringTransforms` logic.

- `inputForm == HighForm` will result in no re-lowering
- `inputForm < HighForm` will result in `IrToWorkingIr`, `ResolveAndCheck`, and `Dedup`

This effectively defines `HighForm` as either (up through) `Dedup` or `ChirrtlToHighFirrtl`.

Consequently, a user writing a `{Low, Mid} -> High` transform may be relying on this re-lowering, e.g., they emit unresolved types.

*Going forward, the dependency API port redefines `HighForm` to mean `Deduped`, but preserves backwards compatibility for transforms that output `HighForm` without overriding the dependency API methods.*

## Low Form Ambiguity

By the current definition of `mergeTransforms` a custom transform will run after the last transform which has an `outputForm` equal to its `inputForm`. However, the last transform which has `outputForm == LowForm` is a function of the chosen compiler/emitter. Concretely a `LowForm` transform, for the chosen compiler, has the following behavior:

- `-X low`: transform runs after `MiddleFirtlToLowFirrtl`
- `-X mverilog`: runs after `MinimumLowFirrtlOptimizations`
- `-X verilog`: runs after `LowFirrtlOptimizations`

This has implications in that a user can write a `LowForm` transform that works with `-X verilog`, but fails for `-X low`. (Consider a transform that will match error on `IsInvalid`. This transform will error for `-X low`, but work for `-X verilog` because `RemoveValidIf` is an optimization transform. h/t @jackkoenig)

*This behavior is preserved in how dependency API dependencies are derived from `inputForm`/`outputForm` for users that do not override dependency API methods.*

## Metadata

**Type of issue**: bug report

**If the current behavior is a bug, please provide the steps to reproduce the problem:**
  * **What is the current behavior?**

A `LowForm` transform will run in different places for different compilers.

  * **What is the expected behavior?**

A `LowForm` should run in the same location, regardless of compiler.

  * **Please tell us about your environment:**
    (examples)

[...]

PR #1784, opened 2020-07-23
Combined Uniquify and LowerTypes pass 

ekiwi-sifive:
### Contributor Checklist

- _[n/a]_ Did you update the FIRRTL spec to include every new feature/behavior?

#### Type of Improvement
 - performance improvement
 - code refactoring

#### API Impact

- currently this just adds a `NewLowerTypes` pass
- we should see if we can just immediately replace `LowerTypes` with this new pass which should be doing strictly more

- this should not impact the generated Verilog
- an exception is the fact that we do better renaming and thus some annotation might not get dropped and thus affect the generated verilog

#### Desired Merge Strategy
- Rebase: You will rebase the PR onto master and it will be merged with a merge commit
- I will fix up the commit history before we merge

- [ ] Did you review?


ekiwi-sifive:
I ran a cold compile benchmark and in general it seems to be slightly faster (but the SD is high as usual)


revision | design | max heap (MiB) | SD | runtime (s) | SD
-- | -- | -- | -- | -- | --
NewLowerTypes | small | 3055 | 98.14 | 48.12 | 3.05
master | small | 3065 | 126.98 | 51.44 | 1.49
NewLowerTypes | medium | 3590 | 85.04 | 112.55 | 2.89
master | medium | 3583 | 95.21 | 120.09 | 29.26

ekiwi-sifive:
In terms of regression testing:

I compiled both the `small` and the `medium` benchmark with the current `master` branch and with the changes in this `PR` and the generated Verilog is exactly the same.

ekiwi-sifive:
The only failing test now is the `LoweringCompilersSpec`

src/main/scala/firrtl/transforms/RemoveWires.scala
```diff
@@ -23,16 +23,20 @@ import scala.util.{Try, Success, Failure}
...
+         Dependency(passes.ResolveKinds),
          Dependency(transforms.RemoveReset),
          Dependency[transforms.CheckCombLoops] )
 
   override def optionalPrerequisites = Seq(Dependency[checks.CheckResets])
 
   override def optionalPrerequisiteOf = Seq.empty
 
-  override def invalidates(a: Transform) = false
+  override def invalidates(a: Transform) = a match {
```
azidar, line 36: Was this change necessary for this PR? Otherwise it should probably be separated into another PR.

src/main/scala/firrtl/passes/NewLowerTypes.scala
```diff
@@ -0,0 +1,439 @@
...
+
+  private trait ParentRef { def ref(name: String, asVecField: Boolean = false): ReferenceTarget }
+  private case class ModuleParentRef(m: ModuleTarget) extends ParentRef {
+    override def ref(name: String, asVecField: Boolean): ReferenceTarget = m.ref(name)
+  }
+  private case class RefParentRef(r: ReferenceTarget) extends ParentRef {
+    override def ref(name: String, asVecField: Boolean): ReferenceTarget =
+      if(asVecField) { r.index(name.toInt) } else { r.field(name) }
+  }
+}
```
azidar, line 439: newline

src/main/scala/firrtl/passes/NewLowerTypes.scala
```diff
@@ -0,0 +1,439 @@
...
+  */
+object NewLowerTypes extends Transform {
+  override def inputForm: CircuitForm = UnknownForm
+  override def outputForm: CircuitForm = UnknownForm
+
+  override def prerequisites: Seq[TransformDependency] = Seq(
+    Dependency(RemoveAccesses), // we require all SubAccess nodes to have been removed
+    Dependency(CheckTypes),     // we require all types to be correct
+    Dependency(InferTypes),     // we require instance types to be resolved (i.e., DefInstance.tpe != UnknownType)
+    Dependency(ExpandConnects)  // we require all PartialConnect nodes to have been expanded
```
azidar, line 27: So does this work prior to expand whens?

src/main/scala/firrtl/passes/NewLowerTypes.scala
```diff
@@ -0,0 +1,439 @@
...
+
+    (newMod, renameMap, memInit)
+  }
+
+  def onPort(p: Port)(symbols: LoweringTable): Seq[Port] = symbols.lower(p)
+
+  def onStatement(s: Statement)(implicit symbols: LoweringTable, memInit: Seq[MemoryInitAnnotation]): Statement = s match {
+    // declarations
+    case d : DefWire =>
+      Block(symbols.lower(d.name, d.tpe, firrtl.WireKind).map{case (name, tpe, _) => d.copy(name=name, tpe=tpe) })
```
azidar, line 96:
```suggestion
      Block(symbols.lower(d.name, d.tpe, firrtl.WireKind).map { case (name, tpe, _) => d.copy(name=name, tpe=tpe) })
```

src/main/scala/firrtl/passes/NewLowerTypes.scala
```diff
@@ -0,0 +1,439 @@
...
+    case d : DefMemory =>
+      // TODO: as an optimization, we could just skip ground type memories here.
+      //       This would require that we don't error in getReferences() but instead return the old reference.
+      val mems = symbols.lower(d)
+      if(mems.length > 1 && memInit.exists(_.target.ref == d.name)) {
+        val mod = memInit.find(_.target.ref == d.name).get.target.encapsulatingModule
+        val msg = s"[module $mod] Cannot initialize memory ${d.name} of non ground type ${d.dataType.serialize}"
+        throw new RuntimeException(msg)
+      }
+      Block(symbols.lower(d))
```
azidar, line 124:
```suggestion
      Block(mems)
```

src/main/scala/firrtl/passes/NewLowerTypes.scala
```diff
@@ -0,0 +1,439 @@
...
+    case d : DefInstance => symbols.lower(d)
+    // connections
+    case Connect(info, loc, expr) =>
+      if(!expr.tpe.isInstanceOf[GroundType]) {
+        throw new RuntimeException(s"NewLowerTypes expects Connects to have been expanded! ${expr.tpe.serialize}")
+      }
+      val rhs = onExpression(expr)
+      // We can get multiple refs on the lhs because of ground-type memory ports like "clk" which can get duplicated.
```
azidar, line 132: Huh, I'd forgotten about this. Nice catch!

src/main/scala/firrtl/passes/NewLowerTypes.scala
```diff
@@ -0,0 +1,439 @@
...
+      val lhs = symbols.getReferences(loc.asInstanceOf[RefLikeExpression])
```
azidar, line 133: I wonder if it would be useful to change the `loc` type in `Connect` to be of `RefLikeExpression`. A future PR perhaps, this is fine as is.

src/main/scala/firrtl/passes/NewLowerTypes.scala
```diff
@@ -0,0 +1,439 @@
...
+      Block(lhs.map(loc => IsInvalid(info, loc)))
+    // others
+    case other => other.mapExpr(onExpression).mapStmt(onStatement)
+  }
+
+  /** Replaces all Reference, SubIndex and SubField nodes with the updated references */
+  def onExpression(e: Expression)(implicit symbols: LoweringTable): Expression = e match {
+    case r: RefLikeExpression =>
+      // When reading (and not assigning to) an expression, we can always just pick the first one.
+      symbols.getReferences(r).head
```
azidar, line 152: Can you update the comment to give an example of when getReferences returns a list of size greater than one? I imagine the memory's clock case is one example.

src/main/scala/firrtl/passes/NewLowerTypes.scala
```diff
@@ -0,0 +1,439 @@
...
+    case other => other.mapExpr(onExpression)
+  }
+}
+
+// Holds the first level of the module-level namespace.
```
azidar, line 157: What does "first level" mean? Please clarify this comment.

src/main/scala/firrtl/passes/NewLowerTypes.scala
```diff
@@ -0,0 +1,439 @@
...
+  }
+
+  private def extractGroundTypeRefString(refs: Seq[ReferenceTarget]): String =
+  if(refs.isEmpty) { "" } else {
+    // Since we depend on ExpandConnects any reference we encounter will be of ground type
+    // and thus the one with the longest access path.
+    refs.reduceLeft((x,y) => if (x.component.length > y.component.length) x else y)
+    // convert references to strings relative to the module
+      .serialize.dropWhile(_ != '>').tail
+  }
```
azidar, line 359:
```suggestion
    if(refs.isEmpty) { "" } else {
      // Since we depend on ExpandConnects any reference we encounter will be of ground type
      // and thus the one with the longest access path.
      refs.reduceLeft((x,y) => if (x.component.length > y.component.length) x else y)
      // convert references to strings relative to the module
        .serialize.dropWhile(_ != '>').tail
    }
```

src/main/scala/firrtl/passes/NewLowerTypes.scala
```diff
@@ -0,0 +1,439 @@
...
+
+  private def destruct(m: ModuleTarget, field: Field, rename: Option[RenameNode]): Seq[(Field, Seq[ReferenceTarget])] =
+    destruct(prefix = "", oldParent = ModuleParentRef(m), oldField = field, isVecField = false, rename = rename)
+
+
+  private def destruct(prefix: String, oldParent: ParentRef, oldField: Field,
```
azidar, line 365: Some scaladoc would be helpful here to understand what's happening, especially in the larger context of where its called.

src/main/scala/firrtl/passes/NewLowerTypes.scala
```diff
@@ -0,0 +1,439 @@
...
+  private def getFields(tpe: Type): Seq[Field] = tpe match {
+    case BundleType(fields) => fields
+    case v : VectorType => vecToBundle(v).fields
+  }
+
+  private def vecToBundle(v: VectorType): BundleType = {
+    BundleType(( 0 until v.size).map(i => Field(i.toString, Default, v.tpe)))
+  }
+
+  private trait ParentRef { def ref(name: String, asVecField: Boolean = false): ReferenceTarget }
```
azidar, line 431: Please add some explanation of why these are necessary, and what they represent.

src/main/scala/firrtl/transforms/RemoveWires.scala
ekiwi-sifive: I did that while debugging. I can move it to a separate PR.

src/main/scala/firrtl/passes/NewLowerTypes.scala
ekiwi-sifive: I haven't tested it but I haven't been able to find a fundamental reason why it shouldn't. As long as all references are lowered to ground types (ExpandConnects) and there are no SubAccess nodes, I think we should be fine.

src/main/scala/firrtl/passes/NewLowerTypes.scala
ekiwi-sifive: great catch!

src/main/scala/firrtl/passes/NewLowerTypes.scala
ekiwi-sifive: I was trying to do that, but it messes with a lot of old code and requires you to insert quite some `asInstanceOf[]` casts. So yeah, maybe in the future.

src/main/scala/firrtl/transforms/RemoveWires.scala
ekiwi-sifive: Now a separate PR: #1797

ekiwi-sifive:
@seldridge , What exactly is the `Legacy Custom Transforms` behavior in `LoweringCompilerSpec` testing?
Is  it ok to add patches there? I am a little hesitant because so far no one else seems to have had to add patches.

src/main/scala/firrtl/passes/NewLowerTypes.scala
```diff
@@ -0,0 +1,454 @@
...
+    destruct(prefix = "", oldParent = ModuleParentRef(m), oldField = field, isVecField = false, rename = rename)
+
+  /** Lowers a field into its ground type fields.
+    * @param prefix carries the prefix of the new ground type name
+    * @param isVecField is used to generate an appropriate old (field/index) reference
+    * @param rename The information from the `uniquify` function is consumed to appropriately rename generated fields.
+    * @return a sequence of ground type fields with new names and, for each field,
+    *         a sequence of old references that should to be renamed to point to the particular field
+    */
+  private def destruct(prefix: String, oldParent: ParentRef, oldField: Field,
```
azidar, line 376: I'd add a comment that this function presumes the fields have already been uniquified.

seldridge:
The `Legacy Custom Transforms` are testing where a custom transform, that uses `inputForm`/`outputForm` to specify dependencies, is placed in the transform sequence. Internally, `inputForm`/`outputForm` are being converted to dependencies and the resulting scheduling behavior needs to match how the old `getLoweringCompilers`/`mergeTransforms` methods worked. 

These tests aren't intended to be patched. (But they might need to be modified if in-tree transforms start specifying very sparse dependencies.)

By example: if you have `inputForm=LowForm`/`outputForm=MidForm`, you'd expect to see the transform run after low form optimizations, and then re-run from mid form to low form optimizations. [The test manually constructs this order and then compares it against the order if you just requested low form optimizations and the custom transform as targets.](https://github.com/freechipsproject/firrtl/blob/master/src/test/scala/firrtlTests/LoweringCompilersSpec.scala#L316)

This is mostly straightforward except for the edge cases identified in: https://github.com/freechipsproject/firrtl/issues/1165. The `outputForm=HighForm` edge cases are tested in `LoweringCompilerSpec.scala`. The edge cases of `inputForm=LowForm` are tested in `CustomTransformSpec.scala`.

src/test/scala/firrtl/passes/LowerTypesSpec.scala
```diff
@@ -0,0 +1,538 @@
...
+    val renames = RenameMap()
+    val mutableSet = scala.collection.mutable.HashSet[String]() ++ namespace
+    val res = DestructTypes.destruct(m, ref, mutableSet, renames)
+    DestructResult(resultToFieldSeq(res), renames)
+  }
+  def resultToFieldSeq(res: Seq[(firrtl.ir.Field, String)]): Seq[String] =
+    res.map(_._1).map(r => s"${r.flip.serialize}${r.name} : ${r.tpe.serialize}")
+  def get(r: RenameMap, m: IsMember): Set[IsMember] = r.get(m).get.toSet
+  protected val m = CircuitTarget("m").module("m")
+}
```
azidar, line 538: newline

src/test/scala/firrtlTests/LoweringCompilersSpec.scala
```diff
@@ -120,6 +120,8 @@ class LoweringCompilersSpec extends FlatSpec with Matchers {
 
     val patched = scala.collection.immutable.TreeMap(m.toArray:_*).values.flatten
 
+    println(patched.map(_.getName).mkString("\n"))
```
azidar, line 123: I presume this println will get removed after the tests pass?

ekiwi-sifive:
Thanks for the detailed explanation Schuyler!

> resulting scheduling behavior needs to match how the old `getLoweringCompilers`/`mergeTransforms` methods worked.

I combined `Uniquify` and `LowerTypes` into a single `NewLowerTypes` pass. How should that be reflected in the tests?

src/test/scala/firrtlTests/LoweringCompilersSpec.scala
ekiwi-sifive: Good find! This was never supposed to be committed!

src/main/scala/firrtl/passes/NewLowerTypes.scala
ekiwi-sifive: The fields haven't really been uniquified quite yet. The uniquifying information has been calculated, but we actually apply it in this function (which is what the `rename` parameter is for).
Is there a way I could clarify that?

src/main/scala/firrtl/passes/NewLowerTypes.scala
```diff
@@ -0,0 +1,454 @@
...
+import scala.collection.mutable
+
+/** Flattens Bundles and Vecs.
+  * - Some implicit bundle types remain, but with a limited depth:
+  *   - the type of a memory is still a bundle with depth 2 (mem -> port -> field), see [[MemPortUtils.memType]]
+  *   - the type of a module instance is still a bundle with depth 1 (instance -> port)
+  */
+object NewLowerTypes extends Transform {
+  override def inputForm: CircuitForm = UnknownForm
+  override def outputForm: CircuitForm = UnknownForm
```
seldridge, line 21: Mixin `with DependencyAPIMigration` and remove deprecated `inputForm`/`outputForm`.

ekiwi-sifive: Will do. I didn't do this initially since the word *Migration* suggested to me that this is only for migrating legacy passes.

seldridge:
The custom transform `inputForm=MidForm`/`outputForm=HighForm` test is doing something weird, but this looks like it's a problem with the `DependencyManager`.

Specifically, after the custom transform runs, the state is high form. `NewLowerTypes` is scheduled to run next so there's a state mismatch. The dependency manager solves a sub-problem to meet the prerequisites of `NewLowerTypes`. However, that is a minimal set of four transforms. The `DependencyManager` then thinks that it's only added those four transforms to the state, but it's actually added a ton of other stuff, like `ToWorkingIR`.

Then, after `NewLowerTypes` runs, the dependency manager re-runs `ToWorkingIR` and a ton of other stuff.

In summary, the `targets` for a dependency manager is the minimal set of things that it does, but it may actually do more work.

src/main/scala/firrtl/passes/NewLowerTypes.scala
```diff
@@ -0,0 +1,466 @@
...
+import firrtl.stage.TransformManager.TransformDependency
+
+import scala.collection.mutable
+
+/** Flattens Bundles and Vecs.
+  * - Some implicit bundle types remain, but with a limited depth:
+  *   - the type of a memory is still a bundle with depth 2 (mem -> port -> field), see [[MemPortUtils.memType]]
+  *   - the type of a module instance is still a bundle with depth 1 (instance -> port)
+  */
+object NewLowerTypes extends Transform with DependencyAPIMigration {
```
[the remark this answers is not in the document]
ekiwi-sifive: The plan was to eventually just name it `LowerTypes` after we make sure that it provides the same public methods as `LowerTypes`.

src/main/scala/firrtl/passes/NewLowerTypes.scala
```diff
@@ -0,0 +1,466 @@
...
+  def loweredName(s: Seq[String]): String = s.mkString(delim)
+
+  override def execute(state: CircuitState): CircuitState = {
+    // When memories are lowered to ground type, we have to fix the init annotation or error on it.
+    val (memInitAnnos, otherAnnos) = state.annotations.partition {
+      case _: MemoryRandomInitAnnotation => false
+      case _: MemoryInitAnnotation => true
+      case _ => false
+    }
+    val memInitByModule = memInitAnnos.map(_.asInstanceOf[MemoryInitAnnotation]).groupBy(_.target.encapsulatingModule)
```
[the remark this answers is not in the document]
ekiwi-sifive: Currently we need to know these annotations in order to throw an error if they are applied to a non-ground type memory (this is a feature that was already in the old `LowerTypes` pass).
I was planning to add custom logic to make these annotations work with non-ground type memories, but that does require me to have information about all the memories that the original memory gets split up into. Not sure how I would get the information in the `duplicate` method of the annotation.

src/main/scala/firrtl/passes/NewLowerTypes.scala
```diff
@@ -0,0 +1,466 @@
+// See LICENSE for license details.
+
+package firrtl.passes
+
+import firrtl.analyses.InstanceKeyGraph
+import firrtl.annotations.{CircuitTarget, MemoryInitAnnotation, MemoryRandomInitAnnotation, ModuleTarget, ReferenceTarget}
+import firrtl.{CircuitForm, CircuitState, DependencyAPIMigration, InstanceKind, Kind, MemKind, PortKind, RenameMap, SymbolTable, Transform, UnknownForm, Utils}
```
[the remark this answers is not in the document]
ekiwi-sifive: I wanted to be verbose about it and its not much of a problem with an IDE. If you want, I can change it to `import firrtl._`.

src/main/scala/firrtl/passes/NewLowerTypes.scala
```diff
@@ -0,0 +1,466 @@
...
+  /** Expands a chain of referential [[firrtl.ir.Expression]]s into the equivalent lowered name
+    * @param e [[firrtl.ir.Expression]] made up of _only_ [[firrtl.WRef]], [[firrtl.WSubField]], and [[firrtl.WSubIndex]]
+    * @return Lowered name of e
+    * @note Please make sure that there will be no name collisions when you use this outside of the context of LowerTypes!
+    */
+  def loweredName(e: Expression): String = e match {
+    case e: Reference => e.name
+    case e: SubField => s"${loweredName(e.expr)}$delim${e.name}"
+    case e: SubIndex => s"${loweredName(e.expr)}$delim${e.value}"
+  }
```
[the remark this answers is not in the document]
ekiwi-sifive: I copied this function from `LowerTypes` because it is used by other passes, so I cannot change the argument type, since I want to try and replace `LowerTypes` without a deprecation cycle. However I think it is a good idea to add a better error message than a match error.

ekiwi-sifive:
> @seldridge this is a good example where external transforms that depend on `Uniquify` or `LowerTypes` may have awkward interactions (although I guess it should at least "work").

The plan is to delete the old `LowerTypes` transform and rename the `NewLowerTypes` to just `RenameTypes`. I just haven't done that yet because I need to go through all the public fields of `LowerTypes` and make sure we provide compatible replacements. `Uniquify` will stay around and be deprecated. So now if you have a custom transform that only depends on `Uniquify`, the deprecated version of `Uniquify` will be run. If your transform depends on both, `LowerTypes` and `Uniquify`, the run of `Uniquify` will be redundant, but in general nothing should break if the new version of `LowerTypes` tries to uniquify again.

Let me know if that transition plan sounds reasonable to you.

seldridge:
@jackkoenig wrote:
> @seldridge this is a good example where external transforms that depend on Uniquify or LowerTypes may have awkward interactions (although I guess it should at least "work"). How hard do you think it would be to support an implements or provides API?

I don't think this is too bad and it's probably the way that we need to go in the mid-term. This is also a mechanism to wholly replace in-tree transforms with custom implementations.

I think this would look like a special transform which, after application, causes a "property" to be applied to the circuit. Dependencies can then be specified either as explicit transforms or as `Provider[A <: Property]`. If it's only requesting a property, then the dependency manager looks around to try and find all transforms that can provide the property. There's then some API for providing an ordering of providers. The dependency manager then either succeeds and will always choose one provider for that property. If it fails to find a provider then it fails. Somewhat annoyingly, I expect this requires using META-INF which isn't amazing.

tl;dr: I think we can do this with no modification to the actual dependency API, just with a new mix-in.

This would be a separate endeavor from this PR.

ekiwi-sifive:
So right now there are four public methods of the old `LowerTypes` pass that I do not intend to provide with the `NewLowerTypes` pass:

```.scala
  def renameExps(renames: RenameMap, n: String, t: Type, root: String): Seq[String]
  def renameExps(renames: RenameMap, n: String, t: Type): Seq[String]
  def renameExps(renames: RenameMap, e: Expression, root: String): Seq[String]
  def lowerTypes(renames: RenameMap, initializedMems: Set[(String, String)])(m: DefModule): DefModule
```

They are never used outside the `LowerTypes` pass in the `firrtl` code base and were probably made public by accident.

Would it be acceptable to drop these methods without a deprecation cycle? This would allow us to just overwrite the old `LowerTypes` pass with the new one.

ekiwi-sifive:
The only test failing now is `LoweringCompilerSpec`, in particular `Legacy Custom Transforms` for `Mid -> High` and `Mid -> Chirrtl`. I think I will need @seldridge 's help to fix that.

azidar:
LGTM.

Issue #1751, opened 2020-07-09
Possible issue with mux verilog emission


albertchen-sifive:
### Checklist

- [x] Did you specify the current behavior?
- [x] Did you specify the expected behavior?
- [x] Did you provide a code example showing the problem?
- [x] Did you describe your environment?
- [x] Did you specify relevant external information?

### What is the current behavior?
for the following firrtl
```
circuit foo :
  module foo :
    output ref : SInt<30>
    ref <= mux(UInt<1>("h0"), SInt<30>("h149b138d"), neg(shr(UInt<31>("h7c0c2fb8"), 2)))
```

the minimum verilog compiler produces
```
module foo_ref(
  output [29:0] ref_
);
  assign ref_ = 1'h0 ? $signed(30'sh149b138d) : $signed(-{$signed(29'h1f030bee)});
endmodule
```

and the verilog compiler with constant propagation produces
```
module foo_custom(
  output [29:0] ref_
);
  assign ref_ = -{$signed(29'h1f030bee)};
endmodule
```

the yosys equivalence test for these circuits fails. I checked that reducing the test cases to just `assign ref_ = $signed(-{$signed(29'h1f030bee)})` and `assign ref_ = -{$signed(29'h1f030bee)` does not pass equivalence. so it seems like the extra `$signed()` is the issue? no idea why though

### What is the expected behavior?
ConstantPropagation should produce equivalent verilog for this firrtl input.

### Steps to Reproduce

run the following test case
```scala
class Foo extends FirrtlFlatSpec {
  private val srcDir = "/constant_propagation_tests"
  private val transforms = Seq(new ConstantPropagation)
  "foo" should "be equivalent" in {
    val input =
      s"""circuit foo :
         |  module foo :
         |    output ref : SInt<30>
         |    ref <= mux(UInt<1>("h0"), SInt<30>("h149b138d"), neg(shr(UInt<31>("h7c0c2fb8"), 2)))
         |""".stripMargin
    firrtlEquivalenceTest(input, transforms)
  }
}
```

### Your environment

- Chisel Verions: 
- OS: 
- Verilator version:

PR #1785, opened 2020-07-23
Fix sign extension issue in Emitter

albertchen-sifive:
fixes https://github.com/freechipsproject/firrtl/issues/1751

Here is what I think is happening based on my interpretation of the verilog spec and the yosys results. I could be completely wrong though.

In the module below, `ref_` is equivalent to `3'sh1`:  `-$signed(2'h3)` evaluates to `2'sh1` and then is sign extended to `3'sh1`.
```verilog
module SignTester_ref(
  output [2:0] ref_
);
  assign ref_ = -$signed(2'h3);
endmodule
```

however in this module, `ref_` is equivalent to `3'sh5`:  in a context-determined expression with width 3, `$signed(2'h3)` would be sign-extended to `3sh7`, but since all operands within `{...}` are self-determined, its width is 2 instead of 3. And since the result of `{...}` is always unsigned the result is zero-extended, so `{$signed(2'h3)}` evaluates to `3'h3`, and when it is negated it becomes `3'sh5` 
```verilog
module SignTester_ref(
  output [2:0] ref_
);
  assign ref_ = -{$signed(2'h3)};
endmodule
```

`DoPrim(Neg, ...)` of a literal is not constant propagated, so in most cases the `verilog` and `mverilog` compilers emit equivalent, incorrect code. However in the case of `Mux`, the true and false cases are wrapped in another `$signed` call.
```verilog
module SignTester_ref(
  output [2:0] ref_
);
  assign ref_ = 1'h0 ? $signed(3'sh0) : $signed(-{$signed(2'h3)});
endmodule
```
In this case `ref_` is equivalent to `3sh'1`, because the argument of `$signed(_)` is self-determined so `{$signed(2'h3)}` does not get zero-extended to `3h'3` and instead evaluates to `2h'3`, which is -1 when converted to a signed number. so the operand of `-` evaluates to `2sh'3` which then gives `2sh'1`. so the outer `$signed` recieves `2'h1` and evaluates to `3'h1` since it is a context-determined expression and `ref_` has width 3.

This PR changes the emission of `Neg(a0)` from `-{a0}` to `-a0`, which allows `a0` to be sign-extended in context-determined expressions rather than always zero-extending.

### Contributor Checklist

#### Type of Improvement

- bug fix

#### API Impact

none

instead of emitting `-{a0}` for the firrtl expression `neg(a0)`, emit `-a0`

#### Desired Merge Strategy

- Squash

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -398,7 +398,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
          error("Verilog emitter does not support SHIFT_RIGHT >= arg width")
        case Shr if c0 == (bitWidth(a0.tpe)-1) => Seq(a0,"[", bitWidth(a0.tpe) - 1, "]")
        case Shr => Seq(a0,"[", bitWidth(a0.tpe) - 1, ":", c0, "]")
-       case Neg => Seq("-{", cast(a0), "}")
+       case Neg => Seq("-", cast(a0))
        case Cvt => a0.tpe match {
          case (_: UIntType) => Seq("{1'b0,", cast(a0), "}")
          case (_: SIntType) => Seq(cast(a0))
--- a/src/test/scala/firrtlTests/ConstantPropagationTests.scala
+++ b/src/test/scala/firrtlTests/ConstantPropagationTests.scala
@@ -1614,4 +1614,14 @@ class ConstantPropagationEquivalenceSpec extends FirrtlFlatSpec {
           |""".stripMargin
      firrtlEquivalenceTest(input, transforms)
    }
+
+  "propagation of signed expressions" should "have the correct signs" in {
+    val input =
+      s"""circuit SignTester :
+         |  module SignTester :
+         |    output ref : SInt<3>
+         |    ref <= mux(UInt<1>("h0"), SInt<3>("h0"), neg(UInt<2>("h3")))
+         |""".stripMargin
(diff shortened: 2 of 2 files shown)


albertchen-sifive:
> This one is a little bit scary, is the implication that this just never worked at all for signed types? Or does it only manifest on signed literals?

I think manifests for all signed types when the lhs has a larger width.

> Does this have any affect for unsigned?

no

albert-magyar:
It seems like the underlying problem here is actually that `neg` is always growing (it adds 1 to the width), whereas the Verilog emission is (among other things) treating it like a wrapping operator. It's easy to forget that `neg` is growing.

albert-magyar:
> It seems like the underlying problem here is actually that `neg` is always growing (it adds 1 to the width), whereas the Verilog emission is (among other things) treating it like a wrapping operator. It's easy to forget that `neg` is growing.

And what I meant to add to that: this makes the solution of dropping the Verilog cat `{}` sound perfectly logical.

albert-magyar:
> > This one is a little bit scary, is the implication that this just never worked at all for signed types? Or does it only manifest on signed literals?
> 
> I think manifests for all signed types when the lhs has a larger width.

Expanding on this, the LHS _always_ has a larger width, so this does indeed mean it was broken for all signed cases.

albertchen-sifive:
@Mergifyio backport 1.2.x. 1.3.x

albertchen-sifive:
@Mergifyio backport 1.2.x

PR #1791, opened 2020-07-25
Add deprecations for more firrtl.Compiler methods

### LINKED ISSUE #1790: Misleading deprecation warning for VerilogCompiler
Working on a `PresetSpec`, I'm getting a deprec warn for the following line: `(new VerilogCompiler).compileAndEmit(CircuitState(parse(input), ChirrtlForm, annos), List.empty)`

```bash
class VerilogCompiler in package firrtl is deprecated (since FIRRTL 1.3): Use stage.{FirrtlStage, FirrtlMain} stage.transforms.Compiler(Dependency[VerilogEmitter])
```
However, when I substitute it with `(new firrtl.stage.transforms.Compiler(Dependency[VerilogEmitter])).compileAndEmit(CircuitState(parse(input), ChirrtlForm, annos), List.empty)`, compiler yields an error `value compileAndEmit is not a member of firrtl.stage.transforms.Compiler`

This should be changed. Deprec warnings should point to valid solutions, which don't end with a compilation errors

seldridge:
Adds missing deprecation warnings to firrtl.Compiler methods: compile
and compileAndEmit. This provides a more informative message to the
user about how to actually migrate. Without this, the user only has
the generic deprecation warning about class Compiler.

This will now produce two deprecation warnings if you use the `Compiler.compileAndEmit` method. Consider:

```scala
class Baz extends FlatSpec with Matchers {
  (new firrtl.VerilogCompiler).compileAndEmit(???, ???)
}
```

With this PR, you now get the following two warnings:

```
[warn] /home/schuyler/repos/github.com/freechipsproject/firrtl/src/test/scala/firrtlTests/Foo.scala:47:32: method compileAndEmit in trait Compiler is deprecated (since FIRRTL 1.3): Migrate to '(new FirrtlStage).execute(args: Array[String], annotations: AnnotationSeq)'.This will be removed in 1.4.
[warn]   (new firrtl.VerilogCompiler).compileAndEmit(???, ???)
[warn]                                ^
[warn] /home/schuyler/repos/github.com/freechipsproject/firrtl/src/test/scala/firrtlTests/Foo.scala:47:15: class VerilogCompiler in package firrtl is deprecated (since FIRRTL 1.3): Use stage.{FirrtlStage, FirrtlMain} stage.transforms.Compiler(Dependency[VerilogEmitter])
[warn]   (new firrtl.VerilogCompiler).compileAndEmit(???, ???)
[warn]               ^
```

Previously, you would only get the latter.

Fixes #1790.

### Contributor Checklist

- [n/a] Did you add at least one test demonstrating the PR?

#### Type of Improvement

- documentation

#### API Impact

None.

None.

#### Desired Merge Strategy

#### Release Notes

None.

- [ ] Did you review?


tampler:
Thanks a lot for this fix!

src/main/scala/firrtl/Compiler.scala
```diff
@@ -540,7 +540,10 @@ trait Compiler extends Transform with DependencyAPIMigration {
     * @param customTransforms Any custom [[Transform]]s that will be inserted
     *   into the compilation process by [[CompilerUtils.mergeTransforms]]
     */
-  @deprecated("Please use compileAndEmit or other compile method instead", "firrtl 1.0")
+  @deprecated(
+    "Migrate to '(new FirrtlStage).execute(args: Array[String], annotations: AnnotationSeq)'." +
+      "This will be removed in 1.4.",
```
ekiwi-sifive, line 545: Not `1.5` since `1.3` is already released? Shouldn't the deprecation go through one full release cycle?

src/main/scala/firrtl/Compiler.scala
```diff
@@ -576,6 +583,10 @@ trait Compiler extends Transform with DependencyAPIMigration {
     *   process by [[CompilerUtils.mergeTransforms]]
     * @return result of compilation
     */
+  @deprecated(
+    "Migrate to '(new FirrtlStage).execute(args: Array[String], annotations: AnnotationSeq)'." +
```
ekiwi-sifive, line 587: Can you clarify what the arguments are to get the behavior of `compile`?

src/main/scala/firrtl/Compiler.scala
seldridge: I was hoping that this would be fine. Since all `Compiler`s are deprecated in 1.3 and to-be-removed in 1.4 (though this may not happen...), I figured it would be fine to retcon deprecations on their methods just to make it more clear about the migration path.

What do you think?

More info: removing `Compiler` is a serious pain and I'm not sure I have the bandwidth before 1.4 goes out. This requires removing `CompilerAnnotation` (which isn't deprecated) to using only `RunFirrtlTransformAnnotation`s. E.g., `-X <compiler>` should map to dependencies and not a compiler.

I'm fine with 1.5. Though, that should probably bump the overall Compiler removal timeline, too.

ekiwi-sifive: Oh, ok, so this PR is mostly intended as a backport for `1.3` to clarify your intention?
Then that's fine!

src/main/scala/firrtl/Compiler.scala
seldridge: I started to write this up, but it's complicated... the correct behavior is really going to be dependent on the child that extends `Compiler`. E.g., `VerilogCompiler` has different arguments than `HighFirrtlCompiler`. We can override the methods in the compiler children with more verbose deprecation warnings that show exactly how to invoke these. 

Alternatively, we can add a link to a gist that shows how to migrate (like with `DependencyAPIMigration`).

I don't personally think that there's huge benefit here as most people that are using the actual compilers are developers. The main API that I think people use is `ChiselStage`/`FirrtlStage`.

What do you think?

ekiwi-sifive: I think the code that I would use to replace this is `new TransformManager(Seq(Dependency(Forms.LowForm), Dependency(CustomTransform)).execute(state)`. Would that be correct?

seldridge: For `LowFirrtlEmitter` that looks perfect.

src/main/scala/firrtl/Compiler.scala
seldridge: Yeah, I was planning to get this into 1.3.3 (I updated the deprecation messages to more accurately indicate "since FIRRTL 1.3.3" for the other two deprecations that are new).

PR #1792, opened 2020-07-26
Added empty annotation to AnnotationSeq

tampler:
Oftentimes, you don't wanna specify any annotation, but you need to use the code, which requires `AnnotationSeq` on input.
An example is a Verilog emitter with the following call:

```scala
class Emiter() {

  private lazy val stage = new chisel3.stage.ChiselStage

  def emit(path: String, ann: AnnotationSeq, form: Form): Object = form match {
    case Verilog => stage.execute((Array("-td", path, "-X", "verilog")), ann)
    case High    => stage.execute((Array("-td", path, "-X", "high")), ann)
    case Low     => stage.execute((Array("-td", path, "-X", "low")), ann)
    case _       => new Exception("Unsuppored output format")
  }
}
```

To call this, one needs to do the following: 
```scala
object Emit extends App {
  val emiter = new Emiter {}
  emiter.emit("out", AnnotationSeq(Seq.empty[Annotation]), Verilog)
}
```
Which can be considerably simplified with the following call: `emiter.emit("out", AnnotationSeq.empty, Verilog)`

So this PR Adds `AnnotationSeq.empty`

### DIFF
--- a/src/main/scala/firrtl/Compiler.scala
+++ b/src/main/scala/firrtl/Compiler.scala
@@ -18,10 +18,13 @@ import firrtl.stage.Forms
 
 /** Container of all annotations for a Firrtl compiler */
 class AnnotationSeq private (private[firrtl] val underlying: List[Annotation]) {
-  def toSeq: Seq[Annotation] = underlying.toSeq
+  def toSeq: Seq[Annotation] = underlying.toSeq  
 }
 object AnnotationSeq {
   def apply(xs: Seq[Annotation]): AnnotationSeq = new AnnotationSeq(xs.toList)
+
+  // Empty Annotation list for integration, debug and testing
+  val empty = Seq.empty[Annotation]
 }
 
 /** Current State of the Circuit


seldridge:
One possible alternative: does `Seq.empty` or `Nil` already fill this role adequately?

For the situation in the example code, either of the following are options:

```scala
emiter.emit("out", Seq.empty, Verilog)
emiter.emit("out", Nil, Verilog)
```

tampler:
`Seq.empty` won't compile, since it requires a type to construct an empty sequence, thus `Seq.empty[T]` should work
`Nil` is smth used in lists and may be confusing to use in annotations.
`AnnotationSeq` is a type and an object. I see a perfect case for an "empty" object, which is similar to an `identity` function, commonly used in FP.

Maybe other opinions should be here. Anyways, `AnnotationSeq.empty` would compile, but fail at runtime. Thus, a common use case is  refactoring and testing

seldridge:
Doesn't the type get inferred for `Seq.empty` in that case?

Something like: https://scastie.scala-lang.org/6xDde5GZTRWwLDWfQB0uNg

tampler:
Oh yeah, Seq.empty indeed gets inferred. Should I close this ?

tampler:
Thanks! Seems to be solved by current approaches ! THanks @seldridge

seldridge:
Sounds good @tampler. I'm glad you have a solution that works. And... I'm not necessarily against this, just seeing if the current approaches are sufficient.

PR #1796, opened 2020-07-28
Scala 2.13 support

ekiwi-sifive:
Based on @tampler's pioneering work!

### Contributor Checklist

- [x] Did you review?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: scala
 sudo: false
 
+jdk: openjdk8
+
 cache:
   directories:
     $HOME/.cache/coursier
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,6 +61,13 @@ jobs:
       script:
         - sbt $SBT_ARGS +unidoc
     - stage: test
+      name: "Tests: FIRRTL (2.13)"
+      workspaces:
+        use: firrtl_build
+      script:
+        - verilator --version
+        - sbt ++2.13.2 $SBT_ARGS test
+    - stage: test
       name: "Tests: FIRRTL (2.12)"
       workspaces:
         use: firrtl_build
--- a/build.sbt
+++ b/build.sbt
@@ -23,12 +23,14 @@ lazy val commonSettings = Seq(
   version := "1.4-SNAPSHOT",
   scalaVersion := "2.12.11",
   crossScalaVersions := Seq("2.13.2", "2.12.11", "2.11.12"),
+  addCompilerPlugin(scalafixSemanticdb),
   scalacOptions := Seq(
     "-deprecation",
     "-unchecked",
     "-language:reflectiveCalls",
     "-language:existentials",
     "-language:implicitConversions",
+    "-Yrangepos",          // required by SemanticDB compiler plugin
   ),
   javacOptions ++= javacOptionsVersion(scalaVersion.value),
   libraryDependencies ++= Seq(
--- a/build.sbt
+++ b/build.sbt
@@ -34,16 +34,13 @@ lazy val commonSettings = Seq(
   name := "firrtl",
   version := "1.4-SNAPSHOT",
   scalaVersion := "2.12.11",
-  crossScalaVersions := Seq("2.12.11", "2.11.12"),
-  addCompilerPlugin(scalafixSemanticdb),
+  crossScalaVersions := Seq("2.13.2", "2.12.11", "2.11.12"),
   scalacOptions := scalacOptionsVersion(scalaVersion.value) ++ Seq(
     "-deprecation",
     "-unchecked",
     "-language:reflectiveCalls",
     "-language:existentials",
     "-language:implicitConversions",
-    "-Yrangepos",          // required by SemanticDB compiler plugin
-    "-Ywarn-unused-import" // required by `RemoveUnused` rule
   ),
   javacOptions ++= javacOptionsVersion(scalaVersion.value),
   libraryDependencies ++= Seq(
--- a/build.sbt
+++ b/build.sbt
@@ -52,6 +52,13 @@ lazy val commonSettings = Seq(
     "org.json4s" %% "json4s-native" % "3.6.8",
     "org.apache.commons" % "commons-text" % "1.8"
   ),
+  // starting with scala 2.13 the parallel collections are separate from the standard library
+  libraryDependencies ++= {
+    CrossVersion.partialVersion(scalaVersion.value) match {
+      case Some((2, major)) if major <= 12 => Seq()
+      case _ => Seq("org.scala-lang.modules" %% "scala-parallel-collections" % "0.2.0")
+    }
+  },
   resolvers ++= Seq(
     Resolver.sonatypeRepo("snapshots"),
     Resolver.sonatypeRepo("releases")
--- a/build.sbt
+++ b/build.sbt
@@ -2,18 +2,6 @@
 
 enablePlugins(SiteScaladocPlugin)
 
-def scalacOptionsVersion(scalaVersion: String): Seq[String] = {
-  Seq() ++ {
-    // If we're building with Scala > 2.11, enable the compile option
-    //  switch to support our anonymous Bundle definitions:
-    //  https://github.com/scala/bug/issues/10047
-    CrossVersion.partialVersion(scalaVersion) match {
-      case Some((2, scalaMajor: Long)) if scalaMajor < 12 => Seq()
-      case _ => Seq("-Xsource:2.11")
-    }
-  }
-}
-
 def javacOptionsVersion(scalaVersion: String): Seq[String] = {
   Seq() ++ {
     // Scala 2.12 requires Java 8, but we continue to generate
@@ -35,7 +23,7 @@ lazy val commonSettings = Seq(
   version := "1.4-SNAPSHOT",
   scalaVersion := "2.12.11",
   crossScalaVersions := Seq("2.13.2", "2.12.11", "2.11.12"),
-  scalacOptions := scalacOptionsVersion(scalaVersion.value) ++ Seq(
+  scalacOptions := Seq(
     "-deprecation",
     "-unchecked",
     "-language:reflectiveCalls",
@@ -169,7 +157,7 @@ lazy val docSettings = Seq(
         }
       s"https://github.com/freechipsproject/firrtl/tree/$branch€{FILE_PATH}.scala"
     }
-  ) ++ scalacOptionsVersion(scalaVersion.value) ++ scalacDocOptionsVersion(scalaVersion.value)
+  ) ++ scalacDocOptionsVersion(scalaVersion.value)
 )
 
 lazy val firrtl = (project in file("."))
--- a/src/main/scala/firrtl/Compiler.scala
+++ b/src/main/scala/firrtl/Compiler.scala
@@ -449,7 +449,7 @@ object CompilerUtils extends LazyLogging {
         case ChirrtlForm =>
           Seq(new ChirrtlToHighFirrtl) ++ getLoweringTransforms(HighForm, outputForm)
         case HighForm =>
-          Seq(new IRToWorkingIR, new ResolveAndCheck, new transforms.DedupModules, new HighFirrtlToMiddleFirrtl) ++
+          Seq(new IRToWorkingIR, new ResolveAndCheck, new firrtl.transforms.DedupModules, new HighFirrtlToMiddleFirrtl) ++
             getLoweringTransforms(MidForm, outputForm)
         case MidForm => Seq(new MiddleFirrtlToLowFirrtl) ++ getLoweringTransforms(LowForm, outputForm)
         case LowForm => throwInternalError("getLoweringTransforms - LowForm") // should be caught by if above
--- a/src/main/scala/firrtl/options/OptionParser.scala
+++ b/src/main/scala/firrtl/options/OptionParser.scala
@@ -11,7 +11,7 @@ case object OptionsHelpException extends Exception("Usage help invoked")
 /** OptionParser mixin that causes the OptionParser to not call exit (call `sys.exit`) if the `--help` option is
   * passed */
 trait DoNotTerminateOnExit { this: OptionParser[_] =>
-  override def terminate(exitState: Either[String, Unit]): Unit = Unit
+  override def terminate(exitState: Either[String, Unit]): Unit = ()
 }
 
 /** OptionParser mixin that converts to [[OptionsException]]
@@ -28,7 +28,7 @@ trait ExceptOnError { this: OptionParser[_] =>
   */
 trait DuplicateHandling extends OptionParser[AnnotationSeq] {
 
-  override def parse(args: Seq[String], init: AnnotationSeq): Option[AnnotationSeq] = {
+  override def parse(args: scala.collection.Seq[String], init: AnnotationSeq): Option[AnnotationSeq] = {
 
     /** Message for found duplicate options */
     def msg(x: String, y: String) = s"""Duplicate $x "$y" (did your custom Transform or OptionsManager add this?)"""
--- a/src/main/scala/firrtl/passes/CheckHighForm.scala
+++ b/src/main/scala/firrtl/passes/CheckHighForm.scala
@@ -96,7 +96,7 @@ trait CheckHighFormLike { this: Pass =>
 
     val intModuleNames = c.modules.view.collect({ case m: Module => m.name }).toSet
 
-    c.modules.view.groupBy(_.name).filter(_._2.length > 1).flatMap(_._2).foreach {
+    c.modules.groupBy(_.name).filter(_._2.length > 1).flatMap(_._2).foreach {
       m => errors.append(new ModuleNameNotUniqueException(m.info, m.name))
     }
 
--- a/src/main/scala/firrtl/passes/RemoveCHIRRTL.scala
+++ b/src/main/scala/firrtl/passes/RemoveCHIRRTL.scala
@@ -10,8 +10,6 @@ import firrtl.Utils._
 import firrtl.Mappers._
 import firrtl.options.Dependency
 
-import scala.collection.mutable
-
 case class MPort(name: String, clk: Expression)
 case class MPorts(readers: ArrayBuffer[MPort], writers: ArrayBuffer[MPort], readwriters: ArrayBuffer[MPort])
 case class DataRef(exp: Expression, source: String, sink: String, mask: String, rdwrite: Boolean)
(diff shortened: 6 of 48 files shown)


ekiwi-sifive:
The crash we are seeing in Travis might be related to this JDK bug: https://bugs.openjdk.java.net/browse/JDK-8208275

At least the crash is at the exact same address in the code.

sequencer:
question on how to PR a PR...
I'm going to add the `build.sc` support on this PR simultaneously.

ekiwi-sifive:
> question on how to PR a PR...

you can make a PR on my fork

>  I'm going to add the `build.sc` support on this PR simultaneously.

awesome!

build.sbt
```diff
@@ -34,16 +22,13 @@ lazy val commonSettings = Seq(
   name := "firrtl",
   version := "1.4-SNAPSHOT",
   scalaVersion := "2.12.11",
-  crossScalaVersions := Seq("2.12.11", "2.11.12"),
-  addCompilerPlugin(scalafixSemanticdb),
```
sequencer, line 38: Why do we remove `scalafixSemanticdb`? I think it's a not related change.

build.sbt
```diff
@@ -34,16 +22,13 @@ lazy val commonSettings = Seq(
...
-  scalacOptions := scalacOptionsVersion(scalaVersion.value) ++ Seq(
+  crossScalaVersions := Seq("2.13.2", "2.12.11", "2.11.12"),
+  scalacOptions := Seq(
     "-deprecation",
     "-unchecked",
     "-language:reflectiveCalls",
     "-language:existentials",
     "-language:implicitConversions",
-    "-Yrangepos",          // required by SemanticDB compiler plugin
-    "-Ywarn-unused-import" // required by `RemoveUnused` rule
```
sequencer, line 46: maybe unrelated too?

build.sbt
```diff
@@ -165,7 +157,7 @@ lazy val docSettings = Seq(
         }
       s"https://github.com/freechipsproject/firrtl/tree/$branch€{FILE_PATH}.scala"
     }
-  ) ++ scalacOptionsVersion(scalaVersion.value) ++ scalacDocOptionsVersion(scalaVersion.value)
```
sequencer, line 168: maybe unrelated too?

src/main/scala/firrtl/Compiler.scala
```diff
@@ -449,7 +449,7 @@ object CompilerUtils extends LazyLogging {
         case ChirrtlForm =>
           Seq(new ChirrtlToHighFirrtl) ++ getLoweringTransforms(HighForm, outputForm)
         case HighForm =>
-          Seq(new IRToWorkingIR, new ResolveAndCheck, new transforms.DedupModules, new HighFirrtlToMiddleFirrtl) ++
+          Seq(new IRToWorkingIR, new ResolveAndCheck, new firrtl.transforms.DedupModules, new HighFirrtlToMiddleFirrtl) ++
```
sequencer, line 452: why this happens? `Compiler` is in firrtl package, and `transform` is in the namespace, I think this line shouldn't be changed.

src/main/scala/firrtl/options/OptionParser.scala
```diff
@@ -28,7 +28,7 @@ trait ExceptOnError { this: OptionParser[_] =>
   */
 trait DuplicateHandling extends OptionParser[AnnotationSeq] {
 
-  override def parse(args: Seq[String], init: AnnotationSeq): Option[AnnotationSeq] = {
+  override def parse(args: scala.collection.Seq[String], init: AnnotationSeq): Option[AnnotationSeq] = {
```
sequencer, line 31: maybe unrelated too?

src/main/scala/firrtl/passes/CheckHighForm.scala
```diff
@@ -96,7 +96,7 @@ trait CheckHighFormLike { this: Pass =>
 
     val intModuleNames = c.modules.view.collect({ case m: Module => m.name }).toSet
 
-    c.modules.view.groupBy(_.name).filter(_._2.length > 1).flatMap(_._2).foreach {
+    c.modules.groupBy(_.name).filter(_._2.length > 1).flatMap(_._2).foreach {
```
sequencer, line 99: note of this: https://www.scala-lang.org/blog/2018/06/13/scala-213-collections.html

src/main/scala/firrtl/passes/RemoveCHIRRTL.scala
```diff
@@ -10,6 +10,8 @@ import firrtl.Utils._
 import firrtl.Mappers._
 import firrtl.options.Dependency
 
+import scala.collection.mutable
+
```
sequencer, line 14: why mutable here? originally it wasn't mutable.

src/main/scala/firrtl/passes/memlib/MemConf.scala
```diff
@@ -22,7 +22,7 @@ object MemPort {
     s.split(",").toSeq.map(MemPort.apply).map(_ match {
       case Some(x) => x
       case _ => throw new Exception(s"Error parsing MemPort string : ${s}")
-    }).groupBy(identity).mapValues(_.size)
+    }).groupBy(identity).mapValues(_.size).toMap
```
sequencer, line 25: should use `groupMap` here in 2.13.

src/main/scala/firrtl/passes/memlib/MemConf.scala
```diff
@@ -57,13 +57,13 @@ object MemConf {
   }
 
   def apply(name: String, depth: BigInt, width: Int, readPorts: Int, writePorts: Int, readWritePorts: Int, maskGranularity: Option[Int]): MemConf = {
-    val ports: Map[MemPort, Int] = (if (maskGranularity.isEmpty) {
-      (if (writePorts == 0) Map.empty[MemPort, Int] else Map(WritePort -> writePorts)) ++
-      (if (readWritePorts == 0) Map.empty[MemPort, Int] else Map(ReadWritePort -> readWritePorts))
+    val ports: Seq[(MemPort, Int)] = (if (maskGranularity.isEmpty) {
+      (if (writePorts == 0) Seq() else Seq(WritePort -> writePorts)) ++
+      (if (readWritePorts == 0) Seq() else Seq(ReadWritePort -> readWritePorts))
```
sequencer, line 62: not related too, but I think it should be `Seq` too

src/main/scala/firrtl/proto/FromProto.scala
```diff
@@ -8,7 +8,9 @@ import java.io.{File, FileInputStream, InputStream}
 import collection.JavaConverters._
 import FirrtlProtos._
 import com.google.protobuf.CodedInputStream
-import Firrtl.Statement.{ReadUnderWrite, Formal}
+import Firrtl.Statement.{Formal, ReadUnderWrite}
+
+import scala.collection.mutable
```
sequencer, line 13: why mutable here?

src/main/scala/firrtl/stage/phases/Compiler.scala
```diff
@@ -111,8 +111,9 @@ class Compiler extends Phase with Translator[AnnotationSeq, Seq[CompilerRun]] {
       c.copy(stateOut = Some(annotationsOut))
     }
 
-    if (b.size <= 1) { b.map(f)         }
-    else             { b.par.map(f).seq }
+    if (b.size <= 1) { b.map(f) } else {
+      collection.parallel.immutable.ParVector(b :_*).par.map(f).seq
```
sequencer, line 115: `import scala.collection.parallel.CollectionConverters._` takes a better look and won't need to change this line?

src/main/scala/firrtl/transforms/PropagatePresetAnnotations.scala
```diff
@@ -71,7 +71,7 @@ class PropagatePresetAnnotations extends Transform with DependencyAPIMigration {
     * @param presetAnnos all the annotations
     * @return updated annotations
     */
-  private def propagate(cs: CircuitState, presetAnnos: Seq[PresetAnnotation]): AnnotationSeq = {
+  private def propagate(cs: CircuitState, presetAnnos: Seq[PresetAnnotation], otherAnnos: Seq[Annotation]): AnnotationSeq = {
```
sequencer, line 74: not a good idea to change API here.

src/test/scala/firrtlTests/ParserSpec.scala
```diff
@@ -21,7 +21,7 @@ class ParserSpec extends FirrtlFlatSpec {
                      "readwriter" -> "c"
                     )
     def fieldsToSeq(m: Map[String, String]): Seq[String] =
-      m map { case (k,v) => s"      ${k} => ${v}" } toSeq
```
sequencer, line 24: no need to change here.

src/test/scala/firrtlTests/interval/IntervalMathSpec.scala
```diff
@@ -19,24 +19,24 @@ class IntervalMathSpec extends FirrtlFlatSpec {
   val DShiftLeftPattern   = """.*output dshl.*<(\d+)>.*""".r
   val DShiftRightPattern  = """.*output dshr.*<(\d+)>.*""".r
   val ArithAssignPattern = """\s*(\w+) <= asSInt\(bits\((\w+)\((.*)\).*\)\)\s*""".r
-  def getBound(bound: String, value: Double): IsKnown = bound match {
```
sequencer, line 22: not related changes?

src/test/scala/firrtlTests/options/PhaseManagerSpec.scala
```diff
@@ -392,7 +392,7 @@ class PhaseManagerSpec extends AnyFlatSpec with Matchers {
 
     /** Convert a Graphviz file to PNG using */
     def maybeToPng(f: File): Unit = try {
-      s"dot -Tpng -O ${f}" !
```
sequencer, line 395: no need to change here.

build.sbt
ekiwi-sifive: There wasn't a `semantic-db` package for `2.13`. We should re-add scalafix before merging this PR, but I do not know how to.

build.sbt
ekiwi-sifive: Should be added back once we get `scalafix` working with `2.13`

src/main/scala/firrtl/Compiler.scala
ekiwi-sifive: It doesn't compile if I do not change it. But yeah, honestly I do not know why...

ekiwi-sifive: Maybe `transforms` somehow gets overwritten

src/main/scala/firrtl/stage/phases/Compiler.scala
ekiwi-sifive: The problem is that `import scala.collection.parallel.CollectionConverters._` isn't available in `2.11`, `2.12` as far as I understand. See https://github.com/scala/scala-library-compat/issues/208

src/main/scala/firrtl/transforms/PropagatePresetAnnotations.scala
ekiwi-sifive: This is a private method, so no externally visible API change. The problem was the way that the annotations were filtered was broken. I think the `partition` pattern is the right way to do it. However, I could be persuaded to filter out the `otherAnnos` inside the `propagate` method.

src/test/scala/firrtlTests/ParserSpec.scala
ekiwi-sifive: I needed to change the `} toSeq` to `}.toSeq` because `2.13` is stricter about postfix notation.
I then also changed the `map` to use the now recommended `.`.

src/test/scala/firrtlTests/interval/IntervalMathSpec.scala
ekiwi-sifive: `Range.Double` was deprecated in `2.12` and removed in `2.13`. Thus I changed it to `Range.BigDecimal` and it seemed silly  to convert from `BigDecimal` to `Double` and back to `BigDecimal`.

src/test/scala/firrtlTests/options/PhaseManagerSpec.scala
ekiwi-sifive: This is the same postfix issue as with the `toSeq` above.

ekiwi-sifive:
Thanks for your review @sequencer !

In general the idea behind this PR is to do the minimal, most straightforward changes to get a `2.13` working in CI.

I would be happy about any follow-up PRs that improve the use of collections and removes the `mapValues(...).toMap` pattern.

ekiwi-sifive:
I think I can fix the semanticdb issue by upgrading to a newer version of scalafix: https://github.com/scalacenter/scalafix/issues/1109

build.sbt
ekiwi-sifive: fixed in https://github.com/freechipsproject/firrtl/pull/1796/commits/9250cd688abb482a9c184f3cbf3bda172ac47050

src/main/scala/firrtl/options/OptionParser.scala
ekiwi-sifive: No. `Seq` used to be `scala.collection.Seq`, now it is `scala.collection.immutable.Seq`. Since we extend the `OptionParser` we need to be explicit about overriding the correct `Seq`.

src/main/scala/firrtl/passes/RemoveCHIRRTL.scala
ekiwi-sifive: It is my understanding that it kind of was: https://underscore.io/blog/posts/2020/05/14/porting-and-seq.html

src/main/scala/firrtl/passes/memlib/MemConf.scala
ekiwi-sifive: Is that backwards compatible? Please make a follow up PR.

src/main/scala/firrtl/passes/memlib/MemConf.scala
ekiwi-sifive: Somehow Scala 2.13 inferred incompatible map types. I decided that creating a Seq of tuples and then doing toMap on it was the cleanest fix.

src/main/scala/firrtl/proto/FromProto.scala
ekiwi-sifive: I will try to change `mutable.Seq` to `scala.collection.Seq`.

src/main/scala/firrtl/passes/RemoveCHIRRTL.scala
ekiwi-sifive: fixed in https://github.com/freechipsproject/firrtl/pull/1796/commits/4265ed79eff1e5106a1dfab25b30466c17839ae3

src/main/scala/firrtl/proto/FromProto.scala
ekiwi-sifive: fixed in https://github.com/freechipsproject/firrtl/pull/1796/commits/4265ed79eff1e5106a1dfab25b30466c17839ae3

build.sbt
jackkoenig: For the record I think we just need to bump the version in `project/plugins.sbt`: https://scalacenter.github.io/scalafix/docs/users/installation.html

ekiwi-sifive: yeah, see https://github.com/freechipsproject/firrtl/pull/1796/commits/9250cd688abb482a9c184f3cbf3bda172ac47050

build.sbt
ekiwi-sifive: fixed in https://github.com/freechipsproject/firrtl/pull/1796/commits/9250cd688abb482a9c184f3cbf3bda172ac47050 and https://github.com/freechipsproject/firrtl/pull/1796/commits/005dd2519ade36a867d5b2dbaa199117860e14ca

src/test/scala/firrtlTests/options/PhaseManagerSpec.scala
jackkoenig: You can fix postfix by just adding a dot

jackkoenig: ie.
```scala
    def maybeToPng(f: File): Unit = try {
      s"dot -Tpng -O ${f}".!
      ...
```

build.sbt
ekiwi-sifive: I removed the `-Xsource` flag in https://github.com/freechipsproject/firrtl/pull/1796/commits/8870b5cc76973e596f1a7c87a4063a3049b858bb as it should not be required for `firrtl` which does not contain any Chisel bundles.

src/test/scala/firrtlTests/options/PhaseManagerSpec.scala
ekiwi-sifive: That is exactly what I did.

ekiwi-sifive: (Somehow github is showing only one line of the change, but if you click on the file name it will show you the red and the green line).

jackkoenig: Lol got it, thanks

.travis.yml
```diff
@@ -60,6 +60,13 @@ jobs:
         use: firrtl_build
       script:
         - sbt $SBT_ARGS +unidoc
+    - stage: test
+      name: "Tests: FIRRTL (2.13)"
+      workspaces:
+        use: firrtl_build
+      script:
+        - verilator --version
+        - sbt ++2.13.2 -J-XX:LoopStripMiningIter=0 $SBT_ARGS test
```
jackkoenig, line 69: do you know what `-J-XX:LoopStripMiningIter=0` is about?

build.sbt
```diff
@@ -2,18 +2,6 @@
...
-    // If we're building with Scala > 2.11, enable the compile option
-    //  switch to support our anonymous Bundle definitions:
-    //  https://github.com/scala/bug/issues/10047
-    CrossVersion.partialVersion(scalaVersion) match {
-      case Some((2, scalaMajor: Long)) if scalaMajor < 12 => Seq()
-      case _ => Seq("-Xsource:2.11")
-    }
-  }
-}
-
```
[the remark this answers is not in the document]
jackkoenig: there are no structural types in FIRRTL (I think) so this was probably just a mistaken copy-paste from Chisel's build.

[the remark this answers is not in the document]
jackkoenig: In any case, in chisel we should use `-language:reflectiveCalls` which is more precise

.travis.yml
ekiwi-sifive: It is supposed to disable an optimization in the JVM JIT, but it doesn't actually fix the Travis run as far as I can tell.
I was hopeful it might be able to fix it because there is a bug that is only fixed in OpenJDK 11.0.3 (and Travis seems to run 11.0.2) which crashes at the same instruction: https://bugs.openjdk.java.net/browse/JDK-8208275

ekiwi-sifive: The flag was added in https://github.com/freechipsproject/firrtl/pull/1796/commits/a025462626452f466fbda147fab9d68c4c9d0435 which I am going to remove before we merge.

ekiwi-sifive:
It seems like the the `WiringSpec` fails because of non-deterministic order probably because  Wiring is iterating over a Map. I will try to make a separate PR to make Wiring (more) deterministic.

src/main/scala/firrtl/options/OptionParser.scala
sequencer: wooo, type system of 2.13 is amazing!

src/main/scala/firrtl/passes/memlib/MemConf.scala
sequencer: yes, `Seq` here is much reasonable!

ekiwi-sifive:
I made a PR for the Wiring non-determinism fixes: https://github.com/freechipsproject/firrtl/pull/1799

Once we know that the non-determinism between `Scala 2.13` and `Scala 2.12` is fixed, we should try to merge #1799 and then I can rebase the `Scala 2.13` PR on top of it.

sequencer:
Sorry for the delay, too busy these days, only have time in the midnight :(
I think this PR is good enough to merge for our current status.

After reviewing @ekiwi’s great work, I realised, like @azidar said, maybe we need to “kill the monkey”, only after we fully remove old versions(like 2.11 or even 2.12), we can finally update our codes with new features in Scala 2.13, or even dotty.

I ask some chisel users, I found most of them never care the Scala version, only one line change in `build.sbt` in their code base. So maybe to release our burden, we can considering drop old Scala version more radically.
So here is just a proposal(maybe we can discuss in new Monday):
1.4: cross 2.11 (deprecate) 2.12 (deprecate) 2.13(support)
1.5: cross <del>2.11</del> <del>2.12</del> 2.13(deprecate) 2.14(support)
1.5: cross <del>2.13</del> 2.14 (deprecate) dotty(support)


As for the strict types(remove `.toMap` and `.toSeq`), it contains a big API change which I think it’s can be done in 1.4, which won’t block our status:
1. updating `IR.scala`
2. fix all tests
3. Add a compatibility objects for the plugins and deprecate in 1.4. Removed in 1.5

PR #1799, opened 2020-07-29
WiringTransform: fix non-determinism

ekiwi-sifive:
While the order of connections does not matter for correctness, we want a stable result.

### Contributor Checklist

#### Type of Improvement
- non-determinism fix

#### API Impact
- changes the signature of `WiringUtils.sinksToSources` to return a Seq of tuples instead of a map
- this did not have any impact on the code inside the `firrtl` compiler, but an external user might have to add a `toMap`

- can change the order in which signals are connected by the `WiringTransform` but now it should be deterministic

#### Desired Merge Strategy
- squash

- [x] Did you review?

### DIFF
--- a/src/main/scala/firrtl/passes/wiring/WiringUtils.scala
+++ b/src/main/scala/firrtl/passes/wiring/WiringUtils.scala
@@ -112,12 +112,25 @@ object WiringUtils {
     * @return a map of sink instance names to source instance names
     * @throws WiringException if a sink is equidistant to two sources
     */
-  def sinksToSources(sinks: Seq[Named],
+  @deprecated("This method can lead to non-determinism in your compiler pass. Use sinksToSourcesSeq instead!", "Firrtl 1.4")
+  def sinksToSources(sinks: Seq[Named], source: String, i: InstanceGraph): Map[Seq[WDefInstance], Seq[WDefInstance]] =
+    sinksToSourcesSeq(sinks, source, i).toMap
+
+  /** Return a map of sink instances to source instances that minimizes
+    * distance
+    *
+    * @param sinks a sequence of sink modules
+    * @param source the source module
+    * @param i a graph representing a circuit
+    * @return a map of sink instance names to source instance names
+    * @throws WiringException if a sink is equidistant to two sources
+    */
+  def sinksToSourcesSeq(sinks: Seq[Named],
                      source: String,
                      i: InstanceGraph):
-      Map[Seq[WDefInstance], Seq[WDefInstance]] = {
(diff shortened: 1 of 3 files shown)


src/main/scala/firrtl/passes/wiring/WiringUtils.scala
```diff
@@ -115,8 +115,9 @@ object WiringUtils {
   def sinksToSources(sinks: Seq[Named],
                      source: String,
                      i: InstanceGraph):
-      Map[Seq[WDefInstance], Seq[WDefInstance]] = {
-    val owners = new mutable.HashMap[Seq[WDefInstance], Vector[Seq[WDefInstance]]]
+      Seq[(Seq[WDefInstance], Seq[WDefInstance])] = {
```
seldridge, line 118: Since the return type is changing, does it make sense to deprecate the old `sinksToSources` and add a new method with a different return type?

ekiwi-sifive: We can. But it means a lot of code duplication (i.e. this PR is going to add > 2x lines).

seldridge: Maybe see if @jackkoenig has an opinion here.

ekiwi-sifive:
There is still some remaining non-determinism exposed by trying to run the `WiringSpec` on Scala `2.13`. I am investigating...

ekiwi-sifive:
So one of the challenges was that `.withDefaultValue` makes the Map no longer be a LinkedMap.

Besides that there seems to have been a bug, where depending on the order some modules got a `wire` instead of an `output` depending on the order of sinks and sources. An `output` can always serve as a `wire`, but a `wire` is not an adequate substitute for an `output`. Thus I now check and only turn things into `wires` if the didn't have a port added already.

src/main/scala/firrtl/passes/wiring/WiringUtils.scala
jackkoenig: It shouldn't require duplication right? Just deprecate `sinkToSources` and have it call the new method that returns a `Seq` then call `.toMap` on it.

ekiwi-sifive: OK!

jackkoenig:
What's the milestone here? This is a bugfix so should it be backported to 1.3.x? 1.2.x?

seldridge:
Bugfix. 1.2.x if it'll go back that far.

PR #1800, opened 2020-07-29
Deprecate InstanceGraph

ekiwi-sifive:
### Contributor Checklist

#### Type of Improvement
- code refactoring
#### API Impact

- deprecates `InstanceGraph`
- adds new methods to `InstanceKeyGraph`

- none

#### Desired Merge Strategy

- rebase

#### Release Notes
- `InstanceGraph` was deprecated in favor of the new `InstanceKeyGraph` and will be removed after the `1.4` release
- [x] Did you review?

### DIFF
--- a/src/main/scala/firrtl/analyses/CircuitGraph.scala
+++ b/src/main/scala/firrtl/analyses/CircuitGraph.scala
@@ -49,7 +49,7 @@ class CircuitGraph private[analyses] (connectionGraph: ConnectionGraph) {
   private val irLookup = connectionGraph.irLookup
 
   // Module/Instance Hierarchy information
-  private lazy val instanceGraph = new InstanceKeyGraph(circuit)
+  private lazy val instanceGraph = InstanceKeyGraph(circuit)
 
   // Per module, which modules does it instantiate
   private lazy val moduleChildren = instanceGraph.getChildInstances.toMap
--- a/src/main/scala/firrtl/analyses/CircuitGraph.scala
+++ b/src/main/scala/firrtl/analyses/CircuitGraph.scala
@@ -3,6 +3,7 @@
 package firrtl.analyses
 
 import firrtl.Kind
+import firrtl.analyses.InstanceKeyGraph.InstanceKey
 import firrtl.annotations.TargetToken.{Instance, OfModule}
 import firrtl.annotations._
 import firrtl.ir.{Circuit, DefInstance}
@@ -48,10 +49,10 @@ class CircuitGraph private[analyses] (connectionGraph: ConnectionGraph) {
   private val irLookup = connectionGraph.irLookup
 
   // Module/Instance Hierarchy information
-  private lazy val instanceGraph = new InstanceGraph(circuit)
+  private lazy val instanceGraph = new InstanceKeyGraph(circuit)
 
   // Per module, which modules does it instantiate
-  private lazy val moduleChildren = instanceGraph.getChildrenInstanceOfModule
+  private lazy val moduleChildren = instanceGraph.getChildInstances.toMap
 
   // Top-level module target
   private val main = ModuleTarget(circuit.main, circuit.main)
@@ -80,7 +81,7 @@ class CircuitGraph private[analyses] (connectionGraph: ConnectionGraph) {
     */
   def absolutePaths(mt: ModuleTarget): Seq[IsModule] = instanceGraph.findInstancesInHierarchy(mt.module).map {
     case seq if seq.nonEmpty => seq.foldLeft(CircuitTarget(circuit.main).module(circuit.main): IsModule) {
-      case (it, DefInstance(_, instance, ofModule, _)) => it.instOf(instance, ofModule)
+      case (it, InstanceKey(instance, ofModule)) => it.instOf(instance, ofModule)
     }
   }
 
@@ -113,9 +114,7 @@ class CircuitGraph private[analyses] (connectionGraph: ConnectionGraph) {
     val leafModule = path.leafModule
     val children = moduleChildren(leafModule)
     val localRefs = localReferences(path, kind)
-    localRefs ++ children.flatMap {
-      case (Instance(inst), OfModule(ofModule)) => deepReferences(kind, path.instOf(inst, ofModule))
-    }
+    localRefs ++ children.flatMap { child => deepReferences(kind, path.instOf(child.name, child.module)) }
   }
 
   /** Return all absolute references to signals of the given kind directly contained in the module
--- a/src/main/scala/firrtl/analyses/InstanceKeyGraph.scala
+++ b/src/main/scala/firrtl/analyses/InstanceKeyGraph.scala
@@ -32,6 +32,9 @@ class InstanceKeyGraph(c: ir.Circuit) {
   private lazy val fullHierarchy: mutable.LinkedHashMap[InstanceKey, Seq[Seq[InstanceKey]]] =
     graph.pathsInDAG(circuitTopInstance)
 
+  /** returns the underlying graph */
+  def getGraph: DiGraph[InstanceKey] = graph
+
   /** maps module names to the DefModule node */
   def moduleMap: Map[String, ir.DefModule] = nameToModule
 
@@ -40,6 +43,24 @@ class InstanceKeyGraph(c: ir.Circuit) {
 
   /** Returns a sequence that can be turned into a map from module name to instances defined in said module. */
   def getChildInstances: Seq[(String, Seq[InstanceKey])] = childInstances
+
+  /** A count of the *static* number of instances of each module. For any module other than the top (main) module,
+    * this is equivalent to the number of inst statements in the circuit instantiating each module,
+    * irrespective of the number of times (if any) the enclosing module appears in the hierarchy.
+    * Note that top module of the circuit has an associated count of one, even though it is never directly instantiated.
+    * Any modules *not* instantiated at all will have a count of zero.
+    */
+  lazy val staticInstanceCount: Map[OfModule, Int] = {
+    val foo = mutable.LinkedHashMap.empty[OfModule, Int]
+    childInstances.foreach {
+      case (main, _) if main == c.main => foo += main.OfModule  -> 1
+      case (other, _)                  => foo += other.OfModule -> 0
+    }
+    childInstances.flatMap(_._2).map(_.OfModule).foreach {
+      mod => foo += mod -> (foo(mod) + 1)
+    }
+    foo.toMap
+  }
 
   /** Finds the absolute paths (each represented by a Seq of instances
     * representing the chain of hierarchy) of all instances of a particular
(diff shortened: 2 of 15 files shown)


ekiwi-sifive:
Once #1799 is merged I should be able to also port the `Wiring` transform to use the new `InstanceKeyGraph`.

ekiwi-sifive:
This now replaces all uses of the old `InstanceGraph` and deprecates it so that we can remove it after `1.4` is released.

ekiwi-sifive:
This is ready for review @jackkoenig , @azidar , @seldridge

src/main/scala/firrtl/passes/wiring/WiringUtils.scala
```diff
@@ -113,8 +115,64 @@ object WiringUtils {
     * @throws WiringException if a sink is equidistant to two sources
     */
   @deprecated("This method can lead to non-determinism in your compiler pass. Use sinksToSourcesSeq instead!", "Firrtl 1.4")
-  def sinksToSources(sinks: Seq[Named], source: String, i: InstanceGraph): Map[Seq[WDefInstance], Seq[WDefInstance]] =
-    sinksToSourcesSeq(sinks, source, i).toMap
+  def sinksToSources(sinks: Seq[Named], source: String, i: InstanceGraph): Map[Seq[WDefInstance], Seq[WDefInstance]] = {
```
azidar, line 118: Is the reason you had to duplicate this function is because you needed the types of the WDefInstances?

src/test/scala/firrtlTests/analyses/InstanceKeyGraphSpec.scala
```diff
@@ -4,10 +4,98 @@ package firrtlTests.analyses
...
+        |  module Child1a :
+        |    skip
+        |  module Child1b :
+        |    skip
+        |  module Child2 :
+        |    skip
+        |""".stripMargin
+
+    val graph = new InstanceKeyGraph(parse(input)).graph.transformNodes(_.module)
+    getEdgeSet(graph) shouldBe Map("Top" -> Set("Child1", "Child2"), "Child1" -> Set("Child1a", "Child1b"), "Child2" -> Set(), "Child1a" -> Set(), "Child1b" -> Set())
```
azidar, line 37: Split over multiple lines.

src/test/scala/firrtlTests/analyses/InstanceKeyGraphSpec.scala
```diff
@@ -4,10 +4,98 @@ package firrtlTests.analyses
...
+        |  module Child2a :
+        |    skip
+        |  module Child2b :
+        |    skip
+        |  module Child3 :
+        |    skip
+        |""".stripMargin
+
+    val graph = new InstanceKeyGraph(parse(input)).graph.transformNodes(_.module)
+    getEdgeSet(graph) shouldBe Map("Top" -> Set("Child1"), "Top2" -> Set("Child2", "Child3"), "Child2" -> Set("Child2a", "Child2b"), "Child1" -> Set(), "Child2a" -> Set(), "Child2b" -> Set(), "Child3" -> Set())
```
azidar, line 65: Split over multiple lines.

azidar:
Please remember to make the `val circuit` in CircuitGraph be a def.

src/main/scala/firrtl/passes/wiring/WiringUtils.scala
ekiwi-sifive: Yes. We can hopefully remove it as soon as 1.4 is released.

src/test/scala/firrtlTests/analyses/InstanceKeyGraphSpec.scala
ekiwi-sifive: Will do!

src/test/scala/firrtlTests/analyses/InstanceKeyGraphSpec.scala
ekiwi-sifive: Will do!

ekiwi-sifive:
> Please remember to make the `val circuit` in CircuitGraph be a def.

I am working on a separate PR for that.

src/main/scala/firrtl/analyses/CircuitGraph.scala
```diff
@@ -48,10 +49,10 @@ class CircuitGraph private[analyses] (connectionGraph: ConnectionGraph) {
   private val irLookup = connectionGraph.irLookup
 
   // Module/Instance Hierarchy information
-  private lazy val instanceGraph = new InstanceGraph(circuit)
+  private lazy val instanceGraph = new InstanceKeyGraph(circuit)
```
jackkoenig, line 52: Can we actually make the constructor for `InstanceKeyGraph` private and instead use a factory method on the companion object? Makes future refactoring easier.

Sorry, this actually should have been a comment on https://github.com/freechipsproject/firrtl/pull/1732

src/main/scala/firrtl/analyses/InstanceKeyGraph.scala
```diff
@@ -52,8 +85,33 @@ class InstanceKeyGraph(c: ir.Circuit) {
...
   }
+
+  /** Given a circuit, returns a map from module name to a map
+    * in turn mapping instances names to corresponding module names
+    */
+  def getChildInstanceMap: mutable.LinkedHashMap[OfModule, mutable.LinkedHashMap[Instance, OfModule]] =
+    mutable.LinkedHashMap(childInstances.map { case (k, v) =>
+      val moduleMap: mutable.LinkedHashMap[Instance, OfModule] = mutable.LinkedHashMap(v.map(_.toTokens):_*)
+      TargetToken.OfModule(k) -> moduleMap
+    }:_*)
```
jackkoenig, line 98: I would rather not return mutable things, now that we have 2.13 support, could we return [`SeqMap`](https://www.scala-lang.org/api/2.13.3/scala/collection/immutable/SeqMap.html) for 2.13 (concretely return `VectorMap`)? This would require specializing for 2.13 and 2.12-, at least on this function

src/main/scala/firrtl/analyses/InstanceKeyGraph.scala
```diff
@@ -52,8 +85,33 @@ class InstanceKeyGraph(c: ir.Circuit) {
...
+
+  /** All modules *not* reachable from the top module of the circuit */
+  def unreachableModules: Seq[OfModule] = cachedUnreachableModules
+
+  /** An [[firrtl.graph.EulerTour EulerTour]] representation of the [[firrtl.graph.DiGraph DiGraph]] */
+  private lazy val tour = EulerTour(internalGraph, circuitTopInstance)
+
+  /** Finds the lowest common ancestor instances for two module names in a design */
+  def lowestCommonAncestor(moduleA: Seq[InstanceKey], moduleB: Seq[InstanceKey]): Seq[InstanceKey] =
+    tour.rmq(moduleA, moduleB)
```
jackkoenig, line 111: I get the value of `InstanceKey` and that this is emulating the API from `InstanceGraph`, just wrapping calls to the `EulerTour`, but this is effectively the same thing as `InstanceTarget` so maybe it's better to share the API there?

src/main/scala/firrtl/analyses/InstanceKeyGraph.scala
```diff
@@ -52,8 +85,33 @@ class InstanceKeyGraph(c: ir.Circuit) {
...
+
+  /** A list of absolute paths (each represented by a Seq of instances) of all module instances in the Circuit. */
+  def fullHierarchy: mutable.LinkedHashMap[InstanceKey, Seq[Seq[InstanceKey]]] = cachedFullHierarchy
```
jackkoenig, line 114: Similarly would rather use immutability here in 2.13

src/main/scala/firrtl/passes/wiring/WiringUtils.scala
```diff
@@ -125,15 +183,12 @@ object WiringUtils {
     * @return a map of sink instance names to source instance names
     * @throws WiringException if a sink is equidistant to two sources
     */
-  def sinksToSourcesSeq(sinks: Seq[Named],
-                     source: String,
-                     i: InstanceGraph):
-      Seq[(Seq[WDefInstance], Seq[WDefInstance])] = {
+  private[firrtl] def sinksToSourcesSeq(sinks: Seq[Named], source: String, i: InstanceKeyGraph):
+      Seq[(Seq[InstanceKey], Seq[InstanceKey])] = {
```
jackkoenig, line 187: The deprecation on `sinksToSources` points to this, but now it's private?

src/main/scala/firrtl/analyses/InstanceKeyGraph.scala
jackkoenig: I would also like comments from @azidar @seldridge and @albert-magyar since this my proposing the first case of this, but I also think we should do this for things like `Record.elements` in Chisel (well changing that from `ListMap` to `SeqMap` in 2.13)

src/main/scala/firrtl/analyses/CircuitGraph.scala
ekiwi-sifive: Sure

src/main/scala/firrtl/analyses/InstanceKeyGraph.scala
ekiwi-sifive: I do not want to specialize things for 2.13 in this PR.

src/main/scala/firrtl/analyses/InstanceKeyGraph.scala
ekiwi-sifive: Yes. But again, we still have to support `2.12` and `2.13` and I think it is beyond the scope of this PR to add special methods for `2.13`.
If you want the return to be immutable, I can return a `Seq[(K,V)]` instead.

src/main/scala/firrtl/passes/wiring/WiringUtils.scala
ekiwi-sifive: Yeah, since no one was using it and it isn't a pretty API at all. If someone comes and complain we can always make it public again.

src/main/scala/firrtl/analyses/InstanceKeyGraph.scala
jackkoenig: Fair, I think we should before we crack 3.4

src/main/scala/firrtl/analyses/InstanceKeyGraph.scala
jackkoenig: The introduction of a replacement API like `InstanceKeyGraph` is the only time we can fix mutable APIs. Specialization for 2.13 is annoying but gives the "best" API, since we could have `SeqMap` there and not have to lose the "map" aspect of it. Better than `Seq` would be `Iterable` because I think we could refine the API in 2.13.

Anyway, I won't insist here, but I think we should really consider not returning mutable types in 2.13 if we can.

src/main/scala/firrtl/passes/wiring/WiringUtils.scala
jackkoenig: I guess the issue is that the deprecation on `sinksToSources` points to it, can you change the message there?

src/main/scala/firrtl/analyses/CircuitGraph.scala
ekiwi-sifive: Done

src/main/scala/firrtl/analyses/InstanceKeyGraph.scala
ekiwi-sifive: I moved the `lowestCommonAncestor` method out of `InstanceKeyGraph` and into the one place where is was used. Thus we do not add unnecessary API cruft.

src/main/scala/firrtl/passes/wiring/WiringUtils.scala
ekiwi-sifive: A clarified the message.

PR #1803, opened 2020-07-29
Fix dshlw in checktypes

chick:
Fixes Problems with CheckTypes when left shifting an Interval
Dshlw can be encountered in CheckTypes and it has similar argument types to Dshl
Added a test that only passes with new CheckTypes

### Contributor Checklist

- [na] Did you update the FIRRTL spec to include every new feature/behavior?

#### Type of Improvement

Bug fix. Dsptools fails in IntervalSpec due to CheckTypes error

#### API Impact

Should have no impact on existing API

No change to backend verilog

#### Desired Merge Strategy
Squash

#### Release Notes
Fixed problem where left shifting and Interval could generated firrtl error

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/passes/CheckTypes.scala
+++ b/src/main/scala/firrtl/passes/CheckTypes.scala
@@ -195,7 +195,7 @@ object CheckTypes extends Pass {
       e.op match {
         case AsUInt | AsSInt | AsClock | AsFixedPoint | AsAsyncReset | AsInterval =>
           // All types are ok
-        case Dshl | Dshr =>
+        case Dshl | Dshr | Dshlw =>
           checkAllTypes(Seq(e.args.head), okUInt=true, okSInt=true,  okClock=false, okFix=true,  okAsync=false, okInterval=true)
           checkAllTypes(Seq(e.args(1)),   okUInt=true, okSInt=false, okClock=false, okFix=false, okAsync=false, okInterval=false)
         case Add | Sub | Mul | Lt | Leq | Gt | Geq | Eq | Neq =>
--- a/src/test/scala/firrtlTests/interval/IntervalSpec.scala
+++ b/src/test/scala/firrtlTests/interval/IntervalSpec.scala
@@ -3,9 +3,11 @@ package interval
 
 import firrtl._
 import firrtl.ir.Circuit
+import firrtl.options.Phase
 import firrtl.passes._
 import firrtl.passes.CheckTypes.InvalidConnect
 import firrtl.passes.CheckWidths.DisjointSqueeze
+import firrtl.stage.{FirrtlCircuitAnnotation, FirrtlStage, Forms}
 import firrtl.testutils.FirrtlFlatSpec
 
 class IntervalSpec extends FirrtlFlatSpec {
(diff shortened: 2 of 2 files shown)


albert-magyar:
It would be better to avoid spreading `Dshlw` all over the codebase.

When you run into the error, what is the full pass ordering?

chick:
@albert-magyar Does PadFile have to change it to a `Dshlw` If I change PadWidths to be
```scala
  // Recursive, updates expression so children exp's have correct widths
  private def onExp(e: Expression): Expression = e map onExp match {
    ...
     case Dshl =>
        // special case as args aren't all same width
//        ex copy (op = Dshlw, args = Seq(fixup(width(ex.tpe))(ex.args.head), ex.args(1)))  THIS LINE IS THE PROBLEM
        ex copy (args = Seq(fixup(width(ex.tpe))(ex.args.head), ex.args(1)))

      case _ => ex
    }
    case ex => ex
  }
```
then I don't need the CheckTypes change and all tests in the src/test still pass.
That's reducing the appearances of `Dshlw`. Can I do that?

albert-magyar:
> @albert-magyar Does PadWidths have to change it to a `Dshlw` If I change PadWidths to be
> 
> ```scala
>   // Recursive, updates expression so children exp's have correct widths
>   private def onExp(e: Expression): Expression = e map onExp match {
>     ...
>      case Dshl =>
>         // special case as args aren't all same width
> //        ex copy (op = Dshlw, args = Seq(fixup(width(ex.tpe))(ex.args.head), ex.args(1)))  THIS LINE IS THE PROBLEM
>         ex copy (args = Seq(fixup(width(ex.tpe))(ex.args.head), ex.args(1)))
> 
>       case _ => ex
>     }
>     case ex => ex
>   }
> ```
> 
> then I don't need the CheckTypes change and all tests in the src/test still pass.
> That's reducing the appearances of `Dshlw`. Can I do that?

`PadWidth`s (in particular, that line) is a currently necessary part of prepping for Verilog emission.

The current flow is that `PadWidths` generates _illegal FIRRTL with internal IR nodes_ (including `Dshlw`) which the Verilog emitter can better translate to Verilog. This means that `PadWidths` should only run as part of the final group of transforms that happen right before emission (these are listed in `Emitter.scala`). This group should _not_ include `CheckTypes` after `PadWidths`, since this will cause the error you're seeing.

Something about the flow is causing `CheckTypes` to run after `PadWidths` (or possibly some other pass is calling functions defined in `CheckTypes`). This is effectively a pass-ordering bug. The introduction of the Dependency API laid the groundwork to fix bugs like these, but they are still possible as long as we are still iteratively improving the dependency specifications of the existing passes (including moving away from `LowForm`, etc).

I don't think that the current `Dshlw` flow is perfect by any means, but it seems to be a long-standing feature that works fine. Improving this whole flow to avoid the custom nodes or adding a general pattern for built-in passes to handle custom nodes would be nice in a perfect world, but that would be a lot of work, and I don't think that `Dshlw` is the appropriate motivating example for that.

seldridge:
I believe that this can be closed in favor of the more fine-grained transform dependency specification approach that Treadle is now using (see: https://github.com/freechipsproject/treadle/pull/238)?

jackkoenig:
We should solve "The dshlw problem" but it's solved sufficiently in Treadle for now and I don't think this is the right general solution

PR #1809, opened 2020-07-30
ir: use Serializer.serialize where possible

ekiwi-sifive:
### Contributor Checklist

- Did you add at least one test demonstrating the PR?
  - no, but there are many tests that are serializing nodes already
  - some comparison tests were run as part of the PR that added the Serializer

#### Type of Improvement
- code refactoring
- code cleanup (we get more of a single source of truth)

#### API Impact

- `Expression` and `Statement` now implement `serialize`, thus if you want to keep a custom serialize around it needs the `override` keyword

- none

#### Desired Merge Strategy

- squash

#### Release Notes

n/a

- [x] Did you review?

### DIFF
--- a/src/main/scala/firrtl/AddDescriptionNodes.scala
+++ b/src/main/scala/firrtl/AddDescriptionNodes.scala
@@ -79,7 +79,7 @@ case class Attribute(string: StringLit) extends Description {
   * @param stmt the encapsulated statement
   */
 private case class DescribedStmt(descriptions: Seq[Description], stmt: Statement) extends Statement with HasDescription {
-  def serialize: String = s"${descriptions.map(_.serialize).mkString("\n")}\n${stmt.serialize}"
+  override def serialize: String = s"${descriptions.map(_.serialize).mkString("\n")}\n${stmt.serialize}"
   def mapStmt(f: Statement => Statement): Statement = f(stmt)
   def mapExpr(f: Expression => Expression): Statement = this.copy(stmt = stmt.mapExpr(f))
   def mapType(f: Type => Type): Statement = this.copy(stmt = stmt.mapType(f))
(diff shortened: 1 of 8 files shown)


ekiwi-sifive:
This PR now also includes some bugfixed to the `ir.Serializer` that I discovered through failing unittests.

azidar:
Can this be 1.3.x? Or is the serializer 1.4?

ekiwi-sifive:
1.4 is good

jackkoenig:
It's not a huge deal, but this is going to break any custom IR nodes. I don't know that it's worth doing this the right way but this kind of change means that any custom IR node cannot possibly compile with both `1.3.x` and `1.4.x`

ekiwi-sifive:
> It's not a huge deal, but this is going to break any custom IR nodes. I don't know that it's worth doing this the right way but this kind of change means that any custom IR node cannot possibly compile with both `1.3.x` and `1.4.x`

What if you use `overwrite def serialize: String = "My custom thing!"`.

jackkoenig:
Ah that's right, I forgot that Scala supports using the `override` keyword even if it's a virtual method. That's less bad.

It's still probably better to do this with a private mixin. Now if you write a custom IR node it's going to compile (not telling you that you need to override `serialize`) and then just crash on serialization, not great.

azidar:
Good point, I didn't catch that. Let's add another PR to add the mixing trait.

ekiwi-sifive:
> Good point, I didn't catch that. Let's add another PR to add the mixing trait.

So you want the trait to override def  serialize and then extend every single node supported by the `Serializer` with this trait?

Please note that custom nodes will fail in `DedupModules` since they won't be supported by the `StructuralHasher`.

jackkoenig:
> So you want the trait to override def serialize and then extend every single node supported by the Serializer with this trait?

That's the best backwards compatible way

> Please note that custom nodes will fail in DedupModules since they won't be supported by the StructuralHasher.

That's fine, custom nodes aren't supposed to leak out of custom transforms. I guess an argument could be that Serialization is a form of "leaking", but the main issue for me is that existing code that works won't compile when it bumps.

Fundamentally, we ought to have sealed the IR and we should consider moving in that direction with special carveouts for custom nodes (eg. `case class CustomStatement(stmt: Any) extends Statement` or something), but we didn't so we have to tread carefully on these kinds of things.

Issue #1161, opened 2019-08-20
CheckHighFormLike: Reference memory is not declared


ekiwi:
Somehow using nodes to address `smem` breaks in  `firrtl.passes.CheckHighFormLike`.

Minimized Chisel3:
```.scala
class regfile extends Module {
  val io = IO(new Bundle{
    val sel = Input(Bool())
    val addr = Input(UInt(5.W))
  })

  val raddr = Mux(io.sel, io.addr, 0.U)
  val memory = SyncReadMem(32, UInt(2.W))
  val rdata = memory.read(raddr)
}
```

Chirrtl:
```
circuit regfile :
  module regfile :
    input clock : Clock
    input reset : UInt<1>
    output io : { flip sel : UInt<1>, flip addr : UInt<5>}
  
    node raddr = mux(io.sel, io.addr, UInt<1>("h0")) @[regfile.scala 30:18]
    smem memory : UInt<2> [32] @[regfile.scala 31:27]
    read mport rdata = memory[raddr], clock @[regfile.scala 32:26]
```

Compiler error + trace:
```
 @[regfile.scala 30:18]: [module regfile] Reference memory is not declared.
firrtl.passes.CheckHighFormLike$UndeclaredReferenceException:  @[regfile.scala 30:18]: [module regfile] Reference memory is not declared.
	at firrtl.passes.CheckHighFormLike.checkHighFormE$1(Checks.scala:158)
	at firrtl.passes.CheckHighFormLike.$anonfun$run$16(Checks.scala:169)
	at firrtl.passes.CheckHighFormLike.$anonfun$run$16$adapted(Checks.scala:169)
	at firrtl.WSubField.foreachExpr(WIR.scala:59)
	at firrtl.traversals.Foreachers$ExprForMagnet$$anon$6.foreach(Foreachers.scala:44)
	at firrtl.traversals.Foreachers$ExprForeach$.foreach$extension(Foreachers.scala:54)
	at firrtl.passes.CheckHighFormLike.checkHighFormE$1(Checks.scala:169)
	at firrtl.passes.CheckHighFormLike.$anonfun$run$16(Checks.scala:169)
	at firrtl.passes.CheckHighFormLike.$anonfun$run$16$adapted(Checks.scala:169)
	at firrtl.WSubField.foreachExpr(WIR.scala:59)
	at firrtl.traversals.Foreachers$ExprForMagnet$$anon$6.foreach(Foreachers.scala:44)
[...]

Issue #913, opened 2018-10-09
smem read port always reads address 0


luismarques:
**Type of issue**: bug report

I'm having problems with `smem` memories. Consider this circuit as an example:

```
circuit Memories :
    module Memories :
        input clock : Clock
        input reset : UInt<1>
        input writeAddr : UInt<2>
        input writeVal : UInt<8>
        input readAddr : UInt<2>
        output readVal : UInt<8>
        readVal is invalid

        smem mem : UInt<8>[3]
        infer mport _T_0 = mem[writeAddr], clock
        _T_0 <= writeVal
        infer mport _T_1 = mem[readAddr], clock
        readVal <= _T_1
```

When I convert this to Verilog (using firrtl git HEAD, commit ed709571876b68e4982d11db15d236752713b6a1), I get the following (trimmed for clarity):

```
  reg [1:0] mem__T_1_addr_pipe_0;
  assign mem__T_1_addr = mem__T_1_addr_pipe_0;
  assign mem__T_1_data = mem[mem__T_1_addr]; // @[:Memories.fir@11.8]
  assign readVal = mem__T_1_data; // @[:Memories.fir@15.8]
  wire  _GEN_0;
  assign _GEN_0 = 1'h0;
  always @(posedge clock) begin
    if (_GEN_0) begin
      mem__T_1_addr_pipe_0 <= readAddr;
    end
  end
```

Because of the `assign _GEN_0 = 1'h0`, the `mem__T_1_addr_pipe_0` is never written to, so the circuit always reads from the memory address 0 (or whatever `mem__T_1_addr_pipe_0` is initialized to).

This looks like a bug, but it's hard to say exactly because `smem` memories are not documented. Any insights into this?

jackkoenig:
This is most certainly a bug, and so far, I'm only more confused, I generated the following with Chisel:

```
circuit SMem :
  module SMem :
    input clock : Clock
    input reset : UInt<1>
    output io : {flip raddr : UInt<2>, rdata : UInt<8>, flip waddr : UInt<2>, flip wdata : UInt<8>}

    smem mem : UInt<8>[3]
    infer mport _T_15 = mem[io.waddr], clock
    _T_15 <= io.wdata
    infer mport _T_16 = mem[io.raddr], clock
    io.rdata <= _T_16      
```
The only difference is an aggregate IO, and yet this **works**:

```verilog
  reg [1:0] mem__T_16_addr_pipe_0;
[...]

Issue #840, opened 2018-07-02
Issue with RemoveChirrtl smem when address is a register


mehnadnerd:
**Type of issue**: bug report

**If the current behavior is a bug, please provide the steps to reproduce the problem:**
  * **If possible, please directly provide the FIRRTL file or a relevant fragment if the file is large**
These two Chisel statements:
`outBundle:= mem.read(currentItem)`
`outBundle := mem.read(Mux(currentItem =/= RegNext(currentItem), currentItem, currentItem))`

And corresponding CHIRRTL:
`read mport _T_428 = mem[currentItem], clock`
`node _T_408 = neq(currentItem, _T_407)
node _T_409 = mux(_T_408, currentItem, currentItem)
read mport _T_410 = mem[_T_409], clock`

Do not result in the same emitted Verilog. In fact, the first leads to an issue where the memory read is enabled only when currentItem changes. However, since there is a one-cycle delay for this mem, this means that the read will only update after the next address is given, not one cycle later!

By adding in the unnecessary Mux, the correct behaviour emerges, which leads me to suspect that it might have to do with optimisation of the read, but looking at the RemoveChirrtl transform, I am unable to find the cause.

  * **Please tell us about your environment:**
    a6

**What is the use case for changing the behavior?**
Circuits working like the documentation suggests

**Impact**: API modification

**Development Phase**: request

**Other information**
#702 #727 both include similar things, although 702 appears to be dealing with the clock rather than the enable. This would hopefully be fixed by 727, although it is important to ensure this bug does not remain in the transition.

I can provide the full chisel/firrtl/verilog to people within BAR if need be, as well as the full trace of the firrtl run. It is quite large (392 MB), but the important part is before and after the RemoveChirrtl transform.

mehnadnerd:
Issue appears to be with using a register as the address, adding in a temporary wire fixes it as well.

mehnadnerd:
Adam's plan: Insert a temporary wire between register and address

albert-magyar:
Closed and absorbed into #727

PR #1821, opened 2020-08-03
Replace CHIRRTL with new behavioral memory statements

albert-magyar:
**Type of change:** bug fix / improvement
**API impact:** API addition
**Backend code-generation impact:** Fixes open issues with CHIRRTL memories and associated enables

This subsumes #1210 

List of issues with CHIRRTL memories:
#286
#512 
#702
#840 
#913 
#1161 

I think it is not ideal to approach this problem from the question of FIRRTL semantics, since the whole point of CHIRRTL was to help support the Chisel semantics. I think it is best to think about code examples in Chisel and what they should do.

Currently, Chisel memory read enables are a complicated mess of the when condition in which `.read` (or `.apply`) is called and the conditions under which the _read address_ is updated. This is also combined with an explicit enable that may be supplied by the user. I propose we scrap the "address update sensitivity" entirely.


spec/spec.tex
```diff
@@ -958,6 +958,99 @@ \subsubsection{Write Under Write Behaviour}
...
+
+In contrast, when the enable signal is provided explicitly, the memory read
+operation itself is conditional on the enable signal.
+
+\begin{lstlisting}
+read mport r = syncmem[cmd.addr], clk, cmd.ren ; conditional read
+when (cmd.ren) :
+  resp.rdata <= r
+\end{lstlisting}
+
```
ekiwi-sifive, line 1007: I think we also need an example where a read port is declared inside a when scope, in order to clarify what happens in that case.

spec/spec.tex
```diff
@@ -958,6 +958,99 @@ \subsubsection{Write Under Write Behaviour}
...
+  w <= cmd.wdata ; conditional write operation
+
+; last-connect semantics apply to inferred writes
+when (cmd.wen && cmd.isClear) :
+  w <= UInt<32>(``h00'') ; different condition
+\end{lstlisting}
+
+Inferred write ports do not support masked writes; therefore, all write operations inherently have
+all subfields enabled. When the conditions are such that no connection to the inferred write port
+occurs, no write operation will occur on that port.
```
ekiwi-sifive, line 1034: Is that in order to simplify the compiler? It seems like for e.g. register updates we do allow last-connect semantics on sub fields.

spec/spec.tex
```diff
@@ -958,6 +958,99 @@ \subsubsection{Write Under Write Behaviour}
...
+
+An inferred write port statement accepts four mandatory arguments: the name of the port, the memory
+to access, the address signal, and the clock. This has the effect of implicitly declaring the
+specified port and connecting the address and clock signals to the corresponding fields (`addr` and
+`clk`) of the new port.
+
+Once an inferred write port is declared, write operations can be performed using the connection
+statement. Intuitively, when the port is used on the left-hand side of an assignment, this
+corresponds to connecting the right-hand side of the assignment to the `data` field of an explicit
+write port and asserting that port's `en` signal. As with any connection, the connection (and
```
ekiwi-sifive, line 1018: Are you allowed to use the port on the rhs?

spec/spec.tex
```diff
@@ -958,6 +958,99 @@ \subsubsection{Write Under Write Behaviour}
...
+
+Inferred write ports do not support masked writes; therefore, all write operations inherently have
+all subfields enabled. When the conditions are such that no connection to the inferred write port
+occurs, no write operation will occur on that port.
+
+\subsubsection{Inferred Memory Accessors}
+
+Once an inferred memory accessor is declared, write operations can also be performed using the connection statement.
+Intuitively, when the port is used on the left-hand side of an assignment, this corresponds to connecting the right-hand side of the assignment to the `data` field of an explicit write port and asserting that port's `en` signal.
+As with any connection, the connection (and therefore the write operation) is conditional on any \verb|when| conditions that appear between the declaration of the inferred write port and the connection statement.
```
ekiwi-sifive, line 1040: Can you add a sentence describing what the difference between a `memaccess` and a `write mport` is?

I have read both sections multiple times and I am still not sure what the exact difference is.

src/main/antlr4/FIRRTL.g4
```diff
@@ -1,4 +1,4 @@
-// See LICENSE for license details.
+// See LICENSE for liAcense details.
```
ekiwi-sifive, line 1: this seems to be a typo

spec/spec.tex
```diff
@@ -1943,12 +2037,13 @@ \subsection{LoFIRRTL}
...
+\item The inferred memory port statements are not used.
 \item All components are connected to exactly once.
 \item All components must be declared with a ground type.
 \item The partial connect statement is not used.
 \end{itemize}
 
-The first three restrictions follow from the fact that any LoFIRRTL circuit is also a legal MidFIRRTL circuit. The additional restrictions give LoFIRRTL a direct correspondence to a circuit netlist.
+The first four restrictions follow from the fact that any LoFIRRTL circuit is also a legal MidFIRRTL circuit. The additional restrictions give LoFIRRTL a direct correspondence to a circuit netlist.
 
 Low level circuit transformations can be conveniently written by first lowering a circuit to its LoFIRRTL form, then operating on the restricted (and thus simpler) subset of constructs. Note that circuit transformations are still free to generate high level constructs as they can simply be lowered again.
```
ekiwi-sifive, line 2049: Are you going to add the new keywords to the firrtl grammar in the appendix?

src/main/antlr4/FIRRTL.g4
```diff
@@ -97,6 +97,8 @@ stmt
   | 'cmem' id ':' type info?
   | 'smem' id ':' type ruw? info?
   | mdir 'mport' id '=' id '[' exp ']' exp info?
+  | 'memaccess' id '=' exp exp ('with' ':' 'en' '=>' exp )? info?
+  | 'memwrite' exp '[' exp ']' '<=' exp exp ('with' ':' 'mask' '=>' exp )? info?
```
ekiwi-sifive, line 101: you never seem to define `memwrite` in the spec

src/main/scala/firrtl/Visitor.scala
```diff
@@ -322,6 +322,10 @@ class Visitor(infoMode: InfoMode) extends AbstractParseTreeVisitor[FirrtlNode] w
         case "smem" =>
           val (tpe, size) = visitCMemType(ctx.`type`())
           CDefMemory(info, ctx.id(0).getText, tpe, size, seq = true, readUnderWrite = visitRuw(Option(ctx.ruw)))
+        case "memaccess" =>
+          DefMemAccess(info, ctx.id(0).getText, visitExp(ctx_exp(0)), visitExp(ctx_exp(1)), visitExp(ctx_exp(2)))
+        case "maskedwrite" =>
+          MemMaskedWrite(info, visitExp(ctx_exp(0)), visitExp(ctx_exp(1)), visitExp(ctx_exp(2)), visitExp(ctx_exp(3)))
```
ekiwi-sifive, line 328: you added the keyword `memwrite` to the grammar, why are you matching on `maskedwrite` here?

albert-magyar:
The spec is not up to date (hence the draft). I will be updating it today.

The bulk of the discussion tomorrow will be on chisel semantics. While the
current CHIRRTL flow somewhat disregards FIRRTL conventions, I think there
is a way to support any chisel behavior easily without abusing FIRRTL
conventions.

Your points are all good ones, though. The current implementation in this
PR resolves the question of CHIRRTL’s differences in “when-sensitivity” and
also allows for masked writes.

PR #1826, opened 2020-08-05
SMT Backend

ekiwi-sifive:
### Contributor Checklist

- **n/a** Did you update the FIRRTL spec to include every new feature/behavior?

#### Type of Improvement
 - new feature/API

#### API Impact
- new emitters: `firrtl.backends.experimental.smt.{Btor2Emitter, SMTLibEmitter}`
- new `-E experimental-smt2` and `-E experimental-btor2` command line options
- new `EmittedSMTModelAnnotation` which serializes to a custom file
- new `StutteringClockTransform` and `GlobalClockAnnotation` to replace clocks with clock enable signals for multi-clock formal verification

- adds new SMTLib and btor2 backend code generation
- no impact on the Verilog generation

#### Desired Merge Strategy
- squash

#### Release Notes
This release of firrtl includes an experimental preview version of a new [SMTLib](smtlib.cs.uiowa.edu) and [Btor2](https://cs.stanford.edu/~niemetz/publication/2018/niemetzpreinerwolfbiere-cav18/) emitter that converts a firrtl module into a format suitable for open source model checkers. The format generally follows the behavior of yosys' `write_smt2` and `write_btor` commands. To generate `btor2` for the module in `m.fir` run `./utils/bin/firrtl -i m.fir -E experimental-btor2`. If you have a design with multiple clocks or an asynchronous reset, try out the new `StutteringClockTransform`. You can designate any input of type `Clock` to be your global simulation clock using the new `GlobalClockAnnotation`. If your toplevel module instantiates submodules, you need to inline them if you want the submodule logic to be included in the formal model. The preview release is intended to help people get started, implementing a formal testing library for Chisel, but is not recommended for end-users just yet.

- [ ] Did you review?


ekiwi-sifive:
This is now based on https://github.com/freechipsproject/firrtl/pull/1835

src/main/scala/firrtl/backends/smt/FirrtlToTransitionSystem.scala
```diff
@@ -0,0 +1,254 @@
...
+  def run(m: ir.Module, warn: String => Unit = defaultWarn, presetRegs: Set[String] = Set()): TransitionSystem = {
+    // first pass over the module to convert expressions; discover state and I/O
+    val scan = new ModuleScanner(warn)
+    m.foreachPort(scan.onPort)
+    // Currently we do not have multi-clock support.
+    assert(scan.clocks.size <=  1, s"Found more than one clock: ${scan.clocks.mkString(", ")}")
+    m.foreachStmt(scan.onStatement)
+
+    // translate into transition system
+    val inputs = scan.inputs.toSeq ++ scan.randoms.values
```
sequencer, line 57: I have a question here: why do we need randoms value for transition system?

src/main/scala/firrtl/backends/smt/FirrtlToTransitionSystem.scala
```diff
@@ -0,0 +1,254 @@
...
+case class Signal(name: String, e: BVExpr) { def toSymbol: BVSymbol = BVSymbol(name, e.width) }
+case class TransitionSystem(
+   name: String, inputs: Seq[BVSymbol], states: Seq[State], signals: Seq[Signal],
+   outputs: Set[String], assumes: Set[String], asserts: Set[String], fair: Set[String],
+   comments: Map[String, String] = Map())
+
+case class TransitionSystemAnnotation(sys: TransitionSystem) extends NoTargetAnnotation
+
+object FirrtlToTransitionSystem extends Transform with DependencyAPIMigration {
+  // TODO: We only really need [[Forms.MidForm]] + LowerTypes, but we also want to fail if there are CombLoops
```
sequencer, line 26: Is possible to make SMT compatible to `High Form`? I guess with this ability, we can use Z3 to prove the correctness between transforms.

ekiwi-sifive: Working on `HighForm` would complicate this pass. However it would be possible to eventually add `HighFirrtl` support, just not in the first iteration.

src/main/scala/firrtl/backends/smt/FirrtlToTransitionSystem.scala
ekiwi-sifive: By "random" I mean "arbitrary".

ekiwi-sifive: E.g., if a wire `is invalid` you will get an arbitrary/random value.

ekiwi-sifive: https://github.com/freechipsproject/firrtl/blob/d94e6678c4c06e26162c57fe3ecbf30b789e43c9/src/test/scala/firrtl/backends/smt/end2end/UndefinedFirrtlSpec.scala#L28

sequencer: Got it, but I think it is still a little strange, that a arbitrary value should not be equal to a specific random value in the model. I mean this "arbitrary" behavior gives specific values to invalid wires rather than a unknown state. I think this should trace back to the x/z state of RTL. Is there any possibility to do so?

ekiwi-sifive: An "arbitrary" value means that any possible value is ok. Thus the translation generates a free variable (an input) that models this behavior. The solver is then free to choose any value that would make the assertion fail. Only if it cannot find one, can you be sure that the assertion will never be violated.

I am not sure what x/z state you are talking about. We don't do tri-state logic and thus there never can be any Zs and firrtl semantics do not specify any Xs and Verilog x-propagation can never precisely model the semantics of firrtl.

ekiwi-sifive:
@seldridge , @jackbackrack  Does anyone of you know a good way to skip the unittests that call out to `z3` if the `z3` binary is not available? I want people to still be able to run `firtl` unittests locally even if they do not have `z3` installed. (ideally it should also be possible to skip the tests that rely on `verilator`)

jackkoenig:
Jim looked at this a long time ago: https://github.com/freechipsproject/chisel3/pull/791

The only trick is make sure tests don't pass, or at least, make sure `sbt test` doesn't pass if dependencies are missing. If people need to run something like `sbt test noz3` or something, that's fine.

PR #1830, opened 2020-08-07
Symbol table

ekiwi-sifive:
Moving the new `SymbolTable` out of https://github.com/freechipsproject/firrtl/pull/1784

### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/Namespace.scala
+++ b/src/main/scala/firrtl/Namespace.scala
@@ -2,6 +2,8 @@
 
 package firrtl
 
+import firrtl.analyses.{NamespaceTable, SymbolTable}
+
 import scala.collection.mutable
 import firrtl.ir._
 
--- a/src/main/scala/firrtl/Namespace.scala
+++ b/src/main/scala/firrtl/Namespace.scala
@@ -50,21 +50,9 @@ class Namespace private {
 object Namespace {
   // Initializes a namespace from a Module
   def apply(m: DefModule): Namespace = {
+    val names = SymbolTable.scanModule(m, new NamespaceTable).getNames
     val namespace = new Namespace
-
-    def buildNamespaceStmt(s: Statement): Seq[String] = s match {
-      case s: IsDeclaration => Seq(s.name)
-      case s: Conditionally => buildNamespaceStmt(s.conseq) ++ buildNamespaceStmt(s.alt)
-      case s: Block => s.stmts flatMap buildNamespaceStmt
-      case _ => Nil
-    }
-    namespace.namespace ++= m.ports map (_.name)
-    m match {
-      case in: Module =>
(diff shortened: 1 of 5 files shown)


ekiwi-sifive:
Some more applications of `SymbolTable`:

- in the `NewLowerTypes` pass (which promoted me to initially write the SymbolTable):
  - we need to scan the module namespace in order to do the uniquification: https://github.com/freechipsproject/firrtl/blob/211cca930a7f99c8c1d68a417c4e1b083db2c208/src/main/scala/firrtl/passes/LowerTypes.scala#L79
  - the custom `LoweringSymbolTable` could now be replaced with the `NamespaceTable` from this PR
- in the `InferTypesFlowsAndKinds` pass that I was working on
  - I made a custom `Lookup` data structure that would hold interned references to all signals in the circuit: https://github.com/ekiwi-sifive/firrtl/blob/combined-infer/src/main/scala/firrtl/passes/InferTypesFlowsAndKinds.scala#L162
  - the `SymbolTable` interface was inspired by this data structure which could now implement the `SymbolTable` interface

ekiwi-sifive:
>     * The `LocalSymbolTable` calling `.toMap` on `getSymbols` is probably unnecessarily inefficient. I think it would be better to either use a `var Map` under the hood or just return `collection.Map` so you can return the mutable one with the API implication that it shouldn't be modified. Another option is to make `LocalSymbolTable` extends `String => Symbol` and just supply the `apply` method (and `.get` and `.getOrElse` forms) and let people do the lookups directly on it

I think I like the apply function idea since that would make sure that no one can iterate over the map. Thanks!



>     * IMO, the most important version of this kind of thing is the live version which could be used by InferTypes and the like, I'd like to see that kind of thing

If I have time I can try to integrate it with `InferTypes`

>     * Another related thing is the type of `Symbol` table to lookup connectivity (called `Netlist` in many transforms). Is this intended to fulfill that kind of roll?

No. The idea is a traditional `SymbolTable` which allows you to look up `Types`, `Kind` and `Flow` of an identifier.
I only specified the minimal API in the trait and the hope is that we will standardize the data access in the future when more passes are using this.

ekiwi-sifive:
Jack, I added a order preserving mix-in (`WithSeq`) as well as a mix-in that allows for fast lookup but prevents you from iterating over the collection, by just implementing `apply(String): Symbol` as you suggested.

The code is getting more complicated though and without a use-case for it in the compiler (for `LocalSymbolTable` and the two new mix-ins) I am not sure whether this is a good idea. Maybe we should slim it down to the subset needed by the new `LowerTypes` pass and then people can further develop the API as needed.

ekiwi-sifive:
I added the suggested improvements to https://github.com/freechipsproject/firrtl/pull/1784 and will merge the `SymbolTable` as part of the new `LowerTypes`.

PR #1833, opened 2020-08-07
Remove LegacyAnnotation and [most] MoultingYaml

jackkoenig:
It has been deprecated since 1.1

This felt good. I couldn't fully delete the moultingyaml dependency because it's still used in memlib. I think json4s actually supports YAML so it may be possible to switch the code without breaking or deprecating the yaml support there.

### Contributor Checklist

#### Type of Improvement

  - code cleanup   

#### API Impact

This removes support for `LegacyAnnotation` and related `YAML` stuff. Some of it is baked in to the `Driver` and related stuff which I left since those are set to be deleted elsewhere, I just elevated some warnings to "errors" although they aren't fatal.

No change

#### Desired Merge Strategy

  - Squash

#### Release Notes

Remove deprecated `LegacyAnnotation` and related YAML support. Also removed deprecated implicit annotation file handling.

- [ ] Did you review?


jackkoenig:
Requires https://github.com/freechipsproject/chisel3/pull/1544 to pass the chisel3 master tests

src/main/scala/firrtl/annotations/Annotation.scala
```diff
@@ -127,130 +127,7 @@ trait MultiTargetAnnotation extends Annotation {
...
-            "This functionality will soon be removed. Please migrate to new annotations."
-          StageUtils.dramaticWarning(msg)
-          warned = true
-        }
-        annox
-      case other => other
-    }
-  }
-}
+object Annotation
```
seldridge, line 130: How would we go about removing this object?

jackkoenig: I don't know why anyone would be relying on it, but we'd have to deprecate. I'm not sure that we should though, it's a reasonable place to put utilities or implicit conversions if needed.

seldridge: Sounds good. Mostly just a curiosity... Yes, this could be reuse for utilities later.

PR #1825, opened 2020-08-05
Sane PadWidths

ekiwi-sifive:
### Contributor Checklist

- [ ] Did you add Scaladoc to every public function/method?
- [x] Did you add at least one test demonstrating the PR?
- [x] Did you delete any extraneous printlns/debugging code?
- [x] Did you specify the type of improvement?
- [x] Did you state the API impact?
- [x] Did you specify the code generation impact?
- [x] Did you request a desired merge strategy?
- [x] Did you add text to be included in the Release Notes for this change?

#### Type of Improvement
- bug fix 
- code refactoring

#### API Impact
- `PadWidths` now doesn't pad `dshl`, `rem`, `mul` and `div`
- the new `PadWidthsVerilog` pass does the padding which violates Firrtl width invariances

#### Backend Code Generation Impact

- none

#### Desired Merge Strategy

- squash

#### Release Notes
- `PadWidths` now generates only valid firrtl

### Reviewer Checklist (only modified by reviewer)
- [ ] Did you add the appropriate labels?
- [ ] Did you mark the proper milestone (1.2.x, 1.3.0, 1.4.0) ?
- [ ] Did you review?
- [ ] Did you check whether all relevant Contributor checkboxes have been checked?
- [ ] Did you mark as `Please Merge`?

PR #1836, opened 2020-08-10
Forms: move Verilog-specific transforms into the emitter

ekiwi-sifive:
I have come to the conclusion that instead of trying to split `PadWidths` into two passes, we should try to make our optimizations work without it.

This PR tries to move all (or at least most) passes that are specific to Verilog emission into the `Verilog` `Forms` and thus into the emitter.

### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/passes/CommonSubexpressionElimination.scala
+++ b/src/main/scala/firrtl/passes/CommonSubexpressionElimination.scala
@@ -12,9 +12,7 @@ object CommonSubexpressionElimination extends Pass {
   override def prerequisites = firrtl.stage.Forms.LowForm ++
     Seq( Dependency(firrtl.passes.RemoveValidIf),
          Dependency[firrtl.transforms.ConstantPropagation],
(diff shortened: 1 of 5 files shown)


ekiwi-sifive:
Screen Shot 2020-08-10 at 15 13 38

ekiwi-sifive:
This seems like a more promising solution compared to https://github.com/freechipsproject/firrtl/pull/1825

ekiwi-sifive:
This seems to break any transform that consumes LowFirrtl and produces LowFirrtl, causing a dependency cycle.

ekiwi-sifive:
I still believe that moving the transforms that are needed for correct Verilog emission into the emitter is the correct step. This way we know that as long as you give the emitter valid LowFirrtl, you will get valid Verilog out, which makes for an easy invariance to work with. All optimization passes should consume any low firrtl and emit valid low firrtl.
The performance of the optimization passes might be different depending on their ordering. But that is ok as long as we don't incorrect results if we change their ordering.

Unfortunately I won't have time to fix this.

PR #1842, opened 2020-08-13
Deprecate support for Scala 2.11

jackkoenig:
Printed as warning in transform, can be suppressed via CLI

I'm working on a companion Chisel PR as well as a documentation PR to fill in that link.

I'm actually not sure how to test this, looking to @seldridge for inspiration.

### Contributor Checklist

#### Type of Improvement

  - code cleanup  

(Deprecating 2.11)

#### API Impact

This deprecates support for Scala 2.11 so it motivates/requires users to migrate their code.

No change

#### Desired Merge Strategy

#### Release Notes

Deprecate support for Scala 2.11

- [ ] Did you review?


jackkoenig:
@seldridge problem I'm having is that the annotation doesn't show up in the output `.anno.json` file, so clearly I'm doing something wrong. Any tips?

On second though, I'm not sure I *want* it to show up. The reason why one would is that if you suppress the warning in Chisel, the suppression will propagate to FIRRTL (if you use split processes to run them), but maybe we want you to have to suppress in both just to be annoying and encourage you to upgrade.

src/main/scala/firrtl/stage/package.scala
```diff
@@ -33,6 +33,7 @@ package object stage {
           case InfoModeAnnotation(i)             => c.copy(infoModeName = i)
           case FirrtlCircuitAnnotation(cir)      => c.copy(firrtlCircuit = Some(cir))
           case a : CompilerAnnotation => logger.warn(s"Use of CompilerAnnotation is deprecated. Ignoring $a") ; c
+          case SuppressScalaVersionWarning       => c
```
seldridge, line 36: It's a-okay to do it this way, but it may make sense to add a `FirrtlOption` for this that can then be viewed as `FirrtlOptions`.

jackkoenig: I'm fine with adding it, I think I'll merge this PR as is but if you think we should add, can open a PR tomorrow

seldridge:
Re: not showing up in the output file -> `FirrtlOption` is `Unserializable`. This is intentional so that options specific to a stage don't leak. However, I think this is more global than that. I think you can just remove that mix-in.

PR #1854, opened 2020-08-20
Async reset tieoff bug

davidbiancolin:
This fixes a bug where, when Async Reset is tied off to a constant, asynchronously reset registers would be emitted with `posedge constant`, which some downstream tools do not accept. This reapplies the same fix used for clocks, but for async resets.

The underlying assumption here is that synchronous reset has been removed, so it's safe to just blindly do this for all registers. Should RemoveResets be added to prerequisite list? 

### Contributor Checklist

- [ x] Did you add Scaladoc to every public function/method?
- [ x] Did you update the FIRRTL spec to include every new feature/behavior?
- [ x] Did you add at least one test demonstrating the PR?
- [ x] Did you delete any extraneous printlns/debugging code?
- [ x] Did you specify the type of improvement?
- [ x] Did you state the API impact?
- [ x] Did you specify the code generation impact?
- [ x] Did you request a desired merge strategy?

#### Type of Improvement

- bug fix

#### API Impact

Transform is renamed but backwards compatible aliases are provided in the firrtl.transforms package object. 

Emits an extra temporary to hold tied-off async resets. 

#### Desired Merge Strategy

#### Release Notes

Fix a bug where tied-off async resets would be emitted as literals in register sensitivity lists.

- [ ] Did you review?


src/main/scala/firrtl/transforms/package.scala
```diff
@@ -0,0 +1,6 @@
+package firrtl
+
+package object transforms {
+  type LegalizeClocksTransform = LegalizeClocksAndAsyncResetTransform
+  val LegalizeClocksTransform = LegalizeClocksAndAsyncResetTransform
```
jackkoenig, line 5:
```suggestion
  @deprecated("Replaced by LegalizeClocksAndAsyncResetTransform", "FIRRTL 1.4.0")
  type LegalizeClocksTransform = LegalizeClocksAndAsyncResetTransform
  @deprecated("Replaced by LegalizeClocksAndAsyncResetTransform", "FIRRTL 1.4.0")
  val LegalizeClocksTransform = LegalizeClocksAndAsyncResetTransform
```

src/main/scala/firrtl/transforms/LegalizeClocksAndAsyncReset.scala
```diff
@@ -33,19 +33,30 @@ object LegalizeClocksTransform {
...
+      case r: DefRegister if (isLiteralExpression(r.clock) || isLiteralExpression(r.reset)) =>
+        val (clockNodeOpt, rxClock) = if (isLiteralExpression(r.clock)) {
+          val node = DefNode(r.info, namespace.newTemp, r.clock)
+          (Some(node), r.copy(clock = WRef(node)))
+        } else {
+          (None, r)
+        }
+        val (resetNodeOpt, rx) = if (isLiteralExpression(r.reset)) {
+          val castToAsync = DoPrim(PrimOps.AsAsyncReset, Seq(r.reset), Nil, AsyncResetType)
+          val node = DefNode(r.info, namespace.newTemp, castToAsync)
```
jackkoenig, line 45: Are we sure that the cast isn't already there and that the emitter isn't just dropping it?

src/main/scala/firrtl/transforms/LegalizeClocksAndAsyncReset.scala
```diff
@@ -33,19 +33,30 @@ object LegalizeClocksTransform {
...
+          (Some(node), rxClock.copy(reset = WRef(node)))
+        } else {
+          (None, rxClock)
+        }
+        Block(clockNodeOpt ++: resetNodeOpt ++: Seq(rx))
```
jackkoenig, line 50: Clever use of `Option` as a `TraversableOnce` 😉

src/main/scala/firrtl/transforms/LegalizeClocksAndAsyncReset.scala
davidbiancolin: I added this because a downstream pass complains that sync reset should have been removed, if i just carry the expression forward into the node. I guess the more correct thing to do is check if theres a cast, and add it if it doesn't already exist.

davidbiancolin: I see. I'll investigate a bit.

PR #1869, opened 2020-08-27
Add --pretty:no-expr-inlining to prevent expression inlining

jackkoenig:
**Note:** this is a backwards incompatible change, but only to a CLI option and associated annotation that is new in `1.4.0-RC1`

Also rename `--Wno-scala-version-warning` to
`--warn:no-scala-version-deprecation` and adopt naming convention where
resulting annotation matches the CLI option

This is motivated partially out of caution (because `InlineBooleanExpressions` is such an intrusive change) and also by the observation that we should just in generation have an option to disable it and related things. The new option also disables CombineCats. I didn't have it apply to `InlineCasts` because those are handled specially by the VerilogEmitter, that transform is not a prettification as much as an "avoid lint warnings", but I could be convinced to make it apply there.

I'm also proposing a new CLI scoping scheme with the new prefixes `--warn:` for warning relation options and `--pretty:` for prettification options. I'm open to changing or dropping these depending on how people feel. The implication here is that we should have other prefixes like:
* `--opt:` for optimization-related things `--no-dce` and `--no-dedup`
* `--check:` for `--no-check-comb-loops`
* `--feature:` or `--func:` for `--no-asa`

There's also the implication that sometimes these disable, sometimes they could enable, we could consider some infrastructure to make it accept `--warn:scala-version-deprecation` which currently would be illegal, but I think that's beyond the scope of this current proposal/PR.

### Contributor Checklist

#### Type of Improvement

 - new feature/API          

#### API Impact

Breaks an API from 1.4.0-RC1 and adds new API to prevent inlining expressions

Gives the user control over if expressions should be inlined

#### Desired Merge Strategy

 - Squash

#### Release Notes

Add `--pretty:no-expr-inlining` to prevent expression inlining and rename `--Wno-scala-version-warning` to `--warn:no-scala-version-deprecation`

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/transforms/InlineBooleanExpressions.scala
+++ b/src/main/scala/firrtl/transforms/InlineBooleanExpressions.scala
@@ -3,6 +3,7 @@
 package firrtl
 package transforms
 
+import firrtl.stage.PrettyNoExprInlining
 import firrtl.annotations.{NoTargetAnnotation, Target}
 import firrtl.annotations.TargetToken.{fromStringToTargetToken, OfModule, Ref}
 import firrtl.ir._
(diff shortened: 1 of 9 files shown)


src/main/scala/firrtl/transforms/InlineBooleanExpressions.scala
```diff
@@ -144,26 +145,33 @@ class InlineBooleanExpressions extends Transform with DependencyAPIMigration {
...
 
   def execute(state: CircuitState): CircuitState = {
-    val dontTouchMap: Map[OfModule, Set[Ref]] = {
-      val refTargets = state.annotations.flatMap {
-        case anno: HasDontTouches => anno.dontTouches
-        case o => Nil
-      }
-      val dontTouches: Seq[(OfModule, Ref)] = refTargets.map {
-        case r => Target.referringModule(r).module.OfModule -> r.ref.Ref
+    val run = !state.annotations.contains(PrettyNoExprInlining)
```
albert-magyar, line 148: I think the option and its overall behavior is great.

My main concern is with the approach of "run the transform and do nothing" vs. "don't run the transform." I think flags like this are a good opportunity to make better use of dependencies (i.e. making sure that optimizations are only ever in the transitive `optionalPrerequisite` closure of non-optimization transforms and removing optimizations from the stage's requested targets).

I don't think this PR is necessarily the right time or place, but it would be great to move toward solving these pass-ordering and pass-enabling problems with dependencies.

albert-magyar:
I think the current CLI for this option is good, and it's great to have this option available sooner rather than later. I'd approve this now, but it still needs a change or two to keep CI happy?

jackkoenig:
Yeah let me fix the issue, it is a weird one (and only on 2.11, ugh)

jackkoenig:
On no it's just a real bug lol, wow, this is why we write tests...

seldridge:
This can be follow-up work, but it might make sense to add helpers for generating the options, e.g.:

```scala
implicit class LongOptionable(a: String) {
  def pretty: String  = s"--pretty:$a"
  def warn: String    = s"--warn:$a"
  def check: String   = s"--check:$a"
  def feature: String = s"--feature:$a"      
}
```

Or:

```scala
sealed trait LongOptionPrefix extends (String => String) { 
  def prefix: String
  override def apply(a: String): String = "--$prefix:$a"
}
object Pretty extends LongOptionPrefix {
  override def prefix = "pretty"
}
```

jackkoenig:
@albert-magyar I think this is good to go

jackkoenig:
@seldridge I think something like that makes sense but I'll leave it for future work

src/test/scala/firrtlTests/InlineBooleanExpressionsSpec.scala
```diff
@@ -276,6 +276,20 @@ class InlineBooleanExpressionsSpec extends FirrtlFlatSpec {
...
+        |    input a : UInt<1>
+        |    input b : UInt<1>
+        |    input c : UInt<1>
+        |    output out : UInt<1>
+        |
+        |    node _T_1 = and(a, b)
+        |    out <= and(_T_1, c)""".stripMargin
+    val result = exec(input, PrettyNoExprInlining :: Nil)
+    (result) should be(parse(input).serialize)
+
```
albert-magyar, line 292:
```suggestion
}

```

albert-magyar (changes requested):

albert-magyar (approved):
LGTM

PR #1835, opened 2020-08-10
stage: allow a RunFirrtlTransformAnnotation(_:Emitter) annotation to be used in place of a CompilerAnnotation

ekiwi-sifive:
### Contributor Checklist

- [x] Did you add Scaladoc to every public function/method?
- [x] Did you update the FIRRTL spec to include every new feature/behavior?
- [x] Did you add at least one test demonstrating the PR?
- [x] Did you delete any extraneous printlns/debugging code?
- [x] Did you specify the type of improvement?
- [x] Did you state the API impact?
- [x] Did you specify the code generation impact?
- [x] Did you request a desired merge strategy?
- [x] Did you add text to be included in the Release Notes for this change?

#### Type of Improvement
 - code refactoring

#### API Impact

- one should now be able to give a `RunFirrtlTransformAnnotation(_:Emitter)` style annotation to the FirrtlStage instead of the deprecated `CompilerAnnotation`

#### Backend Code Generation Impact
-none

#### Release Notes
- `FirrtlStage`: you can now run a specific `Emitter` with a `RunFirrtlTransformAnnotation` instead of using a deprecated `CompilerAnnotation`

### Reviewer Checklist (only modified by reviewer)
- [x] Did you add the appropriate labels?
- [x] Did you mark the proper milestone (1.2.x, 1.3.0, 1.4.0) ?
- [x] Did you review?
- [x] Did you check whether all relevant Contributor checkboxes have been checked?
- [x] Did you mark as `Please Merge`?

PR #1870, opened 2020-08-27
Deprecate CompilerAnnotation

seldridge:
This is a follow-on to https://github.com/freechipsproject/firrtl/pull/1835 to fully deprecate `CompilerAnnotation`. 

This does a couple of things:

1. The `CompilerAnnotation$` companion object is changed to emit `RunFirrtlTransformAnnotation`s instead of `CompilerAnnotation`s. In effect, when you do `-X verilog`, you now get a `RunFirrtlTransformAnnotation(new VerilogEmitter)` as opposed to a `CompilerAnnotation(new VerilogCompiler)`.

2. This adds a `ConvertCompilerAnnotation` phase which is doing an explicit conversion of a `CompilerAnnotation` to a `RunFirrtlTransformAnnotation`. This differs from #1835 which relaxed support so that a `RunFirrtlTransformAnnotation(_ <: Emitter)` could substitute for a `CompilerAnnotation`. This PR goes one step further and forces internal phases to only understand transforms. 

3. Better checks and warnings are now provided inside `ConvertCompilerAnnotation`. The check of a single compiler is migrated from `Checks` here. Additionally, if a `CompilerAnnotation` is converted, a verbose warning is printed telling you exactly how to replace your transform, e.g., you now get:

```
CompilerAnnotation is deprecated since FIRRTL 1.4.0. Please use 'RunFirrtlTransformAnnotation(new firrtl.HighFirrtlEmitter)' instead.
```

Note that you should only now get a warning if you actually use a `CompilerAnnotation`. Any use of `-X` will not produce an error.

### Contributor Checklist

- [nope] Did you add Scaladoc to every public function/method?

#### Type of Improvement

- bug fix

- code refactoring

#### API Impact

None. Removes a warning that the user couldn't do anything to fix.

None.

#### Desired Merge Strategy

#### Release Notes

None.

- [x] Did you review?

### DIFF
--- a/src/main/scala/firrtl/stage/FirrtlAnnotations.scala
+++ b/src/main/scala/firrtl/stage/FirrtlAnnotations.scala
@@ -152,22 +152,23 @@ object FirrtlSourceAnnotation extends HasShellOptions {
   *  - If unset, a [[CompilerAnnotation]] with the default [[VerilogCompiler]]
   * @param compiler compiler name
   */
+@deprecated("Use a RunFirrtlTransformAnnotation targeting a specific Emitter.", "FIRRTL 1.4.0")
 case class CompilerAnnotation(compiler: Compiler = new VerilogCompiler()) extends NoTargetAnnotation with FirrtlOption
 
 object CompilerAnnotation extends HasShellOptions {
 
-  private[firrtl] def apply(compilerName: String): CompilerAnnotation = {
+  private[firrtl] def apply(compilerName: String): RunFirrtlTransformAnnotation = {
     val c = compilerName match {
-      case "none"     => new NoneCompiler()
-      case "high"     => new HighFirrtlCompiler()
-      case "low"      => new LowFirrtlCompiler()
-      case "middle"   => new MiddleFirrtlCompiler()
-      case "verilog"  => new VerilogCompiler()
-      case "mverilog" => new MinimumVerilogCompiler()
-      case "sverilog" => new SystemVerilogCompiler()
+      case "none"     => new ChirrtlEmitter
+      case "high"     => new HighFirrtlEmitter
+      case "low"      => new LowFirrtlEmitter
+      case "middle"   => new MiddleFirrtlEmitter
+      case "verilog"  => new VerilogEmitter
+      case "mverilog" => new MinimumVerilogEmitter
+      case "sverilog" => new SystemVerilogEmitter
       case _          => throw new OptionsException(s"Unknown compiler name '$compilerName'! (Did you misspell it?)")
     }
-    CompilerAnnotation(c)
+    RunFirrtlTransformAnnotation(c)
   }
 
   val options = Seq(
--- a/src/main/scala/firrtl/stage/FirrtlAnnotations.scala
+++ b/src/main/scala/firrtl/stage/FirrtlAnnotations.scala
@@ -155,26 +155,13 @@ object FirrtlSourceAnnotation extends HasShellOptions {
 @deprecated("Use a RunFirrtlTransformAnnotation targeting a specific Emitter.", "FIRRTL 1.4.0")
 case class CompilerAnnotation(compiler: Compiler = new VerilogCompiler()) extends NoTargetAnnotation with FirrtlOption
 
+@deprecated("Use a RunFirrtlTransformAnnotation targeting a specific Emitter.", "FIRRTL 1.4.0")
 object CompilerAnnotation extends HasShellOptions {
-
-  private[firrtl] def apply(compilerName: String): RunFirrtlTransformAnnotation = {
-    val c = compilerName match {
-      case "none"     => new ChirrtlEmitter
-      case "high"     => new HighFirrtlEmitter
-      case "low"      => new LowFirrtlEmitter
-      case "middle"   => new MiddleFirrtlEmitter
-      case "verilog"  => new VerilogEmitter
-      case "mverilog" => new MinimumVerilogEmitter
-      case "sverilog" => new SystemVerilogEmitter
-      case _          => throw new OptionsException(s"Unknown compiler name '$compilerName'! (Did you misspell it?)")
-    }
-    RunFirrtlTransformAnnotation(c)
-  }
 
   val options = Seq(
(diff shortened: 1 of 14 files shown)


src/main/scala/firrtl/stage/phases/ConvertCompilerAnnotations.scala
```diff
@@ -0,0 +1,40 @@
...
+  override def invalidates(a: Phase) = false
+
+  override def transform(annotations: AnnotationSeq): AnnotationSeq = {
+    annotations.collect {
+      case a: CompilerAnnotation => a
+    } match {
+      case a if a.size > 1 =>
+        val (msg, suggest) = (s"""found '${a.mkString(", ")}'""", "use multiple of")
+        throw new OptionsException(
+          s"Zero or more deprecated CompilerAnnotation may be specified, but $msg.".stripMargin
```
jackkoenig, line 27: Should this say "Zero or one"?

src/main/scala/firrtl/stage/phases/ConvertCompilerAnnotations.scala
```diff
@@ -0,0 +1,40 @@
...
+
+import firrtl.AnnotationSeq
+import firrtl.options.{Dependency, OptionsException, Phase}
+import firrtl.stage.{CompilerAnnotation, RunFirrtlTransformAnnotation}
+
+@deprecated(
+  "This only exists to convert deprecated CompilerAnnotations to RunFirrtlTransformAnnotations.",
+  "FIRRTL 1.5.0"
+)
+class ConvertCompilerAnnotations extends Phase {
```
jackkoenig, line 13: Obviously this will add deprecation warnings to normal compilation, is it possible to make it package private and have this still work? Is that a reasonable thing to do?

src/main/scala/firrtl/stage/FirrtlAnnotations.scala
```diff
@@ -152,22 +152,23 @@ object FirrtlSourceAnnotation extends HasShellOptions {
   *  - If unset, a [[CompilerAnnotation]] with the default [[VerilogCompiler]]
   * @param compiler compiler name
   */
+@deprecated("Use a RunFirrtlTransformAnnotation targeting a specific Emitter.", "FIRRTL 1.4.0")
 case class CompilerAnnotation(compiler: Compiler = new VerilogCompiler()) extends NoTargetAnnotation with FirrtlOption
 
 object CompilerAnnotation extends HasShellOptions {
```
jackkoenig, line 158: Should this object be deprecated and it's contents and moved elsewhere?

src/main/scala/firrtl/stage/phases/ConvertCompilerAnnotations.scala
seldridge: I can change it. I had it like this originally and changed it back because my `firrtlTests` test couldn't see it (yes, I should've just changed the test package...).

src/main/scala/firrtl/stage/phases/ConvertCompilerAnnotations.scala
seldridge: Yes, good catch.

src/main/scala/firrtl/stage/FirrtlAnnotations.scala
seldridge: It's slated for removal in 1.5.0, so yes. I'll do that and move the logic into `RunFirrtlTransformAnnotation$`.

src/main/scala/firrtl/stage/phases/ConvertCompilerAnnotations.scala
jackkoenig: Yeah I'd prefer it being private and not deprecated if possible haha, but I feel you

src/main/scala/firrtl/stage/phases/ConvertCompilerAnnotations.scala
seldridge: Fixed in later commits.

src/main/scala/firrtl/stage/phases/ConvertCompilerAnnotations.scala
seldridge: Fixed in later commits.

src/main/scala/firrtl/stage/FirrtlAnnotations.scala
seldridge: Deprecated in later commits.

src/test/scala/firrtl/stage/phases/tests/ConvertCompilerAnnotationsSpec.scala
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  it should "throw an exception if multiple CompilerAnnotations are specified" in new Fixture {
+    val annotations = Seq(
+      CompilerAnnotation(new HighFirrtlCompiler),
+      CompilerAnnotation(new LowFirrtlCompiler)
+    )
+    intercept[OptionsException] {
+      phase.transform(annotations)
+    }.getMessage should include("Zero or more deprecated CompilerAnnotation may be specified")
```
jackkoenig, line 38: I think you forgot to update this test

seldridge: :facepalm:

I never know how to handle it as it's extremely brittle to test the message text. However, exceptions are so non-specific...

Fixed in later commit.

ekiwi:
Did you make sure that specifying only `-X` without `-E` works? What stage will add the `EmitCircuitAnnotation`?

seldridge:
@ekiwi: `firrtl.stage.phases.AddImplicitEmitter` is what is taking care of this. Fortunately, [your edits](https://github.com/freechipsproject/firrtl/pull/1835/files#diff-6d25bb47ac560ee58c4da99f813f8e23) in your original PR are already handling this case nicely.

I added a test to `FirrtlMainSpec` to hit this in a follow-up commit.

PR #1874, opened 2020-08-28
Emitter: add missing parenthesize calls

albertchen-sifive:
Fixes a bug in the Emitter introduced by https://github.com/freechipsproject/firrtl/pull/1817

input:
```
circuit foo :
  module foo :
    input in1 : UInt<1>
    input in2 : UInt<1>
    input in3 : UInt<1>
    output out : UInt<2>

    out <= dshl(in1, or(in2, in3))
```

emits incorrect verilog due to missing parentheses, `in2 | in3` should have parentheses because `|` has lower precedence than `<<`:
```
module foo(
  input        in1,
  input        in2,
  input        in3,
  output [1:0] out
);
  wire [1:0] _GEN_0 = {{1'd0}, in1};
  assign out = _GEN_0 << in2 | in3;
endmodule
```

This PR adds missing calls to `parenthesize` to emit these parentheses.

### Contributor Checklist

- [n/a] Did you add Scaladoc to every public function/method?
- [n/a] Did you update the FIRRTL spec to include every new feature/behavior?
- [x] Did you add at least one test demonstrating the PR?
- [x] Did you delete any extraneous printlns/debugging code?
- [x] Did you specify the type of improvement?
- [x] Did you state the API impact?
- [x] Did you specify the code generation impact?
- [x] Did you request a desired merge strategy?
- [x] Did you add text to be included in the Release Notes for this change?

#### Type of Improvement

- bug fix

#### API Impact

none

#### Backend Code Generation Impact

adds missing parentheses

#### Desired Merge Strategy

- Squash

### Reviewer Checklist (only modified by reviewer)
- [x] Did you add the appropriate labels?
- [x] Did you mark the proper milestone (1.2.x, 1.3.0, 1.4.0) ?
- [x] Did you review?
- [x] Did you check whether all relevant Contributor checkboxes have been checked?
- [ ] Did you mark as `Please Merge`?

PR #1876, opened 2020-08-28
Add custom transforms 'upstream' of emitter annotation in equiv tests

albert-magyar:
**Type of change:** bug fix
**API impact:** none

The LEC environment puts the custom transform annotations later in the annotation sequence than the emitter it selects. Since the transform manager tries to honor this, it ends up running the emitter before the custom transforms if it can.


albert-magyar:
This results in the tests from #1874 and #1875 failing as expected without the appropriate fixes.

PR #1883, opened 2020-09-04
Legalize mem clocks

albert-magyar:
**Type of improvement:** bug fix
**API impact:** none
B**ackend code-generation impact:** FIRRTL will no longer emit `always @(posedge 1'b0)` for synchronous memory ports that have tied-off clocks.

Prior to this PR, `LegalizeClocksAndAsyncResetsTransform` only applied to registers. It makes sense to avoid messing with clocks that just go to submodules or ports, since they are plain wires, but memory port clocks get directly put into sensitivity lists. This PR fixes the potential issue and adds a test that would fail on master.

As with the register clock legalization, this also resolves related issues with emitting things like `always @(posedge ~clk)` that are not handled by Verilator. H/t to @jerryz123 for finding this issue.

### DIFF
--- a/src/main/scala/firrtl/transforms/LegalizeClocksAndAsyncResets.scala
+++ b/src/main/scala/firrtl/transforms/LegalizeClocksAndAsyncResets.scala
@@ -52,6 +52,9 @@ object LegalizeClocksAndAsyncResetsTransform {
           (None, rxClock)
         }
         Block(clockNodeOpt ++: resetNodeOpt ++: Seq(rx))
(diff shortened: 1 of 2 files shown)


jackkoenig:
Is this not needed on `master`?

jackkoenig (approved):

albert-magyar:
> Is this not needed on `master`?

This is needed on `master` -- this PR is against `master`. I marked it as 1.3.x since it should also be backported. I didn't do 1.2.x but I could backport it that far back if need be.

jackkoenig:
Lol duh, I was confused because the PR template looks funny, but maybe you just deleted everything 🧐

PR #1891, opened 2020-09-08
Make StageOption Unserializable

jackkoenig:
These options are generally specific to a stage and thus should not be
propagating across serialization

The `StageOption`s are:
* `TargetDirAnnotation`
* `ProgramArgsAnnotation`
* `InputAnnotationFileAnnotation`
* `OutputAnnotationFileAnnotation`
* `WriteDeletedAnnotation`

`InputAnnotationFileAnnotation` is already [consumed](https://github.com/freechipsproject/firrtl/blob/e420f99d87ece9f56504b3afc2e37d40b6e8c7b1/src/main/scala/firrtl/options/phases/GetIncludes.scala#L34) so the question is if any of these others should propagation across serialization? It does not make sense that providing `-td` to an invocation of Chisel should then be propagated automagically when you call FIRRTL and pass the intermediate annotation file. In general, all of these should be set for each stage by the user as they please.

### Contributor Checklist

#### Type of Improvement

   - code/API cleanup  

#### API Impact

This changes behavior in a way that should generally work better for users. For example, the current behavior has to be [worked around in rocket-chip](https://github.com/chipsalliance/rocket-chip/blob/d71c5a674275676b9c870a20cb0a42cd039a4115/src/main/scala/stage/phases/GenerateFirrtlAnnos.scala#L26) to manually emit the output annotation file so that it doesn't include annotations that would conflict like `TargetDirAnnotation`.

No change

#### Desired Merge Strategy
  - Rebase

#### Release Notes

StageOptions (like TargetDirAnnotation) are no longer serialized in the output annotation file. This improves composability of Stages.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/options/StageAnnotations.scala
+++ b/src/main/scala/firrtl/options/StageAnnotations.scala
@@ -10,7 +10,7 @@ import java.io.File
 
 import scopt.OptionParser
 
-sealed trait StageOption { this: Annotation => }
+sealed trait StageOption extends Unserializable { this: Annotation => }
 
 /** An annotation that should not be serialized automatically [[phases.WriteOutputAnnotations WriteOutputAnnotations]].
   * This usually means that this is an annotation that is used only internally to a [[Stage]].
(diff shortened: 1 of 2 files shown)


seldridge (approved):
lgtm

PR #1896, opened 2020-09-12
Don't use ResolvedAnnotationPaths in ConstProp nor DCE

jackkoenig:
Both use EliminateTargetPaths to duplicate modules based on annotations.
Currently, EliminateTargetPaths API is a little too limited so it
duplicates more than it should which effectively breaks Dedup whenever
DontTouchAnnotations are present.

Also, make ConstProp and DCE treat all HasDontTouches as local
annotations even if they are instance annotations. This is more
conservative but it is generally better to preserve deduplication than
to maximally optimize every instance.

This is a workaround, but it is important in lieu of making `EliminateTargetPaths` work as one would hope. To make `EliminateTargetPaths` work as would be intuitive and useful, I think it's API needs to allow transforms to expose 3 things:
1. Which annotations to be sensitive to (including traits).
2. Which targets to use from those annotations
3. How to dedup those targets

Note that (1) is the current API although I'm not sure if it would work for traits. (2) is important because an annotation that extends `HasDontTouches` may have some targets that are not dontTouched so `EliminateTargetPaths` can't blindly take all `targets`. (3) is the subtle but critical. It differs from how Deduplication should also Deduplicate Annotations. Using `HasDontTouches` as an example, it is reasonable for annotations that are different (and thus shouldn't deduplicate themselves) to nevertheless dont touch targets in differing instances of the same module such that `EliminateTargetPaths` should "deduplicate" the different instance path targets into a local target--thus it shouldn't duplicate those modules.

### Contributor Checklist

#### Type of Improvement

 - bug fix 

The bug is that DontTouchAnnotations break Deduplication (via re-duplication). This is a workaround rather than a fix to the API deficiencies in EliminateTargetPaths

#### API Impact

This is binary incompatible because `ConstProp` and `DeadCodeElimination` no longer implement `ResolvedAnnotationPaths`. Incompatibilities from stuff like this illustrates that we should avoid inheritance for features like this in the future.

Improved deduplication

#### Desired Merge Strategy

  - Squash

#### Release Notes

`ConstantPropagation` and `DeadCodeElimination` no longer implement `ResolvedAnnotationPaths`.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/transforms/ConstantPropagation.scala
+++ b/src/main/scala/firrtl/transforms/ConstantPropagation.scala
@@ -100,7 +100,7 @@ object ConstantPropagation {
 
 }
 
-class ConstantPropagation extends Transform with DependencyAPIMigration with ResolvedAnnotationPaths {
+class ConstantPropagation extends Transform with DependencyAPIMigration {
   import ConstantPropagation._
 
   override def prerequisites =
@@ -123,8 +123,6 @@ class ConstantPropagation extends Transform with DependencyAPIMigration with Res
     case firrtl.passes.Legalize => true
     case _                      => false
   }
-
-  override val annotationClasses: Traversable[Class[_]] = Seq(classOf[DontTouchAnnotation])
 
   sealed trait SimplifyBinaryOp {
     def matchingArgsValue(e: DoPrim, arg: Expression): Expression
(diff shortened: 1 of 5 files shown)


albert-magyar (approved):
LGTM. The one thing I wanted to look at was that maybe there might be non-`Nil` `component` fields (i.e. for ports of instances), but clearly the code before this PR already mapped the `dontTouch`-ed reference targets to ref names.

PR #1906, opened 2020-09-23
Speed up writing CustomFileEmission with buffering

jackkoenig:
Turns out buffering is useful.

For a medium-sized design, this reduces the time to write CHIRRTL from 60 seconds to 200-500 ms.

Adding special handling for common case of `Array[Byte]`* further reduced the time to write the file to ~60-70 ms.

*which comes from having already created a `String` in the `ChirrtlEmitter`, something we should consider doing differently

### Contributor Checklist

- [NA] Did you add at least one test demonstrating the PR?

#### Type of Improvement

  - performance improvement  

#### API Impact

No impact

No impact

#### Desired Merge Strategy

 - Squash

#### Release Notes

Speed up writing of annotations that use `CustomFileEmission`

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/options/phases/WriteOutputAnnotations.scala
+++ b/src/main/scala/firrtl/options/phases/WriteOutputAnnotations.scala
@@ -6,7 +6,7 @@ import firrtl.AnnotationSeq
 import firrtl.annotations.{Annotation, DeletedAnnotation, JsonProtocol}
 import firrtl.options.{CustomFileEmission, Dependency, Phase, PhaseException, StageOptions, Unserializable, Viewer}
 
-import java.io.{File, FileOutputStream, PrintWriter}
+import java.io.{BufferedOutputStream, File, FileOutputStream, PrintWriter}
 
 import scala.collection.mutable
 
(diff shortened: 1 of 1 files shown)


jackkoenig:
This isn't strictly needed for `1.4.0`, I think it'd be safe to squeeze in without another release candidate though. Otherwise it can be part of `1.4.1`, although we should consider doing it relatively quickly in that case.

src/main/scala/firrtl/options/phases/WriteOutputAnnotations.scala
```diff
@@ -37,8 +37,12 @@ class WriteOutputAnnotations extends Phase {
...
         filesWritten.get(canonical) match {
           case None =>
-            val w = new FileOutputStream(filename)
-            a.getBytes.foreach(w.write(_))
+            val w = new BufferedOutputStream(new FileOutputStream(filename))
+            a.getBytes match {
+              case arr: mutable.WrappedArray[Byte] => w.write(arr.array)
+              case other => other.foreach(w.write(_))
+            }
+            w.flush()
```
seldridge, line 45: Is this flush necessary given that the stream is closed on the next line?

seldridge (approved):
Approach looks great! Fix the 2.13 build issue and this is good to go.

src/main/scala/firrtl/options/phases/WriteOutputAnnotations.scala
jackkoenig: I read something on StackOverflow that suggested it was, but further reading indicates it is not, will remove.

jackkoenig:
@seldridge @chick @azidar @albert-magyar Any thoughts about merging PRs right now? This change is reasonable for 1.4.1, or snuck into an RC4 if one becomes needed (🤞 not). It's such a small change I think it could be _snuck_ into 1.4.0 without another RC but perhaps we should not go down that slippery slope.

seldridge:
I'm in favor of just sneaking it in.

PR #1782, opened 2020-07-22
Update negative literal emission

albertchen-sifive:
fixes https://github.com/freechipsproject/firrtl/issues/1755

This changes the way verilog is emitted for minimum value negative literals. in the example
```
module AddTester_ref(
  output [1:0] out
);
  assign out = -1'sh1 - 2'sh1;
endmodule
```
`-1'sh1` evaluates to `2'sh1` since `1'sh1` is already -1. This PR updates the emitter to not prepend the unary `-` to the emitted literal, since the hex value is already correct.

### Contributor Checklist

- [x] Did you add Scaladoc to every public function/method?
- [n/a] Did you update the FIRRTL spec to include every new feature/behavior?
- [x] Did you add at least one test demonstrating the PR?
- [x] Did you delete any extraneous printlns/debugging code?
- [x] Did you specify the type of improvement?
- [x] Did you state the API impact?
- [x] Did you specify the code generation impact?
- [x] Did you request a desired merge strategy?
- [x] Did you add text to be included in the Release Notes for this change?

#### Type of Improvement

- bug fix

#### API Impact

none

#### Backend Code Generation Impact

instead of `-1'h1`, `-2'h2`, `-3'h4`, ... emits `1'h1`, `2'h2`, `3h4`, ...

#### Desired Merge Strategy

- Squash

### Reviewer Checklist (only modified by reviewer)
- [ ] Did you add the appropriate labels?
- [ ] Did you mark the proper milestone (1.2.x, 1.3.0, 1.4.0) ?
- [ ] Did you review?
- [ ] Did you check whether all relevant Contributor checkboxes have been checked?
- [ ] Did you mark as `Please Merge`?

PR #1374, opened 2020-02-10
Fixing lint error: x + -1

azidar:
### Checklist

- [x] Did you specify the type of improvement?
- [x] Did you state the API impact?
- [x] Did you specify the code generation impact?
- [x] Did you request a desired merge strategy?
- [x] Add scaladoc to all public functions
- [x] Add license description on all new files

### Type of Improvement

   - backend code generation 

### Backend Code Generation Impact

Generates lint-clean Verilog for the case: x + -1, where x is anything and 1 is any literal.

Master behavior:
```
input x : SInt<8>
output z : SInt<9>
z <= add(x, SInt(-2))
```
generates
```verilog
assign z = $signed(x) + -8'sh2;
```

After this PR:
```verilog
assign z = $signed(x) - 8'sh2;
```

If the literal is the maximum possible literal, a special case is triggered to properly trim the resulting subtraction.

Input:
```
input x : SInt<2>
output z : SInt<3>
z <= add(x, SInt(-2))
```
now generates (after this PR)
```verilog
assign _GEN_2 = $signed(x) - 3'sh2;
assign z = _GEN_2;
```

### Reviewer Checklist
 - [ ] Did you mark the proper milestone (1.2.x, 1.3.0, 1.4.0) ?
 - [ ] Did you add proper labels?
 - [ ] Did you review?
 - [ ] Did you mark as `Please Merge`?

PR #1915, opened 2020-10-02
Fix "fix" for negative literals > 32 bits

jackkoenig:
Overflow of 32-bit Int would cause any negative literal value equal to
-(2^(width % 32 - 1)) where width >= 32 to be incorrectly inverted.

From the added tests:
```
z <= asSInt(and(x, SInt<34>(-2)))
```
Would compile to
```verilog
assign z = $signed(x) & 34'sh2;
```
This PR fixes it to the correct:
```verilog
assign z = $signed(x) & -34'sh2;
```

This bug was introduced in https://github.com/freechipsproject/firrtl/pull/1374 but would only manifest as an assertion failure. https://github.com/freechipsproject/firrtl/pull/1782 exposed the behavior to all ops (not just addition) in 1.4. Both of those PRs are present on `1.2.x`, so this bugfix needs to be backported.

Note, that https://github.com/freechipsproject/firrtl/pull/1782 causing the inversion to occur for all ops (not just addition) is a little weird, but technically correct. Inverting the max negative literal for a given width overflows back to the negative literal, but I wouldn't be surprised if this is a lint warning. In any case, this PR does **not** fix that (potential) issue, it only fixes the bug.

Please merge commit this to master and then to `1.4.x`, I intend to use this commit in the rocket-chip bump if possible.

### Contributor Checklist

#### Type of Improvement

 - bug fix   

#### API Impact

No impact

No impact

#### Desired Merge Strategy

   - Rebase

#### Release Notes

Fix rare bug where performing an op with a negative literal value equal to -(2^(width % 32 - 1)) where width >= 32 would be emitted incorrectly

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/transforms/FixAddingNegativeLiteralsTransform.scala
+++ b/src/main/scala/firrtl/transforms/FixAddingNegativeLiteralsTransform.scala
@@ -17,7 +17,7 @@ object FixAddingNegativeLiterals {
     * @param width width of the negative number
     * @return maximum negative number
     */
-  def minNegValue(width: BigInt): BigInt = -(1 << (width.toInt - 1))
+  def minNegValue(width: BigInt): BigInt = -(BigInt(1) << (width.toInt - 1))
 
   /** Updates the type of the DoPrim from its arguments (e.g. if is UnknownType)
     * @param d input DoPrim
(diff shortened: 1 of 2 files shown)


aswaterman (approved):
Thanks for the fix.

azidar:
*looks at #1374* gosh darn. Good catch @aswaterman and @jackkoenig

azidar (approved):

aswaterman:
Gotta type “BigInt” three times on the same line for it to take effect, kind of like saying Beetlejuice.

PR #1926, opened 2020-10-13
Make {Stage, FirrtlStage}.run protected

seldridge:
Add the `protected` access modifier to the `Stage.run` method. Carry this through to `FirrtlStage.run` as well.

This enables users to build Stage-based APIs which don't expose the internal `run` method. This should help avoid confusion as there are currently three ways to call a `Stage`:

1. `execute: (Array[String], AnnotationSeq) => AnnotationSeq`
2. `transform: AnnotationSeq => AnnotationSeq`
3. `run: AnnotationSeq => AnnotationSeq`

(1) and (2) are variants of the same thing and provide mandatory `Phase`s that do checks of the annotations, mandatory file reads/writes, and wrap internal `Phase`s. `run` is the internal method that gets packaged in an anonymous `Phase`. Ergo, `run` was never meant to be exposed, but I left it wide open. 😬 

### Contributor Checklist

- [n/a] Did you add at least one test demonstrating the PR?

#### Type of Improvement

- new feature/API

#### API Impact

This removes the `run` method from `Stage` and `FirrtlStage`.

None.

#### Desired Merge Strategy

#### Release Notes

- Remove `run` method from `Stage` and `FirrtlStage` (please use `transform` or `execute`)

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/options/Stage.scala
+++ b/src/main/scala/firrtl/options/Stage.scala
@@ -23,7 +23,7 @@ abstract class Stage extends Phase {
     * @param annotations input annotations
     * @return output annotations
     */
-  def run(annotations: AnnotationSeq): AnnotationSeq
+  protected def run(annotations: AnnotationSeq): AnnotationSeq
 
   /** Execute this stage on some input annotations. Annotations will be read from any input annotation files.
     * @param annotations input annotations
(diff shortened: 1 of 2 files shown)


azidar:
Is this going to affect all subclasses of Stage, because they will have to add the `protected` keyword?

seldridge:
This is binary incompatible, but shouldn't break any code. I.e., users who extend `Stage` can at their discretion add the `protected` modifier, but they don't have to.

Still, this is a 3.5 PR.

azidar (approved):

PR #1928, opened 2020-10-21
Add --dump-on-crash to StageOptions and Shell

timsnyder:
### Contributor Checklist

#### Type of Improvement

 - code refactoring
Moves the code for the annotationSeq trace logger into an AnnotationSeq.serialize method
 - code cleanup
I did a little code gardening to fix a typo in a comment and update a couple of outdated scaladoc blocks.  Not sure how you guys feel about having that stuff mixed in.

- new feature/API 
Adds `--dump-on-crash` see below.

#### API Impact

Adds a new `--dump-on-crash` to `StageOptions` that writes the raw compiler annotations to error.anno.json when an exception is caught in `CatchExceptions`

None known.

#### Desired Merge Strategy

squashing is fine.

#### Future Work
- split the FirrtlCircuitAnnotation out into a .fir file
- add support for encapsulating a CircuitState with a caught exception. For example,
  one could catch in firrtl.Transform#remapAnnotations, augmenting the exception
  with the 'before' and 'after' CircuitState.  Then, CatchExceptions can unpack the
  additional debug info and rethrow original exception.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/Compiler.scala
+++ b/src/main/scala/firrtl/Compiler.scala
@@ -19,6 +19,22 @@ import firrtl.stage.Forms
 /** Container of all annotations for a Firrtl compiler */
 class AnnotationSeq private (private[firrtl] val underlying: List[Annotation]) {
   def toSeq: Seq[Annotation] = underlying.toSeq
+
+  /** Raw serialization of Firrtl compiler annotations for offline debug
+    *
+    * @see [[firrtl.options.phases.WriteOutputAnnotations]] for the canonical mechanism
+    *     of writing [[firrtl.options.StageOptions.annotationFileOut]]
+    */
+  def serialize: String =
+    JsonProtocol
+      .serializeTry(this)
+      .recoverWith {
+        case NonFatal(e) =>
+          val msg = s"Exception thrown during Annotation serialization:\n  " +
+            e.toString.replaceAll("\n", "\n  ")
+          Try(msg)
+      }
+      .get
 }
 object AnnotationSeq {
   def apply(xs: Seq[Annotation]): AnnotationSeq = new AnnotationSeq(xs.toList)
@@ -220,17 +236,7 @@ private[firrtl] object Transform {
 
     logger.info(s"Form: ${after.form}")
     logger.trace(s"Annotations:")
-    logger.trace {
-      JsonProtocol
-        .serializeTry(remappedAnnotations)
-        .recoverWith {
-          case NonFatal(e) =>
-            val msg = s"Exception thrown during Annotation serialization:\n  " +
-              e.toString.replaceAll("\n", "\n  ")
-            Try(msg)
-        }
-        .get
-    }
+    logger.trace(after.annotations.serialize)
 
     logger.trace(s"Circuit:\n${after.circuit.serialize}")
     logger.info(s"======== Finished Transform $name ========\n")
--- a/src/main/scala/firrtl/options/StageAnnotations.scala
+++ b/src/main/scala/firrtl/options/StageAnnotations.scala
@@ -168,3 +168,17 @@ case object WriteDeletedAnnotation extends NoTargetAnnotation with StageOption w
   )
 
 }
+
+/** If this [[firrtl.annotations.Annotation Annotation]] exists in an [[firrtl.AnnotationSeq AnnotationSeq]], then
+  * Dump the raw [[firrtl.AnnotationSeq AnnotationSeq]] to a file on an error
+  */
+case object DumpOnCrashAnnotation extends NoTargetAnnotation with StageOption with HasShellOptions {
+
+  val options = Seq(
+    new ShellOption[Unit](
+      longOption =  "dump-on-crash",
+      toAnnotationSeq = { _ => Seq(this) },
+      helpText = "Dump raw Annotations to error.anno.json on error"
+    )
+  )
+}
--- a/src/main/scala/firrtl/stage/phases/CatchExceptions.scala
+++ b/src/main/scala/firrtl/stage/phases/CatchExceptions.scala
@@ -2,7 +2,10 @@
 
 package firrtl.stage.phases
 
-import firrtl.options.{DependencyManagerException, OptionsException, Phase, PhaseException}
+import java.io.PrintWriter
+import java.lang.AssertionError
+
+import firrtl.options.{DependencyManagerException, OptionsException, Phase, PhaseException, StageOptions, Viewer}
 import firrtl.{
   AnnotationSeq,
   CustomTransformException,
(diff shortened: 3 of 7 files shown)


src/main/scala/firrtl/options/StageAnnotations.scala
```diff
@@ -168,3 +168,17 @@ case object WriteDeletedAnnotation extends NoTargetAnnotation with StageOption w
...
+case object DumpOnCrashAnnotation extends NoTargetAnnotation with StageOption with HasShellOptions {
+
+  val options = Seq(
+    new ShellOption[Unit](
+      longOption =  "dump-on-crash",
+      toAnnotationSeq = { _ => Seq(this) },
+      helpText = "Dump raw Annotations to error.anno.json on error"
+    )
+  )
+}
```
seldridge, line 184: High level: this is an option that is associated with the `firrtl.options` package (which are intended to be common to all Stages, e.g., both `FirrtlStage` and `ChiselStage`). However, this is only handled inside a FIRRTL-specific phase wrapper, `CatchExceptions`. I think this means that this is then really a `FirrtlOption` specific to FIRRTL and not a `StageOption`.

Alternatively, this behavior could be something general to all stages and could be implemented as such.

timsnyder: Thanks for clarifying the difference between `StageOption` and `FirrtlOption`.  It probably makes sense for this to be handled somewhere for `ChiselStage` as well.  I just happened to be. debugging something in firrtl this time.

src/main/scala/firrtl/stage/phases/CatchExceptions.scala
```diff
@@ -32,10 +35,27 @@ class CatchExceptions(val underlying: Phase) extends Phase {
...
+      throw p }
+    case CustomTransformException(cause) => {
+      dumpOnError(underlying, a)
+      throw cause }
+    case e: Exception => {
+      dumpOnError(underlying, a)
+      Utils.throwInternalError(exception = Some(e))}
+  }
+
+  private[CatchExceptions] def dumpOnError(underlying: Phase, a: AnnotationSeq) = {
```
jackkoenig, line 49:
```suggestion
  private def dumpOnError(underlying: Phase, a: AnnotationSeq) = {
```

src/main/scala/firrtl/Compiler.scala
```diff
@@ -19,6 +19,22 @@ import firrtl.stage.Forms
...
+  def serialize: String =
+    JsonProtocol
+      .serializeTry(this)
+      .recoverWith {
+        case NonFatal(e) =>
+          val msg = s"Exception thrown during Annotation serialization:\n  " +
+            e.toString.replaceAll("\n", "\n  ")
+          Try(msg)
+      }
+      .get
```
jackkoenig, line 37: I don't think `AnnotationSeq` should depend on `JsonProtocol`. Instead I'd put this on `JsonProtocol` with a name like `serializeOrErrorMsg` or something.

src/main/scala/firrtl/options/StageAnnotations.scala
jackkoenig: Chisel is a bit more monolithic and doesn't have a way of dumping partial circuits (at least currently), it would be nice though...

src/main/scala/firrtl/stage/phases/CatchExceptions.scala
```diff
@@ -32,10 +35,27 @@ class CatchExceptions(val underlying: Phase) extends Phase {
...
-        _: FirrtlInternalException | _: PhaseException | _: DependencyManagerException) =>
-      throw p
-    case CustomTransformException(cause) => throw cause
-    case e: Exception => Utils.throwInternalError(exception = Some(e))
+        _: FirrtlInternalException | _: PhaseException | _: DependencyManagerException | _: AssertionError) => {
+      dumpOnError(underlying, a)
+      throw p }
+    case CustomTransformException(cause) => {
+      dumpOnError(underlying, a)
+      throw cause }
```
jackkoenig, line 43: I think you're going to need to run Scalafmt: `sbt scalafmtAll`

src/main/scala/firrtl/stage/phases/CatchExceptions.scala
```diff
@@ -32,10 +35,27 @@ class CatchExceptions(val underlying: Phase) extends Phase {
...
+      dumpOnError(underlying, a)
+      Utils.throwInternalError(exception = Some(e))}
+  }
+
+  private[CatchExceptions] def dumpOnError(underlying: Phase, a: AnnotationSeq) = {
+    val sopts = Viewer[StageOptions].view(a)
+
+    if (sopts.dumpOnCrash) {
+      val dumpfile = sopts.getBuildFileName("error", Some(".anno.json"))
+      println(s"Dumping AnnotationSeq to ${dumpfile}")
```
jackkoenig, line 54:
```suggestion
      logger.warn(s"Dumping annotations to ${dumpfile}")
```

src/main/scala/firrtl/stage/phases/CatchExceptions.scala
jackkoenig: Also note that it's not necessary to wrap the right-hand side of a `case` in curly braces, and I guess stylistically discouraged (by me)

jackkoenig (commented):
Thanks for this @timsnyder! This is something I've wanted for a long time so I'm excited to finally have it!

jackkoenig:
I'm not sure why Travis isn't running...

src/main/scala/firrtl/options/StageAnnotations.scala
timsnyder: So, for now should I move this back to being a  `FirrtlOption`?  I had it in there before I got distracted with `Viewer` and `StageOption` and decided to move it:
shiny meme

src/main/scala/firrtl/stage/phases/CatchExceptions.scala
timsnyder: will `sbt scalafmtAll` before next push

timsnyder:
Does this need a test?  If so, can you suggest where I try to add one?

src/main/scala/firrtl/options/StageAnnotations.scala
jackkoenig: Thoughts @seldridge?

seldridge: If it's *not* intended to be common to all stages, then I'd prefer to have it as a `FirrtlOption`. It should be a pretty straight migration, including the shiny typeclass stuff: https://github.com/freechipsproject/firrtl/blob/master/src/main/scala/firrtl/stage/package.scala#L23.

seldridge:
A test could be good to lock in that this does the right thing. 

I'd look at how https://github.com/freechipsproject/firrtl/blob/master/src/test/scala/firrtlTests/stage/FirrtlMainSpec.scala works. You could either piggy-back off of `FirrtlMainTest` and write one of those, checking any expected prints and files, or use the lower-level `catchWrites` API that you can use to trap if a file write occurs and then verify what the filename is.

Issue #1959, opened 2020-11-14
RemoveAccesses failed remove nested SubAccess.index


sequencer:
Seems to be a bug that `RemoveAccesses` failed to remove a nested index.
Here is the smallest example to trigger this bug:
```
circuit Bug :
  module Bug :
    input in0 : UInt<1> [2][2]
    input in1 : UInt<1> [2]
    input in2 : UInt<1> [2]
    output out : UInt<1>
    out <= in0[in1[in2[0]]][in1[in2[1]]]
```
This bug hasn't been spotted for 4 or 5 years. Since most in most cases, `index` has been resolved by previous transforms.
In `Bug` circuit, previous cases failed to resolve, which makes `index` still being a index in `RemoveAccess`, then `LowerTypes` throw it out.
 
### Checklist
- [x] Did you specify the current behavior?
- [x] Did you specify the expected behavior?
- [x] Did you provide a code example showing the problem?
- [x] Did you describe your environment?
- [x] Did you specify relevant external information?

### What is the current behavior?
```
LowerTypes expects all SubAccesses to have been expanded! in1[in2[0]]
```

### What is the expected behavior?
emit verilog successfully.
 
### Steps to Reproduce
https://scastie.scala-lang.org/PSgW9ZHDTUy1Jy2trhbiaQ

### Your environment
Chisel 3.4.0
FIRRTL 1.4.0
OpenJDK 14

ekiwi:
Sound to me like there is a problem in the `RemoveAccesses` pass.

jackkoenig:
Fixed by https://github.com/chipsalliance/firrtl/pull/1984

PR #1984, opened 2020-11-22
Fix subaccess

sequencer:
### Contributor Checklist

#### Type of Improvement
- bug fix

#### API Impact
None

Bug fix for nested `SubAccess`, introduce one more `And` in `firrtl.passes.RemoveAccesses`, but should be eliminated in next optimization passes.
In nested `SubAccess`, fix the bug in #1959 

#### Desired Merge Strategy

#### Release Notes
Bug fix for nested SubAccesses

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/passes/RemoveAccesses.scala
+++ b/src/main/scala/firrtl/passes/RemoveAccesses.scala
@@ -47,8 +47,7 @@ object RemoveAccesses extends Pass {
     *   Seq(Location(a[0], UIntLiteral(0)), Location(a[1], UIntLiteral(1)))
     */
   private def getLocations(e: Expression): Seq[Location] = e match {
-    case e: WRef => create_exps(e).map(Location(_, one))
-    case e: WSubIndex =>
+    case e: SubIndex =>
       val ls = getLocations(e.expr)
       val start = get_point(e)
       val end = start + get_size(e.tpe)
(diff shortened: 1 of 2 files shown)


ekiwi:
Have you considered creating a node for the index expression?
I am just wondering, whether it would make for better firrtl and for a simpler compiler if we would generate nodes for all non-trivial index expressions and then visit them.

sequencer:
I got your point. That would be much better and reasonable to give most of FIRRTL generated expression with a readable name. It will much benefit generated verilog.
Just peek the original implementation before go to bed, I think there is still lacking of a infrastructure to get context for an `Expression`.
For example, most functions in a transform are directly manipulating a small `FirrtlNode`, which don't know where it is, and thus it still lack of enough information to construct its desired name for creating a `DefNode`.
I think we can have a global context cache, something like `CircuitGraph`. Or give every `FirrtlNode` a `parents` and `children` to reflect its own location in the AST, and add another pass to analyse AST, if we need to use the hierarchy of AST tree.

ekiwi:
> That would be much better and reasonable to give most of FIRRTL generated expression with a readable name.

I wasn't too concerned about readable names, for now, adding a new `_GEN_...` node should be fine.

sequencer:
I see, but what benefits can be provided, if this node have no readable name?

src/main/scala/firrtl/passes/RemoveAccesses.scala
```diff
@@ -66,17 +65,27 @@ object RemoveAccesses extends Pass {
...
+          getLocations(index).map {
+            case Location(indexBase, indexGuard) =>
+              Location(
+                exprBase,
+                AND(
+                  AND(
+                    indexGuard,
+                    exprGuard
+                  ),
+                  EQV(
```
azidar, line 80: I think this expression could be cached, but IMO this is such a rare occurrence I don't think its too bad.

azidar (approved):

PR #2024, opened 2020-12-15
Improve performance of LowerTypes renaming

jackkoenig:
In addition to the added tests, I tested this with 3 different designs, diffing the Verilog and output annotation files and saw the results were identical. Also, pretty sure this is correct because I originally used the `InstanceKeyGraph` from _after_ transformation in `LowerTypes` instead of before, and 2/3 of my designs failed.

Also, not very scientific, but a design that was running very slowly during renaming in LowerTypes had it's total FIRRTL runtime reduced from `36m45s` to `20m45s`, so a massive win.

### Contributor Checklist

#### Type of Improvement

   - performance improvement

#### API Impact

This adds some new public utilities

No impact

#### Desired Merge Strategy

  - Squash

#### Release Notes

Improve performance and memory use of `LowerTypes`

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/passes/LowerTypes.scala
+++ b/src/main/scala/firrtl/passes/LowerTypes.scala
@@ -8,8 +8,10 @@ import firrtl.annotations.{
   MemoryInitAnnotation,
   MemoryRandomInitAnnotation,
   ModuleTarget,
-  ReferenceTarget
+  ReferenceTarget,
+  TargetToken
 }
+import TargetToken.{Instance, OfModule}
 import firrtl.{
   CircuitForm,
   CircuitState,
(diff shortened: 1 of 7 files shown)


src/main/scala/firrtl/passes/LowerTypes.scala
```diff
@@ -296,11 +304,11 @@ private object DestructTypes {
...
   def destructInstance(
-    m:         ModuleTarget,
-    instance:  DefInstance,
-    namespace: Namespace,
-    renameMap: RenameMap,
-    reserved:  Set[String]
+    m:           ModuleTarget,
+    instance:    DefInstance,
+    namespace:   Namespace,
+    instRenames: mutable.ListBuffer[(Instance, Instance)],
```
ekiwi, line 310: since you are appending, should this better be an ArrayBuffer?

jackkoenig: `ListBuffer` has constant time for both prepend and append: https://www.scala-lang.org/api/2.12.11/scala/collection/mutable/ListBuffer.html. I use `ListBuffer` whenever I want to mutably build up an immutable `Seq` because `ListBuffer.toList` is free (copy on write to the underlying buffer).

ekiwi: Cool. I learned something!

ekiwi (approved):
LGTM as long as all tests pass.

jackkoenig:
> LGTM as long as all tests pass.

They were passing but failed because I made `RenameMap.fromInstanceRenames` package private which broke a test 🙈 . Fixed now.

src/main/scala/firrtl/passes/LowerTypes.scala
jackkoenig: Yeah it's pretty neat. I think once we're on 2.13 I might shift to using `VectorBuilder` since `Vector` is much better in `2.13.2` and has better memory characteristics than `List`.

PR #2036, opened 2021-01-08
smt: run DeadCodeElimination after PropagatePresetAnnotations

ekiwi:
This enables an important pattern in Chisel:

```scala
val preset = WireInit(false.B.asAsyncReset())
  override def toFirrtl = PresetAnnotation(preset.toTarget)
})
withReset(preset) {
  // ...
}
```

Note that the problematic part is the use of `WireInit` which is required in order not to run afoul of wire initialization checks in the compiler frontend.

### Contributor Checklist

#### Type of Improvement

- bug fix

#### API Impact
- none

- only affects SMT/Btor2 backends, no change to Verilog generaion

#### Desired Merge Strategy

- squash

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/backends/experimental/smt/FirrtlToTransitionSystem.scala
+++ b/src/main/scala/firrtl/backends/experimental/smt/FirrtlToTransitionSystem.scala
@@ -10,18 +10,8 @@ import firrtl.options.Dependency
 import firrtl.passes.PassException
 import firrtl.stage.Forms
 import firrtl.stage.TransformManager.TransformDependency
-import firrtl.transforms.PropagatePresetAnnotations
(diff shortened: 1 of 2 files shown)


ekiwi:
> This may hint that `PropagatePresetAnnotations` is invalidating `DeadCodeElimination`.

I could have also fixed this by only lazily throwing errors. Essentially my problem is that there is a left over AsyncReset expression which I cannot translate. It is never used, so it is not actually a problem for the final translation. However tracking which signals are used would make my translation code harder to write.

jackkoenig (approved):
Looks good, just need to migrate off Travis due to extremely limited open-source credits

PR #2038, opened 2021-01-12
Split More, Necessary Expressions

### LINKED ISSUE #2037: Invalid Verilog if input IR is improperly inlined
The FIRRTL compiler correctly respects self-determined and context-determined widths when inlining expression. However, if an expression is _already_ inlined, it won't be broken back out and you can get buggy Verilog.

There's a full example in [this comment](https://github.com/llvm/circt/issues/394#issuecomment-757653946).

If you start with the following circuit:

```
circuit Foo :
  module Foo :
    input a : UInt<1>
    input inp_6 : UInt<13>
    output tmp13 : SInt<14>
    wire tmp11 : UInt<14>
    tmp11 <= add(inp_6, UInt<12>("h1"))
    node _tmp13_T_1 = mux(a, asSInt(tmp11), asSInt(UInt<13>("h1")))
    tmp13 <= _tmp13_T_1
```

This will inline `tmp11` into:

```verilog
module Foo(
  input         a,
  input  [12:0] inp_6,
  output [13:0] tmp13
);
  assign tmp13 = a ? $signed(inp_6 + 13'h1) : $signed(14'sh1);
endmodule
```

However, if you start with this un-inlined:

```
circuit Foo :
  module Foo :
    input a : UInt<1>
    input inp_6 : UInt<13>
    output tmp13 : SInt<14>
    wire tmp11 : UInt<14>
    tmp11 <= add(inp_6, UInt<12>("h1"))
    node x = asSInt(tmp11)
    node _tmp13_T_1 = mux(a, x, asSInt(UInt<13>("h1")))
    tmp13 <= _tmp13_T_1
```

It won't be inlined:

```verilog
module Foo(
  input         a,
  input  [12:0] inp_6,
  output [13:0] tmp13
);
  wire [13:0] x = inp_6 + 13'h1;
  assign tmp13 = a ? $signed(x) : $signed(14'sh1);
endmodule
```

For safety, expressions may need to be broken out into nodes during parsing. However, this bug should be impossible to hit with the current way that Chisel emits CHIRRTL.

### Checklist

See above.

You should get the second circuit if you input the first circuit.

### Steps to Reproduce

Compile the above code with 1.4.1.

### Your environment

1.4.1

### External Information

See discussion on: https://github.com/llvm/circt/issues/394

seldridge:
Fixes #2037. The example code there will now produce:

```verilog
module Foo(
  input         a,
  input  [12:0] inp_6,
  output [13:0] tmp13
);
  wire [13:0] _GEN_0 = inp_6 + 13'h1;
  assign tmp13 = a ? $signed(_GEN_0) : $signed(14'sh1);
endmodule
```

Change `SplitExpressions` to split more expressions. Previously, the root expression would not be split out, e.g., the true or false condition from a mux which can lead to invalid Verilog like in #2037.

Now, all expressions will be split out unless they are trivial assignments or connects (this transform requires low form so it can safely ignore partial connects). This has the effect of also ensuring idempotency of the transform.

This exposed a weird bug where the invalid types that `VerilogModulusCleanup` produces (it would produce a `bits` op of type `SInt`) would then cause the Verilog emitter to serialize an `UnknownType` string. This PR adds an `asSInt` op and returns the correct types to avoid this.

Note: this may produce slightly more verbose Verilog for situations involving a `rem` op on an `SInt` type which can be cleaned up later with invalidations on inlining, CSE, and DCE. However, that increases the footprint of this PR substantially and changes transform ordering.

### Contributor Checklist

#### Type of Improvement

- bug fix

#### API Impact

None.

This fixes a possible bug in generated code, but may produce additional temporary wires for cases involving `rem` of `SInt` types.

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/passes/SplitExpressions.scala
+++ b/src/main/scala/firrtl/passes/SplitExpressions.scala
@@ -62,7 +62,7 @@ object SplitExpressions extends Pass {
 
       (s match {
         case _: DefNode | _: Connect => s.map(onExp(true))
-        case _                       => s.map(onExp(false))
+        case _ => s.map(onExp(false))
       }) match {
         case x: Block => x.map(onStmt)
         case EmptyStmt => EmptyStmt
--- a/src/main/scala/firrtl/passes/SplitExpressions.scala
+++ b/src/main/scala/firrtl/passes/SplitExpressions.scala
@@ -49,14 +49,24 @@ object SplitExpressions extends Pass {
         case _ => e
       }
 
-      // Recursive. Splits compound nodes
-      def onExp(e: Expression): Expression =
-        e.map(onExp) match {
-          case ex: DoPrim => ex.map(split)
+      /* Split expressions consisting of compound nodes. Do not split the outer expression if it is the RHS of an assignment
+       * (node or connect).  This prevents generating unnecessary nodes and also makes this transform idempotent.
+       */
+      def onExp(e: Expression, isAssign: Boolean): Expression =
+        e.map(onExp(_, false)) match {
+          case ex: DoPrim =>
+            val exx = ex.map(split)
+            isAssign match {
+              case true  => exx
+              case false => split(exx)
+            }
           case ex => ex
         }
 
-      s.map(onExp) match {
+      (s match {
+        case _: DefNode | _: Connect => s.map(onExp(_, true))
+        case _                       => s.map(onExp(_, false))
+      }) match {
         case x: Block => x.map(onStmt)
         case EmptyStmt => EmptyStmt
         case x =>
--- a/src/main/scala/firrtl/passes/SplitExpressions.scala
+++ b/src/main/scala/firrtl/passes/SplitExpressions.scala
@@ -56,10 +56,7 @@ object SplitExpressions extends Pass {
         e.map(onExp(_, false)) match {
           case ex: DoPrim =>
             val exx = ex.map(split)
-            isAssign match {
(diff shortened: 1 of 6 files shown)


src/main/scala/firrtl/passes/SplitExpressions.scala
```diff
@@ -49,14 +49,24 @@ object SplitExpressions extends Pass {
...
+       * (node or connect).  This prevents generating unnecessary nodes and also makes this transform idempotent.
+       */
+      def onExp(e: Expression, isAssign: Boolean): Expression =
+        e.map(onExp(_, false)) match {
+          case ex: DoPrim =>
+            val exx = ex.map(split)
+            isAssign match {
+              case true  => exx
+              case false => split(exx)
+            }
```
jackkoenig, line 62:
```suggestion
            if (isAssign) exx else split(exx)
```

jackkoenig (approved):

jackkoenig:
There appears to be a real failure here. I suspect that this bugfix is exposing other bugs, but it needs digging. Sadly this will miss 3.4.2/1.4.2 but this does merit a closer look.

seldridge:
I wasn't able to reproduce the `sbt test` failure locally for the three failing tests. I am using a newer version of Yosys (0.9+3672) whereas I think the tests are running with 0.8.

The formal equivalence failure may be warranted, but I'd have to investigate.

PR #2057, opened 2021-01-26
Allow Side Effecting Statement to have Names

ekiwi:
### TODO:
- [x] test that a circuit which has a name collision with a named statement is rejected
- [x] test that a circuit which contains a reference to a statement is rejected
- [x] add new feature to the firrtl spec

### Contributor Checklist

#### Type of Improvement
- new feature: named print, stop and verification statements

#### API Impact
- extends Print, Stop and Verification nodes with a `name` field
- statement labels are intentionally kept optional so that old firrtl should parse and work with no problem

- this should not change the Verilog at the moment

#### Desired Merge Strategy

- squash

#### Release Notes
- print, stop, assume, assert and cover statements can now be given a name which allows them to be annotated

- [ ] Did you review?


ekiwi:
Note, this PR is just for public discussion. The code currently does not compile.

spec/spec.tex
```diff
@@ -989,10 +989,15 @@ \subsection{Stops} \label{stop_stmt}
 side-effect-having statements in different modules or with different clocks that trigger
 concurrently is undefined.
 
+The stop statement has an optional name attribute which can be used to
+attach meta data to the statement. The name is part of the module level
```
davidbiancolin, line 993:
```suggestion
attach metadata to the statement. The name is part of the module level
```

spec/spec.tex
```diff
@@ -1004,12 +1009,17 @@ \subsection{Formatted Prints}
 For information about execution ordering of clocked statements with observable environmental side
 effects, see section \ref{stop_stmt}.
 
+The printf statement has an optional name attribute which can be used to
+attach meta data to the statement. The name is part of the module level
```
davidbiancolin, line 1013:
```suggestion
attach metadata to the statement. The name is part of the module level
```

spec/spec.tex
```diff
@@ -1043,6 +1053,11 @@ \subsection{Verification}
 For information about execution ordering of clocked statements with observable environmental side
 effects, see section \ref{stop_stmt}.
 
+Any verification statement has an optional name attribute which can be used to
+attach meta data to the statement. The name is part of the module level
```
davidbiancolin, line 1057:
```suggestion
attach metadata to the statement. The name is part of the module level
```

Just nits.

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -618,23 +620,25 @@ object Stop {
...
-  string: StringLit,
-  args:   Seq[Expression],
-  clk:    Expression,
-  en:     Expression)
+  info:                      Info,
+  string:                    StringLit,
+  args:                      Seq[Expression],
+  clk:                       Expression,
+  en:                        Expression,
+  @since("FIRRTL 1.5") name: String = "")
```
davidbiancolin, line 628: neat.

src/main/scala/firrtl/passes/CheckHighForm.scala
```diff
@@ -270,7 +281,11 @@ trait CheckHighFormLike { this: Pass =>
         case NoInfo => minfo
         case x      => x
       }
-      s.foreach(checkName(info, mname, names))
+      val canBeReference = s match {
+        case _: Stop | _: Print | _: Verification => false
```
davidbiancolin, line 285: Maybe we can have a sealed trait in IR.scala that codifies this property.

src/main/scala/firrtl/transforms/EnsureNamedStatements.scala
```diff
@@ -0,0 +1,42 @@
...
+      val baseName = s.op match {
+        case Formal.Cover  => "cover"
+        case Formal.Assert => "assert"
+        case Formal.Assume => "assume"
+      }
+      s.withName(makeName(namespace, baseName))
+    case other => other.mapStmt(onStmt(namespace))
+  }
+
+  // we are using a slightly different algorithm than the basic namespace
```
davidbiancolin, line 37: Maybe succinctly explain the difference?

davidbiancolin (commented):
Pretty cool, i've been wanting this feature for a while. I think adding the trait might clean up the pattern matching a little bit and make it easier to add other similar nodes in the future, but i'll defer to you. 

What's the chisel strategy going to be for emitting these names?

src/main/scala/firrtl/passes/CheckHighForm.scala
ekiwi: This used to be `HasName` since every `Statement` that had a name would also be able to be referenced.
Should I add a `CanBeReference` trait that extends from `HasName`?

davidbiancolin: Maybe a self-type constraint, to keep the class hierarchy shallow? @jackkoenig has better-informed opinions about this sort of thing

jackkoenig: What if we make the side-effecting statements just extend `HasName` and not `IsDeclaration`? Alternatively, if we want to keep the meaning that `IsDeclaration` means "Annotatable", I think David's idea with a shallower type hierarchy is good so that it'll be more obvious:
```scala
case class DefWire(info: Info, name: String, tpe: Type) extends Statement with IsDeclaration with CanBeReference { ...
```
Verbose, yes, but no tracing through type hierarchies.

src/main/scala/firrtl/transforms/EnsureNamedStatements.scala
```diff
@@ -0,0 +1,43 @@
...
+      s.withName(makeName(namespace, baseName))
+    case other => other.mapStmt(onStmt(namespace))
+  }
+
+  // We are using a slightly different algorithm than the basic namespace.
+  // We want all generated names to end in a number like `assert_0`, `assert_1`, etc.
+  private def makeName(namespace: Namespace, base: String): String = {
+    val name = Iterator.from(0).map(i => s"${base}_$i").find(n => !namespace.contains(n)).get
+    namespace.newName(name)
+  }
```
jackkoenig, line 42:
```suggestion
```
Instead of all of this logic, you could just seed the `Namespace` with the base names to ensure it always starts with `_0`, that's what the [namespace itself does for `_GEN`](https://github.com/chipsalliance/firrtl/blob/6e0e760526090c694ce6507db71122654ffc3000/src/main/scala/firrtl/Namespace.scala#L11).

src/main/scala/firrtl/transforms/EnsureNamedStatements.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  override protected def execute(state: CircuitState): CircuitState = {
+    val c = state.circuit.mapModule(onModule)
+    state.copy(circuit = c)
+  }
+
+  private def onModule(m: DefModule): DefModule = m match {
+    case e:   ExtModule => e
+    case mod: Module =>
+      val namespace = Namespace(mod)
```
jackkoenig, line 20:
```suggestion
      val namespace = Namespace(mod)
      // Ensure we always start with _0 suffix
      Seq("cover", "assert", "assume").foreach(namespace.newName(_))
```

src/test/scala/firrtlTests/CheckSpec.scala
```diff
@@ -536,6 +566,20 @@ class CheckSpec extends AnyFlatSpec with Matchers {
...
+                |circuit test:
+                |  module test:
+                |    input clock: Clock
+                |    output a: UInt<2>
+                |    stop(clock, UInt(1), 1) : hello
+                |    a <= hello
+                |""".stripMargin
+    assertThrows[CheckHighForm.UndeclaredReferenceException] {
+      checkHighInput(src)
+    }
```
jackkoenig, line 580: Should we provide a more specific error for this case?

src/test/scala/firrtlTests/LowerTypesSpec.scala
```diff
@@ -321,6 +321,25 @@ class LowerTypesUniquifySpec extends FirrtlFlatSpec {
...
+        |    printf(clock, UInt(1), "") : a_0_c_
+        |    assert(clock, UInt(1), UInt(1), "") : a__0
+      """.stripMargin
+    val expected = Seq(
+      "node a___0_b = x_0_b",
+      "node a___1_c__1_e = x_1_c__1_e",
+      "node a___1_c_1_e = x_1_c_1_e"
+    )
+
+    executeTest(input, expected)
```
jackkoenig, line 340: I grant it would likely complicate the implementation a lot, but don't you think it'd be better for the actual hardware to win? Probably fine as future work.

jackkoenig (commented):
Looks really good overall, some minor comments.

src/test/scala/firrtlTests/LowerTypesSpec.scala
ekiwi: I would think that a chance of collision here is rather minimal and yes, preferring the hardware and renaming the statement instead would make our "uniquify" algorithm even more complicated than it already is.

src/main/scala/firrtl/transforms/EnsureNamedStatements.scala
ekiwi: I like this idea!

src/main/scala/firrtl/transforms/EnsureNamedStatements.scala
ekiwi: I will change to your suggested solution.

src/main/scala/firrtl/passes/Inline.scala
```diff
@@ -187,7 +187,10 @@ class InlineInstances extends Transform with DependencyAPIMigration with Registe
       renameMap:     RenameMap
     )(s:             Statement
     ): Statement = {
-      def onName(ofModuleOpt: Option[String])(name: String) = {
+      def onName(ofModuleOpt: Option[String])(name: String): String = {
+        // Empty names are allowed for backwards compatibility reasons and
+        // indicate that the entity has essentially no name and thus cannot be prefixed.
+        if (name.isEmpty) return name
```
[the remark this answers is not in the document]
ekiwi: I did not know that it was implemented via exceptions. I thought it would just be a simple control flow operation like an if/else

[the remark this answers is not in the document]
ekiwi: Seems like it isn't always implemented as an exception. Maybe we should add `-Xlint:nonlocal-return` to our `scalac` flags (only supported in Scala 2.13 + apparently)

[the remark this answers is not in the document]
seldridge: Save those early returns for C++... 😛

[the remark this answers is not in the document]
jackkoenig: It looks like the Scala standard library uses `return` sometimes so maybe it's fine. It's probably slow when the return is nested in anonymous functions but fine otherwise.

src/main/scala/firrtl/passes/CheckHighForm.scala
ekiwi: I can add the `CanBeReference` trait. Side-effecting statements should extend `IsDeclaration` since that makes them part of the module level namespace.

src/test/scala/firrtlTests/CheckSpec.scala
ekiwi: I think this should be good enough for now. Very few people actually write firrtl and when you use Chisel, this error case should never happen.

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -660,25 +679,27 @@ object Formal extends Enumeration {
...
-  pred: Expression,
-  en:   Expression,
-  msg:  StringLit)
+  op:                        Formal.Value,
+  info:                      Info,
+  clk:                       Expression,
+  pred:                      Expression,
+  en:                        Expression,
+  msg:                       StringLit,
+  @since("FIRRTL 1.5") name: String = "")
```
jackkoenig, line 688: This is super minor, but I think it'd be nice to have a newline between the `@since` and `name` just so that the types are spaced over so much

jackkoenig (approved):
LGTM

PR #2058, opened 2021-01-28
Stop padding multiply and divide ops

### LINKED ISSUE #2052: (div, cat) mismatch
### Checklist

- [ ] Did you specify relevant external information?

The following FIRRTL program.

```python
; seed: 371
circuit top_mod :
  module top_mod :
    input tmp8: UInt<13>
    input tmp11: UInt<11>
    output tmp40: UInt<24>
    wire tmp28: UInt<13>
    tmp28 <= div(tmp8, UInt<20>(1))
    tmp40 <= cat(tmp11, tmp28)
```

Running with `firrtl-1.5-SNAPSHOT` generates the following Verilog

```verilog
module top_mod(
  input  [12:0] tmp8,
  input  [10:0] tmp11,
  output [23:0] tmp40
);
  wire [19:0] _GEN_0 = {{7'd0}, tmp8};
  wire [19:0] tmp28 = _GEN_0 / 20'h1;
  assign tmp40 = {tmp11,tmp28};
endmodule
```

```verilog
module top_mod(
  input  [12:0] tmp8,
  input  [10:0] tmp11,
  output [23:0] tmp40);

  wire [12:0] tmp28;	// a_top_mod.fir:7:5

  assign tmp28 = tmp8;	// a_top_mod.fir:8:11
  assign tmp40 = {tmp11, tmp28};	// a_top_mod.fir:3:3, :9:14
endmodule
```

Which is not formally equivalent

```
ERROR: Found 11 unproven $equiv cells in 'equiv_status -assert'.
```

### Steps to Reproduce

`./firrtl-1.5-SNAPSHOT -i top_mod.fir`

### Your environment

`Linux linux-sydi 5.9.14-1-default #1 SMP Sat Dec 12 06:57:32 UTC 2020 (c648a46) x86_64 x86_64 x86_64 GNU/Linux`

- Verilator version: `Verilator 4.103 devel rev v4.102-42-g726e78fd`

jackkoenig:
Fixes #2052
Also fixes some Verilog lint issues

A lot of work for a change that is literally deleting what 12 characters?

I'm a bit paranoid about this change so I wrote some formal checks, but notably I'm not checking every width, and I'm relying on the reference Verilog being correct. Any thoughts?

As this is a correctness issue, at least the fix should go all the way back.

### Contributor Checklist

#### Type of Improvement

  - bug fix  
  - backend code generation

#### API Impact

No change

Slightly cleaner emission for multiplication and division

#### Desired Merge Strategy

  - Squash

#### Release Notes

Fix lint errors on multiply and divide operations
Fix incorrect Verilog on multiply or divide followed by concatenation

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/passes/PadWidths.scala
+++ b/src/main/scala/firrtl/passes/PadWidths.scala
@@ -58,7 +58,7 @@ object PadWidths extends Pass {
     case ex: ValidIf => ex.copy(value = fixup(width(ex.tpe))(ex.value))
     case ex: DoPrim =>
       ex.op match {
-        case Lt | Leq | Gt | Geq | Eq | Neq | Not | And | Or | Xor | Add | Sub | Mul | Div | Rem | Shr =>
+        case Lt | Leq | Gt | Geq | Eq | Neq | Not | And | Or | Xor | Add | Sub | Rem | Shr =>
           // sensitive ops
           ex.map(fixup((ex.args.map(width).foldLeft(0))(math.max)))
         case Dshl =>
--- a/src/test/scala/firrtl/testutils/FirrtlSpec.scala
+++ b/src/test/scala/firrtl/testutils/FirrtlSpec.scala
@@ -122,6 +122,48 @@ trait FirrtlRunners extends BackendCompilationUtilities {
     assert(BackendCompilationUtilities.yosysExpectSuccess(customName, refName, testDir, timesteps))
   }
 
+  /** Check equivalence of Firrtl with reference Verilog
+   *
+    * @note the name of the reference Verilog module is grabbed via regex
+    * @param inputFirrtl string containing Firrtl source
+    * @param referenceVerilog Verilog that will be used as reference for LEC
+    * @param timesteps the maximum number of timesteps to consider
+    */
+  def firrtlEquivalenceWithVerilog(
+    inputFirrtl:       String,
+    referenceVerilog:  String,
+    timesteps:         Int = 1
+  ): Unit = {
+    val VerilogModule = """(?s).*module\s(\w+).*""".r
+    val refName = referenceVerilog match {
+      case VerilogModule(name) => name
+      case _ => throw new Exception(s"Reference Verilog must match simple regex! $VerilogModule")
+    }
+    val circuit = Parser.parse(inputFirrtl.split("\n").toIterator)
+    val inputName = circuit.main
+    require(refName != inputName, s"Name of reference Verilog must not match name of input FIRRTL: $refName")
+
+    val testDir = createTestDirectory(inputName + "_equivalence_test")
+
+    val annos = List(
+      TargetDirAnnotation(testDir.toString),
+      InfoModeAnnotation("ignore"),
+      stage.FirrtlCircuitAnnotation(circuit),
+      stage.RunFirrtlTransformAnnotation.stringToEmitter("verilog"),
+      stage.OutputFileAnnotation(inputName)
+    )
+
+    (new firrtl.stage.FirrtlStage).execute(Array(), annos)
+
+    // Write reference
+    val w = new FileWriter(new File(testDir, s"$refName.v"))
+    w.write(referenceVerilog)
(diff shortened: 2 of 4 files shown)


aswaterman (approved):
This does seem like the right fix, but I share your concern.  To gain confidence that the Verilog is doing what you expect, you could consider exhaustively testing that all of these operators do the Firrtl-expected thing, for all inputs up to 7 bits or so (i.e. 49 combinations of widths), and for all values.  Shouldn't take too long to run.

aswaterman:
And consider linting the static code generated from those test runs.

ekiwi:
You could also use `write_smt2` in yosys to get a SMT formula and use  `-E experimental-smt2` to convert the firrtl directly to SMT.

Then you have two formulas and you can ask the SMT solver if there exists any input for which the output is different.

This will probably only work for small-ish bit sizes since divisions are not well supported by SMT solvers.

jackkoenig:
> You could also use `write_smt2` in yosys to get a SMT formula and use `-E experimental-smt2` to convert the firrtl directly to SMT.
> 
> Then you have two formulas and you can ask the SMT solver if there exists any input for which the output is different.
> 
> This will probably only work for small-ish bit sizes since divisions are not well supported by SMT solvers.

Is that checking anything different than what the Yosys equiv check is already doing? I wouldn't be surprised if Yosys isn't doing that under the hood lol.

jackkoenig:
What appears to be a parsing bug in Yosys 0.9 is causing this failure. LEC passes with Yosys 0.8... There hasn't been a release since 0.9 so I guess we can just bump back or pick a commit on master to use. About half of Yosys's commits have happened since 0.9.

ekiwi:
> Is that checking anything different than what the Yosys equiv check is already doing?

Yes. If you do the check only in yosys you do Verilog vs Verilog.
If you do it with the SMT backend in firrtl you do firrtl vs. Verilog.

jackkoenig:
> Yes. If you do the check only in yosys you do Verilog vs Verilog.
> If you do it with the SMT backend in firrtl you do firrtl vs. Verilog.

Fair, in this case I'm actually trying to test the Verilog vs. Verilog. The FIRRTL vs. Verilog does sound useful for other things though.

ekiwi:
For the `div` example the SMT formula looks goot:

```smt2
; yosys:
(define-fun |m#2| ((state |m_s|)) (_ BitVec 4) (bvudiv (|m#0| state) (concat #b00 (|m#1| state)))) ; $div$m.v:6$1_Y

; experimental-smt2
(define-fun z_f ((state m_s)) (_ BitVec 4) (ite (= (y_f state) (_ bv0 2)) (RANDOM.z_f state) (bvudiv (x_f state) ((_ zero_extend 2) (y_f state)))))

```

The only difference is that the `smt` backend of firrtl returns a arbitrary value if `y` is zero.
Both zero extend `y` to be 4 wide and then to the `BVDiv` which results in a 4 bit value.

jackkoenig:
> What appears to be a parsing bug in Yosys 0.9 is causing this failure. LEC passes with Yosys 0.8... There hasn't been a release since 0.9 so I guess we can just bump back or pick a commit on master to use. About half of Yosys's commits have happened since 0.9.

So, Yosys 0.9 has a parser error, Yosys 0.8 completes the RocketCore.fir check in 3m45s with 1.2 GB of memory, Yosys master takes 12m15s with 2.7 GB memory. I'm going to revert to 0.8 on the Docker image.

jackkoenig:
> This does seem like the right fix, but I share your concern. To gain confidence that the Verilog is doing what you expect, you could consider exhaustively testing that all of these operators do the Firrtl-expected thing, for all inputs up to 7 bits or so (i.e. 49 combinations of widths), and for all values. Shouldn't take too long to run.

Agreed, I'm thinking about how to do this best, maybe I'll just write the test up in Chisel, generate the FIRRTL, then commit it as a test. Merging this for now so it can squeeze into 3.4.2, but I think more explicit testing of operator semantics is in order.

PR #2071, opened 2021-02-03
IR: turn some IR nodes into data classes

ekiwi:
As discussed with @jackkoenig , I am changing Stop, Print and Verification to be a data class.
This is done in preparation for a new `name` field in order to preserve backwards compatibility.

### Contributor Checklist

#### Type of Improvement

- code refactoring

#### API Impact

- it should maintain the current API

- none

#### Desired Merge Strategy

- squash

- [ ] Did you review?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -48,7 +48,7 @@ lazy val firrtlSettings = Seq(
     "net.jcazevedo" %% "moultingyaml" % "0.4.2",
     "org.json4s" %% "json4s-native" % "3.6.9",
     "org.apache.commons" % "commons-text" % "1.8",
-    "io.github.alexarchambault" %% "data-class" % "0.2.1",
+    "io.github.alexarchambault" %% "data-class" % "0.2.5",
   ),
   // macros for the data-class library
   libraryDependencies ++= {
--- a/build.sbt
+++ b/build.sbt
@@ -22,6 +22,11 @@ lazy val commonSettings = Seq(
   crossScalaVersions := Seq("2.13.4", "2.12.13", "2.11.12")
 )
 
+lazy val isAtLeastScala213 = Def.setting {
+  import Ordering.Implicits._
+  CrossVersion.partialVersion(scalaVersion.value).exists(_ >= (2, 13))
+}
+
 lazy val firrtlSettings = Seq(
   name := "firrtl",
   version := "1.5-SNAPSHOT",
@@ -42,8 +47,18 @@ lazy val firrtlSettings = Seq(
     "com.github.scopt" %% "scopt" % "3.7.1",
     "net.jcazevedo" %% "moultingyaml" % "0.4.2",
     "org.json4s" %% "json4s-native" % "3.6.9",
-    "org.apache.commons" % "commons-text" % "1.8"
+    "org.apache.commons" % "commons-text" % "1.8",
+    "io.github.alexarchambault" %% "data-class" % "0.2.1",
   ),
+  // macros for the data-class library
+  libraryDependencies ++= {
+    if (isAtLeastScala213.value) Nil
+    else Seq(compilerPlugin("org.scalamacros" % "paradise" % "2.1.1" cross CrossVersion.full))
+  },
+  scalacOptions ++= {
+    if (isAtLeastScala213.value) Seq("-Ymacro-annotations")
+    else Nil
+  },
   // starting with scala 2.13 the parallel collections are separate from the standard library
   libraryDependencies ++= {
     CrossVersion.partialVersion(scalaVersion.value) match {
--- a/src/main/scala/firrtl/ir/IR.scala
+++ b/src/main/scala/firrtl/ir/IR.scala
@@ -636,7 +636,13 @@ object Stop {
   def foreachType(f:   Type => Unit):   Unit = ()
   def foreachString(f: String => Unit): Unit = ()
   def foreachInfo(f:   Info => Unit):   Unit = f(info)
-  def copy(info: Info = info, string: StringLit = string, args: Seq[Expression] = args, clk: Expression = clk, en: Expression = en): Print = {
+  def copy(
+    info:   Info = info,
+    string: StringLit = string,
+    args:   Seq[Expression] = args,
+    clk:    Expression = clk,
+    en:     Expression = en
+  ): Print = {
     Print(info, string, args, clk, en)
   }
 }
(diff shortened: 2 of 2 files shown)


chick (approved):
This looks pretty straightforward to me. 
The @data provided constructor seems to confuse IntelliJ a bit but things compile and run.
You should probably wait for second review from @jackkoenig before merging

ekiwi:
>  The @DaTa provided constructor seems to confuse IntelliJ a bit but things compile and run.

Yes. That is somewhat annoying.



>  You should probably wait for second review from @jackkoenig before merging

I especially would like @jackkoenig to check if there is a way to simplify the additions to the `build.sbt`.

jackkoenig:
As an experiment, I went ahead and backported this and checked binary compatibility against 1.4.2, here were the errors:
```
firrtl: Failed binary compatibility check against edu.berkeley.cs:firrtl_2.12:1.4.2! Found 12 potential problems
 * class firrtl.ir.Print is declared final in current version
   filter with: ProblemFilters.exclude[FinalClassProblem]("firrtl.ir.Print")
 * static method tupled()scala.Function1 in class firrtl.ir.Print does not have a correspondent in current version
   filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("firrtl.ir.Print.tupled")
 * static method curried()scala.Function1 in class firrtl.ir.Print does not have a correspondent in current version
   filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("firrtl.ir.Print.curried")
 * the type hierarchy of object firrtl.ir.Print is different in current version. Missing types {scala.runtime.AbstractFunction5}
   filter with: ProblemFilters.exclude[MissingTypesProblem]("firrtl.ir.Print$")
< and the same for Verification and Stop >
```
These are all good overall, and it should be binary compatible for adding new fields going forward, but it's a bit of a bummer that we can't easily switch to `@data` on a minor version without breaking binary compatibility.

build.sbt
```diff
@@ -42,8 +47,18 @@ lazy val firrtlSettings = Seq(
     "com.github.scopt" %% "scopt" % "3.7.1",
     "net.jcazevedo" %% "moultingyaml" % "0.4.2",
     "org.json4s" %% "json4s-native" % "3.6.9",
-    "org.apache.commons" % "commons-text" % "1.8"
+    "org.apache.commons" % "commons-text" % "1.8",
+    "io.github.alexarchambault" %% "data-class" % "0.2.1",
```
jackkoenig, line 51:
```suggestion
    "io.github.alexarchambault" %% "data-class" % "0.2.5",
```

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -607,8 +608,16 @@ case class Stop(info: Info, ret: Int, clk: Expression, en: Expression)
   def foreachType(f:   Type => Unit):   Unit = ()
   def foreachString(f: String => Unit): Unit = ()
   def foreachInfo(f:   Info => Unit):   Unit = f(info)
+  def copy(info: Info = info, ret: Int = ret, clk: Expression = clk, en: Expression = en): Stop = {
+    Stop(info, ret, clk, en)
+  }
+}
+object Stop {
+  def unapply(s: Stop): Option[(Info, Int, Expression, Expression)] = {
```
jackkoenig, line 616:
```suggestion
  def unapply(s: Stop): Some[(Info, Int, Expression, Expression)] = {
```

This is not something I knew until just today I happened to see a conversation about it on Gitter. In Scala 2.13.4 you'll get an exhaustivity warning unless you return the more specific type: See https://github.com/scodec/scodec/pull/296 and the linked Gitter conversation.

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -627,6 +636,20 @@ case class Print(
...
+    string: StringLit = string,
+    args:   Seq[Expression] = args,
+    clk:    Expression = clk,
+    en:     Expression = en
+  ): Print = {
+    Print(info, string, args, clk, en)
+  }
+}
+object Print {
+  def unapply(s: Print): Option[(Info, StringLit, Seq[Expression], Expression, Expression)] = {
```
jackkoenig, line 650:
```suggestion
  def unapply(s: Print): Some[(Info, StringLit, Seq[Expression], Expression, Expression)] = {
```

src/main/scala/firrtl/ir/IR.scala
```diff
@@ -657,6 +680,21 @@ case class Verification(
...
+    clk:  Expression = clk,
+    pred: Expression = pred,
+    en:   Expression = en,
+    msg:  StringLit = msg
+  ): Verification = {
+    Verification(op, info, clk, pred, en, msg)
+  }
+}
+object Verification {
+  def unapply(s: Verification): Option[(Formal.Value, Info, Expression, Expression, Expression, StringLit)] = {
```
jackkoenig, line 695:
```suggestion
  def unapply(s: Verification): Some[(Formal.Value, Info, Expression, Expression, Expression, StringLit)] = {
```

jackkoenig (changes requested):
build.sbt changes look good; it's the best way to express macros in 2.13 and 2.12 that I've come across.

I have a few requested changes though.

davidbiancolin:
Is there an issue that provides more context for this? Could it be linked in the PR?

ekiwi:
> Is there an issue that provides more context for this? Could it be linked in the PR?

This PR followed from a discussion we had on the UCB-BAR slack in the `chisel-dev` channel.
The most important advantage of `data` classes vs. `case` classes is that `data` classes do not define `unapply` and `copy` methods which are hard/impossible to keep backwards compatible when adding a new field.

jackkoenig:
> Is there an issue that provides more context for this? Could it be linked in the PR?

I'm not sure that there is so we might as well make it here.

Basically, case classes are terrible for binary compatibility if you need to add another field. Adding a field breaks binary compatibility for the `copy` and `unapply` methods. I believe if you implement the `copy` method yourself, the Scala compiler will *not* generate one for you. I think the same is supposed to occur for `unapply`, but I think that is broken in Scala 2.12 despite release notes for [Scala 2.12.12 saying it should work](https://github.com/scala/scala/releases/tag/v2.12.2), here's proof showing it doesn't work in 2.12.12 https://scastie.scala-lang.org/MlNqn8ClQWalnUuZGuZocw (h/t @ekiwi).

davidbiancolin:
Great, thanks @ekiwi & @jackkoenig. And this is in preparation for #2057, which presumably would make it possible to annotate those statements? (Sorry for the tangent.)

jackkoenig:
> Great, thanks @ekiwi & @jackkoenig. And this is in preparation for #2057, which presumably would make it possible to annotate those statements? (Sorry for the tangent.)

Exactly. I think there's a debate to be had if we should use `@data` for everything, or only opportunistically whenever we want to add fields. This is at least letting Kevin do what he wants to accomplish while we mull over the larger policy decision.

One very unfortunate thing is that Scala 3 does not support macro annotations so we will have to use a different approach once we get there, but since the whole "unapply" generation thing is fixed in Scala 2.13, it's probably fixed in Scala 3 as well so presumably we can just use case classes with careful migration.

jackkoenig (approved):
LGTM

PR #2094, opened 2021-03-01
Allow InferReadWrite to combine shared-address R/W ports when appropriate

albert-magyar:
I would categorize this as an "RFC" to see if anyone is interested in expanded readwrite inference. In cases where a read and write port of a `SyncReadMem` share the same clock and address, the semantics of the default `undefined` read-under-write behavior mean that it is legal to combine the ports into an RW port. The definition of this behavior in the spec implies that this is legal even in cases where there is a partial write mask.

**Type of improvement:** enhancement of optional inference pass
**API impact:** none
**Backend code-generation impact:** when `--infer-rw` is enabled, more RW inference may occur
**Release notes:**
The existing, optional InferReadWrite pass may now infer readwrite ports from read/write port pairs that share the same clock and address for undefined read-under-write synchronous-read memories. Previously recognized cases with mutually exclusive read and write enables will continue to be combined as before.

### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/passes/memlib/InferReadWrite.scala
+++ b/src/main/scala/firrtl/passes/memlib/InferReadWrite.scala
@@ -78,6 +78,16 @@ object InferReadWritePass extends Pass {
       case sx                             => sx
     }
 
+  /* If the ports share the same address in an undefined-collision SyncReadMem, reads issued while the write
+   * is enabled are *always* undefined; we may treat the read as if it were gated by the complement of w.en.
+   * Though not a strict requirement, this currently applies only to single-cycle read/write memories.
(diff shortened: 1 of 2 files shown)


ekiwi:
This is very cool!

I think we can prevent users from shooting themselves in the foot by changing the default `ReadUnderWrite` behavior of `SyncReadMem` in chisel to match the way it is normally emitted: `ReadUnderWrite.Now`.
This way users will have to explicitly enable the undefined behavior if they want to take advantage of the optimization. If that is the case, I would be in favor of enabling your proposed optimization by default.

albert-magyar:
Superseded by #2111

PR #2104, opened 2021-03-04
SMT Backend: model Invalid and Division by Zero with DefRandom nodes

ekiwi:
This finally removes all randomization code from the transition
system conversion and into a separate pass using DefRandom nodes.

### Contributor Checklist

#### Type of Improvement

 - code refactoring

#### API Impact

- new `InvalidToRandomPass` and `InvalidToRandomOptions` annotation in the _experimental_ SMT backend

- none

#### Desired Merge Strategy

- squash

#### Release Notes
 n/a

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/backends/experimental/smt/random/InvalidToRandomPass.scala
+++ b/src/main/scala/firrtl/backends/experimental/smt/random/InvalidToRandomPass.scala
@@ -1,0 +1,125 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package firrtl.backends.experimental.smt.random
+
+import firrtl._
+import firrtl.annotations.NoTargetAnnotation
+import firrtl.ir._
+import firrtl.passes._
+import firrtl.options.Dependency
+import firrtl.stage.Forms
+import firrtl.transforms.RemoveWires
+
+import scala.collection.mutable
+
+/** Chooses how to model explicit and implicit invalid values in the circuit */
+case class InvalidToRandomOptions(
+  randomizeInvalidSignals: Boolean = true,
+  randomizeDivisionByZero: Boolean = true)
+    extends NoTargetAnnotation
+
+/** Replaces all explicit and implicit "invalid" values with random values.
+  * Explicit invalids are:
+  * - signal is invalid
+  * - signal <= valid(..., expr)
+  * Implicit invalids are:
+  * - a / b when eq(b, 0)
+  */
+object InvalidToRandomPass extends Transform with DependencyAPIMigration {
+  override def prerequisites = Forms.LowForm
+  // once ValidIf has been removed, we can no longer detect and randomize them
+  override def optionalPrerequisiteOf = Seq(Dependency(RemoveValidIf))
+  override def invalidates(a: Transform) = a match {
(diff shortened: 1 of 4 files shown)


src/main/scala/firrtl/backends/experimental/smt/random/InvalidToRandomPass.scala
```diff
@@ -0,0 +1,128 @@
...
+      }
+      val prefix = other match {
+        case c: Connect => c.loc.serialize.replace('.', '_')
+        case h: HasName => h.name
+        case _ => ""
+      }
+      val rands = mutable.ListBuffer[Statement]()
+      val stmt = other.mapExpr(onExpr(namespace, opt, prefix, info, rands, _)).mapStmt(onStmt(namespace, opt, _))
+      if (rands.isEmpty) { stmt }
+      else { Block(rands.toList :+ stmt) }
```
jackkoenig, line 80:
```suggestion
      else { 
        rands += stmt
        Block(rands.toList)
      }
```
Better to avoid appending to a list. Alternatively can just further nest, eg.
```scala
      else { Block(Block(rands.toList) :: stmt :: Nil)  }
```
The nested-ness will get flattened the next time `.mapStmt` is called on the `Blocks`.

src/main/scala/firrtl/backends/experimental/smt/random/InvalidToRandomPass.scala
```diff
@@ -0,0 +1,128 @@
...
+    namespace: Namespace,
+    prefix:    String,
+    info:      Info,
+    rands:     mutable.ListBuffer[Statement],
+    cond:      Expression,
+    tpe:       Type,
+    value:     Expression,
+    invert:    Boolean
+  ): Expression = {
+    val name = namespace.newName(if (prefix.isEmpty) "invalid" else prefix + "_invalid")
```
jackkoenig, line 114:
```suggestion
  private def makeRand(
    namespace: Namespace,
    prefix:    String,
    info:      Info,
    rands:     mutable.ListBuffer[Statement]
  )(
    cond:      Expression,
    value:     Expression,
    invert:    Boolean
  ): Expression = {
    val tpe = value.tpe
    val name = namespace.newName(if (prefix.isEmpty) "invalid" else prefix + "_invalid")
```

This is minor (since this is private), but I found the uses of this super hard to read, I think it would help to tweak it to make the values you're actually passing (vs. threading through) stand out, maybe via a separate argument list. Also, the `Type` is unnecessary, we can derive that from `value.tpe`.

jackkoenig (approved):
Generally looks good, a couple of suggestions.

src/main/scala/firrtl/backends/experimental/smt/random/InvalidToRandomPass.scala
ekiwi: I moved the namespace, prefix, info and rands into a `ExprCtx` case class (`private` of course!)

src/main/scala/firrtl/backends/experimental/smt/random/InvalidToRandomPass.scala
ekiwi: Done

PR #2105, opened 2021-03-05
SMT: memory port inout fields cannot be used as RHS expressions

ekiwi:
Turns out I screwed up and the input fields of memory ports are not supposed to be read.
This doesn't actually impact the formal backend which is ok with this.
However, this misunderstanding resulted in a bug, where VerilogMemDelays would change the value of the read enable field which the `UndefinedMemoryBehaviorSpecPass` was relying on.

This is a fix to https://github.com/chipsalliance/firrtl/pull/2095

### Contributor Checklist

#### Type of Improvement

- bug fix

#### API Impact

- none

- fix SMT emission which was broken through an interaction between VeriogMemDelays and the Memory Undefined Behavior Pass

#### Desired Merge Strategy
- squash

#### Release Notes
n/a
- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/backends/experimental/smt/random/UndefinedMemoryBehaviorPass.scala
+++ b/src/main/scala/firrtl/backends/experimental/smt/random/UndefinedMemoryBehaviorPass.scala
@@ -390,8 +390,6 @@ private class InstrumentMems(
           val wire = DefWire(info, copyName(signal), signal.tpe)
           declarations += wire
 
-          // println(s"Making: ${wire.name}")
-
           // connect the old expression to the new wire
           val con = Connect(info, Reference(wire), value)
           newStmts.append(con)
--- a/src/main/scala/firrtl/backends/experimental/smt/random/UndefinedMemoryBehaviorPass.scala
+++ b/src/main/scala/firrtl/backends/experimental/smt/random/UndefinedMemoryBehaviorPass.scala
@@ -2,8 +2,7 @@
 
 package firrtl.backends.experimental.smt.random
 
-import firrtl.Utils.{isLiteral, kind, BoolType}
-import firrtl.WrappedExpression.{we, weq}
+import firrtl.Utils.{isLiteral, BoolType}
 import firrtl._
 import firrtl.annotations.NoTargetAnnotation
 import firrtl.backends.experimental.smt._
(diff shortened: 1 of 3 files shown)


src/test/scala/firrtl/backends/experimental/smt/end2end/MemorySpec.scala
```diff
@@ -195,4 +195,46 @@ class MemorySpec extends EndToEndSMTBaseSpec {
...
+    val init = Seq(MemoryScalarInitAnnotation(CircuitTarget(s"ReadEnableTest1").module(s"ReadEnableTest1").ref("m"), 0))
+    // the read port is enabled on even cycles, so on odd cycles we should reliably get zeros
+    test(readEnableSrc("or(not(odd), eq(m.r.data, UInt(0)))", 1), MCSuccess, kmax = 3, annos = init)
+  }
+
+  "a memory with read enable" should "supply invalid data one cycle after en=0" in {
+    val init = Seq(MemoryScalarInitAnnotation(CircuitTarget(s"ReadEnableTest2").module(s"ReadEnableTest2").ref("m"), 0))
+    // the read port is disabled on odd cycles, so on even cycles we should *NOT* reliably get zeros
+    test(readEnableSrc("or(not(even), eq(m.r.data, UInt(0)))", 2), MCFail(1), kmax = 1, annos = init)
+  }
```
jackkoenig, line 239: Why do you do `not(odd)` instead of `even` on the first check, and vice versa on the second?

src/main/scala/firrtl/backends/experimental/smt/random/UndefinedMemoryBehaviorPass.scala
```diff
@@ -383,27 +371,45 @@ private class InstrumentMems(
...
+          // otherwise we make a wire that refelect the value
+          val wire = DefWire(info, copyName(signal), signal.tpe)
+          declarations += wire
 
-    // if the old value is a literal, we just replace all references to it with this literal
-    val oldValue = connects(signal.serialize)
-    if (isLiteral(oldValue)) {
-      println("TODO: better code for literal")
-    }
+          // println(s"Making: ${wire.name}")
```
jackkoenig, line 393: Delete commented code

jackkoenig (approved):
I don't completely understand but from what I do gather, looks good. A couple of comments.

src/test/scala/firrtl/backends/experimental/smt/end2end/MemorySpec.scala
ekiwi: Because I am encoding an implication:
- What I would like to write: `even -> m.r.data == 0`.
- Which mechanically translates to `not(even) || m.r.data == 0`

src/main/scala/firrtl/backends/experimental/smt/random/UndefinedMemoryBehaviorPass.scala
ekiwi: thanks, good catch!

ekiwi: fixed now

PR #2111, opened 2021-03-11
Add -fpga flag to enable FPGA-oriented compilation strategies (currently for memories)

albert-magyar:
**Type of change:** enhancement
**API impact:** API addition + make an internal method private (can keep public in backport).
**Backend code-generation impact:** More readwrite inference. Also, significant changes to emitted Verilog iff new flags set.
**Desired merge strategy:** merge

**Release notes (WIP):**
`FirrtlStage` now supports new flags to incorporate compilation strategies intended to better integrate with downstream FPGA tools. When the `--target:fpga` flag is used, the Verilog emitted for synchronous-read memories will change significantly. However, the default flow will produce Verilog that is identical to head-of-tree before this PR, aside from the purely syntactic change of using more net-declaration assignments.

The existing, optional `InferReadWrite` pass may now infer readwrite ports from read/write port pairs that share the same clock and address for undefined read-under-write synchronous-read memories. Previously recognized cases with mutually exclusive read and write enables will continue to be combined as before.

**Commentary:**

*(This is a draft; it still needs documentation and more tests.)*

After the discussion on #2092 and #2094, I am opening this to subsume both of those PRs and a few related changes. This PR introduces a number of optional compiler strategies that are all enabled by the `--target:fpga` flag:
1. (From #2092) Allowing some synchronous-read memories and readwrite ports to pass through `VerilogMemDelays` without introducing explicit pipeline registers or splitting ports.
2. (From #1111) Lower aggregate-typed memories with always-high masks to packed memories without splitting.
3. (From firesim/firesim#639) Specify that memories with undefined read-under-write behavior should map to microarchitectures characteristic of "read-first" ports by default. This could be further modified by adding heuristics of when a memory is worth changing, but read-first ports provide a good path to Xilinx BRAMs.
4. Enable `InferReadWrite` (which also receives some enhancements; see below).

Most of these are effectively vendor-neutral, as they simply avoid the elaborate and inference-unfriendly permutation of FIRRTL memories and work around the behavioral memory API in Chisel. While the strong preference for read-first ports in (3) is based largely on experience with Xilinx FPGAs, the extremely strict semantics of write-first FIRRTL memories are also not the simplest or most generic pattern for inference to any underlying macro.

Furthermore, these are combined with some other enhancements. The VerilogEmitter is now capable of emitting simple synchronous-read memories intact; however, this will never happen with the default flow, as `VerilogMemDelays` retains its previous, conservative behavior without the `--target:fpga` flag. Finally, when `InferReadWrite` is enabled, some extra matching-address port pairs (from #2094) will be combined. One open question is whether this enhancement should be tied to the `--target:fpga` flag or made available by default as part of `--infer-rw`.

**Milestone:**
This will backport cleanly to `1.4.x` with a couple of minor changes. Notably, this sets up a deprecation for a future change of `VerilogMemDelays` from a `Pass` to a `Transform` so that it can look at annotations. For now, it uses an override of `execute` and a deprecation warning. Since this flow is generally made up of optional features, the consensus in the meeting was that this would make sense as a `1.4.x` milestone if there is general support to add these features as options.

### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
+++ b/src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
@@ -1065,32 +1065,53 @@ class VerilogEmitter extends SeqTransform with Emitter {
           val decl = if (fullSize > (1 << 29)) "reg /* sparse */" else "reg"
           declareVectorType(decl, sx.name, sx.dataType, sx.depth, sx.info)
           initialize_mem(sx, options)
-          if (sx.readLatency != 0 || sx.writeLatency != 1)
+          // Currently, no idiomatic way to directly emit write-first RW ports
+          val hasComplexRW = (sx.readwriters.nonEmpty &&
+            (sx.readLatency != 1 || sx.readUnderWrite == ReadUnderWrite.New))
+          if (sx.readLatency > 1 || sx.writeLatency != 1 || hasComplexRW)
             throw EmitterException(
-              "All memories should be transformed into " +
-                "blackboxes or combinational by previous passses"
+              Seq(
+                s"Memory ${sx.name} is too complex to emit directly.",
+                "Consider running VerilogMemDelays to simplify complex memories.",
+                "Alternatively, add the --repl-seq-mem flag to replace memories with blackboxes."
+              ).mkString(" ")
             )
+          def createMemWire(name: String, tpe: Type, rhs: InfoExpr): Unit = {
+            declare("wire", name, tpe, MultiInfo(sx.info, rhs.info), rhs.expr)
+          }
+
           for (r <- sx.readers) {
             val data = memPortField(sx, r, "data")
             val addr = memPortField(sx, r, "addr")
-            // Ports should share an always@posedge, so can't have intermediary wire
-
-            declare("wire", LowerTypes.loweredName(data), data.tpe, sx.info)
-            declare("wire", LowerTypes.loweredName(addr), addr.tpe, sx.info)
-            // declare("wire", LowerTypes.loweredName(en), en.tpe)
-
-            //; Read port
-            assign(addr, netlist(addr))
-            // assign(en, netlist(en))     //;Connects value to m.r.en
-            val mem = WRef(sx.name, memType(sx), MemKind, UnknownFlow)
-            val memPort = WSubAccess(mem, addr, sx.dataType, UnknownFlow)
+            val en = memPortField(sx, r, "en")
+            val memPort = WSubAccess(WRef(sx), addr, sx.dataType, UnknownFlow)
             val depthValue = UIntLiteral(sx.depth, IntWidth(sx.depth.bitLength))
             val garbageGuard = DoPrim(Geq, Seq(addr, depthValue), Seq(), UnknownType)
 
-            if ((sx.depth & (sx.depth - 1)) == 0)
-              assign(data, memPort, sx.info)
-            else
-              garbageAssign(data, memPort, garbageGuard, sx.info)
+            val clkSource = netlist(memPortField(sx, r, "clk")).expr
+
+            createMemWire(LowerTypes.loweredName(en), en.tpe, netlist(en))
+
+            if (sx.readLatency == 1 && sx.readUnderWrite != ReadUnderWrite.Old) {
+              val InfoExpr(addrInfo, addrDriver) = netlist(addr)
+              declare("reg", LowerTypes.loweredName(addr), addr.tpe, sx.info)
+              initialize(WRef(LowerTypes.loweredName(addr), addr.tpe), zero, zero)
+              update(addr, addrDriver, clkSource, en, addrInfo)
+            } else {
+              createMemWire(LowerTypes.loweredName(addr), addr.tpe, netlist(addr))
+            }
+
+            if (sx.readLatency == 1 && sx.readUnderWrite == ReadUnderWrite.Old) {
+              declare("reg", LowerTypes.loweredName(data), data.tpe, sx.info)
+              initialize(WRef(LowerTypes.loweredName(data), data.tpe), zero, zero)
+              update(data, memPort, clkSource, en, sx.info)
+            } else {
+              declare("wire", LowerTypes.loweredName(data), data.tpe, sx.info)
+              if ((sx.depth & (sx.depth - 1)) == 0)
+                assign(data, memPort, sx.info)
+              else
+                garbageAssign(data, memPort, garbageGuard, sx.info)
+            }
           }
 
           for (w <- sx.writers) {
@@ -1098,31 +1119,41 @@ class VerilogEmitter extends SeqTransform with Emitter {
             val addr = memPortField(sx, w, "addr")
             val mask = memPortField(sx, w, "mask")
             val en = memPortField(sx, w, "en")
-            //Ports should share an always@posedge, so can't have intermediary wire
-            // TODO should we use the info here for anything?
-            val InfoExpr(_, clk) = netlist(memPortField(sx, w, "clk"))
 
-            declare("wire", LowerTypes.loweredName(data), data.tpe, sx.info)
-            declare("wire", LowerTypes.loweredName(addr), addr.tpe, sx.info)
-            declare("wire", LowerTypes.loweredName(mask), mask.tpe, sx.info)
-            declare("wire", LowerTypes.loweredName(en), en.tpe, sx.info)
+            val clkSource = netlist(memPortField(sx, w, "clk")).expr
 
-            // Write port
-            assign(data, netlist(data))
-            assign(addr, netlist(addr))
-            assign(mask, netlist(mask))
-            assign(en, netlist(en))
+            createMemWire(LowerTypes.loweredName(data), data.tpe, netlist(data))
+            createMemWire(LowerTypes.loweredName(addr), addr.tpe, netlist(addr))
+            createMemWire(LowerTypes.loweredName(mask), mask.tpe, netlist(mask))
+            createMemWire(LowerTypes.loweredName(en), en.tpe, netlist(en))
 
-            val mem = WRef(sx.name, memType(sx), MemKind, UnknownFlow)
-            val memPort = WSubAccess(mem, addr, sx.dataType, UnknownFlow)
-            update(memPort, data, clk, AND(en, mask), sx.info)
+            val memPort = WSubAccess(WRef(sx), addr, sx.dataType, UnknownFlow)
+            update(memPort, data, clkSource, AND(en, mask), sx.info)
           }
 
-          if (sx.readwriters.nonEmpty)
-            throw EmitterException(
-              "All readwrite ports should be transformed into " +
-                "read & write ports by previous passes"
-            )
+          for (rw <- sx.readwriters) {
+            val rdata = memPortField(sx, rw, "rdata")
+            val wdata = memPortField(sx, rw, "wdata")
+            val addr = memPortField(sx, rw, "addr")
+            val en = memPortField(sx, rw, "en")
+            val wmode = memPortField(sx, rw, "wmode")
+            val wmask = memPortField(sx, rw, "wmask")
+            val memPort = WSubAccess(WRef(sx), addr, sx.dataType, UnknownFlow)
+
+            val clkSource = netlist(memPortField(sx, rw, "clk")).expr
+
+            createMemWire(LowerTypes.loweredName(wdata), wdata.tpe, netlist(wdata))
+            createMemWire(LowerTypes.loweredName(addr), addr.tpe, netlist(addr))
+            createMemWire(LowerTypes.loweredName(wmode), wmode.tpe, netlist(wmode))
+            createMemWire(LowerTypes.loweredName(wmask), wmask.tpe, netlist(wmask))
+            createMemWire(LowerTypes.loweredName(en), en.tpe, netlist(en))
+
+            declare("reg", LowerTypes.loweredName(rdata), rdata.tpe, sx.info)
+            initialize(WRef(LowerTypes.loweredName(rdata), rdata.tpe), zero, zero)
+            update(rdata, memPort, clkSource, en, sx.info)
+            update(memPort, wdata, clkSource, AND(en, AND(wmode, wmask)), sx.info)
+          }
+
         case _ =>
       }
     }
--- a/src/main/scala/firrtl/passes/memlib/SeparateWriteClocks.scala
+++ b/src/main/scala/firrtl/passes/memlib/SeparateWriteClocks.scala
@@ -1,0 +1,74 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package firrtl.passes
+package memlib
+
+import firrtl._
+import firrtl.ir._
+import firrtl.passes.LowerTypes
+import firrtl.options.{Dependency, OptionsException}
+
+/**
+  * This transform introduces an intermediate wire on the clock field of each write port of synchronous-read memories
+  * that have *multiple* write/readwrite ports and undefined read-under-write collision behavior. Ultimately, the
+  * introduction of these intermediate wires does not change which clock net clocks each port; therefore, the purpose of
+  * this transform is to help generate Verilog that is more amenable to inference of RAM macros with multiple write
+  * ports in FPGA synthesis flows. This change will cause each write and each readwrite port to be emitted in a separate
+  * clocked procedure, yielding multiple benefits:
+  *
+  * 1) Separate write procedures avoid implicitly constraining cross-port read-write and write-write collision behaviors
+  * 2) The preference for separate clocked procedures for each write port is explicitly specified by Intel and Xilinx
+  *
+  * While this feature is not intended to be vendor-specific, inference of *multiple-write* RAM macros from behavioral
+  * Verilog or VHDL requires both advanced underlying RAM primitives and advanced synthesis tools. Currently, mapping
+  * such memories to programmable devices beyond modern Intel and Xilinx architectures can be prohibitive for users.
+  *
+  * Though the emission of separate processes for write ports could be absorbed into the Verilog emitter, the use of a
+  * pure-FIRRTL transform reduces implementation complexity and enhances reliability.
+  */
+class SeparateWriteClocks extends Transform with DependencyAPIMigration {
+  override def prerequisites = Seq(Dependency(passes.RemoveCHIRRTL), Dependency(passes.ExpandConnects))
+  override def optionalPrerequisites = Seq(Dependency[InferReadWrite])
+  override def optionalPrerequisiteOf = Seq(Dependency[SetDefaultReadUnderWrite])
+  override def invalidates(a: Transform): Boolean = a match {
+    case ResolveFlows => true
+    case _            => false
+  }
+
+  private type ExprMap = collection.mutable.HashMap[WrappedExpression, Reference]
+
+  private def onExpr(replaceExprs: ExprMap)(expr: Expression): Expression = expr match {
+    case wsf: WSubField if (replaceExprs.contains(WrappedExpression(wsf))) =>
+      replaceExprs(WrappedExpression(wsf))
+    case e => e.mapExpr(onExpr(replaceExprs))
+  }
+
+  private def isMultiWriteSyncReadUndefinedRUW(mem: DefMemory): Boolean = {
+    (mem.writers.size + mem.readwriters.size) > 1 &&
+    mem.readLatency == 1 && mem.writeLatency == 1 &&
+    mem.readUnderWrite == ReadUnderWrite.Undefined
+  }
+
+  private def onStmt(replaceExprs: ExprMap, ns: Namespace)(stmt: Statement): Statement = stmt match {
+    case mem: DefMemory if isMultiWriteSyncReadUndefinedRUW(mem) =>
+      val clockRefs = (mem.writers ++ mem.readwriters).map { p => MemPortUtils.memPortField(mem, p, "clk") }
+      val clockWireMap = clockRefs.map { pClk =>
+        WrappedExpression(pClk) -> DefWire(mem.info, ns.newName(LowerTypes.loweredName(pClk)), ClockType)
+      }
+      val clockStmts = clockWireMap.flatMap {
+        case (pClk, clkWire) => Seq(clkWire, Connect(mem.info, pClk.e1, Reference(clkWire)))
+      }
+      replaceExprs ++= clockWireMap.map { case (pClk, clkWire) => pClk -> Reference(clkWire) }
+      Block(mem +: clockStmts)
+    case Connect(i, lhs, rhs)        => Connect(i, onExpr(replaceExprs)(lhs), rhs)
+    case PartialConnect(i, lhs, rhs) => PartialConnect(i, onExpr(replaceExprs)(lhs), rhs)
+    case IsInvalid(i, invalidated)   => IsInvalid(i, onExpr(replaceExprs)(invalidated))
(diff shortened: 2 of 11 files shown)


src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
```diff
@@ -1022,71 +1022,105 @@ class VerilogEmitter extends SeqTransform with Emitter {
           }
           instdeclares += Seq(");")
         case sx: DefMemory =>
+          require(sx.writeLatency == 1)
+          require(sx.readLatency <= 1)
```
jackkoenig, line 1026:
```suggestion
```
I'd delete these since they conflict with the exception thrown below. That or the error message in the exception should be migrated here.

src/main/scala/firrtl/stage/FirrtlCompilerModes.scala
```diff
@@ -0,0 +1,26 @@
...
+import firrtl.options.{ShellOption, HasShellOptions}
+
+object OptimizeForFPGA extends HasShellOptions {
+  private val fpgaAnnos = Seq(
+    InferReadWriteAnnotation,
+    RunFirrtlTransformAnnotation(new InferReadWrite),
+    DefaultReadFirstAnnotation,
+    RunFirrtlTransformAnnotation(new SetDefaultReadUnderWrite),
+    RunFirrtlTransformAnnotation(new SimplifyMems),
+    PassthroughSimpleSyncReadMemsAnnotation
```
ekiwi, line 16: Would this be a good place to explain why these passes are important for FPGA emission?
I feel like this would be the first place that I would look if I wanted to understand what the flag does.
Most people doing that will probably have even less of an understanding of what these passes do than me.

src/test/scala/firrtlTests/InferReadWriteSpec.scala
```diff
@@ -177,4 +177,53 @@ circuit sram6t :
...
+    val input = sameAddr("new")
+    val annos = Seq(memlib.InferReadWriteAnnotation)
+    intercept[Exception] {
+      compileAndEmit(CircuitState(parse(input), ChirrtlForm, annos))
+    } match {
+      case CustomTransformException(_: InferReadWriteCheckException) => // success
+      case _ => fail()
+    }
+  }
+
```
jackkoenig, line 228: What about `old` `readUnderWrite`?

src/main/scala/firrtl/stage/FirrtlCompilerModes.scala
```diff
@@ -0,0 +1,26 @@
...
+  )
+   val options = Seq(
+     new ShellOption[Unit](
+       longOption = "optimize-for-fpga",
+       toAnnotationSeq = a => fpgaAnnos,
+       helpText = "Choose compilation strategies that generally favor FPGA targets",
+       shortOption = Some("fpga"),
+     )
```
jackkoenig, line 24: This is a bit bikesheddy, but I proposed a CLI scoping scheme in https://github.com/chipsalliance/firrtl/pull/1869 that we could maybe follow here. Maybe the long option should be something like `--backend:fpga`?

src/main/scala/firrtl/passes/memlib/SetDefaultReadUnderWrite.scala
```diff
@@ -0,0 +1,44 @@
...
+
+package firrtl.passes
+package memlib
+
+import firrtl._
+import firrtl.ir._
+import firrtl.options.{Dependency, OptionsException}
+import firrtl.annotations.NoTargetAnnotation
+
+trait DefaultReadUnderWriteAnnotation extends NoTargetAnnotation
```
jackkoenig, line 12:
```suggestion
sealed trait DefaultReadUnderWriteAnnotation extends NoTargetAnnotation
```

jackkoenig (commented):
I started reviewing before I realized this was a draft, looks good though!

src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
```diff
@@ -1022,71 +1022,105 @@ class VerilogEmitter extends SeqTransform with Emitter {
...
           })
           val decl = if (fullSize > (1 << 29)) "reg /* sparse */" else "reg"
           declareVectorType(decl, sx.name, sx.dataType, sx.depth, sx.info)
           initialize_mem(sx, options)
-          if (sx.readLatency != 0 || sx.writeLatency != 1)
+          if (sx.readLatency > 1 || sx.writeLatency != 1)
             throw EmitterException(
-              "All memories should be transformed into " +
+              "All complex memories should be transformed into " +
                 "blackboxes or combinational by previous passses"
```
ekiwi, line 1037: Could we add more helpful info here?
Like maybe mention `--repl-seq-mem` and `VerilogMemDelay`

src/main/scala/firrtl/stage/FirrtlCompilerModes.scala
albert-magyar: Yes, I totally agree. I still need to add a great deal of ScalaDoc and comments, along with expanding the help text.

src/main/scala/firrtl/stage/FirrtlCompilerModes.scala
seldridge: I'll bikeshed this... Maybe `--target:<asic,fpga>` or `--circuit:<asic,fpga>` (since ASIC's and FPGAs are both integrated circuits). `--backend` makes me think of "compiler back end" which makes me think "VHDL" or "Verilog". 🤷‍♀️

ekiwi: I like `--target:` since it would also work for `simulation`.
So: `--target:<asic,fpga,sim>`

ekiwi:
@albert-magyar, could you add some HighFirrtl circuits which would synthesize differently with your new fpga target?
If you can deliver those examples, I would try to write a test that runs `yosys` for synthesis and checks the results.
While that does not guarantee that things would work with Vivado, we already have `yosys` on our CI image, so using `yosys` would be the most straight forward way to make an end-to-end check.

albert-magyar:
> @albert-magyar, could you add some HighFirrtl circuits which would synthesize differently with your new fpga target?
> If you can deliver those examples, I would try to write a test that runs `yosys` for synthesis and checks the results.
> While that does not guarantee that things would work with Vivado, we already have `yosys` on our CI image, so using `yosys` would be the most straight forward way to make an end-to-end check.

Here is an example that stresses all of the improvements. With the FPGA flag enabled, this will emit a 1024x32b, 2-RW, synchronous-read memory. You will need to target an FPGA architecture that provides underlying memories that can be configured to support two RW ports, so I would recommend starting with the 7-series architecture if you're going to try with Yosys.

```
circuit syncreadmem:
  module syncreadmem:
    input clk: Clock
    input addr_a: UInt<10>
    input we_a: UInt<1>
    input wdata_a: UInt<8>[4]
    output rdata_a: UInt<8>[4]
    input addr_b: UInt<10>
    input we_b: UInt<1>
    input wdata_b: UInt<8>[4]
    output rdata_b: UInt<8>[4]

    mem m:
      data-type => UInt<8>[4]
      depth => 1024
      reader => r_a
      writer => w_a
      reader => r_b
      writer => w_b
      read-latency => 1
      write-latency => 1
      read-under-write => undefined

    m.r_a.clk <= clk
    m.r_a.addr <= addr_a
    m.r_a.en <= UInt(1)
    rdata_a <= m.r_a.data

    m.w_a.clk <= clk
    m.w_a.addr <= addr_a
    m.w_a.en <= we_a
    m.w_a.mask[0] <= UInt(1)
    m.w_a.mask[1] <= UInt(1)
    m.w_a.mask[2] <= UInt(1)
    m.w_a.mask[3] <= UInt(1)
    m.w_a.data <= wdata_a

    m.r_b.clk <= clk
    m.r_b.addr <= addr_b
    m.r_b.en <= UInt(1)
    rdata_b <= m.r_b.data

    m.w_b.clk <= clk
    m.w_b.addr <= addr_b
    m.w_b.en <= we_b
    m.w_b.mask[0] <= UInt(1)
    m.w_b.mask[1] <= UInt(1)
    m.w_b.mask[2] <= UInt(1)
    m.w_b.mask[3] <= UInt(1)
    m.w_b.data <= wdata_b
```

src/main/scala/firrtl/stage/FirrtlCompilerModes.scala
albert-magyar: I made it `--target:fpga`. Note that this is the name of the flag, not a value. This follows the `--pretty:` flag @jackkoenig added earlier, but it might make less sense in a mutually-exclusive context (like target) than in a setting where the flags within the "scope" are orthogonal (like `--pretty:`). Maybe this is an opportunity for a later upstream improvement.

src/main/scala/firrtl/stage/FirrtlCompilerModes.scala
albert-magyar: Added

src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
albert-magyar: Done

src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
albert-magyar: Done

albert-magyar: Done

src/main/scala/firrtl/passes/memlib/SetDefaultReadUnderWrite.scala
albert-magyar: Done

src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
albert-magyar: Done

src/test/scala/firrtlTests/InferReadWriteSpec.scala
albert-magyar: Added that now, thanks

ekiwi:
Thanks for the example @albert-magyar.
Turns out, yosys might not actually be able to infer combined R/W ports atm: https://github.com/chipsalliance/chisel3/issues/1788#issuecomment-797028021

albert-magyar:
> Thanks for the example @albert-magyar.
> Turns out, yosys might not actually be able to infer combined R/W ports atm: [chipsalliance/chisel3#1788 (comment)](https://github.com/chipsalliance/chisel3/issues/1788#issuecomment-797028021)

Thanks for looking into this regardless. It is always nice to try to make this open-source "CI-able."

carlosedp:
@albert-magyar we found another case where FPGA and ASIC generation differs. Reading memory files in SYNTHESYS block.
Take a look at https://github.com/chipsalliance/firrtl/issues/2114

albert-magyar:
> @albert-magyar we found another case where FPGA and ASIC generation differs. Reading memory files in SYNTHESYS block.
> Take a look at #2114

I agree that `readmemX` bitstream initialization is an important feature to support. Making that more robust might be a good feature to add to `--target:fpga` mode in the future should this PR get accepted -- it's already getting quite large as it is.

ekiwi (approved):
I would be happy to merge this as is. But we should wait for @seldridge or @jackkoenig to also give their OK.

albert-magyar:
One note for discussion/review is that the enhancements to `InferReadWrite` (merging same-address ports on undefined read-under-write memories) take effect whenever `--infer-rw` is active and are not tied to the `--target:fpga` flag. Ultimately, this enhanced RW inference can also be useful for a flow involving `--repl-seq-mem` and can help work around the vagaries of behavioral Chisel memories.

Also, I am leaving this as a draft for now, since there is still testing of various cases that I would like to do with Vivado. I am not sure when I will have time to do that.

albert-magyar:
I updated this with some Scaladoc.

src/main/scala/firrtl/passes/memlib/SeparateWriteClocks.scala
```diff
@@ -0,0 +1,71 @@
...
+  override def optionalPrerequisites = Seq(Dependency[InferReadWrite])
+  override def optionalPrerequisiteOf = Seq(Dependency[SetDefaultReadUnderWrite])
+  override def invalidates(a: Transform): Boolean = a match {
+    case ResolveFlows => true
+    case _            => false
+  }
+
+  private type ExprMap = collection.mutable.HashMap[WrappedExpression, Reference]
+
+  private def onExpr(replaceExprs: ExprMap)(expr: Expression): Expression = expr match {
```
ekiwi, line 40: This only really needs to run on the lhs of a connect, right? Since memory port inputs are not allowed to be used as RHS expressions.

src/main/scala/firrtl/passes/memlib/SeparateWriteClocks.scala
```diff
@@ -0,0 +1,71 @@
...
+  * 2) The preference for separate clocked procedures for each write port is explicitly specified by Intel and Xilinx
+  *
+  * While this feature is not intended to be vendor-specific, inference of *multiple-write* RAM macros from behavioral
+  * Verilog or VHDL requires both advanced underlying RAM primitives and advanced synthesis tools. Currently, mapping
+  * such memories to programmable devices beyond modern Intel and Xilinx architectures can be prohibitive for users.
+  *
+  * Though the emission of separate processes for write ports could be absorbed into the Verilog emitter, the use of a
+  * pure-FIRRTL transform reduces implementation complexity and enhances reliability.
+  */
+class SeparateWriteClocks extends Transform with DependencyAPIMigration {
```
ekiwi, line 29: Can you add one simple unittest, that will fail without this pass and succeed once you add this pass?

ekiwi (changes requested):

src/main/scala/firrtl/passes/memlib/SeparateWriteClocks.scala
albert-magyar: Or invalidates or partial connects, but yes, I will change it to be more efficient.

albert-magyar: Done

src/main/scala/firrtl/passes/memlib/SeparateWriteClocks.scala
albert-magyar: Added

albert-magyar:
Thanks for taking a look, @ekiwi. I will join the Monday Chisel-Dev meeting in case anyone wants to discuss or has any thoughts

ekiwi (approved):
:shipit:

ekiwi:
@albert-magyar Did you want to backport this to `1.4`?

albert-magyar:
Thanks all for the discussion today, especially @ekiwi for taking a look.

> @albert-magyar Did you want to backport this to `1.4`?

In general, yes.

There are some open questions for this PR:

1. For the sync-read mems affected by this PR, there are options for read enable style for disabled read ports and read handling of read-under-write ports. I have chosen the following style for an example (read-first RUW, sync-read) memory `mymem` with read port `rp` and readwrite port `rwp`:
```
always @(posedge read_clk) begin
  if (mymem_rp_en)
   mymem_rp_data <= mymem[mymem_rp_addr];
  if (mymem_rwp_en)
    mymem_rwp_rdata <= mymem[mymem_rwp_addr];
  if (mymem_rwp_en && (mymem_rwp_wmode && mymem_rwp_wmask))
    mymem[mymem_rwp_addr] <= mymem_rwp_wdata;
end
```
  * Notably, this turns off read- and port-enables when appropriate but leaves the old data on the output. It also gives the undefined rdata field of the readwrite port the concrete value of the data that would have been read if `wmode` were low. These styles are highly aligned with both Intel and Xilinx guides for best inference, yet they do not require substantive changes to how the emitter emits memories outside of the "FPGA backend" context. A reasonable alternative would be to elide the read-/port-enables for simplicity. Changing the RW-port emission style to "block" the read data from showing up can hurt BRAM inference under Vivado for true dual-port mems. 

2. How far should we backport the `InferReadWrite` improvement (combining shared-address ports for undefined-read-under-write memories)?

3. Does anyone have a pre-existing flow / spare time to test this with Quartus?

schoeberl:
I tried to test it with Intel Quartus, using the following Chisel code:

```
  val mem = SyncReadMem(1024, UInt(8.W))

  io.rdDataA := mem.read(io.addrA)
  when(io.wrEnaA) {
    mem.write(io.addrA, io.wrDataA)
  }
  io.rdDataB := mem.read(io.addrB)
  when(io.wrEnaB) {
    mem.write(io.addrA, io.wrDataB)
  }
```

I am using x.5-SNAPSHOT

Without the "--target:fpga" it generates registers, which we expect. With the option enabled it generates invalid Verilog code:

```
  always @(posedge clock) begin
    if (mem_MPORT_2_en) begin
      mem_MPORT_2_data <= mem[mem_MPORT_2_addr]; // @[Memory.scala 64:24]
    end
  end
  always @(posedge mem_rw_clk) begin
    if (mem_rw_en) begin
      mem_rw_rdata <= mem[mem_rw_addr]; // @[Memory.scala 64:24]
    end
    if (mem_rw_en & (mem_rw_wmode & mem_rw_wmask)) begin
      mem[mem_rw_addr] <= mem_rw_wdata; // @[Memory.scala 64:24]
    end
  end
  always @(posedge mem_rw_0_clk) begin
    if (mem_rw_0_en) begin
      mem_rw_0_rdata <= mem[mem_rw_0_addr]; // @[Memory.scala 64:24]
    end
    if (mem_rw_0_en & (mem_rw_0_wmode & mem_rw_0_wmask)) begin
      mem[mem_rw_0_addr] <= mem_rw_0_wdata; // @[Memory.scala 64:24]
    end
  end
```

Quartus complains:

```
Info (12127): Elaborating entity "TrueDualPortMemory" for the top level hierarchy
Warning (10036): Verilog HDL or VHDL warning at TrueDualPortMemory.v(30): object "mem_rw_rdata" assigned a value but never read
Error (10028): Can't resolve multiple constant drivers for net "mem[0][7]" at TrueDualPortMemory.v(68)
Error (10029): Constant driver at TrueDualPortMemory.v(60)
Error (10028): Can't resolve multiple constant drivers for net "mem[0][6]" at TrueDualPortMemory.v(68)
Error (10028): Can't resolve multiple constant drivers for net "mem[0][5]" at TrueDualPortMemory.v(68)
Error (10028): Can't resolve multiple constant drivers for net "mem[0][4]" at TrueDualPortMemory.v(68)
```

schoeberl:
OK, I will try. BTW, I am wondering why the generated Verilog code has actually 3 read ports and from 3 possible different clocks. One of them, ```mem_rw_data```, is then not used, as we can see in the warning.

schoeberl:
Update: the Verilog error resulted from an error in my Chisel code, as followed:
```
  io.rdDataB := mem.read(io.addrB)
  when(io.wrEnaB) {
    mem.write(io.addrA, io.wrDataB)
  }
```
having the wrong address for the second write port. After fixing it, Quartus compiles but does not infer an on-chip memory, but registers. Will try to propose working Verilog code.

schoeberl:
Quartus needs Verilog that requests newly written data to be returned on the reading. So following code works:
```
  always @(posedge mem_rw_clk) begin
    if (mem_rw_en) begin
      mem_rw_rdata <= mem[mem_rw_addr]; // @[Memory.scala 65:24]
    end
    if (mem_rw_en & (mem_rw_wmode & mem_rw_wmask)) begin
      mem[mem_rw_addr] <= mem_rw_wdata; // @[Memory.scala 65:24]
      mem_rw_rdata <= mem_rw_wdata; // @[Memory.scala 65:24]
    end
  end
  always @(posedge mem_rw_0_clk) begin
    if (mem_rw_0_en) begin
      mem_rw_0_rdata <= mem[mem_rw_0_addr]; // @[Memory.scala 65:24]
    end
    if (mem_rw_0_en & (mem_rw_0_wmode & mem_rw_0_wmask)) begin
      mem[mem_rw_0_addr] <= mem_rw_0_wdata; // @[Memory.scala 65:24]
      mem_rw_0_rdata <= mem_rw_0_wdata; // @[Memory.scala 65:24]
    end
  end
```
The only differences are the following lines added:
```
      mem_rw_rdata <= mem_rw_wdata; // @[Memory.scala 65:24]
```
and
```
      mem_rw_0_rdata <= mem_rw_0_wdata; // @[Memory.scala 65:24]
```

However, a better style would be to have the read on an else branch:
```
	always @ (posedge clk)
	begin // Port a
		if (we_a)
		begin
			ram[addr_a] <= data_a;
			q_a <= data_a;
		end
		else
			q_a <= ram[addr_a];
	end
```

BTW, I am wondering what signal ```mem_rw_en``` and ```mem_rw_0_en``` should be good for. They are tied to a constant, so no problem. However, if we would use an enable, what would be the result of a read? Just for fun, I connected one of them to an input and Quartus complaint with a simple:
```
Error (276001): Cannot synthesize dual-port RAM logic "mem"
```

ekiwi:
> BTW, I am wondering what signal `mem_rw_en` and `mem_rw_0_en` should be good for. They are tied to a constant, so no problem. However, if we would use an enable, what would be the result of a read

According to firrtl semantics, when the `en` on a read port is low, then the `data` on that port in the current cycle is arbitrary. The idea here is to support some kinds of ASIC SRAMs that allow you to save power by using a read enable pin. For FPGAs we probably just want to ignore the read enable (which is a valid implementation of the firrtl semantics).

kammoh:
At least on Xilinx FPGAs the read enable (as currently supported by firrtl) can be inferred for clock gating BRAMs. I'm also using it as an implementation trick to keep the read data value unchanged (which probably doesn't quite match the firrtl semantics but nonetheless the generated Verilog works fine with Vivado).
I really hope we can keep the current behavior with `-fpga`!

ekiwi:
I wish we had a good way to automatically test for regressions in how Quartus/Vivado synthesize our memories.

schoeberl:
Some behavior of hardware can simply not be expressed in Verilog or VHDL. One example is undefined behavior on read during write being undefined for the same address. Bad, but that is what  HW and HW languages are. Keeping read data unchanged when disabling read enable is, I guess, not a strong property of the language and not so easily achieved in HW. What is the implementation in an ASIC RAM? Using read enable to enable a register update for the read address?

I think we would need to reread the Verilog and VHDL spec for this corner case. For other logic when an update depends on an enable signal this would generate latches. Something we would like to avoid.

ekiwi:
> What is the implementation in an ASIC RAM? Using read enable to enable a register update for the read address?

I have no experience with ASIC tapeouts.  I did find this manual for a suite of memory compilers: https://users.ece.cmu.edu/~koopman/ece548/hw/hw5/meml80.pdf On page 17 it describes the interface for a `Single-Port Synchronous RAM Generator`. The generated cell features a `OEN` pin:

> OEN is used to enable/disable the data output driver.

So if we use this cell from a Chisel design, we could wire the `.en` pin of the read port to the `OEN` pin.

ekiwi:
We could also turn off the memory in cycles where both, the read and the write enable are not active:

Screenshot from 2021-05-12 13-49-38

ekiwi:
> . I'm also using it as an implementation trick to keep the read data value unchanged (which probably doesn't quite match the firrtl semantics but nonetheless the generated Verilog works fine with Vivado).

The behavior of retaining the last read value is definitely not guaranteed by the Chisel semantics.

PR #2112, opened 2021-03-11
Fix CSESubAccesses for SubAccesses with flips

jackkoenig:
Fixes a bug in https://github.com/chipsalliance/firrtl/pull/2099 that is causing Chisel CI to fail.

The flow of a LHS SubAccess node may still be SourceFlow if the type of
the Vec element has a flip. Tweak the logic of CSESubAccesses to check
every Expression flow while recursing instead of just the flow of the
final SubAccess.

This did highlight a shortcoming of `CSESubAccesses`. Basically, the transform uses nodes to avoid quadratic behavior of `RemoveAccesses` because nodes are unaffected by `ExpandConnects` and thus it only lowers the `SubAccess` once. See https://github.com/chipsalliance/firrtl/issues/643 for more details on that issue. The shortcoming is that nodes cannot be used to represent bidirectional connections, so `CSESubAccesses` fundamentally cannot handle dynamic indexing of bidirectional Vecs. I did try, but it doesn't really work with the current approach.

I think a better approach is to fix the quadratic behavior in `RemoveAccesses` itself, then have `CSESubAccesses` use wires instead of nodes so that it can also handle bidirectional Vecs. Future work!

### Contributor Checklist

No need for release notes, CSESubAccesses is not yet released.

#### Type of Improvement

 - bug fix   

#### API Impact

No impact

No impact

#### Desired Merge Strategy

 - Squash

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/transforms/CSESubAccesses.scala
+++ b/src/main/scala/firrtl/transforms/CSESubAccesses.scala
@@ -21,7 +21,9 @@ object CSESubAccesses {
     val acc = new mutable.ListBuffer[(SubAccess, Info)]
     def onExpr(outer: Statement)(expr: Expression): Unit = {
       // Need postorder because we want to visit inner SubAccesses first
-      expr.foreach(onExpr(outer))
+      // Stop recursing on any non-Source because flips can make the SubAccess a Source despite the
+      //   overall Expression being a Sink
+      if (flow(expr) == SourceFlow) expr.foreach(onExpr(outer))
       expr match {
         case e: SubAccess if flow(e) == SourceFlow => acc += e -> get_info(outer)
         case _ => // Do nothing
@@ -42,6 +44,8 @@ object CSESubAccesses {
 
   // Replaces all right-hand side SubAccesses with References
   private def replaceOnSourceExpr(replace: SubAccess => Reference)(expr: Expression): Expression = expr match {
+    // Stop is we ever see a non-SourceFlow
+    case e if flow(e) != SourceFlow => e
     // Don't traverse children of SubAccess, just replace it
     // Nested SubAccesses are handled during creation of the nodes that the references refer to
     case acc: SubAccess if flow(acc) == SourceFlow => replace(acc)
--- a/src/test/scala/firrtlTests/transforms/CSESubAccessesSpec.scala
+++ b/src/test/scala/firrtlTests/transforms/CSESubAccessesSpec.scala
@@ -184,4 +184,40 @@ class CSESubAccessesSpec extends FirrtlFlatSpec {
     compile(input) should be(parse(expected).serialize)
   }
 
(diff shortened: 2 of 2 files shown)


src/main/scala/firrtl/transforms/CSESubAccesses.scala
```diff
@@ -42,6 +44,7 @@ object CSESubAccesses {
 
   // Replaces all right-hand side SubAccesses with References
   private def replaceOnSourceExpr(replace: SubAccess => Reference)(expr: Expression): Expression = expr match {
+    case e if flow(e) != SourceFlow => e // Stop is we ever see a non-SourceFlow
```
seldridge, line 47: nit (next pattern has different comment structure):
```suggestion
    // Stop is we ever see a non-SourceFlow
    case e if flow(e) != SourceFlow => e 
```

src/main/scala/firrtl/transforms/CSESubAccesses.scala
```diff
@@ -21,7 +21,9 @@ object CSESubAccesses {
     val acc = new mutable.ListBuffer[(SubAccess, Info)]
     def onExpr(outer: Statement)(expr: Expression): Unit = {
       // Need postorder because we want to visit inner SubAccesses first
-      expr.foreach(onExpr(outer))
+      // Stop recursing on any non-Source because flips can make the SubAccess a Source despite the
+      //   overall Expression being a Sink
+      if (flow(expr) == SourceFlow) expr.foreach(onExpr(outer))
```
seldridge, line 26: To check my understanding... the implication of this is that we won't CSE _all_ subaccesses, and punt on ones which have weird internal flippedness? 

I think this is totally fair.

seldridge (approved):
Awesome. This is a good, minimal solution to a very tricky problem.

src/main/scala/firrtl/transforms/CSESubAccesses.scala
jackkoenig: This comment is really talking about how an Expression that is a Sink may have a SubAccess that is a SourceFlow, eg. from the added tests:
```
input out : { flip foo : UInt<8> }[2]
...
out[idx].foo <= in.foo
```
`out[idx].foo` is a sink, but the `out[idx]` SubAccess is SourceFlow, the `flip` on `foo` is what makes the larger Expression a sink.

Note that this case probably should not ever be handled by this transform, or at least, it would require a lot more sophistication to handle write ports

That being said, this check also stops handling any SubAccesses of bidirectional Vecs. Technically, this transform *could* handle them if running after `ExpandConnects`, but it's a little more complicated and I didn't feel like dealing with it, from the other test I added in this PR:
```
output out : { flip foo : UInt<8>, bar : UInt<8> }[4]
...
out[idx] <= in"
```

PR #2130, opened 2021-03-17
Ensure InlineCasts does not inline complex Expressions

jackkoenig:
Previously, InlineCasts could inline complex (ie. non-cast) Expressions
into other complex Expressions. Now it will only inline so long as there
no more than 1 complex Expression in the current nested Expression.

This fixes the bug that https://github.com/chipsalliance/firrtl/pull/2128 is hitting in CI (the change in #2128 exposes this bug).

These tests probably aren't backportable but the bugfix should be, this is a functional bug.
First test is the functional bug, second is because my original fix caused some Verilog differences, namely the clock in that test would be named `_T_2` which is obviously not what we want. I think code gen would benefit from some "better name" handling that works across casts. Regardless, this PR is a minor bugfix so I wanted it to not cause Verilog diffs if possible, and it doesn't.

The diff looks small but most of the diff is because I added an inner method to `onExpr` to propagate `seenComplexExpr`. I didn't really want to expose that on the public API.

### Contributor Checklist

#### Type of Improvement

- bug fix    

#### API Impact

No impact

No impact

#### Desired Merge Strategy

 - Squash

#### Release Notes

Fix bug where expression inlining could emit Verilog that doesn't match the input FIRRTL's semantics

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/transforms/InlineCasts.scala
+++ b/src/main/scala/firrtl/transforms/InlineCasts.scala
@@ -28,23 +28,30 @@ object InlineCastsTransform {
     * @param expr the Expression being transformed
     * @return Returns expr with [[WRef]]s replaced by values found in replace
     */
-  def onExpr(replace: NodeMap)(expr: Expression): Expression = expr match {
-    // Anything that may generate a part-select should not be inlined!
-    case DoPrim(op, _, _, _) if (isBitExtract(op) || op == Pad) => expr
-    case e =>
-      e.map(onExpr(replace)) match {
-        case e @ WRef(name, _, _, _) =>
-          replace
-            .get(name)
-            .filter(isSimpleCast(castSeen = false))
-            .getOrElse(e)
-        case e @ DoPrim(op, Seq(WRef(name, _, _, _)), _, _) if isCast(op) =>
-          replace
-            .get(name)
-            .map(value => e.copy(args = Seq(value)))
-            .getOrElse(e)
-        case other => other // Not a candidate
-      }
+  def onExpr(replace: NodeMap)(expr: Expression): Expression = {
+    // Keep track if we've seen any non-cast expressions while recursing
+    def rec(hasNonCastParent: Boolean)(expr: Expression): Expression = expr match {
+      // Skip pads to avoid inlining literals into pads which results in invalid Verilog
+      case DoPrim(op, _, _, _) if (isBitExtract(op) || op == Pad) => expr
+      case e =>
+        e.map(rec(hasNonCastParent || !isCast(e))) match {
+          case e @ WRef(name, _, _, _) =>
+            replace
+              .get(name)
+              .filter(isSimpleCast(castSeen = false))
+              .getOrElse(e)
+          case e @ DoPrim(op, Seq(WRef(name, _, _, _)), _, _) if isCast(op) =>
+            replace
+              .get(name)
+              // Only inline the Expression if there is no non-cast parent in the expression tree OR
+              // if the subtree contains only casts and references.
+              .filter(x => !hasNonCastParent || isSimpleCast(castSeen = true)(x))
+              .map(value => e.copy(args = Seq(value)))
+              .getOrElse(e)
+          case other => other // Not a candidate
+        }
+    }
+    rec(false)(expr)
   }
 
   /** Inline casts in a Statement
--- a/src/test/scala/firrtl/testutils/FirrtlSpec.scala
+++ b/src/test/scala/firrtl/testutils/FirrtlSpec.scala
@@ -165,10 +165,14 @@ trait FirrtlRunners extends BackendCompilationUtilities {
 
   /** Compiles input Firrtl to Verilog */
   def compileToVerilog(input: String, annotations: AnnotationSeq = Seq.empty): String = {
+    compileToVerilogCircuitState(input, annotations).getEmittedCircuit.value
+  }
+
+  /** Compiles input Firrtl to Verilog */
+  def compileToVerilogCircuitState(input: String, annotations: AnnotationSeq = Seq.empty): CircuitState = {
     val circuit = Parser.parse(input.split("\n").toIterator)
     val compiler = new VerilogCompiler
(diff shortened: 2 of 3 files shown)


src/main/scala/firrtl/transforms/InlineCasts.scala
```diff
@@ -28,23 +28,30 @@ object InlineCastsTransform {
...
-        case e @ DoPrim(op, Seq(WRef(name, _, _, _)), _, _) if isCast(op) =>
-          replace
-            .get(name)
-            .map(value => e.copy(args = Seq(value)))
-            .getOrElse(e)
-        case other => other // Not a candidate
-      }
+  def onExpr(replace: NodeMap)(expr: Expression): Expression = {
+    // Keep track if we've seen any non-cast expressions while recursing
+    def rec(seenComplexExpr: Boolean)(expr: Expression): Expression = expr match {
```
albert-magyar, line 33:
```suggestion
    def rec(hasNonCastParent: Boolean)(expr: Expression): Expression = expr match {
```

Directly referencing what this means for the structure of the tree might make it clearer.

src/main/scala/firrtl/transforms/InlineCasts.scala
```diff
@@ -28,23 +28,30 @@ object InlineCastsTransform {
...
+        e.map(rec(seenComplex)) match {
+          case e @ WRef(name, _, _, _) =>
+            replace
+              .get(name)
+              .filter(isSimpleCast(castSeen = false))
+              .getOrElse(e)
+          case e @ DoPrim(op, Seq(WRef(name, _, _, _)), _, _) if isCast(op) =>
+            replace
+              .get(name)
+              // Only inline the Expression if we are not complex or if it is not complex
```
albert-magyar, line 47:
```suggestion
              // Only inline the Expression if there is no non-cast parent in the expression tree OR
              // if the subtree contains only casts and references.
```

src/main/scala/firrtl/transforms/InlineCasts.scala
albert-magyar: Keeping the "complex" is fine, I just think that "has<X>Parent" is clearer than "seen."

src/main/scala/firrtl/transforms/InlineCasts.scala
```diff
@@ -28,23 +28,30 @@ object InlineCastsTransform {
...
-            .get(name)
-            .map(value => e.copy(args = Seq(value)))
-            .getOrElse(e)
-        case other => other // Not a candidate
-      }
+  def onExpr(replace: NodeMap)(expr: Expression): Expression = {
+    // Keep track if we've seen any non-cast expressions while recursing
+    def rec(seenComplexExpr: Boolean)(expr: Expression): Expression = expr match {
+      // Anything that may generate a part-select should not be inlined!
+      case DoPrim(op, _, _, _) if (isBitExtract(op) || op == Pad) => expr
```
albert-magyar, line 35: Does this restriction actually serve any purpose? I imagine it might have just prevented a particularly common degenerate case of the bug this PR fixes.

albert-magyar (commented):
I think some Scaladoc/renaming is definitely higher priority than normal here; I wouldn't have imagined that `InlineCasts` inlined things other than casts.

Ultimately, this transform is a lot broader than inlining casts. It takes advantage of the fact that casts are no-ops in the Verilog emitter since all `SIntType` expressions get wrapped in `$signed`, and perhaps a better way to describe it would be stating that it greedily attempts to build larger expression trees that contain at most one expression AST node that is neither a cast nor a reference-like node. I know that renaming is a pain, so we don't necessarily have to do it soon, but I think `InlineAcrossCasts` might make it clearer that the transform inlines things other than casts.

src/main/scala/firrtl/transforms/InlineCasts.scala
jackkoenig: Originally I thought this was a degenerate case but then reasoned myself into thinking it wasn't. Now I agree with you again so I'm going to try deleting this.

src/main/scala/firrtl/transforms/InlineCasts.scala
jackkoenig: `hasNonCastParent` is better I think, will change.

jackkoenig:
> I think some Scaladoc/renaming is definitely higher priority than normal here; I wouldn't have imagined that `InlineCasts` inlined things other than casts.
> 
> Ultimately, this transform is a lot broader than inlining casts. It takes advantage of the fact that casts are no-ops in the Verilog emitter since all `SIntType` expressions get wrapped in `$signed`, and perhaps a better way to describe it would be stating that it greedily attempts to build larger expression trees that contain at most one expression AST node that is neither a cast nor a reference-like node. I know that renaming is a pain, so we don't necessarily have to do it soon, but I think `InlineAcrossCasts` might make it clearer that the transform inlines things other than casts.

I agree with all of this. I'll deprecate and rename in a follow on PR because I want this to backport to 1.2.x.

src/main/scala/firrtl/transforms/InlineCasts.scala
jackkoenig: Alright so this is still required because basically we cannot emit a pad of a literal. This is why we have `legalizePad` which doesn't support cast literals at the moment: https://github.com/chipsalliance/firrtl/blob/94d1bee4c23bd3d8f99dae3ca431ffaa5dc1410d/src/main/scala/firrtl/passes/Legalize.scala#L52

I think the solution is to remove this special case but add `legalizeCast` to ensure that cast literals are always converted.

albert-magyar: Maybe update the comment to say that you skip pads to avoid inlining literals into pads.

jackkoenig: Leaving the suggestion above to future work.

albert-magyar (approved):
This is a strict improvement over the existing pass so LGTM

PR #2157, opened 2021-03-26
Fix RemoveAccesses, delete CSESubAccesses

jackkoenig:
CSESubAccesses was intended to be a simple workaround for a quadratic
performance bug in RemoveAccesses but ended up having tricky corner
cases and was hard to get right. The solution to the RemoveAccesses
bug--quadratic expansion of dynamic indexes of vecs of aggreate
type--turned out to be quite simple and makes CSESubAccesses much less
useful and not worth fixing.

Note that we didn't really have extensive tests of RemoveAccesses so I just sort of converted the CSESubAccesses tests over to a new RemoveAccessesSpec.

### Contributor Checklist

#### Type of Improvement

  - bug fix (by deleting CSESubAccesses)
  - performance improvement (handles cases not handled by CSESubAccesses)

#### API Impact

No impact (CSESubAccesses was never released)

This will likely affect some temporary numbering but should have minimal impact over all.

#### Desired Merge Strategy

  - Squash

#### Release Notes

Fix performance bug in RemoveAccesses

- [ ] Did you review?


src/main/scala/firrtl/passes/RemoveAccesses.scala
```diff
@@ -122,26 +120,26 @@ object RemoveAccesses extends Pass {
...
+              stmts += wire
+              rs.zipWithIndex.foreach {
+                case (x, i) if i < temps.size =>
+                  stmts += IsInvalid(get_info(s), getTemp(i))
+                  stmts += Conditionally(get_info(s), x.guard, Connect(get_info(s), getTemp(i), x.base), EmptyStmt)
+                case (x, i) =>
+                  stmts += Conditionally(get_info(s), x.guard, Connect(get_info(s), getTemp(i), x.base), EmptyStmt)
+              }
+              temp
+          }
```
jackkoenig, line 142: Note this diff is just removing the `match` since we only ever call this on `RefLikeExpressions`, the match is redundant.

src/main/scala/firrtl/passes/RemoveAccesses.scala
```diff
@@ -162,14 +160,23 @@ object RemoveAccesses extends Pass {
...
+          * If we see a RefLikeExpression that contains a SubAccess, we recursively remove
+          * subaccesses from the indices of any SubAccesses, then process modified RefLikeExpression
           */
         def fixSource(e: Expression): Expression = e match {
-          case w: WSubAccess => removeSource(WSubAccess(w.expr, fixSource(w.index), w.tpe, w.flow))
-          //case w: WSubIndex => removeSource(w)
-          //case w: WSubField => removeSource(w)
+          case ref: RefLikeExpression =>
+            if (hasAccess(ref)) removeSource(fixIndices(ref)) else ref
           case x => x.map(fixSource)
```
jackkoenig, line 180: The diff above here is the real meat of the PR, by recursing until you see the `SubAccess` for RHS subaccesses ("Sources"), you end up always expanding the entire aggregate. This changes it to match on the outermost `RefLikeExpression` it can to ensure that we respect any `SubFields` or `SubIndexes` when doing the expansion. This is the fix.

chick (approved):
lgtm

azidar (approved):

PR #2166, opened 2021-03-31
Add memory initialization options for synthesis

### LINKED ISSUE #2114: readmem block is not picked up by synthesis tools
### Checklist

- [ ] Did you describe your environment?
- [ ] Did you specify relevant external information?

In the recently implemented inline readmem annotation, in some cases synthesis tools do not pick-up the readmem statement when placed inside the `ifndef SYNTHESYS` like it currently is.

```verilog
...
`ifndef SYNTHESIS
`ifdef FIRRTL_BEFORE_INITIAL
`FIRRTL_BEFORE_INITIAL
`endif
initial begin
  `ifdef RANDOMIZE
    `ifdef INIT_RANDOM
      `INIT_RANDOM
    `endif
    `ifndef VERILATOR
      `ifdef RANDOMIZE_DELAY
        #`RANDOMIZE_DELAY begin end
      `else
        #0.002 begin end
      `endif
    `endif
`ifdef RANDOMIZE_REG_INIT
  _RAND_0 = {1{`RANDOM}};
  ram_MPORT_addr_pipe_0 = _RAND_0[10:0];
  _RAND_1 = {1{`RANDOM}};
  ram_MPORT_1_addr_pipe_0 = _RAND_1[10:0];
`endif // RANDOMIZE_REG_INIT
  `endif // RANDOMIZE
  $readmemh("insns.hex", ram);
end // initial
`ifdef FIRRTL_AFTER_INITIAL
`FIRRTL_AFTER_INITIAL
`endif
`endif // SYNTHESIS
endmodule
```

Verilator for example works correctly but Yosys does not.

The following Verilog, works correctly with the readmem outside the `ifndef`. A new `initial begin ... end` pair must be used.

```verilog
...
`ifndef SYNTHESIS
`ifdef FIRRTL_BEFORE_INITIAL
`FIRRTL_BEFORE_INITIAL
`endif
initial begin
  `ifdef RANDOMIZE
    `ifdef INIT_RANDOM
      `INIT_RANDOM
    `endif
    `ifndef VERILATOR
      `ifdef RANDOMIZE_DELAY
        #`RANDOMIZE_DELAY begin end
      `else
        #0.002 begin end
      `endif
    `endif
`ifdef RANDOMIZE_REG_INIT
  _RAND_0 = {1{`RANDOM}};
  ram_MPORT_addr_pipe_0 = _RAND_0[10:0];
  _RAND_1 = {1{`RANDOM}};
  ram_MPORT_1_addr_pipe_0 = _RAND_1[10:0];
`endif // RANDOMIZE_REG_INIT
  `endif // RANDOMIZE
end // initial
`ifdef FIRRTL_AFTER_INITIAL
`FIRRTL_AFTER_INITIAL
`endif
`endif // SYNTHESIS
initial begin
  $readmemh("insns.hex", ram);
end
endmodule
```

### Steps to Reproduce

Create a simple memory module with the load annotation.
Check that the synthesis tool does not load the memory from file.

I'm already looking into providing a fix.

carlosedp:
This PR adds options for memory initialization inside or outside the
`ifndef SYNTHESIS` block.

Fixes #2114.

### Contributor Checklist

#### Type of Improvement

- new feature/API

#### API Impact

This new feature adds memory initialization options for having the `readmem` blocks inside or outside the existing `ifndef SYNTHESIS`.

In the future, this parameter can be used by the `-target:fpga` PR (https://github.com/chipsalliance/firrtl/pull/2111) that is in development.

This changes the generated Verilog code based on the parameters:

If using `MemoryNoSynthInit` (current default):

```verilog
...
initial begin
  `ifdef RANDOMIZE
    `ifdef INIT_RANDOM
      `INIT_RANDOM
    `endif
    `ifndef VERILATOR
      `ifdef RANDOMIZE_DELAY
        #`RANDOMIZE_DELAY begin end
      `else
        #0.002 begin end
      `endif
    `endif
  `endif // RANDOMIZE
  $readmemh("sample.hex", mem);
end // initial
`ifdef FIRRTL_AFTER_INITIAL
`FIRRTL_AFTER_INITIAL
`endif
`endif // SYNTHESIS
endmodule
```

If using `MemorySynthInit`:

```verilog
initial begin
  `ifdef RANDOMIZE
    `ifdef INIT_RANDOM
      `INIT_RANDOM
    `endif
    `ifndef VERILATOR
      `ifdef RANDOMIZE_DELAY
        #`RANDOMIZE_DELAY begin end
      `else
        #0.002 begin end
      `endif
    `endif
  `endif // RANDOMIZE
end // initial
`ifdef FIRRTL_AFTER_INITIAL
`FIRRTL_AFTER_INITIAL
`endif
`endif // SYNTHESIS
  initial begin
    $readmemh("sample.hex", mem);
  end
endmodule
```

#### Desired Merge Strategy

#### Release Notes

This PR adds two annotations to determine how `readmem` statements are generated in Verilog code.
By default, it is placed inside the `ifndef SYNTHESIS` block which gets ignored by some FPGA tools.
One can add either `MemorySynthInit` or `MemoryNoSynthInit` (which is the default if not defined) annotation to change the behaviour and have the `readmem` outside this `ifndef` block.

In the future, MemorySynthInit could be set as default for FPGA targets (`-target:FPGA`).

TBD

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/annotations/MemoryInitAnnotation.scala
+++ b/src/main/scala/firrtl/annotations/MemoryInitAnnotation.scala
@@ -52,3 +52,8 @@ case class MemoryFileInlineAnnotation(
   override def initValue:    MemoryInitValue = MemoryFileInlineInit(filename, hexOrBinary)
   override def isRandomInit: Boolean = false
 }
+
+/** Defines the memory initialization target, inside or outside the `ifndef SYNTHESIS` block */
+trait MemoryInitTargetAnnotation extends NoTargetAnnotation
+case object MemoryNoSynthInit extends MemoryInitTargetAnnotation
+case object MemorySynthInit extends MemoryInitTargetAnnotation
--- a/src/main/scala/firrtl/annotations/MemoryInitAnnotation.scala
+++ b/src/main/scala/firrtl/annotations/MemoryInitAnnotation.scala
@@ -53,7 +53,8 @@ case class MemoryFileInlineAnnotation(
   override def isRandomInit: Boolean = false
 }
 
-/** Defines the memory initialization target, inside or outside the `ifndef SYNTHESIS` block */
-trait MemoryInitTargetAnnotation extends NoTargetAnnotation
-case object MemoryNoSynthInit extends MemoryInitTargetAnnotation
-case object MemorySynthInit extends MemoryInitTargetAnnotation
+/** Initializes the memory inside the `ifndef SYNTHESIS` block (default) */
+case object MemoryNoSynthInit extends NoTargetAnnotation
+
+/** Initializes the memory outside the `ifndef SYNTHESIS` block */
+case object MemorySynthInit extends NoTargetAnnotation
--- a/src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
+++ b/src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
@@ -7,7 +7,14 @@ import firrtl.PrimOps._
 import firrtl.Utils._
 import firrtl.WrappedExpression._
 import firrtl.traversals.Foreachers._
-import firrtl.annotations.{CircuitTarget, MemoryLoadFileType, ReferenceTarget, SingleTargetAnnotation}
+import firrtl.annotations.{
+  CircuitTarget,
+  MemoryLoadFileType,
+  MemoryNoSynthInit,
+  MemorySynthInit,
+  ReferenceTarget,
+  SingleTargetAnnotation
+}
 import firrtl.passes.LowerTypes
 import firrtl.passes.MemPortUtils._
 import firrtl.stage.TransformManager
@@ -482,6 +489,15 @@ class VerilogEmitter extends SeqTransform with Emitter {
     def getConnectEmissionOption(target: ReferenceTarget): ConnectEmissionOption =
       connectEmissionOption(target)
 
+    // Defines the memory initialization based on the annotation
+    // Defaults to having the memories inside the `ifndef SYNTHESIS` block
+    def emitMemoryInitAsNoSynth: Option[Boolean] = {
+      annotations.collectFirst {
+        case MemoryNoSynthInit => Option(true)
+        case MemorySynthInit   => Option(false)
+      }.getOrElse(Option(true))
+    }
+
     private val emissionAnnos = annotations.collect {
       case m: SingleTargetAnnotation[ReferenceTarget] @unchecked with EmissionOption => m
     }
@@ -605,6 +621,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
     val asyncInitials = ArrayBuffer[Seq[Any]]()
     // memories need to be initialized even when randomization is disabled
     val memoryInitials = ArrayBuffer[Seq[Any]]()
+    val randomMemoryInitials = ArrayBuffer[Seq[Any]]()
     val simulates = ArrayBuffer[Seq[Any]]()
     val formals = mutable.LinkedHashMap[Expression, ArrayBuffer[Seq[Any]]]()
 
@@ -861,7 +878,14 @@ class VerilogEmitter extends SeqTransform with Emitter {
             case MemoryLoadFileType.Binary => "$readmemb"
             case MemoryLoadFileType.Hex    => "$readmemh"
           }
-          memoryInitials += Seq(s"""$readmem("$filename", ${s.name});""")
+          if (emissionOptions.emitMemoryInitAsNoSynth.get) {
+            memoryInitials += Seq(s"""$readmem("$filename", ${s.name});""")
+          } else {
+            val inlineLoad = s"""initial begin
+                                |    $readmem("$filename", ${s.name});
+                                |  end""".stripMargin
+            memoryInitials += Seq(inlineLoad)
+          }
       }
     }
 
@@ -1140,7 +1164,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
         emit(Seq(tab, "end"))
       }
 
-      if (initials.nonEmpty || ifdefInitials.nonEmpty || memoryInitials.nonEmpty) {
+      if (initials.nonEmpty || ifdefInitials.nonEmpty || memoryInitials.nonEmpty || randomMemoryInitials.nonEmpty) {
         emit(Seq("// Register and memory initialization"))
         emit(Seq("`ifdef RANDOMIZE_GARBAGE_ASSIGN"))
         emit(Seq("`define RANDOMIZE"))
@@ -1158,7 +1182,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
         emit(Seq("`define RANDOM $random"))
         emit(Seq("`endif"))
         // the initvar is also used to initialize memories to constants
-        if (memoryInitials.isEmpty) emit(Seq("`ifdef RANDOMIZE_MEM_INIT"))
+        if (randomMemoryInitials.isEmpty) emit(Seq("`ifdef RANDOMIZE_MEM_INIT"))
         // Since simulators don't actually support memories larger than 2^31 - 1, there is no reason
         // to change Verilog emission in the common case. Instead, we only emit a larger initvar
         // where necessary
@@ -1169,7 +1193,7 @@ class VerilogEmitter extends SeqTransform with Emitter {
           val width = maxMemSize.bitLength - 1 // minus one because [width-1:0] has a width of "width"
           emit(Seq(s"  reg [$width:0] initvar;"))
         }
-        if (memoryInitials.isEmpty) emit(Seq("`endif"))
+        if (randomMemoryInitials.isEmpty) emit(Seq("`endif"))
         emit(Seq("`ifndef SYNTHESIS"))
         // User-defined macro of code to run before an initial block
         emit(Seq("`ifdef FIRRTL_BEFORE_INITIAL"))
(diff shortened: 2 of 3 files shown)


carlosedp:
I'd like some feedback on this approach and if it's the correct way to do it.

I successfully got the results by changing the parameter on `def initValue: MemoryInitEmissionValue = MemoryNoSynthInit` but don't know how to make it change based on different generation parameters (like the one currently in dev in PR #2111.

src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
```diff
@@ -458,6 +458,8 @@ class VerilogEmitter extends SeqTransform with Emitter {
   private[firrtl] class EmissionOptions(annotations: AnnotationSeq) {
     // Private so that we can present an immutable API
     private val memoryEmissionOption = new EmissionOptionMap[MemoryEmissionOption](MemoryEmissionOptionDefault)
+    private val memoryInitEmissionOption =
+      new EmissionOptionMap[MemoryInitEmissionOption](MemoryInitBlockEmissionOption)
```
jackkoenig, line 462: Is the intent for this to be more of a global option or controllable _per memory_? `EmissionOptionMaps` have machinery for handling different options at a granularity of per component (in this case per memory), but I think that's overkill for what you're doing.

Since `getMemoryInitEmissionOption` doesn't take a `ReferenceTarget` as an option, I think we can simplify this a bit and reduce the public API exposure.

src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
```diff
@@ -467,6 +469,9 @@ class VerilogEmitter extends SeqTransform with Emitter {
     def getMemoryEmissionOption(target: ReferenceTarget): MemoryEmissionOption =
       memoryEmissionOption(target)
 
+    def getMemoryInitEmissionOption(): MemoryInitEmissionOption =
+      MemoryInitBlockEmissionOption
+
```
jackkoenig, line 474:
```suggestion
    def emitMemoryInitAsNoSynth: Boolean =
      annotations.collectFirst { case SomeAnnotationIndicatingWhatToDo => true }.getOrElse(false)
      MemoryInitBlockEmissionOption
      
```
I think you can then just add a `case object SomeAnnotationIndicatingWhatToDo extends NoTargetAnnotation` (pick a better name) and get rid of the `MemoryInitEmissionValue` and `MemoryInitEmissionOption` stuff above. This would make it pretty easy to control whether the emission style.

Minor style nit but `()` at the end of a 0-arity method implies it mutates something which this does not.

carlosedp: I also thought it became a little convoluted (just a little? 😂).
Gonna try to simplify it.

src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
ekiwi: I also think this should be a global option. (i.e. a NoTargetAnnotation)

src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
jackkoenig: This relates to the comments above on `private val memoryInitEmissionOption`, I would make this:
```scala
// Somewhere else
case object EmitMemoryInitBlockOutsideIfDefAnnotation extends NoTargetAnnotation
...

// Now in this code block
lazy val emitMemoryInitAsNoSynth: Boolean = annotations.collectFirst {
  case EmitMemoryInitBlockOutsideIfDefAnnotation => true
}.getOrElse(false)
// Then you can just use this value elsewhere in the emitter
```

@ekiwi thoughts on the annotation name / style? It could be a `case class` with a Boolean argument or as I suggested, default to inside the ifndef SYNTHESIS and emit outside of it if the annotation is provided

carlosedp:
I think is is much cleaner and allows defining the behaviour by using an annotation in the Chisel code.

```scala
    override def toFirrtl =
      firrtl.annotations.MemorySynthInit
      // or
      // firrtl.annotations.MemoryNoSynthInit
  })

  val mem = SyncReadMem(words, UInt(bitWidth.W))
  if (memoryFile.trim().nonEmpty) {
    println(s"  Load memory file: " + memoryFile)
    loadMemoryFromFileInline(mem, memoryFile)
  }
```

src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
ekiwi: We can do a single case object. Maybe something like `EmitMemoryInitForSynthesis` or `SynthesizeMemoryInit`.

carlosedp: I added both cases so one could define in the code which anno to use.

src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
```diff
@@ -1200,13 +1223,19 @@ class VerilogEmitter extends SeqTransform with Emitter {
         for (x <- initials) emit(Seq(tab, x))
         for (x <- asyncInitials) emit(Seq(tab, x))
         emit(Seq("  `endif // RANDOMIZE"))
-        for (x <- memoryInitials) emit(Seq(tab, x))
+
+        if (emissionOptions.emitMemoryInitAsNoSynth.get) {
+          for (x <- memoryInitials) emit(Seq(tab, x))
+        }
```
jackkoenig, line 1229: You definitely don't want to do a `.get`, you'll throw an exception if neither annotation is provided.

src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
```diff
@@ -482,6 +489,15 @@ class VerilogEmitter extends SeqTransform with Emitter {
...
       connectEmissionOption(target)
 
+    // Defines the memory initialization based on the annotation
+    // Defaults to having the memories inside the `ifndef SYNTHESIS` block
+    def emitMemoryInitAsNoSynth: Option[Boolean] = {
+      annotations.collectFirst {
+        case MemoryNoSynthInit => Option(true)
+        case MemorySynthInit   => Option(false)
+      }.getOrElse(Option(true))
+    }
```
jackkoenig, line 499: You're always making sure it's `Some` anyway, might as well just return the `Boolean`.
```suggestion
    def emitMemoryInitAsNoSynth: Boolean = {
      annotations.collectFirst {
        case MemoryNoSynthInit => true
        case MemorySynthInit   => false
      }.getOrElse(true)
    }
```

src/main/scala/firrtl/annotations/MemoryInitAnnotation.scala
```diff
@@ -52,3 +52,8 @@ case class MemoryFileInlineAnnotation(
   override def initValue:    MemoryInitValue = MemoryFileInlineInit(filename, hexOrBinary)
   override def isRandomInit: Boolean = false
 }
+
+/** Defines the memory initialization target, inside or outside the `ifndef SYNTHESIS` block */
+trait MemoryInitTargetAnnotation extends NoTargetAnnotation
+case object MemoryNoSynthInit extends MemoryInitTargetAnnotation
+case object MemorySynthInit extends MemoryInitTargetAnnotation
```
jackkoenig, line 59:
```suggestion
sealed trait MemoryInitTargetAnnotation extends NoTargetAnnotation
case object MemoryNoSynthInit extends MemoryInitTargetAnnotation
case object MemorySynthInit extends MemoryInitTargetAnnotation
```
You don't use `MemoryInitTargetAnnotation` for anything, is it really needed?

src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
jackkoenig: Also, I'm ambivalent if we provide just 1 annotation or both, but if we provide both, we need to check somewhere that the user doesn't provide both. It's not great waiting till the VerilogEmitter to check, but it's also possible that a custom transform could erroneously inject one that contradicts the one provided by the user so this is probably the right place to check.

src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
jackkoenig: I see now that you're providing a default, let's just drop the `Option` on the return.

carlosedp: Done!

src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
carlosedp: Don't know why it was complaining and wanted an Option[Bool] before.
Fixed.

src/main/scala/firrtl/annotations/MemoryInitAnnotation.scala
carlosedp: Really not needed. I was doing a `match case` before.. but not anymore.
Removed.

src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
```diff
@@ -482,6 +489,15 @@ class VerilogEmitter extends SeqTransform with Emitter {
...
       connectEmissionOption(target)
 
+    // Defines the memory initialization based on the annotation
+    // Defaults to having the memories inside the `ifndef SYNTHESIS` block
+    def emitMemoryInitAsNoSynth: Boolean = {
+      annotations.collectFirst {
+        case MemoryNoSynthInit => true
+        case MemorySynthInit   => false
+      }.getOrElse(true)
+    }
```
jackkoenig, line 499:
```suggestion
      val annos = annotations.collect { case a @ (MemoryNoSynthInit | MemorySynthInit) => a }
      annos match {
        case Seq() => true
        case Seq(MemoryNoSynthInit) => true
        case Seq(MemorySynthInit) => false
        case other =>
          throw new FirrtlUserException("There should only be at most one memory initialization option annotation, got $other")
      }
    }
```
I'm not sure of the best way to express this, but we need to check that the user doesn't provide contradictory options, here I'm checking only 1 is specified but we could also `.distinct` the `annos` to allow more than 1 of the same kind.

carlosedp: Yes, that's much better. I think for now there is no need for multiple values. Would over-complicate things.

src/main/scala/firrtl/annotations/MemoryInitAnnotation.scala
```diff
@@ -52,3 +52,7 @@ case class MemoryFileInlineAnnotation(
   override def initValue:    MemoryInitValue = MemoryFileInlineInit(filename, hexOrBinary)
   override def isRandomInit: Boolean = false
 }
+
+/** Defines the memory initialization target, inside or outside the `ifndef SYNTHESIS` block */
```
ekiwi, line 56: You should add a scala doc comment to each annotation describing what each one individually does.

ekiwi (dismissed):

src/main/scala/firrtl/annotations/MemoryInitAnnotation.scala
carlosedp: Added!

carlosedp:
Also I tested that the `MemorySynthInit`/`MemoryNoSynthInit` annotation doesn't need to be in the memory module itself. It can be in the Topmodule for example and works fine.

jackkoenig (approved):

PR #2175, opened 2021-04-08
smt: use existing bitWidth API

edwardcwang:
#### Type of Improvement

refactoring/documentation

#### API Impact
no impact
should have no impact
#### Desired Merge Strategy
squash
#### Release Notes
Refactor SMT backend to use the existing public API `firrtl.bitWidth`.
Also adds scaladoc to `firrtl.bitWidth`.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/Utils.scala
+++ b/src/main/scala/firrtl/Utils.scala
@@ -51,7 +51,18 @@ object getWidth {
   def apply(e: Expression): Width = apply(e.tpe)
 }
 
+/**
+  * Helper object for computing the width of a firrtl type.
+  */
 object bitWidth {
+
+  /**
+    * Compute the width of a firrtl type.
+    * For example, a Vec of 4 UInts of width 8 should have a width of 32.
+    *
+    * @param dt firrtl type
+    * @return Width of the given type
+    */
   def apply(dt:           Type): BigInt = widthOf(dt)
   private def widthOf(dt: Type): BigInt = dt match {
     case t: VectorType => t.size * bitWidth(t.tpe)
--- a/src/main/scala/firrtl/backends/experimental/smt/FirrtlExpressionSemantics.scala
+++ b/src/main/scala/firrtl/backends/experimental/smt/FirrtlExpressionSemantics.scala
@@ -8,19 +8,10 @@ import firrtl.PrimOps
 import firrtl.passes.CheckWidths.WidthTooBig
 
 private trait TranslationContext {
-  def getReference(name: String, tpe: ir.Type): BVExpr = BVSymbol(name, FirrtlExpressionSemantics.getWidth(tpe))
+  def getReference(name: String, tpe: ir.Type): BVExpr = BVSymbol(name, firrtl.bitWidth(tpe).toInt)
 }
 
 private object FirrtlExpressionSemantics {
-  def getWidth(tpe: ir.Type): Int = tpe match {
-    case ir.UIntType(ir.IntWidth(w))   => w.toInt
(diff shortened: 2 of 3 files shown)


src/main/scala/firrtl/Utils.scala
```diff
@@ -51,7 +51,17 @@ object getWidth {
   def apply(e: Expression): Width = apply(e.tpe)
 }
 
+/**
+ * Helper object for computing the width of a firrtl type.
+ */
 object bitWidth {
```
ekiwi, line 57: This is great! Thanks for discovering this function. I did not know that it existed.

src/main/scala/firrtl/backends/experimental/smt/Btor2Serializer.scala
```diff
@@ -48,15 +48,15 @@ private class Btor2Serializer private () {
...
 
-  private def t(width: Int): Int = bitVecTypeCache.getOrElseUpdate(width, line(s"sort bitvec $width"))
+  private def t(width: BigInt): Int = bitVecTypeCache.getOrElseUpdate(width, line(s"sort bitvec $width"))
 
   // bit vector expression serialization
   private def s(expr: BVExpr): Int = expr match {
     case BVLiteral(value, width) => lit(value, width)
     case BVSymbol(name, _)       => symbols(name)
-    case BVExtend(e, 0, _)       => s(e)
+    case BVExtend(e, by, _) if by == 0 => s(e)
```
ekiwi, line 59: This is why I did not want to use BigInt: They do not work well for pattern matching.

src/main/scala/firrtl/backends/experimental/smt/FirrtlExpressionSemantics.scala
```diff
@@ -8,19 +8,10 @@ import firrtl.PrimOps
 import firrtl.passes.CheckWidths.WidthTooBig
 
 private trait TranslationContext {
-  def getReference(name: String, tpe: ir.Type): BVExpr = BVSymbol(name, FirrtlExpressionSemantics.getWidth(tpe))
+  def getReference(name: String, tpe: ir.Type): BVExpr = BVSymbol(name, firrtl.bitWidth(tpe))
```
ekiwi, line 11: This is good

ekiwi (changes requested):
Replacing the custom `getWidth` function with one from the central utilities file is great. I did not know about it when writing the backend.

I am skeptical about the benefit of using BigInt for bit vector widths since they make pattern matching harder for no apparent benefit. The SMT Expression library has a type system that is completely separate from the firrtl type system (for one we do not distinguish SInt and UInt), and thus it does not necessarily have to follow the firrtl decision to use BigInt for widths.

Of course, if you can show me a use case where you even get close to the limit of an `Int` I am happy to make the change to BigInt.

src/main/scala/firrtl/backends/experimental/smt/Btor2Serializer.scala
edwardcwang: No worries, that's fine - will put it back to Int. I don't have a strong opinion either way. Was just noting that for some reason the rest of FIRRTL uses BigInt for widths.

src/main/scala/firrtl/Utils.scala
edwardcwang: Real credit goes to @jackkoenig  :)

ekiwi (approved):
LGTM, you just need to run `sbt scalafmtAll` to make the CI happy.

edwardcwang:
(useful note, `mill firrtl[2.12.13].reformat` also seems to do reformatting)

PR #2177, opened 2021-04-13
Add indent parameter to Serializer.serialize()

jared-barocsi:
#### Type of Improvement

Code refactoring

#### API Impact

No impact

Should not impact code generation

#### Desired Merge Strategy

Squash

#### Release Notes

Add indentation parameter to `Serializer.serialize()` to indent `FirrtlNode` strings by a certain amount.

- [x] Did you review?

### DIFF
--- a/src/main/scala/firrtl/ir/Serializer.scala
+++ b/src/main/scala/firrtl/ir/Serializer.scala
@@ -11,7 +11,8 @@ object Serializer {
   val Indent = "  "
 
   /** Converts a `FirrtlNode` into its string representation with
-    * default indentation. */
+    * default indentation.
+    */
   def serialize(node: FirrtlNode): String = {
     serialize(node, 0)
   }
--- a/src/main/scala/firrtl/ir/Serializer.scala
+++ b/src/main/scala/firrtl/ir/Serializer.scala
@@ -10,29 +10,29 @@ object Serializer {
   val NewLine = '\n'
   val Indent = "  "
 
-  /** Converts a `FirrtlNode` into its string representation
-   * with *zero* indents (old behavior for binary compatibility) */
+  /** Converts a `FirrtlNode` into its string representation with
+    * default indentation. */
   def serialize(node: FirrtlNode): String = {
     serialize(node, 0)
   }
 
   /** Converts a `FirrtlNode` into its string representation. */
-  def serialize(node: FirrtlNode, indents: Int): String = {
+  def serialize(node: FirrtlNode, indent: Int): String = {
     val builder = new StringBuilder()
     node match {
-      case n: Info        => s(n)(builder, indents)
-      case n: StringLit   => s(n)(builder, indents)
-      case n: Expression  => s(n)(builder, indents)
-      case n: Statement   => s(n)(builder, indents)
-      case n: Width       => s(n)(builder, indents)
-      case n: Orientation => s(n)(builder, indents)
-      case n: Field       => s(n)(builder, indents)
-      case n: Type        => s(n)(builder, indents)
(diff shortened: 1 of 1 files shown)


src/main/scala/firrtl/ir/Serializer.scala
```diff
@@ -10,24 +10,29 @@ object Serializer {
...
 
-  /** Converts a `FirrtlNode` into its string representation. */
+  /** Converts a `FirrtlNode` into its string representation
+   * with *zero* indents (old behavior for binary compatibility) */
   def serialize(node: FirrtlNode): String = {
+    serialize(node, 0)
+  }
+
+  /** Converts a `FirrtlNode` into its string representation. */
+  def serialize(node: FirrtlNode, indents: Int): String = {
```
mwachs5, line 20:
```suggestion
  def serialize(node: FirrtlNode, indents: Int = 0): String = {
```
Can't we do away with the above version by just adding a default param? Does that still match binary compatibility? (I don't actually know the ansewr this is a weak suggestion)

jared-barocsi: The default parameter approach was my first attempt at solving this, but I'm not sure how it would be compiled into bytecode. My intuition tells me that the serialize with the default parameter will still be a two-parameter function (with some fancy magic to "insert" the 0 as an argument). So if we swap the firrtl binary then anything else that tries to run that will get a complaint that the one-parameter serialize doesn't exist.

mwachs5: looks like your intution is right: https://github.com/jatcwang/binary-compatibility-guide#dont-adding-parameters-with-default-values-to-methods

Not sure what the firrtl philosophy is on making this breaking change on master and then fixing in the backport version?

jackkoenig: I think just having 2 is fine.

I'm going to be super bikesheddy here though and suggest naming the parameter just `indent`:
```suggestion
  def serialize(node: FirrtlNode, indent: Int): String = {
```

Doing a bit of search on other projects, singular tense is more common for this kind of parameter.

src/main/scala/firrtl/ir/Serializer.scala
```diff
@@ -10,24 +10,29 @@ object Serializer {
   val NewLine = '\n'
   val Indent = "  "
 
-  /** Converts a `FirrtlNode` into its string representation. */
+  /** Converts a `FirrtlNode` into its string representation
+   * with *zero* indents (old behavior for binary compatibility) */
```
jackkoenig, line 14:
```suggestion
  /** Converts a `FirrtlNode` into its string representation with default indentation
    */
```
A couple of things here:
1. Use ScalaDoc style (stars align with the rightmost star from the start of the comment
2. Slight tweak to phrasing, the behavior here didn't change so no reason to bring attention to it. There also is indentation if you pass something with things scoped inside it (like a Module or a When), so slightly more precise phrasing.

src/main/scala/firrtl/ir/Serializer.scala
```diff
@@ -10,29 +10,29 @@ object Serializer {
   val NewLine = '\n'
   val Indent = "  "
 
-  /** Converts a `FirrtlNode` into its string representation
-   * with *zero* indents (old behavior for binary compatibility) */
+  /** Converts a `FirrtlNode` into its string representation with
+    * default indentation. */
```
mwachs5, line 14:
```suggestion
    * default indentation.
    */
```

jackkoenig (approved):

ekiwi:
Congratulations on your first firrtl PR @jared-barocsi :tada:

PR #2189, opened 2021-04-17
Don't use declaration-assigns for wires representing mem ports

albert-magyar:
**Type of improvement:** bug-fix
**API impact:** none
**Backend code-generation impact:** Avoids using declaration-assignments for "intermediate" wires created to represent the various fields of memory ports. Since these wires are create when the memory is declared, they may appear before the definition of whatever happens to drive them.

### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
+++ b/src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
@@ -1076,8 +1076,11 @@ class VerilogEmitter extends SeqTransform with Emitter {
                 "Alternatively, add the --repl-seq-mem flag to replace memories with blackboxes."
               ).mkString(" ")
             )
-          def createMemWire(name: String, tpe: Type, rhs: InfoExpr): Unit = {
-            declare("wire", name, tpe, MultiInfo(sx.info, rhs.info), rhs.expr)
+          def createMemWire(firrtlRef: Expression, rhs: InfoExpr): Unit = {
+            // Don't use declaration-assignment, since this assignment might be emitted earlier than the
+            // actual connection to the memory port field in the source FIRRTL
+            declare("wire", LowerTypes.loweredName(firrtlRef), firrtlRef.tpe, MultiInfo(sx.info, rhs.info))
+            assign(firrtlRef, rhs)
           }
 
           for (r <- sx.readers) {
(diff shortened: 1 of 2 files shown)


albert-magyar:
Here is a minimal example of a circuit that would cause this Verilog declaration-use issue before this PR. This PR fixes it. I will add this as some kind of test as part of this PR.

```
File Edit Options Buffers Tools Help                                                                                                                       
circuit wdata_driver_decl_after_mem:
  module wdata_driver_decl_after_mem:
    input clk: Clock
    input addr: UInt<10>
    input wdata: UInt<8>[2]

    mem m:
      data-type => UInt<16>
      depth => 1024
      writer => w
      read-latency => 1
      write-latency => 1

    m.w.clk <= clk
    m.w.addr <= addr
    m.w.en <= UInt(1)
    m.w.mask <= UInt(1)
    node wdata_catted = cat(wdata[1], wdata[0])
    node wdata_actual = wdata_catted
    m.w.data <= wdata_actual
```

ekiwi:
> This PR fixes it. I will add this as some kind of test as part of this PR.

I wonder if we could add some more strict `verilator` linting to the regression tests. I am surprised that this problem wasn't caught in CI because it seems to also affect rocket chip. Apparent `verilator` accepts use before declaration?

ekiwi:
I cannot get Verilator to complain about this problem. Not with `-Wall` and not by setting the `default_nettype` to `none`.

albert-magyar:
Default nettype only comes into play when an undeclared identifier appears on the *lhs* of an assign.

albert-magyar:
The change that caused the use-before-declaration was basically just a silly attempt at improving style that I tacked on to the actual substantive changes to the emitter in #2111. Most consumers of Verilog don't require declaration-before-rhs-reference for non-default nets since they try to be "declarative," but it's definitely better to emit universally acceptable code from a compiler.

@ekiwi @jackkoenig, since there is basically no way to write a test for this that will run in CI, can we just merge this?

ekiwi (approved):
>  since there is basically no way to write a test for this that will run in CI, can we just merge this?

:ship:

ekiwi:
> Most consumers of Verilog don't require declaration-before-rhs-reference for non-default nets since they try to be "declarative," but it's definitely better to emit universally acceptable code from a compiler.

Just FYI, the tool that was complaining about this is VCS.

jackkoenig:
This should have been backported right? Needed for https://github.com/chipsalliance/firrtl/pull/2172?

PR #2191, opened 2021-04-17
Memlib Refactor

sequencer:
This PR depends on #2199
PLCT intern @sinofp is working on rewriting `vlsi_mem_gen` into a Firrtl Transform.
This PR refactor Memlib to pave the path for him to add Annotations for memlib.

1. remove -c option in --replSeqMem console output(this not used anymore in firrtl codebase)
1. replace ConfWriter with MemLibOutConfigFileAnnotation, ConfWriter should be removed in next FIRRTL verision, since it's not private.
1. deprecate CreateMemoryAnnotation, inline which in ReplSeqMem.
1. fix ReplSeqMemTests, checkMemConf will read from correspond annotations.

Generally this PR try to replace
https://github.com/chipsalliance/firrtl/blob/bf1cf3d2db49195d031f89594baebcc9f307659e/src/main/scala/firrtl/passes/memlib/ReplaceMemTransform.scala#L135-L145
with Annotation, to make memlib transforms be able to digest Annotation and generate another Annotation, rather than directly write the conf file.
### Contributor Checklist

#### Type of Improvement

- code refactoring

#### API Impact
Deprecate CreateMemoryAnnotations, ConfWriter

None

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/passes/memlib/ReplaceMemMacros.scala
+++ b/src/main/scala/firrtl/passes/memlib/ReplaceMemMacros.scala
@@ -122,12 +122,13 @@ class ReplaceMemMacros extends Transform with DependencyAPIMigration {
     })
   )
 
-  def memToBundle(s: DefAnnotatedMemory) = BundleType(
+  private def memToBundle(s: DefAnnotatedMemory) = BundleType(
     s.readers.map(Field(_, Flip, rPortToBundle(s))) ++
       s.writers.map(Field(_, Flip, wPortToBundle(s))) ++
       s.readwriters.map(Field(_, Flip, rwPortToBundle(s)))
   )
-  def memToFlattenBundle(s: DefAnnotatedMemory) = BundleType(
+
+  private def memToFlattenBundle(s: DefAnnotatedMemory) = BundleType(
     s.readers.map(Field(_, Flip, rPortToFlattenBundle(s))) ++
       s.writers.map(Field(_, Flip, wPortToFlattenBundle(s))) ++
       s.readwriters.map(Field(_, Flip, rwPortToFlattenBundle(s)))
@@ -137,7 +138,7 @@ class ReplaceMemMacros extends Transform with DependencyAPIMigration {
     *  The wrapper module has the same type as the memory it replaces
     *  The external module
     */
-  def createMemModule(
+  private def createMemModule(
     m:                    DefAnnotatedMemory,
     wrapperName:          String,
     annotatedMemoriesBuffer: ListBuffer[DefAnnotatedMemory]
@@ -167,19 +168,19 @@ class ReplaceMemMacros extends Transform with DependencyAPIMigration {
 
   // TODO(shunshou): get rid of copy pasta
   // Connects the clk, en, and addr fields from the wrapperPort to the bbPort
-  def defaultConnects(wrapperPort: WRef, bbPort: WSubField): Seq[Connect] =
+  private def defaultConnects(wrapperPort: WRef, bbPort: WSubField): Seq[Connect] =
     Seq("clk", "en", "addr").map(f => connectFields(bbPort, f, wrapperPort, f))
 
   // Generates mask bits (concatenates an aggregate to ground type)
   // depending on mask granularity (# bits = data width / mask granularity)
-  def maskBits(mask: WSubField, dataType: Type, fillMask: Boolean): Expression =
+  private def maskBits(mask: WSubField, dataType: Type, fillMask: Boolean): Expression =
     if (fillMask) toBitMask(mask, dataType) else toBits(mask)
 
-  def adaptReader(wrapperPort: WRef, bbPort: WSubField): Seq[Statement] =
+  private def adaptReader(wrapperPort: WRef, bbPort: WSubField): Seq[Statement] =
     defaultConnects(wrapperPort, bbPort) :+
       fromBits(WSubField(wrapperPort, "data"), WSubField(bbPort, "data"))
 
-  def adaptWriter(wrapperPort: WRef, bbPort: WSubField, hasMask: Boolean, fillMask: Boolean): Seq[Statement] = {
+  private def adaptWriter(wrapperPort: WRef, bbPort: WSubField, hasMask: Boolean, fillMask: Boolean): Seq[Statement] = {
     val wrapperData = WSubField(wrapperPort, "data")
     val defaultSeq = defaultConnects(wrapperPort, bbPort) :+
       Connect(NoInfo, WSubField(bbPort, "data"), toBits(wrapperData))
@@ -194,7 +195,7 @@ class ReplaceMemMacros extends Transform with DependencyAPIMigration {
     }
   }
 
-  def adaptReadWriter(wrapperPort: WRef, bbPort: WSubField, hasMask: Boolean, fillMask: Boolean): Seq[Statement] = {
+  private def adaptReadWriter(wrapperPort: WRef, bbPort: WSubField, hasMask: Boolean, fillMask: Boolean): Seq[Statement] = {
     val wrapperWData = WSubField(wrapperPort, "wdata")
     val defaultSeq = defaultConnects(wrapperPort, bbPort) ++ Seq(
       fromBits(WSubField(wrapperPort, "rdata"), WSubField(bbPort, "rdata")),
@@ -216,7 +217,7 @@ class ReplaceMemMacros extends Transform with DependencyAPIMigration {
   private type NameMap = collection.mutable.HashMap[(String, String), String]
 
   /** Construct NameMap by assigning unique names for each memory blackbox */
-  def constructNameMap(namespace: Namespace, nameMap: NameMap, mname: String)(s: Statement): Statement = {
+  private def constructNameMap(namespace: Namespace, nameMap: NameMap, mname: String)(s: Statement): Statement = {
     s match {
       case m: DefAnnotatedMemory =>
         m.memRef match {
@@ -228,7 +229,7 @@ class ReplaceMemMacros extends Transform with DependencyAPIMigration {
     s.map(constructNameMap(namespace, nameMap, mname))
   }
 
-  def updateMemStmts(
+  private def updateMemStmts(
     namespace:            Namespace,
     nameMap:              NameMap,
     mname:                String,
@@ -256,7 +257,7 @@ class ReplaceMemMacros extends Transform with DependencyAPIMigration {
     case sx => sx.map(updateMemStmts(namespace, nameMap, mname, memPortMap, memMods, annotatedMemoriesBuffer))
   }
 
-  def updateMemMods(
+  private def updateMemMods(
     namespace:            Namespace,
     nameMap:              NameMap,
     memMods:              Modules,
--- a/src/main/scala/firrtl/passes/memlib/ReplaceMemMacros.scala
+++ b/src/main/scala/firrtl/passes/memlib/ReplaceMemMacros.scala
@@ -13,6 +13,8 @@ import firrtl.annotations._
 import firrtl.stage.Forms
 import wiring._
 
+import scala.collection.mutable.ListBuffer
+
 /** Annotates the name of the pins to add for WiringTransform */
 case class PinAnnotation(pins: Seq[String]) extends NoTargetAnnotation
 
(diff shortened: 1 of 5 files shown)


src/main/scala/firrtl/passes/memlib/ReplaceMemMacros.scala
```diff
@@ -24,8 +24,9 @@ object ReplaceMemMacros {
   * This will not generate wmask ports if not needed.
   * Creates the minimum # of black boxes needed by the design.
   */
-class ReplaceMemMacros(writer: ConfWriter) extends Transform with DependencyAPIMigration {
-
+class ReplaceMemMacros extends Transform with DependencyAPIMigration {
+  val defAnnotatedMemories: collection.mutable.ListBuffer[DefAnnotatedMemory] =
+    collection.mutable.ListBuffer[DefAnnotatedMemory]()
```
jackkoenig, line 29: This shouldn't be public, and also must be cleared each time the transform is run. Ideally we wouldn't have a class field for this information (arguments are better), but that obviously would be a much more intrusive change.

ekiwi: > This shouldn't be public, and also must be cleared each time the transform is run.

Are you telling me that if my Transform is a class, only a single instance is created and could be used for multiple calls to `execute`? If that is the case, shouldn't all transforms be `object` instead of `class`?

seldridge: Yes, the dependency manager will only construct one instance of a class if it needs to run multiple times. Maybe this wasn't the best idea, but the intent was to reuse existing transforms to solve lowerings due to invalidations. This can be a source of bugs (like this) where there is state in the transform.

src/main/scala/firrtl/passes/memlib/ReplaceMemMacros.scala
```diff
@@ -276,7 +275,10 @@ class ReplaceMemMacros(writer: ConfWriter) extends Transform with DependencyAPIM
       seq ++ memMods.collect {
         case m: ExtModule => SinkAnnotation(ModuleName(m.name, CircuitName(c.main)), pin)
       }
-    } ++ state.annotations
+    } ++ state.annotations.map {
+      case MemLibOutConfigFileAnnotation(f, _) => MemLibOutConfigFileAnnotation(f, defAnnotatedMemories.toList)
+      case a                                   => a
+    }
```
sequencer, line 281: As for option 2:
I think I can add clean up at line 282 here.

src/main/scala/firrtl/passes/memlib/ReplaceMemMacros.scala
ekiwi: > Maybe this wasn't the best idea, but the intent was to reuse existing transforms to solve lowerings due to invalidations.

I see. I guess that would mean that eventually we just migrate all transforms to be `object` and then maybe we can deprecate `Dependency[..]` and leave `Dependency(...)` as the only option.

(_Note_:  sorry for hijacking this thread)

seldridge: (2) may also require defining a critical region to serialize accesses to the mutable internals. I think there can be bugs when working in a parallel environment using the same stage. This can happen in tests.

@ekiwi: That seems very reasonable, though a long, painful deprecation road. :)

sequencer: > (2) may also require defining a critical region to serialize accesses to the mutable internals. I think there can be bugs when working in a parallel environment using the same stage. This can happen in tests.

Got it, I will directly refactor it ;)

> (Note: sorry for hijacking this thread)

Haha, no worries, I think this did help me understand what I should do for this PR.

src/main/scala/firrtl/passes/memlib/ReplaceMemMacros.scala
```diff
@@ -24,8 +26,7 @@ object ReplaceMemMacros {
   * This will not generate wmask ports if not needed.
   * Creates the minimum # of black boxes needed by the design.
   */
-class ReplaceMemMacros(writer: ConfWriter) extends Transform with DependencyAPIMigration {
-
+class ReplaceMemMacros extends Transform with DependencyAPIMigration {
```
jackkoenig, line 29: Your deprecation warnings say things are deprecated in 1.4 (to be removed in 1.5), but this is a pretty substantial binary compatibility change. I'm not arguing against the change but it may be tough to do them in a way that is reasonable to backport.

src/main/scala/firrtl/passes/memlib/ReplaceMemTransform.scala
```diff
@@ -70,19 +73,76 @@ class ConfWriter(filename: String) {
...
+      bitWidth(m.dataType).toInt,
+      m.readers.length,
+      m.writers.length,
+      m.readwriters.length,
+      m.maskGran.map(_.toInt)
+    ).toString
+  }.mkString("\n").getBytes
+}
+
+trait HasAnnotatedMemories {
```
jackkoenig, line 100: Why make this a trait (especially unsealed) rather than just methods on the class?

More problematic, I don't love the mutability here. Why not have 2 different annotations (shared functionality in a trait perhaps 😉), 1 before the information is evaluated and 1 after?

src/main/scala/firrtl/passes/memlib/ReplaceMemMacros.scala
```diff
@@ -136,7 +137,11 @@ class ReplaceMemMacros(writer: ConfWriter) extends Transform with DependencyAPIM
     *  The wrapper module has the same type as the memory it replaces
     *  The external module
     */
-  def createMemModule(m: DefAnnotatedMemory, wrapperName: String): Seq[DefModule] = {
+  def createMemModule(
+    m:           DefAnnotatedMemory,
+    wrapperName: String,
+    insertFunc:  DefAnnotatedMemory => Unit
+  ): Seq[DefModule] = {
```
jackkoenig, line 144: As noted in other comments, this isn't binary compatible. But when making such changes, I'd prefer if the new methods are private since in many of these cases, it doesn't seem like a useful public API (making it easier to make such changes in the future).

jackkoenig (changes requested):
I like the overall idea, would like some clarification about the plan: is this just slated for 1.5 in which case we can backport some deprecations and then break APIs or should we instead work on making things binary compatible (honestly, doesn't seem worth it)?

I also really don't like the mutability of `MemLibOutConfigFileAnnotation`

sequencer:
So generally I'd like to make a binary incompatible change in this PR, since `memlib` is pretty old, most of its API are public and too hard to maintain the compatibility here.
I'll make most of functions in these transforms private, and of cause make sure chipyard is not using those APIs.
After finishing those binary incompatible changes, I'll submit for review, and PR another deprecation only PR.

> More problematic, I don't love the mutability here. Why not have 2 different annotations (shared functionality in a trait perhaps wink), 1 before the information is evaluated and 1 after?

I'll create 2 annotations, one annotation for collect `DefAnnotatedMemory` with a `ListBuffer`, and another annotation with another transform to covert this annotation to annotations which will emit files.

src/main/scala/firrtl/passes/memlib/ReplaceMemTransform.scala
sequencer: I removed this API and added `private[memlib] case class AnnotatedMemoriesCollectorAnnotation() extends NoTargetAnnotation`, I don't like mutable neither, but I think in this transform we must have a mutable thing to store those `DefAnnotatedMemory`, during walking the entire circuit: originally it was the config writer, in b385574a, it was `HasAnnotatedMemories`, and in 7f82ea10, it is `AnnotatedMemoriesCollectorAnnotation`, I use `insert` and `done` function to make sure it only can be written during walking the circuit at first time, so I think it's safe here.

sequencer:
@jackkoenig Need review again :)
After you think modifications are ok, I will start to add private keyword to make make those functions as private as possible.

src/main/scala/firrtl/passes/memlib/ReplaceMemMacros.scala
sequencer: I will make these changes 1.5 only and not to backport, after approve, I will start to create another PR to add deprecations o memlib.

src/main/scala/firrtl/passes/memlib/ReplaceMemMacros.scala
```diff
@@ -242,18 +248,24 @@ class ReplaceMemMacros(writer: ConfWriter) extends Transform with DependencyAPIM
...
   }
 
-  def updateMemMods(namespace: Namespace, nameMap: NameMap, memMods: Modules)(m: DefModule) = {
+  def updateMemMods(
+    namespace:  Namespace,
+    nameMap:    NameMap,
+    memMods:    Modules,
+    insertFunc: DefAnnotatedMemory => Unit
+  )(m:          DefModule
+  ) = {
```
jackkoenig, line 265: I think these methods should be private (since we're changing them anyway, might as well take the opportunity).

Also, rather than using `insertFunc` to mutate an Annotation, why not just pass a `ListBuffer[DefAnnotatedMemory]` here and get rid of `AnnotatedMemoriesCollectorAnnotation` altogether? It's all private, I don't think there's any need for that annotation when we could just use the mutable collection directly and construct `MemLibOutConfigFileAnnotation` at the end.

sequencer: > get rid of `AnnotatedMemoriesCollectorAnnotation`

I personal prefer using an standalone Annotation to store this mutable data,so in the `DumpMemoryAnnotations` Transform, we can directly convert this annotation to other annotations, in current case only the configure file, in the next PRs, `InlinedBlackBoxes` can be converted to.
That’s the reason why I add this intermediate annotation.

sequencer: And another question is if we use mutable collection here, when I switch to dependency API, will it be thread safe? As @seldridge said, 
> I think there can be bugs when working in a parallel environment using the same stage. This can happen in tests.

this is the reason I put mutable thing in annotations, which makes transform fully immutable.

sequencer: > I think these methods should be private (since we're changing them anyway, might as well take the opportunity).

Sure, I’ll add private keywords when finishing review for a easy rebasing flow.

ekiwi: > And another question is if we use mutable collection here, when I switch to dependency API, will it be thread safe?

You just need to create one instance of the collection for every call to the `execute` function. Like do not make it a member of the class, instead create it in `execute` and give a reference to it to every function that needs to mutate.

sequencer: Got it, fixed :)

sequencer: Hi, @jackkoenig 
> Also, rather than using insertFunc to mutate an Annotation, why not just pass a ListBuffer[DefAnnotatedMemory] here and get rid of AnnotatedMemoriesCollectorAnnotation altogether? 

Done. I have removed that annotation. But still have the `AnnotatedMemoriesAnnotation` For transform to other annotations. Is that ok? As @ekiwi suggested, mutable is in execute function now. 

> It's all private, I don't think there's any need for that annotation when we could just use the mutable collection directly and construct MemLibOutConfigFileAnnotation at the end.

Since after this PR, I wanna transform to Verilog blackboxes annotations, thus create the intermediate annotation, and consume it in next transform.

jackkoenig (approved):

mwachs5:
@sequencer , the behavior after this change is that `ReplSeqMem` removes the `ReplSeqMemAnnotation`s (replacing them with `MemLibOutConfigFileAnnotation`). Is this intentional? If not we should fix it. If so, we should really put somethign in the release notes to this effect as its a pretty breaking change for anyone using `ReplSeqMemAnnotation`s for other purposes

sequencer:
I thought `ReplSeqMemAnnotations` is not removed? I kept it unchanged, `ReplSeqMemAnnotations` will be convert to `MemLibOutConfigFileAnnotation` in `CreateMemoryAnnotations` pass. What's the different behavior?

sequencer:
btw, XiangShan projects has switched to new memlib pass in this PR: https://github.com/OpenXiangShan/XiangShan/pull/1000/files, hope this helps

PR #2200, opened 2021-04-26
module-name-prefix FIRRTL pass

kenzhang82:
### Contributor Checklist

- [ ] Did you review?


kenzhang82:
This is the PR for creating a new FIRRTL pass to prefix module name to help with the issue [here](https://github.com/chipsalliance/chisel3/issues/1059)

I am trying to implement and test this at the moment, my code looks like the following, note that I am doing it from firrtl repo, so probably worth of creating another issue in there:
```
package firrtl.passes

import firrtl._
import firrtl.options.{Dependency, RegisteredTransform, ShellOption}
import firrtl.transforms.DedupModules
import firrtl.annotations.NoTargetAnnotation
import firrtl.stage.RunFirrtlTransformAnnotation

////////////////
// Pass
////////////////

/** Specifies a global prefix for all module names. */
case class ModulePrefix(prefix: String = "abcdefg") extends NoTargetAnnotation


class PrefixModulesPass extends Transform with DependencyAPIMigration with RegisteredTransform {
  // we run after deduplication to save some work
  override def prerequisites = Seq(Dependency[DedupModules])

  // we do not invalidate the results of any prior passes
  override def invalidates(a: Transform) = false

  val options = Seq(
    new ShellOption[String](
      longOption = "module-name-prefix",
      toAnnotationSeq = (a: String) =>
        Seq(ModulePrefix("abcdefg"), RunFirrtlTransformAnnotation(new PrefixModulesPass)),
      helpText = "Global prefix to every modules",
      shortOption = Some("prefix"),
      helpValueName = Some("<prefix>")
    )
  )

  override protected def execute(state: CircuitState): CircuitState = {
    val prefixes = state.annotations.collect{ case a: ModulePrefix => a.prefix }.distinct
    prefixes match {
      case Seq() =>
        logger.info("[PrefixModulesPass] No ModulePrefix annotation found.")
        state
      case Seq("") => state
      case Seq(prefix) =>
        val c = state.circuit.mapModule(onModule(_, prefix))
        state.copy(circuit = c.copy(main = prefix + c.main))
      case other =>
        throw new PassException(s"[PrefixModulesPass] found more than one prefix annotation: $other")
    }
  }

  private def onModule(m: ir.DefModule, prefix: String): ir.DefModule = m match {
    case e : ir.ExtModule => e.copy(name = prefix + e.name)
    case mod: ir.Module =>
      val name = prefix + mod.name
      val body = onStmt(mod.body, prefix)
      mod.copy(name=name, body=body)
  }

  private def onStmt(s: ir.Statement, prefix: String): ir.Statement = s match {
    case i : ir.DefInstance => i.copy(module = prefix + i.module)
    case other => other.mapStmt(onStmt(_, prefix))
  }
}
```

Then I register this transformation by appending `firrtl.passes.PrefixModulesPass` inside `src/main/resources/META-INF/firrtl.options.RegisteredTransform`.

Then I run `sbt assembly` to compile everything, this compiled without any problem, I can see this CLI option when I do:
```
./utils/bin/firrtl --help
FIRRTL Transform Options
  --no-dce                 Disable dead code elimination
  --no-check-comb-loops    Disable combinational loop checking
  -fil, --inline <circuit>[.<module>[.<instance>]][,...]
                           Inline selected modules
  -clks, --list-clocks -c:<circuit>:-m:<module>:-o:<filename>
                           List which signal drives each clock of every descendent of specified modules
  --no-asa                 Disable assert submodule assumptions
  --no-constant-propagation
                           Disable constant propagation elimination
  -prefix, --module-name-prefix <prefix>
                           Global prefix to every modules
```

However, when I run an example circuit to perform the transformation:
```
./utils/bin/firrtl -td regress -i regress/RocketCore.fir --module-name-prefix abcdefg
```
The output Verilog doesn't have any of the prefixes I specify. Note that I am hardcoding the prefix to start with, it would be great to somehow pass this from the command line. Any idea what is going wrong here? @seldridge @ekiwi @jackkoenig

kenzhang82:
Some debugging shows that FIRRTL doesn't recognize any annotation:
```
[DCE] node MorseJtagModule._tapIO_T_7
[DCE] node AppsMacSystem.debug_reset_syncd
Main module will be renamed. Renaming circuit: '~AppsMacSystem' -> ['~AppsMacSystem']
----------------------------------------------------------

Time: 3229.3 ms
======== Finished firrtl.VerilogEmitter ========
======== Starting firrtl.passes.ModuleNamePrefix ========
[ModuleNamePrefixAnnotation] No ModulePrefix annotation found.
```

How do I pass annotation to test this transform?

kenzhang82:
More progress here, I ran:
```
./utils/bin/firrtl -td regress -i regress/RocketCore.fir -prefix "TestPrefix" --log-level debug
```
and I see:
```
Time: 2891.6 ms
======== Finished firrtl.VerilogEmitter ========
======== Starting firrtl.passes.ModuleNamePrefixTransform ========
[ModuleNamePrefixAnnotation] Actually doing something?
prefix we got is [TestPrefix]
```
So it went to actually transform the circuit I reckon, as per:
```
  override protected def execute(state: CircuitState): CircuitState = {
    val annos = state.annotations.collect{ case a: ModuleNamePrefixAnnotation => a.prefix }.distinct
    annos match {
      case Seq() =>
        logger.info("[ModuleNamePrefixAnnotation] No ModuleNamePrefixAnnotation annotation found.")
        state
      case Seq("") =>
        logger.info("[ModuleNamePrefixAnnotation] Empty string received?")
        state
      case Seq(prefix) =>
        logger.info("[ModuleNamePrefixAnnotation] Actually doing something?")
        println(s"prefix we got is [$prefix]")
        val c = state.circuit.mapModule(onModule(_, prefix))
        state.copy(circuit = c.copy(main = prefix + c.main))
      case other =>
        throw new PassException(s"[ModuleNamePrefixAnnotation] found more than one prefix annotation: $other")
    }
  }
```

However the output verilog doesn't show any difference, any idea? @sequencer @jackkoenig @edwardcwang @ekiwi

kenzhang82:
In case anyone is still wondering if there is a solution, here is transform that Chipyard folks are using for tapeout
https://github.com/ucb-bar/barstools/blob/master/tapeout/src/main/scala/barstools/tapeout/transforms/AddSuffixToModuleNames.scala

I think we should ask Chipyard folks to upstream this. I am closing this PR for now.

sequencer:
Sorry for the late replay.
AFAIK, SiFive and UCB has their own transform for this feature, I personally think this can be upstreamed, if you are interested in making this feature upstreamed, I'm willing to help:)

kenzhang82:
The link I posted above is pretty much the same feature as in this PR, I think it would be better for the maintainers of FIRRTL and the barstool repo to upstream this (it would be much smoother process I reckon).

kenzhang82:
Giving a second thought I think we should take the initiative to upstream this, I do have another similar feature request for the memlib pass to be upstreamed too, I will put them together.

src/main/scala/firrtl/passes/ModuleNamePrefix.scala
```diff
@@ -0,0 +1,57 @@
...
+  override protected def execute(state: CircuitState): CircuitState = {
+    val annos = state.annotations.collect{ case a: ModuleNamePrefixAnnotation => a.prefix }.distinct
+    annos match {
+      case Seq() =>
+        logger.info("[ModuleNamePrefixAnnotation] No ModuleNamePrefixAnnotation annotation found.")
+        state
+      case Seq("") => state
+      case Seq(prefix) =>
+        val c = state.circuit.mapModule(onModule(_, prefix))
+        state.copy(circuit = c.copy(main = prefix + c.main))
```
seldridge, line 52: This needs to return a `RenameMap` with the updated names. Otherwise, annotations will get out of date.

src/main/scala/firrtl/passes/ModuleNamePrefix.scala
```diff
@@ -0,0 +1,57 @@
...
+      helpText = "Add global prefix to every verilog module (default: \"\")",
+      shortOption = Some("prefix"),
+      helpValueName = Some("<prefix>")
+    )
+  )
+
+  private def onStmt(s: ir.Statement, prefix: String): ir.Statement = s match {
+    case i : ir.DefInstance => i.copy(module = prefix + i.module)
+    case other => other.mapStmt(onStmt(_, prefix))
+  }
```
seldridge, line 33: Annoyingly the instance cannot be blindly renamed here. Consider the situation of:

```
circuit Foo:
  module Bar:
    skip
  module Prefix_Bar
    skip
  module Foo:
    inst bar1 of Bar
    inst bar2 of Prefix_Bar
```

This will need to track what previous modules have been renamed in some datastructure (a `RenameMap` would work) and then query that to know how to rename the instances.

src/main/scala/firrtl/passes/ModuleNamePrefix.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  override protected def execute(state: CircuitState): CircuitState = {
+    val annos = state.annotations.collect{ case a: ModuleNamePrefixAnnotation => a.prefix }.distinct
+    annos match {
+      case Seq() =>
+        logger.info("[ModuleNamePrefixAnnotation] No ModuleNamePrefixAnnotation annotation found.")
+        state
+      case Seq("") => state
+      case Seq(prefix) =>
+        val c = state.circuit.mapModule(onModule(_, prefix))
```
seldridge, line 51: Related to the comment on instance renaming... Because FIRRTL allows use-before-def of instances, you can do the correct instance renaming by:
1. Topologically sorting the modules
2. Reversing the sort (so its now leaves to roots)
3. Renaming modules and instances
4. Converting back to the original order

This enables you to rename modules in the circuit namespace and guarantee that you'll rename them before you ever see an instance of them.

There is a related pass for renaming everything in the circuit which shows how to do this dance: https://github.com/chipsalliance/firrtl/blob/master/src/main/scala/firrtl/transforms/ManipulateNames.scala#L449

seldridge (commented):
There are some annoying logistics that you have to do for a pass like this. Comments inline about that.

An alternative way of doing this that I think should be strongly considered is to have a transform that is really just wrapping the existing `ManipulateNames` transform. I think you can do module renaming by building a `ManipulateNamesAllowList` annotation that targets all modules. Then just rely on that to get the renaming correct (and _not_ have to worry about all the logistical headaches associated with name manipulation).

One thing to consider is that some people want prefixes while some people want suffixes. `ManipulateNames` is [already abstract in a `(String, Namespace) => Option[String]` manipulate method](https://github.com/chipsalliance/firrtl/blob/master/src/main/scala/firrtl/transforms/ManipulateNames.scala#L181) which means it's trivial to provide concrete implementations of both. This already lets you build more complicated renaming pretty simply, e.g., [the `LetterCaseTransform` stuff](https://github.com/chipsalliance/firrtl/blob/cd845bdbfea0c09e9edbf61e651ede5197d8a084/src/main/scala/firrtl/features/LetterCaseTransform.scala#L10).

seldridge:
Also, I'd like to get _something_ upstreamed here! This is a pretty fundamental thing. Thanks for hacking on this. :+1:

src/main/scala/firrtl/passes/ModuleNamePrefix.scala
ekiwi: Why is that the case? This transform is written with the assumption that _every_ module gets prefixed. So in this case you would just end up with a circuit like this:

```
circuit Prefix_Foo:
  module Prefix_Bar:
    skip
  module Prefix_Prefix_Bar
    skip
  module Prefix_Foo:
    inst bar1 of Prefix_Bar
    inst bar2 of Prefix_Prefix_Bar
```

src/main/scala/firrtl/passes/ModuleNamePrefix.scala
ekiwi: Since we do not get any name collisions by construction (see my previous comment), the renaming is a completely deterministic/pure function and you should be able to run this is any order. If we want more features than just prefixing every module, we are probably going to end up with something more like the ManipulateNames transform that you mentioned.

src/main/scala/firrtl/passes/ModuleNamePrefix.scala
seldridge: Ack, you're right.

If the transform is blindly doing the prefixing and cannot deviate from it, then it's not an error.

While extreme scope creep, it would be ideal if a transform like this was idempotent. That would require a formulation like what I'm proposing where the instance cannot be renamed blindly.

ekiwi: If we remove the prefix annotation after applying the prefixing and just do nothing if there is no annotation, it would become idempotent, right?

ekiwi:
One question that I think needs to be answered here is:
What is the desired behavior for the toplevel (aka main) module?
- prefix toplevel module
- do not prefix toplevel module
- make this an option

src/main/scala/firrtl/passes/ModuleNamePrefix.scala
seldridge: Fair. It's idempotent from that perspective. However, the intent of this may be that every Verilog module will have a prefix. If a later transform adds a module, that is an invalidation of module prefixing which can't be enforced by re-running with this approach. It's a definition of what this transform actually means. Is it a static application of a prefix or is it a guarantee that modules have a prefix in Verilog emission?

kenzhang82:
Thanks for the advice and comments @seldridge @ekiwi, I will amend the PR accordingly and let you guys know when they are good for review.

src/main/scala/firrtl/passes/ModuleNamePrefix.scala
seldridge: To clarify: I'm good with this as blindly renaming. This would fill an API hole. If we need something more, we can update it in the future.

src/main/scala/firrtl/passes/ModuleNamePrefix.scala
seldridge: Keeping this simple would make sense. Good analysis, @ekiwi.

ekiwi:
@kenzhang82 Could you please run `sbt scalafmtAll` and commit + push any changes? The CI is complaining that the code isn't properly formatted.

kenzhang82:
Yes sure @ekiwi

kenzhang82:
> Yes sure @ekiwi

And done, can you check if it is roughly ok now? Sorry I don't have time yet to amend the PR, do you think we should merge this in or wait for me to amend it further as per the comments above?

ekiwi:
> And done, can you check if it is roughly ok now? Sorry I don't have time yet to amend the PR, do you think we should merge this in or wait for me to amend it further as per the comments above?

I will try to get @seldridge or @jackkoenig to look at this. Since I am not really a user of this API, I am not sure what the requirements are.

PR #2202, opened 2021-04-27
Rewrite vlsi_mem_gen into a Firrtl Transform

sinofp:
1. Add a function called `vlsiMemGen`, most of it is the same as the original Python code except there’s no mutable.
2. `DumpMemoryAnnotations` will call `vlsiMemGen` when it sees `GenVerilogMemBehaviorModelAnno`. It will produce separate files for every sequential memory, using `BlackBoxInlineAnno`.
3. Add a command line option “--gen-mem-verilog” to trigger `GenVerilogMemBehaviorModelAnno`.
4. Add a simple test function to test the Verilog file is emitted. However, it didn’t check the Verilog code because it’s complicated to transform the code back to MemConf.

### Contributor Checklist

#### Type of Improvement

- new feature/API

#### API Impact

This adds `GenVerilogMemBehaviorModelAnno` to extend existing `ReplSeqMem` & `DumpMemoryAnnotations`, add an option to emit Verilog files for memories.

None

#### Desired Merge Strategy

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/passes/memlib/DumpMemoryAnnotations.scala
+++ b/src/main/scala/firrtl/passes/memlib/DumpMemoryAnnotations.scala
@@ -41,7 +41,7 @@ class DumpMemoryAnnotations extends Transform with DependencyAPIMigration {
     val addrWidth = math.max(math.ceil(math.log(depth) / math.log(2)).toInt, 1)
     val readPorts = ports.filter(port => port.portType == ReadPort || port.portType == ReadWritePort)
 
-    def genPortSpec(port: Port): Seq[String] = {
+    val portSpec = ports.flatMap(port =>
       Seq(s"input ${port.prefix}clk", s"input [${addrWidth - 1}:0] ${port.prefix}addr", s"input ${port.prefix}en") ++
         ((port, masked) match {
           case (Port(prefix, ReadPort), _)      => Seq(s"output [${width - 1}:0] ${prefix}data")
@@ -62,10 +62,9 @@ class DumpMemoryAnnotations extends Transform with DependencyAPIMigration {
               s"output [${width - 1}:0] ${prefix}rdata"
             )
         })
-    }
-    val portSpec = ports.flatMap(genPortSpec)
+    )
 
-    def genDecl(): Seq[String] = readPorts.flatMap(port =>
+    val decl = readPorts.flatMap(port =>
       Seq(s"reg reg_${port.prefix}ren;", s"reg [${addrWidth - 1}:0] reg_${port.prefix}addr;")
     ) ++ Seq(
       s"reg [${width - 1}:0] ram [${depth - 1}:0];",
@@ -79,35 +78,34 @@ class DumpMemoryAnnotations extends Transform with DependencyAPIMigration {
       "  end",
       "`endif"
     )
-    val decl = genDecl()
 
-    def genSequential(port: Port): Seq[String] = {
-      def genReadSequential(en: String) = Seq(
-        s"always @(posedge ${port.prefix}clk)",
-        s"  reg_${port.prefix}ren <= $en;",
-        s"always @(posedge ${port.prefix}clk)",
-        s"  if ($en) reg_${port.prefix}addr <= ${port.prefix}addr;"
+    val sequential = {
+      def genReadSequential(en: String, prefix: String): Seq[String] = Seq(
+        s"always @(posedge ${prefix}clk)",
+        s"  reg_${prefix}ren <= $en;",
+        s"always @(posedge ${prefix}clk)",
+        s"  if ($en) reg_${prefix}addr <= ${prefix}addr;"
       )
-      def genWriteSequential(en: String, inputData: String): Seq[String] = Seq(
-        s"always @(posedge ${port.prefix}clk)",
+      def genWriteSequential(en: String, prefix: String, inputData: String): Seq[String] = Seq(
+        s"always @(posedge ${prefix}clk)",
         s"  if ($en) begin"
       ) ++ (0 until maskSeg).map { i =>
-        val mask = if (masked) s"if (${port.prefix}mask[$i]) " else ""
+        val mask = if (masked) s"if (${prefix}mask[$i]) " else ""
         val ram_range = s"${(i + 1) * maskGran - 1}:${i * maskGran}"
-        s"    ${mask}ram[${port.prefix}addr][$ram_range] <= ${port.prefix}$inputData[$ram_range];"
+        s"    ${mask}ram[${prefix}addr][$ram_range] <= ${prefix}$inputData[$ram_range];"
       } ++ Seq("  end")
-
-      port.portType match {
-        case ReadPort  => genReadSequential(port.prefix + "en")
-        case WritePort => genWriteSequential(port.prefix + "en", "data")
-        case ReadWritePort =>
-          genReadSequential(s"${port.prefix}en && !${port.prefix}wmode") ++
-            genWriteSequential(s"${port.prefix}en && ${port.prefix}wmode", "wdata")
-      }
+      ports.flatMap(port =>
+        port.portType match {
+          case ReadPort  => genReadSequential(port.prefix + "en", port.prefix)
+          case WritePort => genWriteSequential(port.prefix + "en", port.prefix, "data")
+          case ReadWritePort =>
+            genReadSequential(s"${port.prefix}en && !${port.prefix}wmode", port.prefix) ++
+              genWriteSequential(s"${port.prefix}en && ${port.prefix}wmode", port.prefix, "wdata")
+        }
+      )
     }
-    val sequential = ports.flatMap(genSequential)
 
-    def genCombinational(port: Port): Seq[String] = {
+    val combinational = readPorts.flatMap { port =>
       val data = port.prefix + (if (port.portType == ReadWritePort) "rdata" else "data")
       Seq(
         "`ifdef RANDOMIZE_GARBAGE_ASSIGN",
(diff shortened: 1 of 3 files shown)


edwardcwang:
FYI, https://github.com/ucb-bar/barstools has some of this already. Another question is whether this should live in FIRRTL core or in a separate utils library.

sequencer:
This transform is a rewrite to [rocket-chip script](https://github.com/chipsalliance/rocket-chip/blob/master/scripts/vlsi_mem_gen), and it's really necessary for rocket-chip [emulation flows](https://github.com/chipsalliance/rocket-chip/blob/c6b41bf5918bbafc7dccee70e208352e029d9505/emulator/Makefrag-verilator#L27-L30).
I didn't find a blackbox generation script in barstool, maybe the closest implementation is [this](https://github.com/ucb-bar/plsi-mdf/blob/4be9b173647c77f990a542f4eb5f69af01d77316/scalalib/src/main/scala/SRAM.scala)?
My intention was removing external dependency to SRAM replace flows, so I prefer that making this one live in FIRRTL.(so no more dependency hell)
Before this PR: FIRRTL generates a memconf file, then pass it to external script(vlsi-mem-gen), and generate a blackbox file.
After this PR: FIRRTL generates these blackbox directly, and we can get rid of that external script.
So I think this PR should be good :)

colinschmidt:
I think keeping this external to allow users to do something like rocket-chip or something like chipyard is better than further tying FIRRTL to rocket-chip.

sequencer:
yes, I also keep original flow intact.
So as you can see this PR is only API addition without any modification to original APIs.
And vlsi-mem-gen is not only useful for rocketchip based projects.

jackkoenig:
While I generally push for things to be external, I think the fact that we have repl-seq-mems but require the user to write their own script or go find one in rocket-chip or chipyard to use it is a bit weird. Having at least some implementation built-in makes sense for better out-of-box experience, but then more complex use cases remain external.

sequencer:
By the way, I think this default implementation should be a good example to other implementations, although I think this implementation still needs some intensive review.

src/main/scala/firrtl/passes/memlib/DumpMemoryAnnotations.scala
```diff
@@ -13,16 +15,149 @@ class DumpMemoryAnnotations extends Transform with DependencyAPIMigration {
...
+      "  integer initvar;",
+      "  initial begin",
+      "    #`RANDOMIZE_DELAY begin end",
+      s"    for (initvar = 0; initvar < $depth; initvar = initvar+1)",
+      s"      ram[initvar] = {${(width - 1) / 32 + 1} {$$random}};"
+    ) ++ readPorts.map(port => s"    reg_${port.prefix}addr = {${(addrWidth - 1) / 32 + 1} {$$random}};") ++ Seq(
+      "  end",
+      "`endif"
+    )
+    val decl = genDecl()
```
sequencer, line 82: inline this `def`

src/main/scala/firrtl/passes/memlib/DumpMemoryAnnotations.scala
```diff
@@ -13,16 +15,149 @@ class DumpMemoryAnnotations extends Transform with DependencyAPIMigration {
...
+          case (Port(prefix, ReadWritePort), true) =>
+            Seq(
+              s"input ${prefix}wmode",
+              s"input [${maskSeg - 1}:0] ${prefix}wmask",
+              s"input [${width - 1}:0] ${prefix}wdata",
+              s"output [${width - 1}:0] ${prefix}rdata"
+            )
+        })
+    }
+    val portSpec = ports.flatMap(genPortSpec)
```
sequencer, line 66: inline this `def`

src/main/scala/firrtl/passes/memlib/DumpMemoryAnnotations.scala
```diff
@@ -13,16 +15,149 @@ class DumpMemoryAnnotations extends Transform with DependencyAPIMigration {
...
+
+      port.portType match {
+        case ReadPort  => genReadSequential(port.prefix + "en")
+        case WritePort => genWriteSequential(port.prefix + "en", "data")
+        case ReadWritePort =>
+          genReadSequential(s"${port.prefix}en && !${port.prefix}wmode") ++
+            genWriteSequential(s"${port.prefix}en && ${port.prefix}wmode", "wdata")
+      }
+    }
+    val sequential = ports.flatMap(genSequential)
```
sequencer, line 108: inline this `def`

src/main/scala/firrtl/passes/memlib/DumpMemoryAnnotations.scala
```diff
@@ -13,16 +15,149 @@ class DumpMemoryAnnotations extends Transform with DependencyAPIMigration {
...
+        "  end",
+        "`endif",
+        s"always @(posedge ${port.prefix}clk) ${port.prefix}random <= {${Seq.fill((width - 1) / 32 + 1)("$random").mkString(", ")}};",
+        s"assign $data = reg_${port.prefix}ren ? ram[reg_${port.prefix}addr] : ${port.prefix}random[${width - 1}:0];",
+        "`else",
+        s"assign $data = ram[reg_${port.prefix}addr];",
+        "`endif"
+      )
+    }
+    val combinational = readPorts.flatMap(genCombinational)
```
sequencer, line 129: inline this `def`

src/main/scala/firrtl/passes/memlib/DumpMemoryAnnotations.scala
```diff
@@ -13,16 +15,149 @@ class DumpMemoryAnnotations extends Transform with DependencyAPIMigration {
...
+
+    val body = if (genBlackBox) "" else s"""
+                                           |  ${decl.mkString("\n  ")}
+                                           |  ${sequential.mkString("\n  ")}
+                                           |  ${combinational.mkString("\n  ")}""".stripMargin
```
sequencer, line 134: I think `decl`, `sequential`, `combinational` can be `String` rather than `Seq[String]`

src/main/scala/firrtl/passes/memlib/DumpMemoryAnnotations.scala
```diff
@@ -13,16 +15,149 @@ class DumpMemoryAnnotations extends Transform with DependencyAPIMigration {
...
+
+    (name, s"""
+              |module $name(
+              |  ${portSpec.mkString(",\n  ")}
+              |);
+              |
+              |$body
+              |
+              |endmodule""".stripMargin)
```
sequencer, line 143: Add Verilog comments in the header, since original `config` is not elaborated, need contain `DefAnnotatedMemory` informations

src/main/scala/firrtl/passes/memlib/ReplaceMemTransform.scala
```diff
@@ -131,6 +134,20 @@ class ReplSeqMem extends SeqTransform with HasShellOptions with DependencyAPIMig
...
+      longOption = "gen-mem-verilog",
+      toAnnotationSeq = (a: String) =>
+        Seq(
+          a match {
+            case "blackbox" => GenVerilogMemBehaviorModelAnno(genBlackBox = true)
+            case _          => GenVerilogMemBehaviorModelAnno(genBlackBox = false)
+          },
+          RunFirrtlTransformAnnotation(new ReplSeqMem)
+        ),
+      helpText = "Blackbox and emit a Verilog file for each sequential memory",
```
sequencer, line 148: Blackbox and emit a Verilog behavior model for each sequential memory.

src/test/scala/firrtlTests/ReplSeqMemTests.scala
```diff
@@ -45,6 +45,22 @@ class ReplSeqMemSpec extends SimpleTransformSpec {
...
+      val annos = Seq(GenVerilogMemBehaviorModelAnno(genBlackBox = genBlackBox)) ++ additionalAnnos
+      val res = compileAndEmit(CircuitState(parse(input), ChirrtlForm, annos))
+      // Check correctness of firrtl
+      parse(res.getEmittedCircuit.value)
+      // Check the emitted Verilog
+      mems.foreach { mem =>
+        val file = new java.io.File(mem.name + ".v")
+        require(file.exists(), s"${file.getName} should be emitted!")
+        // todo check content (using verilator?)
+        file.delete()
```
sequencer, line 59: compiling by verilator to check verilog is correct.

sequencer (changes requested):

src/main/scala/firrtl/passes/memlib/DumpMemoryAnnotations.scala
sinofp: I think changing them to `String` will break the consistency of the code. `portSpec` and `decl` are all generated lines, why should one be `Seq[String]`, and another be `String`?

src/main/scala/firrtl/passes/memlib/ReplaceMemTransform.scala
sinofp: Fixed in [`e0859e8` (#2202)](https://github.com/chipsalliance/firrtl/pull/2202/commits/e0859e84b3f507f47390b5b6838cb1b3fab45cdb)

src/test/scala/firrtlTests/ReplSeqMemTests.scala
sinofp: Is `--lint-only` enough? I see they using it in `lintVerilog`. If it is enough, then [`e0859e8` (#2202)](https://github.com/chipsalliance/firrtl/pull/2202/commits/e0859e84b3f507f47390b5b6838cb1b3fab45cdb) fixed this.

src/main/scala/firrtl/passes/memlib/DumpMemoryAnnotations.scala
sinofp: Fixed in e0859e8 (#2202)

sequencer (approved):

sequencer:
Test successfully with RC, let's ship it.

sequencer:
Congratulations to @sinofp first PR!

PR #2212, opened 2021-05-03
Implement MFC-style source locator compression

jared-barocsi:
#### Type of Improvement

code refactoring

#### API Impact

No impact

This only affects the commented source locators in the generated Verilog code. No functional impact.

#### Desired Merge Strategy

Squash

#### Release Notes

Implement MFC-style locator compression in the Scala FIRRTL Compiler
- This is to reduce the redundancy in file info comments (// @[source.file line:col]), by merging together
  locators that share the same file and line numbers.
- For example, long comments like `// @[A 23:45 A 83:44 A 83:72 A 102:23 B 18:1 B 26:12]` may be compressed
  into a more readable comment like `// @[A 23:45 83:{44,72} 102:23 B 18:1 26:12]`.
- For now, this algorithm is only attempted whenever source locators strictly follow the `source.file line:col` format.

- [ ] Did you review?


seldridge:
@jared-barocsi: This looks cool. Can you run an `sbt scalafmtAll` on this? (The reason why CI is failing.)

ekiwi (changes requested):
Thanks for this PR. There are a couple of things that need to be done, before we can merge it:

1) this needs a couple of unit-tests that fail without this feature and pass once you add it.
2) it seems like this would actually fail if you use the source locator format that we currently have in the firrtl spec:
Screenshot from 2021-05-04 09-41-05

So you would at least have to update the spec to contain the format that you are assuming here and clarify that this format will be required from now on.

On that note: it there a reason you decided to just parse the string inside the `split` function? Are you expecting a use-case, where people use "non-conforming" file-info strings and just never call the `split` function? If we start requiring a certain format for file-info strings (which I am in favor of!) we might want to consider doing that parsing as part of the firrtl-ir parser and then we could keep that information as `file`, `line`, `column` inside the IR node.
What do you think?

jared-barocsi:
@ekiwi:

My impression with this compression algorithm is that it would be another step to conduct when emitting Verilog in particular. The issue was introduced to me as the existing .v output containing a lot of redundancy in its source locators, as compared to another approach that tries to reduce this redundancy as much as possible. As such, I assumed that all of these file-infos will always be in the form of `{file.scala} {line number}:{column number}` because they would be automatically generated while compiling Scala code into Verilog.

As for `split`, I mainly wanted a clean-looking function that would split the file name, line number and column number from the String and return these as a tuple, and it made sense for it to be entirely contained within `FileInfo`. But I agree that this issue would be better addressed with a proper data structure rather than parsing a wrapped String. Perhaps I should do some work in a separate PR to upgrade the firrtl spec and turn `FileInfo` into the data structure we would want first?

ekiwi:
> As such, I assumed that all of these file-infos will always be in the form of `{file.scala} {line number}:{column number}` because they would be automatically generated while compiling Scala code into Verilog.

We would like to be able to convert any "valid" firrtl to Verilog. So if your `flatten` function gets called by default when emitting Verilog, that would probably fail.

One good way to test this would be to add a unittest that takes the example firrtl from the spec that I posted above and to just translate it to Verilog. That way we can see if your new implementation of `flatten` will blow up on that kind of firrtl.

jared-barocsi:
@ekiwi, I've made some changes to how greedily the algorithm tries to compress these FileInfos. Now, it only does it to any FileInfo which matches the "file line:col" format, and otherwise it simply passes over any other FileInfos that don't meet that criteria. Thus, all valid firrtl code should compile now - for instance, I've added a test that uses the FileInfo format in the firrtl spec, and the algorithm is capable of ignoring it.

Most of this work might end up being technical debt as I think FileInfos having their own structure makes sense.

ekiwi (changes requested):
Looks good to me, but before we merge this I have a question:

1) Is this compression supposed to only be used in the Verilog emission, or also when serializing Firrtl?
  a) if the answer is no, it would be great if you could add a test that makes sure that when serializing to Firrtl, the file info is not changed
  b) if the answer is yes, then you need to make sure that your algorithm can deal with its own output since someone might serialize to firrtl, parse the generated firrtl string and then try to generate Verilog

jared-barocsi:
I believe the algorithm acts whenever the firrtl gets serialized in some way. I've added a test to simply check if the algorithm ignores its own output, which it should if it doesn't match the regular expression.

ekiwi (approved):
LGTM

ekiwi:
>     1. a) if the answer is no, it would be great if you could add a test that makes sure that when serializing to Firrtl, the file info is not changed

turns out that the info is actually changed when serializing: https://github.com/chipsalliance/firrtl/pull/2253

PR #2240, opened 2021-05-21
WiringTransform: cannot run after RemoveWires

ekiwi:
The problem this is trying to address is that I sometimes get pass orderings where `RemoveWires` runs before the `Wiring` transform and that gets a wiring target turned from a `Wire` into a `Node` which `Wiring` then fails to deal with properly.

### Contributor Checklist

#### Type of Improvement
 - bug fix

#### API Impact

- none

- none
#### Desired Merge Strategy

- squash

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/passes/wiring/WiringTransform.scala
+++ b/src/main/scala/firrtl/passes/wiring/WiringTransform.scala
@@ -40,7 +40,9 @@ class WiringTransform extends Transform with DependencyAPIMigration {
 
   override def prerequisites = Forms.MidForm
   override def optionalPrerequisites = Seq.empty
-  override def optionalPrerequisiteOf = Forms.MidEmitters
+  override def optionalPrerequisiteOf = Forms.MidEmitters ++
+    // once wire targets are turned into nodes, our logic to wire them up no longer works
+    Seq(Dependency[firrtl.transforms.RemoveWires])
 
   private val invalidates = Forms.VerilogOptimized.toSet -- Forms.MinimalHighForm
   override def invalidates(a: Transform): Boolean = invalidates(Dependency.fromTransform(a))


ekiwi:
This should fix https://github.com/chipsalliance/firrtl/issues/2208

src/main/scala/firrtl/passes/wiring/WiringTransform.scala
```diff
@@ -39,7 +39,8 @@ case class SinkAnnotation(target: Named, pin: String) extends SingleTargetAnnota
 class WiringTransform extends Transform with DependencyAPIMigration {
 
   override def prerequisites = Forms.MidForm
-  override def optionalPrerequisites = Seq.empty
+  // once wire targets are turned into nodes, our logic to wire them up no longer works
+  override def optionalPrerequisites = Seq(Dependency[firrtl.transforms.RemoveWires])
   override def optionalPrerequisiteOf = Forms.MidEmitters
```
seldridge, line 44: I think this is supposed to be added to `optionalPrerequisiteOf` and not `optionalPrerequisites`? (The idea is to inject a prerequisite of `WiringTransform` into `RemoveWires` so that the `WiringTransform` runs before `RemoveWires`, right?)

ekiwi: oh yeah, sorry

ekiwi: (surprised that all our tests still pass)

seldridge (approved):
Looks great! Thanks for fixing this, @ekiwi.

ekiwi:
Thanks! Let's not backport this, just in case it has unintended consequences or breaks someone's flow.

PR #2243, opened 2021-05-21
Replace mem macros renaming

albertchen-sifive:
### Contributor Checklist

#### Type of Improvement

- bug fix

This change makes `ReplaceMemMacros` rename replaced memory ReferenceTargets to InstanceTargets that target the memory blackbox e.g. `~Top|Mod>mem -> ~Top|Mod/mem:mem/mem_ext:mem_ext`.

Chisel emits ReferenceTargets for memories. and because `ReplaceMemMacros` didn't update these ReferenceTargets to instance targets when it replaced them, these targets would never get updated because downstream transforms would only record InstanceTarget renames instead of ReferenceTarget renames.

#### API Impact

no API changes, but does make `ResolveMemoryReference` delete `NoDedupMemAnnotation`s after it runs. This is needed because `SingleTargetAnnotation[ComponentName]` will throw an exception if its target is renamed to an instance. i think ideally it would extend `SingleTargetAnnotation[IsComponent]` instead which would allow it to be renamed from a reference to instance, but that would break the existing api.

none

#### Desired Merge Strategy

- Squash

#### Release Notes

`ReplaceMemMacros` will now rename replaced memory `ReferenceTarget`s to `InstanceTarget`s that target the memory blackbox.

- [ ] Did you review?

### DIFF
--- a/src/test/scala/firrtlTests/ReplSeqMemTests.scala
+++ b/src/test/scala/firrtlTests/ReplSeqMemTests.scala
@@ -14,8 +14,12 @@ import firrtl.util.BackendCompilationUtilities.loggingProcessLogger
 import scala.sys.process._
 
 object ReplSeqMemSpec {
-  private case class DummyAnnotation(target: Target) extends SingleTargetAnnotation[Target] {
-    override def duplicate(n: Target): Annotation = DummyAnnotation(n)
+  private case class DummyAnno(targets: CompleteTarget*) extends Annotation {
+    override def update(renames: RenameMap): Seq[Annotation] = {
+      Seq(DummyAnno(targets.flatMap { t =>
+        renames.get(t).getOrElse(Seq(t))
+      }: _*))
+    }
   }
 }
 
@@ -642,12 +646,15 @@ circuit CustomMemory :
     val confLoc = "ReplSeqMemTests.confTEMP"
     val annos = Seq(
       ReplSeqMemAnnotation.parse("-c:CustomMemory:-o:" + confLoc),
-      DummyAnnotation(CircuitTarget("CustomMemory").module("CustomMemory").ref("mem_0")),
-      DummyAnnotation(CircuitTarget("CustomMemory").module("CustomMemory").ref("mem_1"))
+      DummyAnno(
+        CircuitTarget("CustomMemory").module("CustomMemory").ref("mem_0"),
+        CircuitTarget("CustomMemory").module("CustomMemory").ref("mem_1")
+      )
     )
     val res = compileAndEmit(CircuitState(parse(input), ChirrtlForm, annos))
-    (res.annotations.collect {
-      case DummyAnnotation(t) => t
+    (res.annotations.flatMap {
+      case a: DummyAnno => a.targets
+      case _ => Seq.empty
     }) should be(
       Seq(
         CircuitTarget("CustomMemory")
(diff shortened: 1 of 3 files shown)


albertchen-sifive:
does the `--` method on `LinkedHashSet` not preserve order in 2.13.4? was seeing an error where my test failed on 2.13.4 but not 2.12.13 because some annos got reordered https://github.com/chipsalliance/firrtl/runs/2659521855. changing `resSet -- inSet` to `resSet.filter(!inSet(_))` here https://github.com/chipsalliance/firrtl/blob/master/src/main/scala/firrtl/Compiler.scala#L245 fixes the reordering issue for me locally.  I changed my test anno in this PR from `DummyAnno(target: Target)` to `DummyAnno:(targets: Seq[Target])` as a workaround.

jackkoenig:
> does the -- method on LinkedHashSet not preserve order in 2.13.4? 

If it doesn't (and you make a compelling case) that is pretty concerning.

src/test/scala/firrtlTests/ReplSeqMemTests.scala
```diff
@@ -608,4 +619,53 @@ circuit Top :
...
+    read mport _T_17 = mem_0[io.rAddr], clock
+    read mport _T_19 = mem_1[io.rAddr], clock
+    io.dO <= and(_T_17, _T_19)
+    when io.wEn :
+      write mport _T_18 = mem_0[io.wAddr], clock
+      write mport _T_20 = mem_1[io.wAddr], clock
+      _T_18 <= io.dI
+      _T_20 <= io.dI
+      skip
+"""
```
jackkoenig, line 642:
```suggestion
    val input = """
      |circuit CustomMemory :
      |  module CustomMemory :
      |    input clock : Clock
      |    input reset : UInt<1>
      |    output io : {flip rClk : Clock, flip rAddr : UInt<3>, dO : UInt<16>, flip wClk : Clock, flip wAddr : UInt<3>, flip wEn : UInt<1>, flip dI : UInt<16>}

      |    io is invalid
      |    smem mem_0 : UInt<16>[7]
      |    smem mem_1 : UInt<16>[7]
      |    read mport _T_17 = mem_0[io.rAddr], clock
      |    read mport _T_19 = mem_1[io.rAddr], clock
      |    io.dO <= and(_T_17, _T_19)
      |    when io.wEn :
      |      write mport _T_18 = mem_0[io.wAddr], clock
      |      write mport _T_20 = mem_1[io.wAddr], clock
      |      _T_18 <= io.dI
      |      _T_20 <= io.dI
      |""".stripMargin
```

src/test/scala/firrtlTests/ReplSeqMemTests.scala
```diff
@@ -608,4 +619,53 @@ circuit Top :
...
+        CircuitTarget("CustomMemory")
+          .module("CustomMemory")
+          .instOf("mem_0", "mem_0")
+          .instOf("mem_0_ext", "mem_0_ext"),
+        CircuitTarget("CustomMemory")
+          .module("CustomMemory")
+          .instOf("mem_1", "mem_0")
+          .instOf("mem_0_ext", "mem_0_ext")
+      )
+    )
```
jackkoenig, line 669:
```suggestion
    val resAnnos = res.annotations.collect { case a: DummyAnno => a.targets }
    val expected = Seq(
      CircuitTarget("CustomMemory")
        .module("CustomMemory")
        .instOf("mem_0", "mem_0"
        .instOf("mem_0_ext", "mem_0_ext"),
      CircuitTarget("CustomMemory")
        .module("CustomMemory")
        .instOf("mem_1", "mem_0")
        .instOf("mem_0_ext", "mem_0_ext")
    )
    resAnnos should be(expected)
```

jackkoenig (approved):
This makes me a little nervous as a backport but I do agree it's an important bug fix, so I think we should just do it.

PR #2244, opened 2021-05-21
Optimize Annotation.getTargets

jackkoenig:
Inspired by https://github.com/chipsalliance/firrtl/pull/2241

Let's make sure `.getTargets` doesn't stack overflow in practice and also be faster.

### Contributor Checklist

#### Type of Improvement

 - performance improvement  

#### API Impact

No impact

No impact

#### Desired Merge Strategy

  - Squash
 
#### Release Notes

Improve performance of `Annotation.getTargets`

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/annotations/Annotation.scala
+++ b/src/main/scala/firrtl/annotations/Annotation.scala
@@ -5,6 +5,8 @@ package annotations
 
 import firrtl.options.StageUtils
 
+import scala.collection.Traversable
+
 case class AnnotationException(message: String) extends Exception(message)
 
 /** Base type of auxiliary information */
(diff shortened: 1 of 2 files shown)


src/main/scala/firrtl/annotations/Annotation.scala
```diff
@@ -23,25 +25,28 @@ trait Annotation extends Product {
     * @param ls
     * @return
     */
-  private def extractComponents(ls: scala.collection.Traversable[_]): Seq[Target] = {
-    ls.collect {
+  private def extractComponents(ls: Traversable[_]): Traversable[Target] = {
```
seldridge, line 28: This could use an `@tailrec` decorator to ensure that it's working correctly.

seldridge (approved):

src/main/scala/firrtl/annotations/Annotation.scala
seldridge: Errr... I guess that won't work given the flatmap is doing work after the recursive call. Disregard.

jackkoenig: Yeah it's not actually tail recursive, this could technically stack overflow it would just be very hard, you'd have to have something like `case class MyAnno(xs: List[List[List[List[...]]]]) extends Annotation { ...`. Still possible so I'll soften my original "never stack overflows", but this is *much* faster and won't stack overflow on a simple `List[...]`.

jackkoenig:
Lol it StackOverflowed in Scala 2.13.4... wat.

ekiwi (approved):
This is nice! (assuming it does not overflow again on Scala `2.13`)

jackkoenig:
> This is nice! (assuming it does not overflow again on Scala 2.13)

I think it was due to the deprecation of the methods to convert from `Iterator` to `Iterable` (and questionable implementations of those methods). I worked around the issue by moving `Traversable` above `Product` (so that `List` hits there instead of on `Product` which will be a perf boost for the entire Scala collections library anyway). This isn't _great_ but it works. I can't really do better and backport this because Scala 2.11 support on 1.4.x, can do better on master though.

PR #2278, opened 2021-06-21
Fix VerilogMemDelays use before declaration

jackkoenig:
The pass injects pipe registers immediately after the declaration of the
memory. This can be problematic if the clock for the associated memory
port is defined after the declaration of the memory. For any memory port
clocks that are driven by non-ports, we now inject a wire before the
pipe register declarations to be sure there are no
use-before-declaration issues.

### Contributor Checklist

#### Type of Improvement

   - bug fix  

#### API Impact

No impact

This may reorder emitted Verilog for pipe registers in VerilogMemDelays--however it likely won't since registers are emitted separately in Verilog anyway.

#### Desired Merge Strategy

- Squash

#### Release Notes

Fix FIRRTL crash when VerilogMemDelays is used with memories that use clocks declared after the memory

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/passes/memlib/VerilogMemDelays.scala
+++ b/src/main/scala/firrtl/passes/memlib/VerilogMemDelays.scala
@@ -56,6 +56,14 @@ object MemDelayAndReadwriteTransformer {
   private val metaChars = raw"[\[\]\.]".r
   private def flatName(e: Expression) = metaChars.replaceAllIn(e.serialize, "_")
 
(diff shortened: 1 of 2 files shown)


seldridge (approved):
Fix looks good. Nice test case (super easy to understand what the problem is).

jackkoenig:
@seldridge fairly substantial change, now instead of just putting the registers at the end, we inject a Wire for the clock iff the clock for the associated memory port is driven by non-port-kind things.

seldridge (approved):

PR #2286, opened 2021-06-30
Fix memory annotation deduplication

jared-barocsi:
### Contributor Checklist

#### Type of Improvement
- Bug fix
- Backend code generation

#### API Impact

No changes to any API

`MemoryInitAnnotations` and its subclasses are 're-localized' after module deduplication, which will result in the correct Verilog being emitted for memories with these annotations.

#### Desired Merge Strategy

Squash and merge

- [ ] Did you review?


src/main/scala/firrtl/transforms/DedupAnnotations.scala
```diff
@@ -0,0 +1,131 @@
...
+
+/** Deduplicates memory annotations
+  */
+class DedupAnnotationsTransform extends Transform with DependencyAPIMigration {
+
+  override def prerequisites = firrtl.stage.Forms.LowForm
+
+  override def optionalPrerequisites = Nil
+
+  override def optionalPrerequisiteOf = firrtl.stage.Forms.BackendEmitters
```
seldridge, line 121: I think this actually has no `prerequisites` and no `optionalPrerequisiteOf`.

src/main/scala/firrtl/stage/Forms.scala
```diff
@@ -84,7 +84,8 @@ object Forms {
       Dependency(passes.PadWidths),
       Dependency(passes.memlib.VerilogMemDelays),
       Dependency(passes.SplitExpressions),
-      Dependency[firrtl.transforms.LegalizeAndReductionsTransform]
+      Dependency[firrtl.transforms.LegalizeAndReductionsTransform],
+      Dependency[firrtl.transforms.DedupAnnotationsTransform]
```
seldridge, line 88: I think this should actually be part of `ChirrtlForm` and `EliminateTargetPaths` invalidates it. Also, any transform which implements the `ResolvedAnnotationPaths` trait would now invalidate annotation deduplication.

Does that make sense to you?

seldridge (commented):
Awesome. Thanks for tackling this, @jared-barocsi!

I have some thoughts about the dependencies for this transform included inline.

src/main/scala/firrtl/stage/Forms.scala
jared-barocsi: Yes, that does make sense. I noticed that the actually deduplication passes themselves were in `HighForm`, but I was previously pointed to `VerilogMinimumOptimized` being where I would put this pass. Clearly that approach ends up breaking some tests, along with `LowFormMinimumOptimized`, so I'll try and get some more clarification on where it should belong.

seldridge: Also, any change to `Forms.scala` will cause the `LoweringCompilersSpec` to break. That is a super-legacy thing that is hard-coding the transform order _before_ the Dependency API was added. That test needs to be updated if anything in `Forms.scala` changes or if dependencies change. 

(It's expected that that will break with the change you made here! That will need a patch.)

jared-barocsi: It seems that adding `DedupAnnotationsTransform` to `Deduped` which is then appended to `HighForm` does not require a change to `LoweringCompilersSpec`

src/main/scala/firrtl/transforms/DedupAnnotations.scala
```diff
@@ -0,0 +1,131 @@
...
+    // For every annotation, search for an appropriate group to append it to.
+    annotations.map(annotation => {
+      // If we previously added an annotation to the empty group, append a new empty group to the list
+      if (annotationCheckGroups.tail.nonEmpty) annotationCheckGroups = annotationCheckGroups :+ Seq()
+
+      // Append the annotation to the first matching group (this can be the empty group itself)
+      annotationCheckGroups = annotationCheckGroups.map(group =>
+        if (annotationBelongsInGroup(annotation, group)) group :+ annotation else group
+      )
+
```
jackkoenig, line 92: This logic is a little mindbending (and n^3 complexity where `n` is the number of annotations involved in this deduplication--to be fair, would be n^2 if it weren't for the list append on `annotationCheckGroups`). I'm also hoping that the code we end up here is easier to generalize.

I'd prefer an approach that is sort of in the spirit of [`groupBy`](https://www.scala-lang.org/api/2.12.13/scala/collection/Seq.html#groupBy[K](f:A=%3EK):scala.collection.immutable.Map[K,Repr]). I would first iterate on all annotations to see which ones "can dedup", then group them to see which "should dedup", then finally determine if they "do dedup". "do dedup" also needs an `InstanceGraph` to determine that an annotation applies to *all* instances of the module before actually doing the deduplication (otherwise it would be illegal), and we need a test for this, eg. 2 out of 3 instances have an annotation that dedups and the 3rd doesn't have an annotation at all, can't dedup in that case.

I'm thinking something like:

```scala
// Utility class for deduping annotations
// Deduped should == any other version of the same annotation 
// This type isn't quite right but it's something like this
private case class AnnoDedup[A, B <: Annotation](shouldDedupKey: A, deduped: B, orig: B)

// Does this utility already exist?
private def toLocal(rt: ReferenceTarget): ReferenceTarget =
  if (rt.isLocal) rt else rt.copy(module = rt.encapsulatingModule, rt.path = Nil)

// TODO: All of this can be folded into the annotation types
// May also apply to generic annotations: add *something* to the Annotation trait class
// Returns `Option` because it isn't implemented for all annotations
// Turns an Annotation into a Dedupable representation
def dedupable(annotation: Annotation): Option[AnnoDedup] = annotation match {
  // Example of what the implementation looks like for a particular annotation
  case a @ MemoryScalarInitAnnotation(target, value) =>
    val localTarget = toLocal(target)
    val shouldDedup = (localTarget, value)
    val deduped = a.copy(target = localTarget)
    AnnoDedup(shouldDedup, deduped, a)
  case _ => None
}

// ***** Step 1. Can Dedup *****
val resultAnnos = mutable.ArrayBuffer.empty[Annotation] // Annotations that aren't involved in the dedup
val canDedup = mutable.ArrayBuffer.empty[AnnoDedup]
annotations.foreach { anno =>
  dedupable(anno) match {
    case None => otherAnnos += anno
    case Some(d) => canDedup += d
  }
}

// ***** Step 2. Should Dedup *****
// This is a sketch, probably wrong
val groups = canDedup.groupBy(_.shouldDedupKey)
                     .mapValues(_.map(_._2)) // boilerplate due to lack of .groupMap in Scala 2.11 and 2.12

// ***** Step 3. Does Dedup *****
// This is a sketch, probably wrong, also need an InstanceGraph check
val groups.map { case (key, values) => 
  val deduped = values.map(_.deduped).distinct
  if (deduped.size == 1 && somehowCheckItsTheSameForAllInstancesOfTheSameModule) {
    resultAnnos += deduped.head
  } else {
    resultAnnos += values.map(_.orig)
  }
}
```

jackkoenig (commented):
This is a good start, thanks for the working test!

I want to take this in a slightly different direction, my suggestion comment ended up getting really long so it might be helpful to meet to discuss it.

src/main/scala/firrtl/annotations/MemoryInitAnnotation.scala
```diff
@@ -25,20 +26,29 @@ case class MemoryRandomInitAnnotation(target: ReferenceTarget) extends MemoryIni
...
 }
 
 /** Initialize all entries of the `target` memory with the scalar `value`. */
 case class MemoryScalarInitAnnotation(target: ReferenceTarget, value: BigInt) extends MemoryInitAnnotation {
   override def duplicate(n: ReferenceTarget): Annotation = copy(n)
   override def initValue:    MemoryInitValue = MemoryScalarInit(value)
   override def isRandomInit: Boolean = false
+  override def dedup: Option[((ReferenceTarget, Any), Annotation, ReferenceTarget)] = Some(
+    ((Utils.localizeTarget(target), value), copy(target = Utils.localizeTarget(target)), target)
+  )
```
jackkoenig, line 41: I think these overrides need to keep it private:
```suggestion
  override private[firrtl] def dedup: Option[((ReferenceTarget, Any), Annotation, ReferenceTarget)] = Some(
    ((Utils.localizeTarget(target), value), copy(target = Utils.localizeTarget(target)), target)
  )
```

src/main/scala/firrtl/annotations/Annotation.scala
```diff
@@ -38,6 +38,18 @@ trait Annotation extends Product {
...
+    * deduplicated annotation's "dedup key", the deduplicated [[Annotation]], and the
+    * [[firrtl.annotations.ReferenceTarget ReferenceTarget]](s) to the annotated objects.
+    *
+    * If two absolute instances of this [[Annotation]] would deduplicate to the same
+    * local form, both of their "dedup key"s must be equivalent.
+    *
+    * Returning None signifies this annotation will not deduplicate.
+    * @return
+    */
+  private[firrtl] def dedup: Option[((ReferenceTarget, Any), Annotation, ReferenceTarget)] = None
```
jackkoenig, line 52:
```suggestion
  private[firrtl] def dedup: Option[(Any, Annotation, ReferenceTarget)] = None
```

Since `(ReferenceTarget, Any)` is itself an `Any`, I think the key should just be `Any`.

src/main/scala/firrtl/transforms/DedupAnnotations.scala
```diff
@@ -0,0 +1,102 @@
...
+    absoluteTarget: ReferenceTarget)
+  private object DedupableRepr {
+    def apply(annotation: Annotation): Option[DedupableRepr] = annotation.dedup match {
+      case Some((dedupKey, dedupedAnno, absoluteTarget)) =>
+        Some(new DedupableRepr(dedupKey, dedupedAnno, annotation, absoluteTarget))
+      case _ => None
+    }
+  }
+
+  type InstancePath = Seq[(TargetToken.Instance, TargetToken.OfModule)]
```
jackkoenig, line 43:
```suggestion
  private type InstancePath = Seq[(TargetToken.Instance, TargetToken.OfModule)]
```

src/test/scala/firrtlTests/MemoryInitSpec.scala
```diff
@@ -213,13 +214,127 @@ class MemInitSpec extends FirrtlFlatSpec {
...
+
+  "MemoryFileInlineAnnotation" should "emit $readmemh in deduped module" in {
+    val annos = Seq(
+      MemoryFileInlineAnnotation(child1MRef, filename = "text.hex"),
+      MemoryFileInlineAnnotation(child2MRef, filename = "text.hex")
+    )
+    val result = compile(dedupTest, annos)
+    result should containLine("""$readmemh("text.hex", """ + dedupedRef.name + """);""")
+  }
 }
```
jackkoenig, line 338: I'd like to see a negative test for when not all of the instances have the annotation and the annotations thus don't dedup. I don't think you need it for every single annotation type, but for at least one of them.

jackkoenig (changes requested):
This is really great, excellent work @jared-barocsi!

I left some comments inline and 1 more thing I'd like to see is the Verilog Emitter to error if the annotations were unable to be deduped. I think that logic would live here: https://github.com/chipsalliance/firrtl/blob/1f7faaa36816fc145a39bb7e736fd5ab739112c4/src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala#L465

src/main/scala/firrtl/Utils.scala
```diff
@@ -355,6 +355,9 @@ object Utils extends LazyLogging {
     onExp(expression)
     ReferenceTarget(main, module, Nil, ref, tokens.toSeq)
   }
+  def localizeTarget(target: ReferenceTarget) =
+    if (target.isLocal) target else target.copy(module = target.encapsulatingModule, path = Nil)
+
```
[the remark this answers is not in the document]
davidbiancolin: +1 for moving into RT

src/main/scala/firrtl/annotations/Annotation.scala
davidbiancolin: Should this be private to firrtl? I may be wrong but it seems to me to be pretty important to annotation writers.

It would be useful to have a generic mixin that automagically defines this, by grabbing all the parameters of the case class and filtering out the target.

src/main/scala/firrtl/Utils.scala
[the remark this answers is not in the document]
davidbiancolin: Actually, is this not just `pathlessTarget`?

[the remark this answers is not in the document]
jackkoenig: In fact it is, @jared-barocsi can you just use `.pathlessTarget` and maybe add some ScalaDoc to that describing what it does with an example?

src/main/scala/firrtl/annotations/Annotation.scala
jackkoenig: I totally agree this could be important, I don't want it to be public *yet*. I mainly want to get this fix in and released as 1.4.4 while we discuss what the user-facing API should be.

jackkoenig (approved):
LGTM! Great work!

kammoh:
The problem I reported [here](https://matrix.to/#/!jSbgpvvpsPbLpsLZOC:matrix.org/$iC4I7jhwt-26Xu8AL9SVF1FPRew41nQHWmvGmWXZ0xg?via=matrix.org&via=gitter.im) doesn't seem to be resolved. 

`AttributeAnnotation` disappears when more than one instance of the module `Ram1R1W` exist:
https://scastie.scala-lang.org/epf6LnZkRNiXxq0vYbez4Q

@jackkoenig Do you think it could be a different issue?

jared-barocsi:
@kammoh, `AttributeAnnotation` needs a deduplication implementation of its own. I can follow this up with a secondary PR to resolve the issue.

seldridge:
Would you be able to provide a default implementation for all `SingleTargetAnnotation` (and possibly `MultiTargetAnnotation`)?  That should catch most of this stuff.

jared-barocsi:
> Would you be able to provide a default implementation for all `SingleTargetAnnotation` (and possibly `MultiTargetAnnotation`)? That should catch most of this stuff.

I don't think this is possible because the tuple ('dedup key') that's used to sort all of the annotations into deduplicatable groups contains the annotation's value, and that different across implementations of `SingleTargetAnnotation` (for example, `MemoryFileInlineAnnotation` has a `filename`, `MemoryArrayAnnotation` has an array `values`, `AttributeAnnotation` has the `description` string).

`MultiTargetAnnotation` requires an additional patch to the transform so that it can correctly check each target against the circuit's `InstanceGraph`.

seldridge:
🤔 I think what you may need here is to define "annotation value equality".  Given both the existing `duplicate` method (explaining how to copy the annotation with a new target) and a way to compare two annotations, ignoring their targets, you may be able to avoid the value issue.  For `SingleTargetAnnotation`, I think "annotation value equality" could be derived as equality of all product members that are not the target.  For other annotations, a user may need to define this.

I think you could then group all annotations that are "annotation value equal" and, for their targets, deduplicate those and generate new annotations.  I.e., produce fewer annotations or annotations with simpler targets.

I'm just spitballing here and trying to reuse as much work as possible. 😄

What do you think?

I admit `MultiTargetAnnotation` is a pain, but is fortunately much rarer.  No need for a solution there.

davidbiancolin:
+1 for @seldridge's idea, but I think [Jack was concerned](https://github.com/chipsalliance/firrtl/pull/2286#discussion_r666588512) about the API leaking out before a longer-term annotation deduplication strategy is worked out. At very least, it could be done for SingleTargetAnnotations and kept package private for now. 

I'd also like see inline attributes fixed in the next 1.4.x release.

PR #2302, opened 2021-07-26
ir: make HashCode.toHashString public

ekiwi:
This will allow chiseltest to save the hash code to disk for the purpose of caching simulation binaries.

### Contributor Checklist

#### Type of Improvement
- new feature/API

#### API Impact

- `HashCode` produced by ir.StructuralHash now has a public `str` method which returns a string representation of the hash

- none
#### Desired Merge Strategy
- squash
- 
- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/ir/StructuralHash.scala
+++ b/src/main/scala/firrtl/ir/StructuralHash.scala
@@ -115,16 +115,20 @@ object StructuralHash {
 }
 
 trait HashCode {
-  protected val str: String
-  override def hashCode(): Int = str.hashCode
+
+  /** String representation of the hash code.
+    * Two instances of [[HashCode]] are equal if and only if their toHashString values are equal.
+    */
+  def toHashString: String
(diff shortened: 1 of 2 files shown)


src/main/scala/firrtl/ir/StructuralHash.scala
```diff
@@ -115,7 +115,9 @@ object StructuralHash {
 }
 
 trait HashCode {
-  protected val str: String
+
+  /** string representation of the hash code */
+  def str: String
```
jackkoenig, line 120: Why not `override def toString` instead of calling it `str`?

ekiwi: Because this is a _special_ String :)
It essentially satisfies the property that if (and only if) the two hash codes are the same, the two `str` are the same. This isn't generally true for `toString`.

Do you want me to rename it to `toHashString` or something like that?

jackkoenig (approved):

PR #2308, opened 2021-08-02
Skip Formal CI checks via Github Actions not commit message

jackkoenig:
https://github.com/chipsalliance/firrtl/pull/2307 is failing due to awkwardness in the Formal CI checks. The old way of skipping them doesn't seem to be working so perhaps we should just do this using Labels anyway.

### Contributor Checklist

- [ish] Did you add at least one test demonstrating the PR?

#### Type of Improvement

  - CI enhancement

#### API Impact

No impact

No impact

#### Desired Merge Strategy

  - Squash (Definitely squash, I don't want to rebase because the commit history is the "test" of this PR)

- [ ] Did you review?


jackkoenig:
The commit history here (that will squash away) shows that this works:
* fa71964 - Had [intentionally] incorrect logic to skip the formal checks on all non-push events (including PRs)
* 15f7f4d - Has correct logic and doesn't skip because this PR lacked the label at the time
* b288632 - With the label, the formal CI checks are correct skipped

jackkoenig:
Marked for backporting to `1.4.x` so that we can get this fixed for https://github.com/chipsalliance/firrtl/pull/2307. 

IIRC, Mergify isn't allowed to open PRs that touch the `.github/workflows` directory so I'll probably have to manually backport.

seldridge (approved):
Nice!  Way better.

jackkoenig:
One thing I don't love is that it's putting "Skipped" on `all_tests_passed`, so this may actually not actually work as is. I might have to move the skip into just the actual step inside `formal equivalence`. This would have the slightly negative consequence of having a green checkmark on `formal equivalence` even though it was skipped, but the label should make that obvious enough and we definitely want `all tests passed` to be green.

PR #2312, opened 2021-08-03
Require Andr, Orr, Xorr, Neg to have one operand

seldridge:
Fix an OG bug where Andr, Orr, and Xorr would accept an arbitrary number
of operands.  Verilog emission doesn't support this and will silently
drop all operands after the first.  E.g., "andr(a, b)" would emit as
"&a".  After this commit, "andr(a, b)" will be rejected by checking
passes.

For archaeological purposes, this appears to have been the behavior
dating back to when this was added in d2d3260a.

Signed-off-by: Schuyler Eldridge <schuyler.eldridge@sifive.com>

### Example

Start with:

```scala
circuit Foo :
  module Foo :
    input a: UInt<1>
    input b: UInt<1>
    output c: UInt<1>

    c <= orr(a, b)
```

If you compile this you get:

```verilog
module Foo(
  input   a,
  input   b,
  output  c
);
  assign c = |a;
endmodule
```

This patch causes:

```verilog
Exception in thread "main" firrtl.passes.CheckHighFormLike$IncorrectNumArgsException: : [module Foo] Primop orr requires 1 expression arguments.
```

### DIFF
--- a/src/main/scala/firrtl/passes/CheckHighForm.scala
+++ b/src/main/scala/firrtl/passes/CheckHighForm.scala
@@ -187,7 +187,7 @@ trait CheckHighFormLike { this: Pass =>
         case AsInterval =>
           correctNum(Option(1), 3)
         case Andr | Orr | Xorr | Neg =>
-          correctNum(None, 0)
+          correctNum(Option(1), 0)
       }
     }
 


ekiwi (approved):
Good find!

edwardcwang:
Does the spec already require only one operand?

ekiwi:
> Does the spec already require only one operand?

Yes. Afaik there are no primops with a variable number of arguments.

seldridge:
@edwardcwang: It's in the spec already as having one argument:

Screen Shot 2021-08-03 at 11 13 29

Also, it is entirely impossible to generate something like `andr(a, b)` from Chisel since reduction ops are methods on `UInt` and `Bool`. This can only happen if you manually craft input FIRRTL text. (I was expecting the parser would outright reject this, but it doesn't.)

jackkoenig (approved):

PR #2318, opened 2021-08-05
Turn RenameMap into a trait and add MutableRenameMap

jackkoenig:
The mutability of the RenameMap API has long been a wart. It also has
made it impossible to provide alternative implementations that do not
have a underlying map. By changing RenameMap to a [sealed] trait, it
will become possible to provide optimized implementations for different
purposes.

Possible optimized `RenameMaps`:
* `LowerTypesRenameMap` (not having to expand out every possible rename)
* `ModuleRenameMap` -- See `RenameMap. fromInstanceRenames`, a new version could lazily expand instead of having to create a Map
* `ComponentRenameMap` -- Maybe a simplified `RenameMap` that only allows for renaming within modules, could have much simpler logic than the current `RenameMap`

**Important Note for Backporting**

My plan for making this binary compatible in the backport (prototyped in [this branch](https://github.com/chipsalliance/firrtl/tree/renamemap-deprecations)) is to provide deprecated versions of public transform methods that currently take a `RenameMap` argument and then call the new version that takes a `MutableRenameMap` (doing an `.asInstanceOf` cast). For this to work, `MutableRenameMap` must be the only concrete implementation of `RenameMap`, so I must change the implementations of `RenameMap.++` and `.andThen` back to how they work for `MutableRenameMap`. I'm only noting this because it might be easy to forget but is very important.

### Contributor Checklist

#### Type of Improvement

- code refactoring  
 - code cleanup  
 - new feature/API    

#### API Impact

This breaks the `RenameMap` API. Most transform writers should do a search and replace `RenameMap` for `MutableRenameMap` (and import `firrtl.renamemap.MutableRenameMap`).

Any annotations taking advantage of the mutable methods on `RenameMap` will break.

No impact

#### Desired Merge Strategy

- Squash

#### Release Notes

Turn `firrtl.RenameMap` into a trait with an immutable API. Transform writers using the mutable methods should instead us `firrtl.renamemap.MutableRenameMap`. Annotations that use the mutable `RenameMap` methods in their `update` methods can no longer do that.

- [ ] Did you review?

### DIFF
--- a/fuzzer/src/main/scala/firrtl/FirrtlEquivalenceTest.scala
+++ b/fuzzer/src/main/scala/firrtl/FirrtlEquivalenceTest.scala
@@ -13,6 +13,7 @@ import firrtl.annotations.{Annotation, CircuitTarget, ModuleTarget, Target}
 import firrtl.ir.Circuit
 import firrtl.options.Dependency
 import firrtl.options.phases.WriteOutputAnnotations
+import firrtl.renamemap.MutableRenameMap
 import firrtl.stage.{FirrtlCircuitAnnotation, InfoModeAnnotation, OutputFileAnnotation, TransformManager}
 import firrtl.stage.Forms.{VerilogMinimumOptimized, VerilogOptimized}
 import firrtl.transforms.{InlineBooleanExpressions, ManipulateNames}
@@ -36,7 +37,7 @@ object FirrtlEquivalenceTestUtils {
         case _: CircuitTarget => true
         case _: Target => false
       }
-      val renames = RenameMap()
+      val renames = MutableRenameMap()
       val circuitx = run(state.circuit, renames, block, allow)
       state.copy(circuit = circuitx, renames = Some(renames))
     }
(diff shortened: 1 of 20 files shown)


jackkoenig:
The diff for `RenameMap.scala` is an absolute disaster, it might be worth just looking at the new file, and maybe generating the API docs locally and taking a look.

seldridge (dismissed):
Nice.  This looks like a great improvement, Jack, and will open up some flexibility on how the rename map data structure works internally in the future.

I only gave this a very cursory review judging from the fact that is is mostly just a structure refactor to create the trait `RenameMap` with all existing functionality of that moved into `MutableRenameMap`.  If I'm mis-understanding this and/or should take a more detailed review, let me know. 😅

jackkoenig:
I've rebased this on master but I'd like to hear from @davidbiancolin before I merge this. I **could** instead do the deprecation in 3.5.0-RC2. It's a bit harder from my development perspective but I'm feeling like this may be too painful for users who do actually write firrtl transforms since the deprecation will be in yet-unreleased 3.4.5.

PR #2319, opened 2021-08-05
Fix Specification Memory Port Types

seldridge:
Correct incorrect type specified for memories in the FIRRTL
specification.  This is important because the memory type determines
what is a legal bundle to try to connect to a memory port.

I based this off of FIRRTL accepting the following circuit:

```scala
    circuit MemOrder:
      module MemOrder:
        input r: {addr : UInt<3>, en : UInt<1>, clk : Clock, flip data : UInt<1>}
        input w: {addr : UInt<3>, en : UInt<1>, clk : Clock, data : UInt<1>, mask : UInt<1>}
        input rw: {addr : UInt<3>, en : UInt<1>, clk : Clock, flip rdata : UInt<1>, wmode : UInt<1>, wdata : UInt<1>, wmask : UInt<1>}

        mem memory:
          data-type => UInt<1>
          depth => 8
          reader => r
          writer => w
          readwriter => rw
          read-latency => 1
          write-latency => 1
          read-under-write => undefined

        memory.r <= r
        memory.w <= w
        memory.rw <= rw
```

Signed-off-by: Schuyler Eldridge <schuyler.eldridge@sifive.com>


ekiwi:
From the compiler perspective, the bundle type is coded up here: https://github.com/chipsalliance/firrtl/blob/master/src/main/scala/firrtl/passes/memlib/MemUtils.scala#L66

I assume you spec update matches that, right?

seldridge:
> From the compiler perspective, the bundle type is coded up here: https://github.com/chipsalliance/firrtl/blob/master/src/main/scala/firrtl/passes/memlib/MemUtils.scala#L66
> 
> I assume you spec update matches that, right?

I had only empirically tested this. However, the spec now matches that exactly. 👀

ekiwi (approved):
LGTM :ship:

PR #2331, opened 2021-08-18
Add RTLIL Backend.

nxmq:
### Contributor Checklist

#### Type of Improvement

- backend code generation
- new feature/API

#### API Impact

This PR adds the names "experimental-rtlil" and "rtlil" to the list of valid emitters, and also adds the annotations required to utilize the emitter. These are the only changes to the public-facing API.

This PR introduces an entirely new code generation backend, targeting the textual representation of the RTLIL IR format used by the Yosys open synthesis tool. As such, it should have minimal effects on the existing Verilog backend, as the only change made which could affect that backend was making verilog-specific memory lowering an optional dependency of the CombineCats pass, instead of a required one.

#### Desired Merge Strategy

Rebase please.

#### Release Notes
- Adds code generation backend targeting RTLIL IR for use with the Yosys open synthesis tool.

- [x] Did you review?


src/main/scala/firrtl/Emitter.scala
```diff
@@ -146,6 +149,8 @@ object EmitAllModulesAnnotation extends HasShellOptions {
               RunFirrtlTransformAnnotation(new SystemVerilogEmitter),
               EmitAllModulesAnnotation(classOf[SystemVerilogEmitter])
             )
+          case "experimental-rtlil" | "rtlil" =>
+            Seq(RunFirrtlTransformAnnotation(Dependency(RtlilEmitter)), EmitCircuitAnnotation(RtlilEmitter.getClass))
```
ekiwi, line 153: Is your backend capable of emitting one file per module? If not, you shouldn't add it here.
The `smt2` backend for example always produces a single file which is why it is only supported by the `-E`  option.

src/main/scala/firrtl/Emitter.scala
```diff
@@ -60,6 +61,8 @@ object EmitCircuitAnnotation extends HasShellOptions {
             Seq(RunFirrtlTransformAnnotation(Dependency(Btor2Emitter)), EmitCircuitAnnotation(Btor2Emitter.getClass))
           case "experimental-smt2" | "smt2" =>
             Seq(RunFirrtlTransformAnnotation(Dependency(SMTLibEmitter)), EmitCircuitAnnotation(SMTLibEmitter.getClass))
+          case "experimental-rtlil" | "rtlil" =>
```
ekiwi, line 64: For now, only `experimental-rtlil`. The SMT backend went through a year of updates until we decided it was time to allow the use without the `experimental-` prefix. Essentially by forcing users to acknowledge that this is experimental we reserve the right to radically change or remove the backend in the future.

src/main/scala/firrtl/stage/FirrtlAnnotations.scala
```diff
@@ -192,6 +193,7 @@ object RunFirrtlTransformAnnotation extends HasShellOptions {
       case "verilog"  => new VerilogEmitter
       case "mverilog" => new MinimumVerilogEmitter
       case "sverilog" => new SystemVerilogEmitter
+      case "experimental-rtlil" | "rtlil" => RtlilEmitter
```
ekiwi, line 196: I don't think this is needed here. Afaik this is only used by the old `CompilerAnnotation`.

src/main/scala/firrtl/transforms/CombineCats.scala
```diff
@@ -63,11 +63,11 @@ class CombineCats extends Transform with DependencyAPIMigration {
   override def prerequisites = firrtl.stage.Forms.LowForm ++
     Seq(
       Dependency(passes.RemoveValidIf),
-      Dependency(firrtl.passes.memlib.VerilogMemDelays),
       Dependency(firrtl.passes.SplitExpressions)
     )
 
-  override def optionalPrerequisites = Seq(Dependency[firrtl.transforms.ConstantPropagation])
+  override def optionalPrerequisites = Seq(Dependency(firrtl.passes.memlib.VerilogMemDelays),
```
ekiwi, line 69: This change is fine, shouldn't mess anything up for the Verilog emission :+1:

src/main/scala/firrtl/Emitter.scala
ekiwi: I just saw that you support `EmitAllModulesAnnotation` in your emitter code. You probably just forgot to add it here.

src/main/scala/firrtl/backends/experimental/rtlil/RtlilEmitter.scala
```diff
@@ -0,0 +1,888 @@
...
+import firrtl.passes.LowerTypes
+import firrtl.passes.MemPortUtils.memPortField
+import firrtl.stage.TransformManager
+
+import scala.collection.mutable
+import scala.collection.mutable.ArrayBuffer
+import scala.language.postfixOps
+
+
+object RtlilEmitter extends SeqTransform with Emitter with DependencyAPIMigration {
```
ekiwi, line 21: Since this is experimental, please make the `RtlilEmitter` object `private [firrtl]`.

src/main/scala/firrtl/backends/experimental/rtlil/RtlilEmitter.scala
```diff
@@ -0,0 +1,888 @@
...
+              EmittedRtlilModuleAnnotation(EmittedRtlilModule(module.name, writerToString(writer), outputSuffix))
+            )
+          case _ => None
+        }
+      case _ => Seq()
+    }
+    state.copy(annotations = newAnnos ++ state.annotations)
+  }
+
+  class RtlilRender(description:      Seq[Description],
```
ekiwi, line 87: Internal classes should always be `private`

src/main/scala/firrtl/backends/experimental/rtlil/RtlilEmitter.scala
```diff
@@ -0,0 +1,888 @@
...
+            val (a, b) = set.toSeq match {
+              case Seq(x, y) => (x, y)
+            }
+            attachSynAssigns += Seq("connect ", SrcInfo(a,sx.info).str_rep, " ", SrcInfo(b,sx.info).str_rep)
+          }
+        case _ =>
+        }
+      }
+
+    def emit_rtlil(): DefModule = {
```
ekiwi, line 787: all internal functions should be `private`

ekiwi: (and by internal I mean anything that isn't called from outside the class)

ekiwi (changes requested):
This looks awesome. I added some small initial feedback.
Before this gets merged we need some tests. You can look at https://github.com/chipsalliance/firrtl/tree/master/src/test/scala/firrtl/backends/experimental/smt for inspiration.

A minimal smoke test could be to run your backend on all our regression firrtl files and then to just read the result with yosys to make sure it is at least valid RTLIL. Like: https://github.com/chipsalliance/firrtl/blob/master/src/test/scala/firrtl/backends/experimental/smt/end2end/SMTCompilationTest.scala

Thanks!

nxmq:
> This looks awesome. I added some small initial feedback.
> Before this gets merged we need some tests. You can look at https://github.com/chipsalliance/firrtl/tree/master/src/test/scala/firrtl/backends/experimental/smt for inspiration.
> 
> A minimal smoke test could be to run your backend on all our regression firrtl files and then to just read the result with yosys to make sure it is at least valid RTLIL. Like: https://github.com/chipsalliance/firrtl/blob/master/src/test/scala/firrtl/backends/experimental/smt/end2end/SMTCompilationTest.scala
> 
> Thanks!

One test I did use throughout development was using yosys's `equiv_check` and related features to verify equivalence between the circuit as emitted by the rtlil backend and the circuit as emitted by the verilog backend. Unfortunately, this test takes about 5m to run on even a moderately complex circuit, so I feel hesitant about including something with such a long runtime into the tests for firrtl. Should I add an `equiv_check` based test, or a simple "does yosys consume this" test?

ekiwi:
> One test I did use throughout development was using yosys's `equiv_check` and related features to verify equivalence between the circuit as emitted by the rtlil backend and the circuit as emitted by the verilog backend. Unfortunately, this test takes about 5m to run on even a moderately complex circuit, so I feel hesitant about including something with such a long runtime into the tests for firrtl. Should I add an `equiv_check` based test, or a simple "does yosys consume this" test?

I would go for "does yosys consume this" for all regression designs and then a `equiv_check` for a very small design.

ekiwi:
To solve the CI issues, you just need to run:
```
sbt scalafmtAll
```

ekiwi:
Not saying you need to do this to get the PR merged, but just as another idea on how we could test this backend:
You could use cxxrtl to generate an executable simulation with yosys and then you could write some executable unittests similar to what we do with Verilator to test the Verilog emission.
- [executable memory init test](https://github.com/chipsalliance/firrtl/blob/765af619af24fccd2dea748e95d70c9c6e8ea12b/src/test/scala/firrtlTests/MemoryInitSpec.scala#L359)
- [SimpleExecutionTest](https://github.com/chipsalliance/firrtl/blob/1f7faaa36816fc145a39bb7e736fd5ab739112c4/src/test/scala/firrtlTests/execution/SimpleExecutionTest.scala)

ekiwi:
Looks like some of your code doesn't compile on scala 2.13. To debug that you can do `sbt ++2.13.4 compile`

ekiwi:
@nxmq  We discussed this PR in our dev meeting and we are happy to merge it once the tests pass.

I assume that the reason the CI fails right now is that we might be using an older version of `yosys`. Could you add some output to stdout to debug what is going wrong with yosys?

ekiwi:
Could you also add some of your formal equivalence check tests? Maybe marked as `ignore` for now? We would be interested in running them from time to time (but maybe not on every PR).

ekiwi:
This PR might help us get a newer yosys version: https://github.com/chipsalliance/firrtl/pull/2365

src/main/scala/firrtl/Emitter.scala
```diff
@@ -242,3 +247,9 @@ case class EmittedVerilogCircuitAnnotation(value: EmittedVerilogCircuit)
     extends EmittedCircuitAnnotation[EmittedVerilogCircuit]
 case class EmittedVerilogModuleAnnotation(value: EmittedVerilogModule)
     extends EmittedModuleAnnotation[EmittedVerilogModule]
+
+final case class EmittedRtlilCircuit(name: String, value: String, outputSuffix: String) extends EmittedCircuit
```
ekiwi, line 251: please move these classes into the `firrtl.backends.experimental.rtlil` package similar to how we did it for the SMT backend : https://github.com/chipsalliance/firrtl/blob/a921e1230b389be87d993f3016cb46174b1ebfad/src/main/scala/firrtl/backends/experimental/smt/SMTEmitter.scala#L46

src/main/scala/firrtl/stage/FirrtlAnnotations.scala
```diff
@@ -221,15 +222,16 @@ object RunFirrtlTransformAnnotation extends HasShellOptions {
...
+      case "none"               => new ChirrtlEmitter
+      case "mhigh"              => new MinimumHighFirrtlEmitter
+      case "high"               => new HighFirrtlEmitter
+      case "low"                => new LowFirrtlEmitter
+      case "middle"             => new MiddleFirrtlEmitter
+      case "verilog"            => new VerilogEmitter
+      case "mverilog"           => new MinimumVerilogEmitter
+      case "sverilog"           => new SystemVerilogEmitter
+      case "experimental-rtlil" => new RtlilEmitter
+      case _                    => throw new OptionsException(s"Unknown compiler name '$a'! (Did you misspell it?)")
```
ekiwi, line 234: Please undo your changes to this file. This code is for the (soon to be) deprecated `-X` option. New emitters should only use the `-E` option.

src/main/scala/firrtl/transforms/CombineCats.scala
```diff
@@ -63,11 +63,11 @@ class CombineCats extends Transform with DependencyAPIMigration {
   override def prerequisites = firrtl.stage.Forms.LowForm ++
     Seq(
       Dependency(passes.RemoveValidIf),
-      Dependency(firrtl.passes.memlib.VerilogMemDelays),
       Dependency(firrtl.passes.SplitExpressions)
     )
 
-  override def optionalPrerequisites = Seq(Dependency[firrtl.transforms.ConstantPropagation])
+  override def optionalPrerequisites =
+    Seq(Dependency(firrtl.passes.memlib.VerilogMemDelays), Dependency[firrtl.transforms.ConstantPropagation])
```
ekiwi, line 70: :+1:

ekiwi (changes requested):
Two small changes and then I think this new experimental backend is ready to :ship:

ekiwi (approved):
Awesome! Thank you very much for your patience and your hard work!

Looks like this change will make it into the upcoming release and it should be part of the snapshot in 1-2h. I would be happy to review any followup PRs if you see that something is missing or anything needs to be fixed.

PR #2344, opened 2021-08-31
Multi protobuf module emission and consumption

jared-barocsi:
### Contributor Checklist

#### Type of Improvement

- new feature/API 
#### API Impact

No API impact

Can change ordering of modules. `MultiInfo`s do not serialize properly because of placeholder code in the protobuf emitter.

#### Desired Merge Strategy

- Squash and merge

#### Release Notes

Implement multiple protobuf emission and consumption in the FIRRTL compiler
- `-p` option exports all modules into individual protobuf files, similar to `-e`.
- `-I` option takes a directory of FIRRTL files, deserializes them and combines them back into a single circuit. Currently supports only the protobuf files emitted by `-p`

- [ ] Did you review?


src/main/scala/firrtl/Emitter.scala
```diff
@@ -151,6 +151,43 @@ object EmitAllModulesAnnotation extends HasShellOptions {
       helpText = "Run the specified module emitter (one file per module)",
       shortOption = Some("e"),
       helpValueName = Some("<chirrtl|high|middle|low|verilog|mverilog|sverilog>")
+    ),
+    new ShellOption[String](
+      longOption = "emit-module-protobufs",
```
jackkoenig, line 156:
```suggestion
      longOption = "emit-modules-protobuf",
```
Slight tweak to make the option match the pattern of `-e` and follow the style of `emit-circuit` and `emit-circuit-protobuf`

src/main/scala/firrtl/Utils.scala
```diff
@@ -978,6 +978,37 @@ object Utils extends LazyLogging {
...
+      // Left means "lone module", Right means has ExtModules
+      val module: Option[Module] = {
+        val found: Seq[Module] = modules.collect { case m: Module => m }
+        assert(found.size <= 1)
+        found.headOption
+      }
+      val extModules: Seq[ExtModule] = modules.collect { case e: ExtModule => e }.distinct
+
+      if (extModules.isEmpty) Seq(Left(module.get)) else (module ++: extModules).map(Right(_))
+    }
```
jackkoenig, line 993: I think this logic is close but not quite right. Given the following circuit:

```
circuit top :
  module child :
    output out : UInt<8>
    out <= UInt(0)
  module top :
    output out : UInt<8>
    inst c of child
    out <= c.out
```

This if split into multiple files would give:

```
; child.fir
circuit child :
  module child :
    output out : UInt<8>
    out <= UInt(0)
```
```
; top.fir
circuit top :
  extmodule child :
    output out : UInt<8>

  module top :
    output out : UInt<8>
    inst c of child
    out <= c.out
```
Thus when we do the `groupByIntoSeq` we'll have `extmodule child` and `module child` together passed to this function. This function needs to eliminate the `extmodule` by seeing that it's implemented by the associated `Module`.

Now I will grant that down below we have:
```scala
val res = deduped.collect { case Right(m: Module) => m }
```

Which effectively does what I'm saying should be done up here in this function. That is actually wrong because you could truly have an `extmodule` incorrectly getting deleted, consider:

```
circuit top :
  extmodule foo :
    output out : UInt<8>

  module top :
    output out : UInt<8>
    inst c of foo
    out <= c.out
```
I think that the `dedup.collect` below will incorrectly just delete `foo`.

src/main/scala/firrtl/proto/FromProto.scala
```diff
@@ -33,6 +35,22 @@ object FromProto {
...
+    * @param dir directory containing ProtoBuf representation(s)
+    * @return Deserialized FIRRTL Circuit
+    */
+  def fromDirectory(dir: String): ir.Circuit = {
+    val d = new File(dir)
+    val fileList = if (d.exists && d.isDirectory) {
+      d.listFiles.filter(_.isFile).toList
+    } else {
+      List[File]()
+    }
```
jackkoenig, line 49: Rather than returning an empty list if there's something wrong with `dir`, I think it should be an error.

src/main/scala/firrtl/stage/FirrtlAnnotations.scala
```diff
@@ -64,6 +64,39 @@ object FirrtlFileAnnotation extends HasShellOptions {
...
+case class FirrtlDirectoryAnnotation(dir: String) extends NoTargetAnnotation with CircuitOption {
+
+  def toCircuit(info: Parser.InfoMode): FirrtlCircuitAnnotation = {
+    val circuit =
+      try {
+        proto.FromProto.fromDirectory(dir)
+      } catch {
+        case a @ (_: FileNotFoundException | _: NoSuchFileException) =>
+          throw new OptionsException(s"Directory '$dir' not found! (Did you misspell it?)", a)
+      }
```
jackkoenig, line 80: This seems out of sync with the fact that `fromDirection` isn't erroring if there's something wrong with `dir`. I think this is right but `fromDirectory` needs to throw such exceptions.

jackkoenig (commented):
Generally looks good although a few minor issues. Also obviously needs tests but it is a Draft for now 🙂

src/main/scala/firrtl/stage/FirrtlAnnotations.scala
```diff
@@ -64,6 +64,41 @@ object FirrtlFileAnnotation extends HasShellOptions {
 
 }
 
+/** Read a directory of FIRRTL files or ProtoBufs
+  *  - set with `-I/--input-directory`
+  * @param file input filename
```
mwachs5, line 69: this `@param` needs updating, shouldn't this be `dir`?

mwachs5: also unclear the comment a little -- so the same argument `-I` can be used for a directory of both firrtl files or protobuf files? can both be in the same directory?

src/main/scala/firrtl/stage/FirrtlAnnotations.scala
```diff
@@ -64,6 +64,41 @@ object FirrtlFileAnnotation extends HasShellOptions {
...
+
+}
+
+object FirrtlDirectoryAnnotation extends HasShellOptions {
+
+  val options = Seq(
+    new ShellOption[String](
+      longOption = "input-directory",
+      toAnnotationSeq = a => Seq(FirrtlDirectoryAnnotation(a)),
+      helpText = "A directory of FIRRTL files",
```
mwachs5, line 94: can these be either `.fir` or `.pb` files? A mixture?

src/main/scala/firrtl/stage/FirrtlAnnotations.scala
```diff
@@ -64,6 +64,41 @@ object FirrtlFileAnnotation extends HasShellOptions {
...
+/** Read a directory of FIRRTL files or ProtoBufs
+  *  - set with `-I/--input-directory`
+  * @param file input filename
+  */
+case class FirrtlDirectoryAnnotation(dir: String) extends NoTargetAnnotation with CircuitOption {
+
+  def toCircuit(info: Parser.InfoMode): FirrtlCircuitAnnotation = {
+    val circuit =
+      try {
+        proto.FromProto.fromDirectory(dir)
```
mwachs5, line 76: this seems to be only reading protobuf, should the `FirrtlDirectoryAnnotation` be named `FirrtlProtobufDirectoryAnnoatation`?

jackkoenig: In the case of the analogous `FirrtlFileAnnotation`, we support both `.pb` and `.fir` with the same annotation, I think we should do the analogous thing here. Even if we're not implementing reading directories of `.fir` files in this PR, we should keep the name of the annotation general.

mwachs5: ah, ok, so maybe the scaladoc at the top should be updated to reflect the "FIRRTL files" is "TODO"

src/main/scala/firrtl/stage/FirrtlAnnotations.scala
jackkoenig: We should figure that out. I think to start we should make it be only `.fir` or only `.pb` files. Note that `-i` actually doesn't care about file extensions, it just tries to deserialize as protobuf and if it fails it deserializes as `.fir` but I think it's okay to require correct file extensions here.

src/test/scala/firrtlTests/stage/FirrtlMainSpec.scala
```diff
@@ -324,6 +334,41 @@ class FirrtlMainSpec
...
+      And("some input multi-module FIRRTL IR")
+      val inputFile: Array[String] = {
+        val in = new File(td.dir, c.main)
+        val pw = new PrintWriter(in)
+        pw.write(c.input)
+        pw.close()
+        Array("-i", in.toString)
+      }
+
+      When("the user tries to compile to multiple Protocol Buffers in the target directory")
```
jackkoenig, line 352:
```suggestion
      When("the user tries to emit a circuit to multiple Protocol Buffer files in the target directory")
```

src/test/scala/firrtlTests/stage/FirrtlMainSpec.scala
```diff
@@ -324,6 +334,41 @@ class FirrtlMainSpec
...
+      }
+
+      When("the user compiles the Protobufs to a single High FIRRTL IR")
+      f.stage.main(
+        Array("-I", td.buildDir.toString, "-X", "none", "-E", "high", "-td", td.buildDir.toString, "-o", "Foo")
+      )
+
+      Then("one single High FIRRTL file should be emitted")
+      new File(td.buildDir + "/Foo.hi.fir") should (exist)
+    }
```
jackkoenig, line 370: Can you check that the `.pb` you emit and then read back in is the same? I think if you change the initial emission to `-p chirrtl` (instead of `-p high`) you should be able to just check `==` of the `Circuit` objects.

jackkoenig: It also would be helpful if the example circuit had an extmodule that we check specifically for because I believe that was broken in an earlier version of this PR.

src/main/scala/firrtl/backends/proto/ProtoBufEmitter.scala
```diff
@@ -59,10 +62,47 @@ sealed abstract class ProtoBufEmitter(prereqs: Seq[TransformDependency])
...
+      case m: Module =>
+        val instModules = collectInstantiatedModules(m, modMap)
+        val extModules = instModules.map {
+          case Module(info, name, ports, _) => ExtModule(info, name, ports, name, Seq.empty)
+          case ext: ExtModule => ext
+        }
+        val newCircuit = Circuit(m.info, extModules :+ m, m.name)
+        Annotation.ProtoBufSerialization(newCircuit, Some(outputSuffix))
+    }
+  }
```
jackkoenig, line 90: Given that this is almost a copy-paste of the same method in `FirrtlEmitter`, can we instead factor them out into a common utility that both call? https://github.com/chipsalliance/firrtl/blob/de56a19b0a240e39366dc2d979ec05c65e0ada63/src/main/scala/firrtl/backends/firrtl/FirrtlEmitter.scala#L22

Preferable `private[firrtl]` so that we don't leak it into the public API but still would be better to not have the code duplication.

src/main/scala/firrtl/Utils.scala
```diff
@@ -978,6 +978,62 @@ object Utils extends LazyLogging {
...
+  def orderAgnosticEquality(a: Circuit, b: Circuit): Boolean =
+    a.copy(modules = a.modules.sortBy(_.name)) == b.copy(modules = b.modules.sortBy(_.name))
+
+  /** Combines several separate circuit modules (typically emitted by -e or -p compiler options) into a single circuit */
+  def combine(circuits: Seq[Circuit]): Circuit = {
+    def dedup(modules: Seq[DefModule]): Seq[Either[Module, DefModule]] = {
+      // Left means module with no ExtModules, Right means child modules or lone ExtModules
+      val module: Option[Module] = {
+        val found: Seq[Module] = modules.collect { case m: Module => m }
+        assert(found.size <= 1)
```
jackkoenig, line 1004:
```suggestion
        assert(found.size <= 1, s"Module definitions should have unique names, found ${found.size} definitions named ${found.head.name}")
```

src/main/scala/firrtl/Utils.scala
```diff
@@ -978,6 +978,62 @@ object Utils extends LazyLogging {
...
+    val grouped: Seq[(String, Seq[DefModule])] = groupByIntoSeq(circuits.flatMap(_.modules))({
+      case mod: Module    => mod.name
+      case ext: ExtModule => ext.defname
+    })
+    val deduped: Iterable[Either[Module, DefModule]] = grouped.flatMap { case (_, insts) => dedup(insts) }
+
+    // 2. Determine top
+    val top = {
+      val found = deduped.collect { case Left(m) => m }
+      assert(found.size == 1)
```
jackkoenig, line 1030:
```suggestion
      assert(found.size == 1, s"There should only be 1 top module, got: ${found.map(_.name).mkString(", ")}")
```

src/test/scala/firrtlTests/UtilsSpec.scala
```diff
@@ -46,4 +46,174 @@ class UtilsSpec extends AnyFlatSpec {
...
+         |  module Child:
+         |    output foo: UInt<32>
+         |
+         |    skip
+         |""".stripMargin
+
+    combineTest(input, output)
+  }
+
+  "combine" should "not dedup lone ExtModules" in {
```
jackkoenig, line 163:
```suggestion
  "combine" should "support lone ExtModules" in {
```

jackkoenig (approved):
Looks good and a nice suite of tests! Just some suggestions for improved error messages then let's get this merged.

PR #2355, opened 2021-09-11
Remove BlackBoxSourceHelper from ReplaceMemTransform

jackkoenig:
BlackBoxSourceHelper should only run late in compilation to allow
transforms to tweak its behavior (eg. changing BlackBoxTargetDirAnno).

Reverts a subtle behavior change in https://github.com/chipsalliance/firrtl/pull/2202, see test that captures the functionality I am trying to maintain. SiFive hit this in our internal flow where we emit different BlackBoxes to different directories so we need `BlackBoxSourceHelper` to only run during or after Verilog emission.

This really falls out of the fact that `BlackBoxSourceHelper` exists at all rather than using `CustomFileEmission` to emit the various files.

### Contributor Checklist

Fixes unreleased bug

#### Type of Improvement

  - bug fix     

#### API Impact

This enshrines the FIRRTL 1.4.x behavior of letting late running transforms tweak the behavior of `BlackBoxSourceHelper`.

No impact

#### Desired Merge Strategy

 - Squash:

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/passes/memlib/ReplaceMemTransform.scala
+++ b/src/main/scala/firrtl/passes/memlib/ReplaceMemTransform.scala
@@ -9,7 +9,6 @@ import firrtl.annotations._
 import firrtl.options.{CustomFileEmission, Dependency, HasShellOptions, ShellOption}
 import firrtl.passes.wiring._
 import firrtl.stage.{Forms, RunFirrtlTransformAnnotation}
-import firrtl.transforms.BlackBoxSourceHelper
 
 import java.io.{CharArrayWriter, PrintWriter}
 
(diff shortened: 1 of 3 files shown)


mwachs5 (commented):
Did this test actually fail before the fix? wouldn't you (maybe) need a SeqMem or other thing handled by the ReplSeqMems pass to trigger this, or is the fact that ReplSeqMems runs at all enough?

jackkoenig:
> Did this test actually fail before the fix? wouldn't you (maybe) need a SeqMem or other thing handled by the ReplSeqMems pass to trigger this, or is the fact that ReplSeqMems runs at all enough?

It did indeed fail, just running ReplSeqMems currently runs BlackBoxSourceHelper

mwachs5 (approved):
I can approve this but I don't think I can actually check the boxes (my perms seem a bit strange on this repo)

davidbiancolin (approved):
LGTM

PR #2372, opened 2021-09-28
TopWiring: don't try to wire Print, etc statements

mwachs5:
### Contributor Checklist

#### Type of Improvement

- bug fix   

#### API Impact

Prints and other non-`CanBeReferenced` statements will not try to be top-wired. They can't be because they generally do not have `.name`s (`""` is not a valid name).

This makes some verilog work that currently would not be able to be generated.

#### Desired Merge Strategy

#### Release Notes

Fix bug in TopWiringTransform for handling new Printf, Verif, Stop etc statements.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/transforms/TopWiring.scala
+++ b/src/main/scala/firrtl/transforms/TopWiring.scala
@@ -69,7 +69,7 @@ class TopWiringTransform extends Transform with DependencyAPIMigration {
   )(s:             Statement
   ): Statement = s match {
     // If target wire, add name and size to to sourceMap
-    case w: IsDeclaration =>
+    case w: IsDeclaration with CanBeReferenced =>
       if (sourceList.keys.toSeq.contains(ComponentName(w.name, currentmodule))) {
         val (isport, tpe, prefix) = w match {
           case d: DefWire     => (false, d.tpe, sourceList(ComponentName(w.name, currentmodule)))


mwachs5:
This should only go to 1.5

mwachs5:
Also questionable is this a bug, or should peopel be adding names to their prints that they want to move? Why do prints have names anyway...

davidbiancolin:
> Also questionable is this a bug, or should people be adding names to their prints that they want to move? Why do prints have names anyway...

It makes them reference-able from annotations, amoungst other things.

ekiwi:
Could you provide a regression test for this bug, @mwachs5 ?
I am curious to see how that code would ever be triggered.

ekiwi:
> It makes them reference-able from annotations, amoungst other things.

Also note that by default Chisel 3.5 should be naming all printf, stop, assert, assume and cover statements.

mwachs5:
> Also note that by default Chisel 3.5 should be naming all printf, stop, assert, assume and cover statements.

What do you mean "by default"? All of their default names are `""`... what am I missing?

ekiwi:
> What do you mean "by default"? All of their default names are `""`... what am I missing?

The current `RC-1` is missing this PR which made all statements named: https://github.com/chipsalliance/chisel3/pull/2089

Before that commit only `printf` and the statements in `experimental.verification` would be named.
https://scastie.scala-lang.org/yoolKgTPTEmvKuaWJS5Xrw

mwachs5:
> The current RC-1 is missing this PR which made all statements named: chipsalliance/chisel3#2089

Ok. So you think this firesim workaround would not be necessary once we get to RC2:
https://github.com/firesim/firesim/pull/836/commits/347e7e4f21e86f851ec30f71ead15c3e27c70ed6

ekiwi:
> Ok. So you think this firesim workaround would not be necessary once we get to RC2:

I think the firesim case is different. If you want to add a new statement in a firrtl pass, you should give it a name, just like you need to name any registers or memories that you add. I was just pointing out that with `RC-2` every "verification" statement should have a name generated by Chisel.

mwachs5:
Ok, so my original question again... do we want TopWiring to do stuff to print/verif nodes (those that are not `CanBeReferenced`? If this is just user error and the ComponentName was flagging the user error as intended, I can close this PR.

ekiwi:
I do not know enough about what `TopWiring` does to know whether we need this fix or not.
I think an example that would fail without the fix would be nice either way since it would help us catch any future regressions.

davidbiancolin:
I think a reasonable, low effort thing to do is reject unnamed statements here:
https://github.com/chipsalliance/firrtl/pull/2372/files#diff-8e48008febbd4ce8c153d3d958b027f87c73d8076dc734c79fff7242d17337beR73

And then let the following case statements croak on unsupported statement types. This should preserve the existing behavior of the pass while giving marginally better error reporting.

PR #2375, opened 2021-09-29
verilog: add comment to named statements

ekiwi:
I need this feature in order to more easily translate a failing assertion in Verilator into a firrtl statement name.
Besides that, I think this will help developers map the generated Verilog to the input firrtl or Chisel code.

### Contributor Checklist

#### Type of Improvement
- backend code generation (non-functional!)

#### API Impact
- no API change

- if a statement has a name, it will be included as a comment in the generated Verilog

#### Desired Merge Strategy
- squash

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
+++ b/src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
@@ -931,15 +931,23 @@ class VerilogEmitter extends SeqTransform with Emitter {
       addFormalStatement(formals, clk, en, stmt, info, msg)
     }
 
-    def formalStatement(op: Formal.Value, cond: Expression): Seq[Any] = {
-      Seq(op.toString, "(", cond, ");")
+    private def namedStatementComment(name: String): String = if (name.isEmpty) ""
+    else {
+      s" // ${m.name}.${name}"
     }
 
-    def stop(ret: Int): Seq[Any] = Seq(if (ret == 0) "$finish;" else "$fatal;")
+    def formalStatement(op: Formal.Value, cond: Expression, name: String): Seq[Any] = {
+      Seq(op.toString, "(", cond, ");", namedStatementComment(name))
+    }
 
-    def printf(str: StringLit, args: Seq[Expression]): Seq[Any] = {
+    def stop(ret: Int, name: String): Seq[Any] = {
+      val function = if (ret == 0) "finish" else "fatal"
+      Seq("$" + function + ";" + namedStatementComment(name))
+    }
+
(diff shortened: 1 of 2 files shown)


mwachs5:
I'm a bit confused what you mean. What is the default behavior for statements coming from e.g.:

```
import chisel3._

assert(a == b, "a must be b")

```

what about for:
```
import Chisel3._

assert(a == b, "a must be b")
```

I think suddenly adding comments to user's verilog is a pretty invasive change, not sure if this is opt-in or not, by using a new `import` path?

edwardcwang:
You meant `import Chisel._` instead of `import Chisel3._`?

mwachs5:
> You meant import Chisel._ instead of import Chisel3._?

Yes fixed

ekiwi:
If you have something like:

```.scala
class MyModule extends Module {
  // ...
  assert(a == b).suggestName("my_assertion")
}
```

The resulting Verilog with the current `master` branch contains something like:

```.v
if (1'h1) begin
  assert(_T_1);
end
```

With this PR it would get a comment like:

```.v
if (1'h1) begin
  assert(_T_1); // MyModule.my_assertion
end
```

mwachs5:
what If I do NOT `.suggestName(...)` (I rarely do...)

ekiwi:
> what If I do NOT `.suggestName(...)` (I rarely do...)

`suggestName` working with `assert` is a new `3.5` feature. If you do not use it, the assert just gets the default name (+ whatever the plugin does).

mwachs5:
right, but now I have new comments in my code i did not have before (which while admittedly useful, might not be what I want).

For example those comments will now need to be obfuscated by an obfuscation path that did not have to worry about them before.

Is there any way this feature could be opt-in via an annotation or cmd line arg?

ekiwi:
> Is there any way this feature could be opt-in via an annotation or cmd line arg?

If this addition proves to be controversial, I am going to re-consider it. Since this is a non-functional change, I did not see it as more than a trivial change. I will change the PR back to a draft.

> For example those comments will now need to be obfuscated by an obfuscation pass that did not have to worry about them before.

If you maintain an obfuscation pass, I would recommend that you add awareness of statement names when you upgrade to the `X.5` release, otherwise you could easily leak the names of statements, e.g., when you emit firrtl.

mwachs5:
> If you maintain an obfuscation pass, I would recommend that you add awareness of statement names when you upgrade to the X.5 release, otherwise you could easily leak the names of statements, e.g., when you emit firrtl.

yes, that is a fair point

Issue #2384, opened 2021-10-11
Add option to ignore undeserializable annotations (basically pass them through)


jackkoenig:
### Checklist

- [ ] Did you write out a description of the feature you want to see?
- [ ] Did you look around for any related features?
- [ ] Did you specify relevant external information?

jackkoenig:
@davidbiancolin has noted that package-level allow and disallow lists may be useful here.

PR #2387, opened 2021-10-13
Let firrtl based applications run despite loading unknown annotations

chick:
An application like barstools may contain a main that loads an annotations file containing
annotation classes that are not on it's classpath. This change allows unknown annotations
to be preserved by wrapping them in a UnrecognizedAnnotation.
This feature can be enabled via an AllowUnrecognizedAnnotations

Addresses issue #2384

### Contributor Checklist

#### Type of Improvement

New feature with a minor API addition

#### API Impact

This new feature is opt in. It adds an Annotation and a CLI flag. Allows Stage subClasses to load annotation files with unrecognized Annotations.
Error messages produced when unrecognized class occur and this feature is not enabled have changed slightly and show class names for all unrecognized annotations. These messages appear via Error level logging.

No change to backend code generation

#### Desired Merge Strategy

Squash: The PR will be squashed and merged (choose this if you have no preference.

#### Release Notes
It is possible to disable errors caused by JSON annotation files containing unrecognized annotations. This situation
can occur when, for example applications like barstools with no code level dependency on rocket chip encounters
a rocket chip annotations in an annotation file. This change carries the annotations forward while ignoring their content. 

- [ ] Did you review?


src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -288,15 +305,63 @@ object JsonProtocol {
...
+        try {
+          val result = Class.forName(x)
+          result
+        } catch {
+          case _: java.lang.ClassNotFoundException =>
+            classNotFoundBuildingLoaded = true // tells us which Exception to throw in recovery
+            // Found an annotation we don't recognize, So add UnrecognizedAnnotation to `loaded`
+            Class.forName("firrtl.annotations.UnrecognizedAnnotation")
+        }
+      result
```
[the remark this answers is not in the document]
chick: will do

src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -288,15 +305,63 @@ object JsonProtocol {
...
+    // I don't much like this var here, but it has made it much simpler
+    // to maintain backward compatibility with the exception test structure
+    var classNotFoundBuildingLoaded = false
     val classes = findTypeHints(annos, true)
-    val loaded = classes.map(Class.forName(_))
+    val loaded = classes.map { x =>
+      val result =
+        try {
+          val result = Class.forName(x)
+          result
```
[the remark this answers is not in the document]
chick: will do

src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -4,14 +4,18 @@ package firrtl
...
 
 import scala.util.{Failure, Success, Try}
-
 import org.json4s._
 import org.json4s.native.JsonMethods._
 import org.json4s.native.Serialization
 import org.json4s.native.Serialization.{read, write, writePretty}
 
+import scala.collection.mutable
+import scala.collection.mutable.ArrayBuffer
```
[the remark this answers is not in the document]
chick: will do

src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -266,9 +278,14 @@ object JsonProtocol {
...
 
+  /** Deserialize JSON input into a Seq[Annotation]
+    *
+    * @param in JsonInput, can be file or string
+    * @return
+    */
   def deserialize(in: JsonInput): Seq[Annotation] = deserializeTry(in).get
 
-  def deserializeTry(in: JsonInput): Try[Seq[Annotation]] = Try({
+  def deserializeTry(in: JsonInput): Try[Seq[Annotation]] = Try {
```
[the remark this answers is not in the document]
chick: I don't see the paren after the `Try` not sure why, I'll call this done

src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -288,15 +305,63 @@ object JsonProtocol {
...
     implicit val formats = jsonFormat(loaded)
-    read[List[Annotation]](in)
-  }).recoverWith {
+    try {
+      read[List[Annotation]](in)
+    } catch {
+      case e: org.json4s.MappingException =>
+        // If we get here, the build `read` failed to process an annotation
+        // So we will map the annos one a time, wrapping the JSON of the unrecognized annotations
+        var exceptionList = new ArrayBuffer[Exception]()
```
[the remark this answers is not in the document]
chick: Agreed, will do

src/main/scala/firrtl/annotations/JsonProtocol.scala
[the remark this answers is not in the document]
jackkoenig: This isn't a comment about the deleted parentheses (I have no strong opinion about the stylistic choice either way), but rather that we need another argument to this function in case the flag was passed on the command line.

src/main/scala/firrtl/annotations/Annotation.scala
```diff
@@ -165,3 +166,5 @@ object Annotation
 case class DeletedAnnotation(xFormName: String, anno: Annotation) extends NoTargetAnnotation {
   override def serialize: String = s"""DELETED by $xFormName\n${anno.serialize}"""
 }
+
```
mwachs5, line 169: can this have scaladoc added?

src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -22,43 +25,70 @@ trait HasSerializationHints {
...
       extends CustomSerializer[Class[_ <: Transform]](format =>
         (
-          { case JString(s) => Class.forName(s).asInstanceOf[Class[_ <: Transform]] },
-          { case x: Class[_] => JString(x.getName) }
+          {
+            case JString(s) => Class.forName(s).asInstanceOf[Class[_ <: Transform]]
+          },
+          {
+            case x: Class[_] => JString(x.getName)
+          }
```
jackkoenig, line 39: This change just seems to be adding [unnecessary] whitespace? Why?

jackkoenig: Same for many below, seems to make it verbose for not any real readability benefit. It also distracts from the PR with lots of unrelated diffs.

src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -277,6 +419,18 @@ object JsonProtocol {
...
+
+    /* Tries to extract class name from the mapping exception */
+    def getAnnotationNameFromMappingException(mappingException: MappingException): String = {
+      try {
+        val GetClassPattern(name) = mappingException.getMessage
+        name
+      } catch {
+        case _: Exception =>
+          mappingException.getMessage
+      }
```
jackkoenig, line 431:
```suggestion
      mappingException.getMessage match {
        case GetClassPattern(name) => name
        case other                                => other
      }
```
Simpler and more canonical style.

src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -75,135 +105,238 @@ object JsonProtocol {
...
+          {
+            case JString(s) => Parser.parseType(s).asInstanceOf[GroundType]
+          },
+          {
+            case tpe: GroundType => JString(tpe.serialize)
+          }
+        )
+      )
+
+  class UnrecognizedAnnotationSerializerSerializer
```
jackkoenig, line 331: Why does this have `Serializer` twice? Isn't it just an "UnrecognizedAnnotationSerializer"?

src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -217,7 +350,8 @@ object JsonProtocol {
       new LoadMemoryFileTypeSerializer + new IsModuleSerializer + new IsMemberSerializer +
       new CompleteTargetSerializer + new TypeSerializer + new ExpressionSerializer +
       new StatementSerializer + new PortSerializer + new DefModuleSerializer +
-      new CircuitSerializer + new InfoSerializer + new GroundTypeSerializer
+      new CircuitSerializer + new InfoSerializer + new GroundTypeSerializer +
+      new UnrecognizedAnnotationSerializerSerializer
```
jackkoenig, line 354: Note that this matches on every possibly annotation. I suspect that whatever is last will always be the lowest priority, but are we absolutely certain of that fact? I think there should at minimum be a comment here to prevent someone from adding a Serializer after it which would never be used because this one will always match.

src/main/scala/firrtl/annotations/JsonProtocol.scala
jackkoenig: If you're trying to say that it only serializes (and doesn't deserialize) I'd name it something like `UnrecognizedAnnotationSerializerOnly` or something.

src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -288,27 +442,86 @@ object JsonProtocol {
...
     implicit val formats = jsonFormat(loaded)
-    read[List[Annotation]](in)
-  }).recoverWith {
+    try {
+      read[List[Annotation]](in)
+    } catch {
+      case e: org.json4s.MappingException =>
+        // If we get here, the build `read` failed to process an annotation
+        // So we will map the annos one a time, wrapping the JSON of the unrecognized annotations
+        var exceptionList = new mutable.ArrayBuffer[String]()
```
jackkoenig, line 470:
```suggestion
        val exceptionList = new mutable.ArrayBuffer[String]()
```
It's already mutable, it should be a val.

src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -288,27 +442,86 @@ object JsonProtocol {
...
+        val firrtlAnnos = annos.map { jsonAnno =>
+          try {
+            jsonAnno.extract[Annotation]
+          } catch {
+            case mappingException: org.json4s.MappingException =>
+              exceptionList += getAnnotationNameFromMappingException(mappingException)
```
jackkoenig, line 476:
```suggestion
              exceptionList += jsonAnno.getClass.getName
```
I believe this will do the same thing without extra regex right?

src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -288,27 +442,86 @@ object JsonProtocol {
...
     val classes = findTypeHints(annos, true)
-    val loaded = classes.map(Class.forName(_))
+    val loaded = classes.map { x =>
+      try {
+        Class.forName(x)
+      } catch {
+        case _: java.lang.ClassNotFoundException =>
+          classNotFoundBuildingLoaded = true // tells us which Exception to throw in recovery
+          // Found an annotation we don't recognize, So add UnrecognizedAnnotation to `loaded`
+          Class.forName("firrtl.annotations.UnrecognizedAnnotation")
```
jackkoenig, line 460: Why are we adding `UnrecognizedAnnotation` here? I believe this basically doesn't do anything. Why not just change to a `flatMap` and do `None` here?

src/test/scala/firrtlTests/AnnotationTests.scala
```diff
@@ -564,7 +564,7 @@ class JsonAnnotationTests extends AnnotationTests {
     val manager = setupManager(Some(anno))
 
     the[Exception] thrownBy Driver.execute(manager) should matchPattern {
-      case InvalidAnnotationFileException(_, _: AnnotationClassNotFoundException) =>
+      case InvalidAnnotationFileException(_, _: UnrecogizedAnnotationsException) =>
     }
   }
```
jackkoenig, line 569: I suspect, but am not certain, that this change will change the error you receive if you have an unserializable annotation (eg. an annotation that just wraps a random non-case class). Can you add a test here showing what happens in that case?

jackkoenig (changes requested):

src/test/scala/firrtlTests/annotationTests/UnrecognizedAnnotationSpec.scala
```diff
@@ -0,0 +1,344 @@
...
+        "high",
+        "-o",
+        fileNames.firrtlOutput,
+        "--output-annotation-file",
+        fileNames.outputAnnotations
+      )
+  }
+}
+
+object UnrecognizedAnnotationTextGenerator {
```
jackkoenig, line 173: I get using real test cases but this is like 170 lines of stuff that doesn't really add anything. I think it's be better just to make up 1 or two small, artificial annotations and leave it at that.

src/main/scala/firrtl/annotations/Annotation.scala
chick: I could but shouldn't an unrelated tests like this go in a separate PR?

src/main/scala/firrtl/annotations/JsonProtocol.scala
chick: Yes, of course.

src/main/scala/firrtl/annotations/JsonProtocol.scala
chick: One serializer should be enough

src/main/scala/firrtl/annotations/JsonProtocol.scala
chick: yup

src/main/scala/firrtl/annotations/JsonProtocol.scala
chick: This change did not work. In fact blew up with `Reporter completed abruptly with an exception after receiving event`

src/main/scala/firrtl/annotations/JsonProtocol.scala
chick: I found this to be necessary to get things to work. I tried using flatMap several different ways and it just didn't work right. I used several variants of 
```scala
    val loaded = classes.flatMap { x =>
      try {
        Some(Class.forName(x))
      } catch {
        case _: java.lang.ClassNotFoundException =>
          classNotFoundBuildingLoaded = true 
          None
      }
    }
```
 but always got some form of 
```
[error] /Users/chickm/SiFive/code/firrtl/src/main/scala/firrtl/annotations/JsonProtocol.scala:331:33: no type parameters for method flatMap: (f: String => scala.collection.GenTraversableOnce[B])(implicit bf: scala.collection.generic.CanBuildFrom[List[String],B,That])That exist so that it can be applied to arguments (String => Iterable[Class[?0]] forSome { type ?0 })
```

src/test/scala/firrtlTests/annotationTests/UnrecognizedAnnotationSpec.scala
chick: Sure. Reduced the cruft to just the class names

src/main/scala/firrtl/annotations/Annotation.scala
jackkoenig: @chick I think you may have responded to the wrong cmoment

src/main/scala/firrtl/annotations/JsonProtocol.scala
jackkoenig: Works for me but since this is only for error messages, I guess a regex is fine. We should at least make the APIs private though.

src/main/scala/firrtl/annotations/JsonProtocol.scala
jackkoenig: Type inference is struggling, it works if you give it a hint:
```scala
    val loaded = classes.flatMap { x =>
      (try {
        Some(Class.forName(x))
      } catch {
        case _: java.lang.ClassNotFoundException =>
          classNotFoundBuildingLoaded = true 
          None
      }): Option[Class[_]]
    }
```
That being said the current code is probably fine, likely marginally slower than flattening them out but I doubt it's even measurable.

jackkoenig (changes requested):
A couple of minor requests but otherwise this looks good to :shipit:

src/main/scala/firrtl/annotations/Annotation.scala
chick: @jackkoenig @mwachs5 I meant to say, adding scaladoc to pre-existing code that is unrelated to this PR should go on a separate PR. But I'm happy to add some. Any thoughts on what it should say, I feel like DeletedAnnotations are heading towards deprecation and using them, particularly looking inside them, is an anti-pattern

src/main/scala/firrtl/annotations/JsonProtocol.scala
chick: The hint solved it, thanks

src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -277,6 +300,15 @@ object JsonProtocol {
...
         )
     }
+
+    /* Tries to extract class name from the mapping exception */
+    def getAnnotationNameFromMappingException(mappingException: MappingException): String = {
+      mappingException.getMessage match {
+        case GetClassPattern(name) => name
+        case other                 => other
+      }
+    }
```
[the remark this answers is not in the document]
chick: Good idea.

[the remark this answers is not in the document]
chick: This is an internal method to def deserializeTry, I don't think private works here

src/main/scala/firrtl/annotations/JsonProtocol.scala
```diff
@@ -22,7 +26,9 @@ trait HasSerializationHints {
 /** Wrapper [[Annotation]] for Annotations that cannot be serialized */
 case class UnserializeableAnnotation(error: String, content: String) extends NoTargetAnnotation
 
-object JsonProtocol {
+object JsonProtocol extends LazyLogging {
+  val GetClassPattern = "[^']*'([^']+)'.*".r
```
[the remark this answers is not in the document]
chick: Will do

src/test/scala/firrtlTests/AnnotationTests.scala
chick: Does the second to last test in AnnotationTests demonstrate this?

jackkoenig (approved):

PR #2393, opened 2021-10-23
Modify and optimize performance of propagate annotations

jackkoenig:
This is basically just deletes all of the logic we had determining which annotations were added or deleted between transforms. This does not change any Scala APIs, but it very much changes what sorts of DeletedAnnotations are created. The deleted behavior could be useful for debugging and technically *could* be relied upon someone, but my hunch is that no one cares and we're harming performance and burning carbon for nothing. If it turns out someone relies on this, I think we should add it back under some sort of "debug" annotation.

This change only really affects annotation-heavy flows. I found that with small designs with moderate numbers of annotations, it improves performance ~2-7%. For large designs with tons of annotations however, it improved performance by **34%** (15m45 -> 10m20s)

This is a bit aggressive, but I think it's worth it.

### Contributor Checklist

#### Type of Improvement

 - performance improvement            
 - code refactoring                   
 - code cleanup                    

#### API Impact

Annotations that are deleted between transforms are no longer automatically captured as `DeletedAnnotations`

No direct impact

#### Desired Merge Strategy

 - Squash

#### Release Notes

When a transform deletes an `Annotation`, a `DeletedAnnotation` is no longer created. Furthermore, debug-mode logging no longer captures any information about annotation changes between transforms. If any of this behavior is desired, please file an issue to restore it under a debug flag.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/Compiler.scala
+++ b/src/main/scala/firrtl/Compiler.scala
@@ -17,11 +17,11 @@ import firrtl.stage.Forms
 import firrtl.transforms.DedupAnnotationsTransform
 
 /** Container of all annotations for a Firrtl compiler */
-class AnnotationSeq private (private[firrtl] val underlying: List[Annotation]) {
-  def toSeq: Seq[Annotation] = underlying.toSeq
+class AnnotationSeq private (underlying: Seq[Annotation]) {
+  def toSeq: Seq[Annotation] = underlying
 }
 object AnnotationSeq {
-  def apply(xs: Seq[Annotation]): AnnotationSeq = new AnnotationSeq(xs.toList)
+  def apply(xs: Seq[Annotation]): AnnotationSeq = new AnnotationSeq(xs)
 }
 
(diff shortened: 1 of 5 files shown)


jackkoenig:
I've revived this PR, it should be passing tests now. I know this is a bit of a change, but I think it's worth including in 1.5.0-RC2 and then if people have issues with it we can either revert or add an option to include the old `DeletedAnnotation` behavior.

ekiwi (approved):
LGTM

I was never a fan of the "DeletedAnnotation". Seemed like a memory leak to me.

jackkoenig:
Lets go for it. We can revert or fix later if it's an issue

jackkoenig:
I believe this PR has exposed a previously hidden bug, you can see it manifest in the chisel3 integration tests: https://github.com/chipsalliance/chisel3/runs/4563876893?check_suite_focus=true

It also manifests in chiseltest's own tests... I'm looking into it.

PR #2397, opened 2021-10-27
Add CustomRadixGTKWaveAnnotation to generate Tcl script for GTKWave

sequencer:
depends on #2434 
GTKWave will invoke script to generate readable signal below: 
telegram-cloud-photo-size-1-4913540722787985897-y

### Contributor Checklist

#### Type of Improvement
- new feature/API

#### API Impact

Add API `CustomRadix{Def,Apply}Annotation` `CustomRadixGTKWaveAnnotation` and `CustomRadixConfigFileAnnotation`,
Add command line API: `--wave-viewer-script`

None

#### Desired Merge Strategy

#### Release Notes
None

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/stage/FirrtlCli.scala
+++ b/src/main/scala/firrtl/stage/FirrtlCli.scala
@@ -4,7 +4,7 @@ package firrtl.stage
 
 import firrtl.options.Shell
 import firrtl.passes.CommonSubexpressionElimination
-import firrtl.transforms.NoCircuitDedupAnnotation
+import firrtl.transforms.{CustomRadixTransform, NoCircuitDedupAnnotation}
 
 /** [[firrtl.options.Shell Shell]] mixin that provides command line options for FIRRTL. This does not include any
   * [[firrtl.options.RegisteredLibrary RegisteredLibrary]] or [[firrtl.options.RegisteredTransform RegisteredTransform]]
@@ -28,7 +28,8 @@ trait FirrtlCli { this: Shell =>
     OptimizeForFPGA,
     CurrentFirrtlStateAnnotation,
     CommonSubexpressionElimination,
-    AllowUnrecognizedAnnotations
+    AllowUnrecognizedAnnotations,
+    CustomRadixTransform
   )
     .map(_.addOptions(parser))
 
--- a/src/main/scala/firrtl/transforms/CustomRadixTransform.scala
+++ b/src/main/scala/firrtl/transforms/CustomRadixTransform.scala
@@ -63,6 +63,67 @@ case class CustomRadixConfigFileAnnotation(
   }.getBytes
 }
 
+/** Dumps a Tcl script for GTKWave on Custom Radix, using GTKWave's `setCurrentTranslateEnums` and `installFileFilter`
+  * If you want to generate the script through [[AnnotationSeq]], use:
+  * {{{
+  *   Seq(
+  *     GTKWaveCustomRadixScriptAnnotation(Seq(), Seq()),
+  *     RunFirrtlTransformAnnotation(Dependency(CustomRadixTransform))
+  *   )
+  * }}}
+  *
+  * You can pass
+  * {{{
+  *   Array("--wave-viewer-script", "gtkwave")
+  * }}}
+  * to `args` for the sake of ease.
+  *
+  * @param signals which alias contains which signals, the signals should be converted from ReferenceTarget to String
+  * @param filters sequence of [[CustomRadixDefAnnotation]], the name should match [[signals]].map(_._1)
+  */
+case class GTKWaveCustomRadixScriptAnnotation(
+  signals: Seq[(AliasName, Seq[String])],
+  filters: Seq[CustomRadixDefAnnotation])
+    extends CustomRadixScriptAnnotation {
+  def waveViewer: String = "gtkwave"
+  def suffix:     Option[String] = Some(".tcl")
+
+  private def toStr(bigInt: BigInt, width: Int, base: Int) =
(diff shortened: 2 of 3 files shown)


src/main/scala/firrtl/annotations/SignalTranslateAnnotation.scala
```diff
@@ -0,0 +1,6 @@
+package firrtl.annotations
+
+case class SignalTranslateAnnotation(target: ReferenceTarget, definition: Seq[(BigInt, String)], width: Int)
+    extends SingleTargetAnnotation[ReferenceTarget] {
+  override def duplicate(n: ReferenceTarget): Annotation = this.copy(n)
+}
```
sequencer, line 6: Review this in #2408

jackkoenig (commented):
It might be nice to generate a data file instead of the GTK `.tcl` and then instead provide a static `.tcl` file that reads the data file. This might extend better to other tools since someone using something proprietary could write their own `.tcl` to read the same data file.

Another option is to look at generate `.gtkw` files directly. You can see what these are like by manually creating the mapping inside of an instance of GTKWave and then saving it to see what it looks like for the mappings you've created.

ekiwi:
I still think this should be part of Chisel and not firrtl. Generally I would prefer if chisel users never need to import things from the firrtl package. The firrtl backend should remain an implementation detail since eventually one might be able to use Chisel with CIRCT and no dependency on firrtl. So everything that users are expected to use directly should be in a `chisel3` (sub)package.

sequencer:
> I still think this should be part of Chisel and not firrtl. Generally I would prefer if chisel users never need to import things from the firrtl package.

Chisel `StrongEnum` will depend on `CustomRadix` Annotations, and `CustomRadix` doesn't provide any chisel-related API. So I think this is reasonable to provide this as a standalone API in FIRRTL. 

> The firrtl backend should remain an implementation detail since eventually one might be able to use Chisel with CIRCT and no dependency on firrtl.

I think in the long-term decision, as @seldridge said, FIRRTL should improve the IR support to custom radix(enum), and directly emit correspond IR to FIRRTL/CIRCT.

> So everything that users are expected to use directly should be in a chisel3 (sub)package.

Yes, user won't touch these APIs, they only use `StrongEnum`, and `StrongEnum` will generate correspond API, if user doesn't provide a `CustomRadixConfigFileAnnotation`, the `CustomRadixTransform` won't execute and there will be no behavior changes.

In general, I think in this PR, we only leverage the dependency from Chisel to FIRRTL, original `StrongEnumAnnotation` doesn't do any thing, so I think it's reasonable to re-implement this.

sequencer:
> It might be nice to generate a data file instead of the GTK .tcl and then instead provide a static .tcl file that reads the data file. This might extend better to other tools since someone using something proprietary could write their own .tcl to read the same data file.

We have tried, tcl sucks :(
So we think a trade-off might be good: providing a JSON output by default, if other users prefer directly using the raw data. If user explicitly selected the `gtkwave` via args, the correspond `tcl` script will be emitted. 

> Another option is to look at generate .gtkw files directly. You can see what these are like by manually creating the mapping inside of an instance of GTKWave and then saving it to see what it looks like for the mappings you've created.

I investigated this, and found this might not be a good solution: one GTKWave session will only have one `gtkw` file, which makes it impossible to let user using their own `gtkw`(if they have), so executing a `tcl` might be a reasonable solution instead.

jackkoenig:
I'm fine with this being in firrtl but I agree with @ekiwi that the user API needs to be fully in chisel3. The issue is the annotation to specify what file you want emitted. I think we can have a different annotation in chisel3 that turns into one of these firrtl annotations

sequencer:
In my point of view, The reason I think this API should exist in FIRRTL is:
1. These codes doesn't depend on any Chisel codes.
2. User API is exposed via command-line args: (`--wave-viewer-script`, `gtkwave`), which can be directly used by any Chisel users. we can add docs to [`docs`](https://github.com/chipsalliance/chisel3/blob/master/docs/src/explanations/chisel-enum.md)
3. If user choose to use annotation to generate the gtkwave tcl: they need to append these annotations:
```
Seq(
  firrtl.transforms.RunFirrtlTransformAnnotation(Dependency(CustomRadixTransform)),
  firrtl.transforms.GTKWaveCustomRadixScriptAnnotation(Seq.empty, Seq.empty)
)
```
AFAIK, This pattern is universally used in Chisel designs, for example:
if user needs to select output dir: `firrtl.stage.TargetDirAnnotation` is required.
if user needs to infer read-write of SRAM: `firrtl.passes.memlib.InferReadWriteAnnotation, firrtl.passes.memlib.RunFirrtlTransformAnnotation(new InferReadWrite)` are required
if user needs to change log level: `logger.LogLevelAnnotation`(which is vended by firrtl)
Those APIs are APIs in my daily uses, I usually import them from firrtl package. So I design this API via this guideline: common user can always access APIs from command-line without importing anything from firrtl. If user choose to explicit use the Annotation in their API, it's OK to add API which belongs to firrtl.

jackkoenig:
@sequencer does this need to be rebased since https://github.com/chipsalliance/firrtl/pull/2434 was merged?

sequencer:
Yes , it will be rebased to master. I thought this won’t block 3.5-RC2? It
it did I’ll work on it today.

sequencer:
I’m still finding a better way to invoke gtkwave ;p

jackkoenig:
>  I thought this won’t block 3.5-RC2

Don't worry, it's not blocking RC2! I've just been looking through PRs so thought I'd follow up!

sequencer:
Ok! I’ll finish it this weekend ;p

PR #2404, opened 2021-11-03
Add SuggestSeqMemName annotation to rename replaced blackboxes

jared-barocsi:
This PR implements an annotation (`SuggestSeqMemNameAnnotation`) to rename a memory blackbox to a user-specified name instead of the default `{mem name}_ext`.

### Contributor Checklist

#### Type of Improvement

- new feature/API

#### API Impact

No new API

Memory blackboxes get renamed with th

#### Desired Merge Strategy

Squash and merge

#### Release Notes
Implement a mechanism to rename blackboxes generated by `ReplSeqMems` with a user-specified name

- [ ] Did you review?

### DIFF
--- a/src/test/scala/firrtlTests/ReplSeqMemTests.scala
+++ b/src/test/scala/firrtlTests/ReplSeqMemTests.scala
@@ -693,4 +693,37 @@ circuit Top :
          |""".stripMargin
     compileAndEmit(CircuitState(parse(input), ChirrtlForm))
   }
+
+  "ReplSeqMem" should "rename blackbox instances to suggested names" in {
+    val input =
+      """
+        |circuit CustomMemory :
+        |  module CustomMemory :
+        |    input clock : Clock
+        |    input reset : UInt<1>
+        |    output io : {flip rClk : Clock, flip rAddr : UInt<3>, dO : UInt<16>, flip wClk : Clock, flip wAddr : UInt<3>, flip wEn : UInt<1>, flip dI : UInt<16>}
+        |    io is invalid
+        |    smem mem_0 : UInt<16>[7]
+        |    read mport _T_17 = mem_0[io.rAddr], clock
+        |    io.dO <= _T_17
+        |    when io.wEn :
+        |      write mport _T_18 = mem_0[io.wAddr], clock
+        |      _T_18 <= io.dI
+        |""".stripMargin
+    val mems = Set(
+      MemConf("renamed_mem", 7, 16, Map(WritePort -> 1, ReadPort -> 1), None)
+    )
+    val confLoc = "ReplSeqMemTests.confTEMP"
+    val annos = Seq(
+      ReplSeqMemAnnotation.parse("-c:CustomMemory:-o:" + confLoc),
+      SuggestSeqMemNameAnnotation(
+        CircuitTarget("CustomMemory")
+          .module("CustomMemory")
+          .ref("mem_0"),
+        "renamed_mem"
+      )
+    )
+    val res = compileAndEmit(CircuitState(parse(input), ChirrtlForm, annos))
+    checkMemConf(res, mems)
+  }
 }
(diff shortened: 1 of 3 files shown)


jackkoenig (approved):
This looks great! Just a couple of minor nits

seldridge:
As an alternative approach... would it make sense to just keep the user-provided name of the memory for the blackbox and change the name of the blackbox that gets generated?

mwachs5 (commented):
does this change the instance names as well as the module names?

src/test/scala/firrtlTests/ReplSeqMemTests.scala
```diff
@@ -693,4 +693,37 @@ circuit Top :
...
+      SuggestSeqMemNameAnnotation(
+        CircuitTarget("CustomMemory")
+          .module("CustomMemory")
+          .ref("mem_0"),
+        "renamed_mem"
+      )
+    )
+    val res = compileAndEmit(CircuitState(parse(input), ChirrtlForm, annos))
+    checkMemConf(res, mems)
+  }
```
jackkoenig, line 728: This test is showing the memory name in `MemConf`, but it would be really helpful if it could also show what the instance and module names are for the wrapper module, as well as the instance of the blackbox.

jackkoenig (changes requested):
> As an alternative approach... would it make sense to just keep the user-provided name of the memory for the blackbox and change the name of the blackbox that gets generated?

@seldridge can you clarify this? You referred to the name of the blackbox twice.

> does this change the instance names as well as the module names?

@mwachs5 great question, @jared-barocsi see my new comment on the test

The main goal here is that we want to be able to prefix the names of the modules that are creates by ReplSeqMems, it also seemed like some user control over the naming made sense. We want this to compose with a built-in prefixing API in Chisel3 so maybe some feedback in that context would be helpful as well.

src/test/scala/firrtlTests/ReplSeqMemTests.scala
mwachs5: is the goal to change the instance name? I dont think in general we *want* to, but what it was doing before that the instance name was the same as the blackbox name... which IMHO is strange but that's the existing behavior

seldridge:
@jackkoenig wrote:
> @seldridge can you clarify this? You referred to the name of the blackbox twice.

Ack. Fixed above. 

It seems like a cleaner API is to have the memory blackbox inherit the name of the original memory (`foo` not `foo_ext`) and to instead mangle the name of the _wrapping_ module (e.g., use `foo_wrapper` instead of `foo`).  Then Chisel is in complete control of the final memory name and there is not a need for another API to control the specific naming behavior of a specific, optional pass.

src/test/scala/firrtlTests/ReplSeqMemTests.scala
jackkoenig: I would say we do **not** want to change the instance name

mwachs5: I guess by "change" there is "same as it was before this PR" or "change based on what the person passes in"... which do you mean we do NOT want to do

jackkoenig:
> It seems like a cleaner API is to have the memory blackbox inherit the name of the original memory (foo not foo_ext) and to instead mangle the name of the wrapping module (e.g., use foo_wrapper instead of foo). Then Chisel is in complete control of the final memory name and there is not a need for another API to control the specific naming behavior of a specific, optional pass.

I think this is a reasonable suggestion, but it either doesn't fulfill the intended use case or would be a bit weird. The Prefixing API needs to make sure that all modules use the prefix, so this includes both the wrapper module and the blackbox. Using your suggestion, that means Chisel would have to give the prefix to the _instance name_ of the memory which is a bit odd, I think that should still be derived from the name of the `val`. That or with your suggestion, users who want prefixing to affect repl seq mem would have to wrap every SyncReadMem in basically `prefix(Module.currentModulePrefix)` which wouldn't really work when trying to reuse other people's code (eg. Queue). 

This is, of course, dancing around the need for packages/namespaces, and it's basically a huge pain that we can prefix all modules in Chisel (via https://github.com/chipsalliance/chisel3/pull/2155) but then ReplSeqMems decides to just screw everything up.

mwachs5:
> that means Chisel would have to give the prefix to the instance name of the memory which is a bit odd, 

We should NOT give the prefix to the instance name of the memory.

PR #2408, opened 2021-11-08
implement SignalTranslateAnnotation

sequencer:


src/main/scala/firrtl/annotations/SignalTranslateAnnotation.scala
```diff
@@ -0,0 +1,6 @@
+package firrtl.annotations
+
+case class SignalTranslateAnnotation(target: ReferenceTarget, definition: Seq[(BigInt, String)], width: Int)
+    extends SingleTargetAnnotation[ReferenceTarget] {
+  override def duplicate(n: ReferenceTarget): Annotation = this.copy(n)
+}
```
seldridge, line 6: Couple of different thoughts...

The formulation of this is attaching full information about the enumerated type to every use of it, yes?  An alternative formulation would be to use two annotations:

1) Enum definition annotation which defines the `Seq[(BigInt, String)]` mapping.
2) A second annotation, on each bind, that associates that value with the enum definition.

I don't know exactly what this would look like, but that would avoid the redundancy of the `Seq[(BigInt, String)]` on every use.

Why does this need to encode the width? Can that field be dropped? That information should be implicit in the FIRRTL IR (or an unknown width if the signal has an unknown width).

seldridge (commented):
High level, seems fine. One alternative design point to consider in the review.

seldridge:
Going one step further... it would really be nice if instead of more annotations, we expanded FIRRTL IR to add support for user-defined, enumerated types. 😅

ekiwi (changes requested):
This needs scaladoc (unless you can make it package private) as well as some unit tests. As is, I don't really understand what the intended use-case for this is.

sequencer:
merged to #2397

PR #2427, opened 2021-11-21
[deprecation clean up] remove firrtl.stage.transforms.CheckScalaVersion.

sequencer:
### Contributor Checklist

- [ ] Did you review?


jackkoenig (changes requested):
This is just deprecated as of 1.5 so we should wait till 1.6 to remove it. We can merge this PR after the 3.5.0 release.

sequencer:
We have dropped 2.11 in 1.5, maybe this is a dead code since 1.5?

jackkoenig:
> We have dropped 2.11 in 1.5, maybe this is a dead code since 1.5?

It is dead code in 1.5, but it wasn't dead in 1.4. Thus if someone were referring to this code in their 1.4 code, upgrading to 1.5 will just break it. Thus instead we deprecate in 1.5 and remove in 1.6.

sequencer:
> It is dead code in 1.5, but it wasn't dead in 1.4. Thus if someone were referring to this code in their 1.4 code, upgrading to 1.5 will just break it. Thus instead we deprecate in 1.5 and remove in 1.6.

Got it! So in this PR, I think we can delete references to `CheckScalaVersion`, and keep `CheckScalaVersion` object as a user API?

sequencer:
Another idea is back port deprecation to 1.4 branch and remove this code in master, will this be OK?

jackkoenig:
> Got it! So in this PR, I think we can delete references to CheckScalaVersion, and keep CheckScalaVersion object as a user API?

We should just keep the deprecated API as is in 1.5 and as soon as 1.5.0 is released we can merge this PR for release in 1.6

> Another idea is back port deprecation to 1.4 branch and remove this code in master, will this be OK?

No, because the code isn't dead in 1.4, it's perfectly legal for a user to being using it so they shouldn't see deprecation warnings.

PR #2434, opened 2021-12-04
Implement CustomRadixTransform for wave viewers

sequencer:
This PR implements:

1. Add CustomRadix{Def,Apply}Annotation to define and apply custom radix.
2. Add CustomRadixConfigFileAnnotation to output a JSON config file so users can generate scripts on their own.

IMHO, I don't this this should be added to Chisel. However if @ekiwi really don't like this in FIRRTL, I can make a PR in Chisel as well. And splitting GTKWave/Verdi plugin out of FIRRTL/Chisel upstream implementation. The main purpose of this PR is speeding up chipsalliance/chisel3#2239 which is necessary to get rid of the bad legacy codes of Enum related Annotation.
 
### Contributor Checklist

#### Type of Improvement
- new feature/API

#### API Impact
Add new Annotation to define custom radix for being able to add name to signal values.

None

#### Desired Merge Strategy

#### Release Notes
None

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/transforms/CustomRadixTransform.scala
+++ b/src/main/scala/firrtl/transforms/CustomRadixTransform.scala
@@ -7,7 +7,6 @@ import firrtl.annotations.{Annotation, NoTargetAnnotation, ReferenceTarget, Sing
 import firrtl.options.{CustomFileEmission, Dependency, HasShellOptions, ShellOption}
 import firrtl.stage.TransformManager.TransformDependency
 import firrtl.stage.{Forms, RunFirrtlTransformAnnotation}
-import firrtl.transforms.CustomRadixTransform.{AliasName, AliasValue}
 import firrtl.{AnnotationSeq, CircuitState, DependencyAPIMigration, Transform}
 
 /** Contains a static map from signal value(BigInt) to signal name(String)
@@ -17,22 +16,16 @@ import firrtl.{AnnotationSeq, CircuitState, DependencyAPIMigration, Transform}
   * @param filters a sequence of translation filter
   * @param width width of this alias
   */
-case class CustomRadixDefAnnotation(name: AliasName, filters: Seq[AliasValue], width: Int) extends NoTargetAnnotation
+case class CustomRadixDefAnnotation(name: String, filters: Seq[(BigInt, String)], width: Int) extends NoTargetAnnotation
 
 /** A annotation making a ReferenceTarget to be a specific [[CustomRadixDefAnnotation]].
   *
   * @param target the ReferenceTarget which the alias applied to
   * @param name the identifier for the alias
   */
-case class CustomRadixApplyAnnotation(target: ReferenceTarget, name: AliasName)
+case class CustomRadixApplyAnnotation(target: ReferenceTarget, name: String)
     extends SingleTargetAnnotation[ReferenceTarget] {
   override def duplicate(n: ReferenceTarget): Annotation = this.copy(n)
-}
-
-private[firrtl] trait CustomRadixScriptAnnotation extends NoTargetAnnotation with CustomFileEmission {
-  def waveViewer: String
-
-  final def baseFileName(annotations: AnnotationSeq): String = "custom_radix_" + waveViewer
 }
 
 /** Dumps a JSON config file for custom radix. Users can generate script using the emitted file.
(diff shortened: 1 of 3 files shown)


src/main/scala/firrtl/transforms/CustomRadixTransform.scala
```diff
@@ -0,0 +1,109 @@
...
+          .split(',')
+          .map {
+            case "json" | "" => CustomRadixConfigFileAnnotation(Seq.empty, Seq.empty)
+          }
+          .toSeq
+      },
+      helpText = "<json>, you can combine them like 'json', pass empty string will generate json",
+      shortOption = None
+    )
+  )
```
sequencer, line 86: If FIRRTL doesn't like GTKWave tcl generator, this should be refactored.

src/main/scala/firrtl/transforms/CustomRadixTransform.scala
```diff
@@ -0,0 +1,109 @@
...
+case class CustomRadixApplyAnnotation(target: ReferenceTarget, name: AliasName)
+    extends SingleTargetAnnotation[ReferenceTarget] {
+  override def duplicate(n: ReferenceTarget): Annotation = this.copy(n)
+}
+
+private[firrtl] trait CustomRadixScriptAnnotation extends NoTargetAnnotation with CustomFileEmission {
+  def waveViewer: String
+
+  final def baseFileName(annotations: AnnotationSeq): String = "custom_radix_" + waveViewer
+}
```
sequencer, line 36: If FIRRTL doesn't like GTKWave tcl generator, this should be refactored.

ekiwi:
I am OK with this if @jackkoenig can review.

src/main/scala/firrtl/transforms/CustomRadixTransform.scala
```diff
@@ -0,0 +1,109 @@
...
+    )
+  }.getBytes
+}
+
+/** A Transform that generate scripts or config file for Custom Radix */
+object CustomRadixTransform extends Transform with DependencyAPIMigration with HasShellOptions {
+
+  private[firrtl] type AliasName = String
+  private[firrtl] type AliasValue = (BigInt, String)
+
```
jackkoenig, line 71: Not sure that these should be private, I'm curious what it looks like if a user not in package `firrtl` tries to construct or process a `CustomRadixDefAnnotation`

src/main/scala/firrtl/transforms/CustomRadixTransform.scala
```diff
@@ -0,0 +1,109 @@
...
+case class CustomRadixDefAnnotation(name: AliasName, filters: Seq[AliasValue], width: Int) extends NoTargetAnnotation
+
+/** A annotation making a ReferenceTarget to be a specific [[CustomRadixDefAnnotation]].
+  *
+  * @param target the ReferenceTarget which the alias applied to
+  * @param name the identifier for the alias
+  */
+case class CustomRadixApplyAnnotation(target: ReferenceTarget, name: AliasName)
+    extends SingleTargetAnnotation[ReferenceTarget] {
+  override def duplicate(n: ReferenceTarget): Annotation = this.copy(n)
```
jackkoenig, line 29: I think these generally look good. I'm a little nervous about the linking between the annotations via a String name since we don't really have namespacing support here, but I don't have a good solution in mind so I guess it's fine.

src/main/scala/firrtl/transforms/CustomRadixTransform.scala
jackkoenig, line 36: I don't think this trait is providing that much value (literally just adding `custom_radix_` prefix to the wave viewer name, so is it really worth having?

src/main/scala/firrtl/transforms/CustomRadixTransform.scala
```diff
@@ -0,0 +1,109 @@
...
+    import org.json4s.JsonDSL.WithBigDecimal._
+    import org.json4s.native.JsonMethods._
+    val aliasMap = signals.toMap
+    pretty(
+      render(
+        filters.map(a =>
+          a.name -> (("width" -> a.width) ~ ("values" -> a.filters.map {
+            case (int, str) => ("digit" -> int) ~ ("alias" -> str)
+          }) ~ ("signals" -> aliasMap(a.name)))
+        )
```
jackkoenig, line 60:
```suggestion
        filters.map { a =>
          val values = a.filters.map { case (int, str) =>
            ("digit" -> int) ~
            ("alias" -> str
          }
          a.name -> (
            ("width" -> a.width) ~
            ("values" -> values) ~
            ("signals" -> aliasMap(a.name))
          )
        }
```
I find this code very hard to read, so a extra val and some newlines would help (assuming scalafmt doesn't just destroy it lol)

jackkoenig (approved):
Generally looks good! A few comments but approving, please consider the comments before merging.

src/main/scala/firrtl/transforms/CustomRadixTransform.scala
sequencer: Signal has name space support here, while `name`(which should be the radix type name, and this won't be renamed) doesn't have. This is dangerous, if user defines multiple `CustomRadixDefAnnotation` with same `name`, I added check inside transforms to forbid this

src/main/scala/firrtl/transforms/CustomRadixTransform.scala
sequencer: OK, removed.

src/main/scala/firrtl/transforms/CustomRadixTransform.scala
sequencer: OK :)

PR #2440, opened 2021-12-09
Fix width of signed addition when input to mux

### LINKED ISSUE #2439: Truncating the result of signed addition
### Checklist

```
circuit Ops:
  module Ops:
    input sel: UInt<4>
    input is: SInt<8>
    output os: SInt<9>

    os <= SInt(0)

    when eq(sel, UInt(0)):
      os <= add(is, is)
```
 produces:
 ```verilog
module ops(
  input  [3:0] sel,
  input  [7:0] is,
  output [8:0] os
);
  assign os = sel == 4'h0 ? $signed($signed(is) + $signed(is)) : $signed(9'sh0); // @[]
endmodule
```
 
The expression `$signed($signed(is) + $signed(is))` performs the addition at 8 bits instead of 9.

The following two lines of verilog are not equivalent:
```verilog
  assign os = $signed($signed(is) + $signed(is))
  assign os = $signed(is) + $signed(is);
```

---

See https://www.edaplayground.com/x/jbLq to play around with this.

```verilog
module Foo(
  input        sel,
  input  [3:0] is,
  output [4:0] out1,
  output [4:0] out2
);
  assign out1 = $signed(is) + $signed(is);
  assign out2 = sel ? $signed($signed(is) + $signed(is)) : $signed(9'sh0);
endmodule
```
```
    for (int i = 0; i < 4'hf; ++i) begin
    is = i;
    #1;
    $display("%x + %x = %x vs. %x", is, is, out1, out2);
    end
```

```
Compiler version Q-2020.03-SP1-1; Runtime version Q-2020.03-SP1-1;  Dec  9 16:17 2021
00, 00
02, 02
04, 04
06, 06
08, 18
0a, 1a
0c, 1c
0e, 1e
10, 00
12, 02
14, 04
16, 06
18, 18
1a, 1a
1c, 1c
           V C S   S i m u l a t i o n   R e p o r t 
Time: 15 ns
```
---

When the addition is not used in another expression, the outer `$signed()` cast is not present:
```
circuit Ops:
  module Ops:
    input sel: UInt<4>
    input is: SInt<8>
    output os: SInt<9>

    os <= add(is, is)
```
produces:
```
module Ops(
  input  [3:0] sel,
  input  [7:0] is,
  output [8:0] os
);
  assign os = $signed(is) + $signed(is);
endmodule
```

The addition should be performed at 9 bits.

### Steps to Reproduce

Above.

jackkoenig:
Extremely focused fix for #2439. I'm not sure if this is the best fix but I think the test is pretty good at least so wanted to put this out there.

Fixes #2439

h/t to @youngar for finding an OG bug.

I'm wondering if it would be better to just remove the casts in the VerilogEmitter, do we ever actually need to cast the inputs to a mux? https://github.com/chipsalliance/firrtl/blob/4347f57b7ef201251fb2cd463124f2d7cea369f8/src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala#L182
That's just a lot scarier of a change 😨 

I'm backporting this as far as it will go...

### Contributor Checklist

#### Type of Improvement

  - bug fix        

#### API Impact

No impact

Minimal, will split out signed addition into another Statement if it is a direct input to a mux

#### Desired Merge Strategy

 - Squash

#### Release Notes

Fix width of signed addition when it is an input to a mux.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/passes/SplitExpressions.scala
+++ b/src/main/scala/firrtl/passes/SplitExpressions.scala
@@ -27,6 +27,14 @@ object SplitExpressions extends Pass {
     case _ => false
   }
 
+  private def isSignedArithmetic(e: Expression): Boolean = e match {
+    case DoPrim(PrimOps.Add, _, _, _: SIntType) => true
+    case DoPrim(PrimOps.Sub, _, _, _: SIntType) => true
+    case DoPrim(PrimOps.Mul, _, _, _: SIntType) => true
+    case DoPrim(PrimOps.Div, _, _, _: SIntType) => true
+    case _ => false
+  }
+
   private def onModule(m: Module): Module = {
     val namespace = Namespace(m)
     def onStmt(s: Statement): Statement = {
@@ -53,6 +61,10 @@ object SplitExpressions extends Pass {
       def onExp(e: Expression): Expression =
         e.map(onExp) match {
           case ex: DoPrim => ex.map(split)
+          // Arguably we should be splitting all Mux expressions but this has a negative impact on
+          // Verilog, instead this is a focused fix for
+          // https://github.com/chipsalliance/firrtl/issues/2439
+          case ex: Mux if isSignedArithmetic(ex.tval) || isSignedArithmetic(ex.fval) => ex.map(split)
           case ex => ex
         }
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -90,7 +90,10 @@ jobs:
         # By having this "if" here, this job returns success so that all_tests_passed will succeed too
         if: github.event_name == 'pull_request' &&
             ! contains(github.event.pull_request.labels.*.name, 'Skip Formal CI')
-        run: ./.run_formal_checks.sh ${{ matrix.design }}
+        run: |
+          echo ${{ github.event_name }}
+          echo ${{ github.event.pull_request.labels }}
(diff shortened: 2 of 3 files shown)


src/main/scala/firrtl/passes/SplitExpressions.scala
```diff
@@ -27,6 +27,11 @@ object SplitExpressions extends Pass {
     case _ => false
   }
 
+  private def isSignedAdd(e: Expression): Boolean = e match {
+    case DoPrim(PrimOps.Add, _, _, _: SIntType) => true
+    case _ => false
+  }
```
seldridge, line 33: This may be prefetching, too far, but maybe we want to do this for any arithmetic op of `SIntType`? E.g., I can produce a similar error with subtraction:

https://www.edaplayground.com/x/WAHv

jackkoenig: Definitely not prefetching too far. If you can produce the same error with other ops then we should fix those as well. I'm leaning toward the VerilogEmitter change instead since it should fix the whole class of issues but it makes me a little nervous.

seldridge (approved):

jackkoenig:
Skipping formal CI because it *should* fail

jackkoenig:
I added the label a bit late. Formal equiv for Ops failed (as it should because we're fixing it), now rerunning.

PR #2449, opened 2021-12-22
disable random should not contain useInitAsPreset.

sequencer:
My bad, I mistakenly introduced `useInitAsPreset = true` in `disableMemRandomization` console option.
This will introduce a critical bug:
https://scastie.scala-lang.org/ZRHZGEs1TsGoIdku7niUog

This is because use `useInitAsPreset` for `reg b`, while no init logic.
### Contributor Checklist

#### Type of Improvement
- bug fix

#### API Impact
none

bug fix for Verilog with `disableMemRandomization` option.

#### Desired Merge Strategy

#### Release Notes
None

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/Emitter.scala
+++ b/src/main/scala/firrtl/Emitter.scala
@@ -200,7 +200,7 @@ object EmitAllModulesAnnotation extends HasShellOptions {
         s.split(",")
           .map {
             case "disableMemRandomization" =>
-              CustomDefaultRegisterEmission(useInitAsPreset = true, disableRandomization = true)
+              CustomDefaultRegisterEmission(useInitAsPreset = false, disableRandomization = true)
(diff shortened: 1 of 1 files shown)


jackkoenig (approved):
LGTM! Important fix for 3.5.0

Andful:
This feels a bit like avoiding the problem. Correct me if I am wrong but in the bug here (in the linked snipped of code) is that `reg  b = b` where it should be `reg b = `1'b0` as in the chisel code we wrote `val b = RegInit(false.B)`.

jackkoenig:
> This feels a bit like avoiding the problem. Correct me if I am wrong but in the bug here (in the linked snipped of code) is that `reg b = b` where it should be `reg b = 1'b0` as in the chisel code we wrote `val b = RegInit(false.B)`.

See my reply to your related StackOverflow question: https://stackoverflow.com/a/70840128/2483329

PR #2465, opened 2022-01-17
Add MLIREmitter for FIRRTL

sequencer:
This is a WIP work.
It allows FIRRTL directly emit MLIR without parsing by MFC.

### Contributor Checklist

#### Type of Improvement
- new feature/API

#### API Impact
Add MLIREmitter

None

#### Desired Merge Strategy

#### Release Notes
A MLIREmitter is added for directly emitting MLIR file for MFC.

- [ ] Did you review?


jackkoenig:
Why do we need this instead of just letting CIRCT parse the `.fir` and `.anno.json`?

sequencer:
> Why do we need this instead of just letting CIRCT parse the .fir and .anno.json?

Currently, CIRCT can directly parse fir+annos. However I think this PR is still necessary:
1. When users try to implementing some custom IR which will be emit to MLIR, This Emitter can be leveraged without adding new annotations, it can be serialized to MLIR directly, and append to correct place, that what annotation cannot achieve. This can also empower designing MLIR with SFC.
2. If MFC parse FIRRTL and annotation together, it will localize the annotation to correspond `Target`, it is essentially a list to list traverse issue, when circuit is really big, and annotation are too much. Using `MLIREmitter` may speed up MFC(but I admit this will introduce FIRRTL's runtime.), but in the future, if we are able to annotate to local IRs directly, it can be a real speed up.
3. Basically, a lot of Annotations should be convert to local IR, we can use this to sort out how to improve FIRRTL Spec by adding those Annotations directly to FIRRTL IR to align the implementation between MFC and SFC.

darthscsi:
I'm in general a fan of bypassing fir files, though there are compatibility concerns.  Circt is also a fast moving target and isn't likely at the point it's going to provide a stable IR to out-of-tree clients.

> 1. When users try to implementing some custom IR which will be emit to MLIR, This Emitter can be leveraged without adding new annotations, it can be serialized to MLIR directly, and append to correct place, that what annotation cannot achieve. This can also empower designing MLIR with SFC.

There isn't a good reason the fir itself can't also contain new operations.  In fact we plan to as many operations are passed as annotations and this gets cumbersome.  The fir spec either needs to evolve or interesting uses will just start using a super-set.  Circt/FIR is quite happy, and planning, on supporting a superset of the fir spec.

> 2. If MFC parse FIRRTL and annotation together, it will localize the annotation to correspond `Target`, it is essentially a list to list traverse issue, when circuit is really big, and annotation are too much. Using `MLIREmitter` may speed up MFC(but I admit this will introduce FIRRTL's runtime.), but in the future, if we are able to annotate to local IRs directly, it can be a real speed up.

Yea, annotation parsing is large and slow.  It can, and the patches to do so are partially done, be made much faster and parallel.  Much of the work right now is finding IR representations of the annotation information so we can eliminate annotations early on in the compiler.  This would be even more effective if those operations came out of chisel directly rather than bouncing through annotations (see 1 and the superset idea).

> 3. Basically, a lot of Annotations should be convert to local IR, we can use this to sort out how to improve FIRRTL Spec by adding those Annotations directly to FIRRTL IR to align the implementation between MFC and SFC.

That is currently the plan in Circt/FIR.  As we figure out efficient, generic IR representations we intend to support them on the parser side so custom FIRRTL flavors can use them directly.  A few things come to mind:

a) Value-observability.  Here, Circt has a probe operation which captures a value without assigning a name.  Given the way FIRRTL works, this may not make sense to port.  But this gives us a data-flow sink which can be referred to by symbol (name) without imposing as many constraints as a node or wire does.  This is (or is planned) to be used for Binds, debug info, XMRs, and some internal annotations.

b) verification constructs.  Here, we've seen a lot of need for more interesting verification ops.  We have them in the Circt IR, but not in  a form we can parse from a fir file yet (right now they have to be pattern-matched from control flow structure + annotations + ops).

c) verbatim verilog.  Circt has verbatim verilog expressions and statements.  These are a great escape hatch (inline assembly).  A number of (at least internal) firrtl passes get around this by just printing verilog on the side, rather than storing it in the IR.  This seems to make the compiler pipeline fork several times as side pipelines do custom verilog emission.  Inline verilog would solve this.  (It's already being done in serveral ways, might as well formalize and normalize it).

d) Strict connects.  A fair amount of complexity comes from the broad semantics of connect (partial/not, structural/field-name, truncation, extension).  Circt has toyed with adding a very strict (types match, full) connect op and using that internally exclusively after the input code has been legalized to it's semantics.  This would reduce the complexity of down-stream passes.   There may be some value in directly encoding this.

e) binds.  Circt directly models remote-instantiation.  This is used extensively in SiFive's internal transforms and not having it in the IR causes a lot of annotations.  We were previously modeling this as an annotation instance op, but are going to migrate it to a bind+probe.

f) parameters.  @seldridge is adding support for parameterized modules.  Circt supports this in the hw/sv dialects, but it would be very helpful to support in firrtl too, if nothing else to better mix with external modules.  It is useful to do this even if Chisel doesn't yet support parameterization, as several passes (namely dedup) can be a lot more powerful if they can introduce trivial parameters.

ekiwi:
> The fir spec either needs to evolve or interesting uses will just start using a super-set. Circt/FIR is quite happy, and planning, on supporting a superset of the fir spec.

I think everyone agrees that there are some useful extensions that should be made to the firrtl spec. I think the general process for that would be to write up an RFC PR that changes the spec document. In general it is also nice to have a prototype implementation for the Scala firrtl compiler, but if you aren't familiar with the code base, a textual description of the new feature should be enough. It would also help if we could dedicate a chisel dev meeting (Mon/Tue at 11am) to talk about the new proposed feature.

PR #2468, opened 2022-01-21
Fix faulty MemorySynthInit behavior

johnsbrew:
- Fix & test MemorySynthInit behavior with MemoryArrayInitAnnotation and MemoryScalarInitAnnotation.
Add test case for MemoryRandomInitAnnotation which is, on the contrary, expected not to leak any randomization statement in synthesis context.

- Refactor MemoryInitSpec for improved results readability

### Context:

#2166  introduced `MemorySynthInit` annotation to control whether statement generated with `Memory*InitAnnotation` (emitted within initial begin block in verilog) should be guarded with `ifndef SYNTHESIS` or not.
Unfortunately only one configuration (`MemoryFileInlineAnnotation`) has been tested while the others have been generating incorrect verilog statements (`MemoryArrayInitAnnotation` and `MemoryScalarInitAnnotation`).
Example: 
https://scastie.scala-lang.org/rqS4FWY1SsaZKO4N49D9oA
produces illegal verilog:
```verilog
module Example(
// ...
reg [3:0] rom [0:4];
// ...
`endif // SYNTHESIS
  rom[0] = 0;
  rom[1] = 1;
  rom[2] = 3;
  rom[3] = 7;
  rom[4] = 15;
endmodule
```

while the following is expected (this PR)
```verilog
module Example(
// ...
reg [3:0] rom [0:4];
// ...
`endif // SYNTHESIS
initial begin
  rom[0] = 0;
  rom[1] = 1;
  rom[2] = 3;
  rom[3] = 7;
  rom[4] = 15;
end
endmodule
```

### Contributor Checklist

- [x] Did you add Scaladoc to every public function/method?
- [x] Did you update the FIRRTL spec to include every new feature/behavior?
- [x] Did you add at least one test demonstrating the PR?
- [x] Did you delete any extraneous printlns/debugging code?
- [x] Did you specify the type of improvement?
- [x] Did you state the API impact?
- [x] Did you specify the code generation impact?
- [x] Did you request a desired merge strategy?
- [x] Did you add text to be included in the Release Notes for this change?

#### Type of Improvement
bug fix

#### API Impact
none

#### Backend Code Generation Impact
none on currently valid verilog

#### Desired Merge Strategy
rebase

#### Release Notes
Fix & test MemorySynthInit behavior with MemoryArrayInitAnnotation and MemoryScalarInitAnnotation.
Add test case for MemoryRandomInitAnnotation which is, on the contrary, expected not to leak any randomization statement in synthesis context.

[...]

PR #2396, opened 2021-10-27
Disable random init

sequencer:
@SharzyL will work on this for PLCT internship contribution. Looking forward to his great job!
resolves chipsalliance/chisel3#1444

### Contributor Checklist

- [ ] Did you add Scaladoc to every public function/method?
- [ ] Did you update the FIRRTL spec to include every new feature/behavior?
- [ ] Did you add at least one test demonstrating the PR?
- [ ] Did you delete any extraneous printlns/debugging code?
- [ ] Did you specify the type of improvement?
- [ ] Did you state the API impact?
- [ ] Did you specify the code generation impact?
- [ ] Did you request a desired merge strategy?
- [ ] Did you add text to be included in the Release Notes for this change?

### Reviewer Checklist (only modified by reviewer)
- [ ] Did you add the appropriate labels?
- [ ] Did you mark the proper milestone (1.2.x, 1.3.0, 1.4.0) ?
- [ ] Did you review?
- [ ] Did you check whether all relevant Contributor checkboxes have been checked?
- [ ] Did you mark as `Please Merge`?

PR #2473, opened 2022-01-25
Avoid using self reference to signal no init

Andful:
This fixes the bug in issue #2449.  
I thought these changes may or may not come in useful given the discussion in the issue so I decided to just let you decide and put this pull request here.  
But if there is an interest in supporting `useInitAsPreset` I am willing to make further adjustments and add a test.

### Contributor Checklist

#### API Impact

It avoids using `init` as self referential to signal no init.

It fixes the issue in #2449.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/backends/experimental/rtlil/RtlilEmitter.scala
+++ b/src/main/scala/firrtl/backends/experimental/rtlil/RtlilEmitter.scala
@@ -376,7 +376,7 @@ private[firrtl] class RtlilEmitter extends SeqTransform with Emitter with Depend
         case e =>
           Seq(Seq(tabs, "assign ", regTempName, " ", SrcInfo(e, info).str_rep))
       }
-      if (weq(init, r)) { // Synchronous Reset
+      if (reset.tpe != AsyncResetType) { // Synchronous Reset
         val InfoExpr(info, e) = netlist(r)
         processes += Seq(info)
         processes += Seq("wire ", r.tpe, " ", regTempName)
@@ -387,7 +387,6 @@ private[firrtl] class RtlilEmitter extends SeqTransform with Emitter with Depend
         processes += Seq(tab, tab, "update ", SrcInfo(r).str_rep, " ", regTempName)
         processes += Seq("end")
       } else { // Asynchronous Reset
-        assert(reset.tpe == AsyncResetType, "Error! Synchronous reset should have been removed!")
         val tv = init
         val InfoExpr(finfo, fv) = netlist(r)
         processes += Seq(finfo)
--- a/src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
+++ b/src/main/scala/firrtl/backends/verilog/VerilogEmitter.scala
@@ -781,11 +781,10 @@ class VerilogEmitter extends SeqTransform with Emitter {
           }
         case e => Seq(Seq(tabs, r, " <= ", e, ";", info))
       }
-      if (weq(init, r)) { // Synchronous Reset
+      if (reset.tpe != AsyncResetType) { // Synchronous Reset
         val InfoExpr(info, e) = netlist(r)
         noResetAlwaysBlocks.getOrElseUpdate(clk, ArrayBuffer[Seq[Any]]()) ++= addUpdate(info, e, Seq.empty)
       } else { // Asynchronous Reset
-        assert(reset.tpe == AsyncResetType, "Error! Synchronous reset should have been removed!")
         val tv = init
         val InfoExpr(info, fv) = netlist(r)
         asyncResetAlwaysBlocks += (
--- a/src/main/scala/firrtl/transforms/RemoveReset.scala
+++ b/src/main/scala/firrtl/transforms/RemoveReset.scala
@@ -71,13 +71,13 @@ object RemoveReset extends Transform with DependencyAPIMigration {
           reg
         /* A register is initialized to an invalid expression */
         case reg @ DefRegister(_, _, _, _, _, init) if invalids.contains(we(init)) =>
-          reg.copy(reset = Utils.zero, init = WRef(reg))
+          reg.copy(reset = Utils.zero)
         case reg @ DefRegister(_, rname, _, _, Utils.zero, _) =>
-          reg.copy(init = WRef(reg)) // canonicalize
+          reg // canonicalize
         case reg @ DefRegister(info, rname, _, _, reset, init) if reset.tpe != AsyncResetType =>
           // Add register reset to map
           resets(rname) = Reset(reset, init, info)
-          reg.copy(reset = Utils.zero, init = WRef(reg))
+          reg.copy(reset = Utils.zero)
         case reg @ DefRegister(info, rname, _, _, reset, init) if reset.tpe == AsyncResetType =>
           asyncResets(rname) = Reset(reset, init, info)
           reg
--- a/src/test/scala/firrtlTests/transforms/RemoveResetSpec.scala
+++ b/src/test/scala/firrtlTests/transforms/RemoveResetSpec.scala
@@ -112,7 +112,7 @@ class RemoveResetSpec extends FirrtlFlatSpec with GivenWhenThen {
     outputState shouldNot containTree { case Connect(_, WRef("foo_b", _, _, _), Mux(_, _, _, _)) => true }
   }
 
-  it should "canvert a reset wired to UInt<0> to a canonical non-reset" in {
+  it should "convert a reset wired to UInt<0> to a canonical non-reset" in {
     Given("foo's reset is connected to zero")
     val input =
       """|circuit Example :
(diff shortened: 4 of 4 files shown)


jackkoenig:
@sequencer can you take a look at this?

sequencer:
Sure! I think this should be a bug fix(which might block 3.5.1) I will take a look at this today.

jackkoenig (changes requested):
Actually I don't think this is right. A self-initialization value is the canonical way within the FIRRTL IR to mean "no reset" so it shouldn't be changed. If there is a bug here, it's in emitters misinterpreting that.

johnsbrew:
As far as I understand, there is no bug here, only related to https://github.com/chipsalliance/firrtl/pull/2449#issuecomment-1020595519

(anyone to have a look at #2468 btw ?)

Andful:
> Actually I don't think this is right. A self-initialization value is the canonical way within the FIRRTL IR to mean "no reset" so it shouldn't be changed. If there is a bug here, it's in emitters misinterpreting that.

Yes I did notice that. I think using reset as the "0" literal would be a more elegant solution. Both `useInitAsPreset` and `PresetRegAnnotation` rely on the `init` value as a preset.  
This becomes a problem when you have a register to both have a preset and a possibly a reset signal (which is the case for `useInitAsPreset`). If the compiler notices that there is no reset it will modify `init` which also is its preset value.

Andful:
But I don't know how involved would be to use the literal "0" to signal no reset. From what I saw only `RtlilEmitter` and `VerilogEmitter` rely on it.

Andful:
@johnsbrew The bug is not user facing. The bug is that the class [`CustomDefaultRegisterEmission`](https://github.com/chipsalliance/firrtl/blob/b027eb466b033a0a9d229f19feb931ddb292a9fa/src/main/scala/firrtl/EmissionOption.scala#L15) has an option `useInitAsPreset` which does not make sense if the compiler changes `init` if it sees that a register is never resetted.

ekiwi:
> But I don't know how involved would be to use the literal "0" to signal no reset. From what I saw only `RtlilEmitter` and `VerilogEmitter` rely on it.

There are other passes that rely on this. For example some of my own passes use the `init` value to see if the register has an init value or not.

> @johnsbrew The bug is not user facing. The bug is that the class [`CustomDefaultRegisterEmission`](https://github.com/chipsalliance/firrtl/blob/b027eb466b033a0a9d229f19feb931ddb292a9fa/src/main/scala/firrtl/EmissionOption.scala#L15) has an option `useInitAsPreset` which does not make sense if the compiler changes `init` if it sees that a register is never resetted.

If the register is properly connected to a reset that is marked with the `PresetAnnotation`, then the compiler will keep the `init` value around. The same is true for async registers. Only for synchronous reset / no reset registers will `init` be replaced with a reference to the register.

ekiwi:
@Andful : Do you have a concrete example of a Chisel circuit that could demonstrate the bug you are trying to fix? I am not quite sure what the problem is and maybe, once we all understand what exactly the behavior is that you are trying to change, we can think of a better way to solve this.

Andful:
@ekiwi The example in the pull request #2449. That was the very reason why `useInitAsPreset` was set to `false`.

ekiwi:
> @ekiwi The example in the pull request #2449. That was the very reason why `useInitAsPreset` was set to `false`.

But #2449 fixed that bug, no?

Andful:
I mean, the bug is not user facing anymore but if you set `useInitAsPreset` to true it might emit malformed verilog. If `init` is possibly self referencing at any time `useInitAsPreset` should never be set to true.

ekiwi:
> I mean, the bug is not user facing anymore but if you set `useInitAsPreset` to true it might emit malformed verilog. If `init` is possibly self referencing at any time `useInitAsPreset` should never be set to true.

Could we fix this maybe by removing the `useInitAsPreset` option from the constructor? It seems to me like it sneaked into an unrelated PR and maybe should never have ended up there.

Andful:
> > I mean, the bug is not user facing anymore but if you set `useInitAsPreset` to true it might emit malformed verilog. If `init` is possibly self referencing at any time `useInitAsPreset` should never be set to true.
> 
> Could we fix this maybe by removing the `useInitAsPreset` option from the constructor? It seems to me like it sneaked into an unrelated PR and maybe should never have ended up there.

Yes. But generally I think a preset is useful information for a circuit.

sequencer:
In my understanding:
**the semantic of `Preset` is register/memory state before reset.**
So there will be 3 different options each preset:
1. Preset to 1
2. Preset to 0
3. Random

Current implementation to Preset is adding an `initial` field(defined IEEE1364-2005: 9.9.1 Initial construct) but don't use `#` to construct delay, and so does `RANDOMIZE`.
So I think we cannot have both random and preset in the same time.

So the issue becomes:
we can disable random and set preset a register, but this seems to be an Emitter bug.

ekiwi:
> Yes. But generally I think a preset is useful information for a circuit.

We do have a way to make registers with a "preset" and that is to use the `PresetAnnotation`. I don't think we ever intended to support an alternative way. https://github.com/chipsalliance/firrtl/pull/2396 was supposed to allow users to disable the randomization logic and was not intended to change anything about "preset" registers.

sequencer:
> #2396 was supposed to allow users to disable the randomization logic and was not intended to change anything about "preset" registers.

Actually random is another type of "preset", so both "preset" and "random" is not possible.

sequencer:
Doesn't define the preset value, https://scastie.scala-lang.org/ZRHZGEs1TsGoIdku7niUog, if it is defined, will it behave correctly? @ekiwi 
If true, I think we only need add some logic to emitter: 
if a register is preset, we should disable random for it, and assert the preset value exist.

Andful:
@sequencer In this [example](https://scastie.scala-lang.org/XWRlEiEHRjuHYIUgQjbBbQ) (provided by @jackkoenig) does not initialize `context`as random so I would say it is working correctly. So I guess currently only `useInitAsPreset` is dubious and should be removed.

johnsbrew:
@Andful 

### A bit of context about `Reset`, `AsyncReset` and `Preset`

I do understand your point, the main issue you raise here is that firrtl does not consider preset as first class citizen.
For the record I was in favor of such an approach when I initially proposed the support of these presets #991 

After discussion, @jackkoenig & @seldridge point of view was the following:
- chisel is aimed at providing 100% re-usable hardware generators
- preset are a FPGA-only feature *(also meaningful in simulation)*
- hence, preset cannot be a first class citizen alongside synchronous and asynchronous resets 

It was then decided to provide an implementation of preset (which is the current version) based on annotating AsyncReset
- convienient because AsyncResets already enforce static init values
- *de facto* preventing the ability to use multiple resets for the the same register.

As of today, there is for example no way to generate the following verilog, and this is unlikely to happen anytime soon.
```verilog
    reg triple_reset = <preset_init>;
    always @ (posedge clock or areset) begin
      if (areset) begin
        triple_reset <= <async_init>;
      end else begin
        if (sreset) begin
          triple_reset <= <sync_init>;
        end else begin
          triple_reset <= normal_value;
        end
      end
    end

```
Note that the main restriction is the incompatibility of `AsyncReset` with `Preset`, as synchronous resets can always be written in plain Chisel:

```scala
class Async(forcePreset: Boolean=false) extends Module with RequireAsyncReset {
  val sreset = IO(Input(Bool())) 

  val i = IO(Input(Bool()))
  val o = IO(Output(Bool())) 

  val r = RegInit(false.B)
  when(sreset) {
    r := true.B
  } .otherwise {
    r := i
  }
  o := r

  // then, one might want to convert AsyncReset to Preset
  if(forcePreset) {
      def toFirrtl = firrtl.annotations.PresetAnnotation(reset.toTarget)
    }) 
  }
}
```
see https://scastie.scala-lang.org/z6Ch0u5ASueDaTTbacjhCQ for resulting verilog

### `EmissionOptions` trait vs `--emission-options=` command-line

I first introduced `EmissionOptions`with `useInitAsPreset` and `disableRandomization` in #1050 
https://github.com/chipsalliance/firrtl/pull/1050/files#diff-acc9dea5ae74bda279f6f83be79e549dba609d7ac04306426395f0dccd693e16

It was an attempt to partially customized the emission, intended to be used by transforms adding annotations.
Theses options add 2 features on register emission
- `useInitAsPreset` => used to convert an asynchronously reset RegInit into a preset RegInit
- `disableRandomization` => used to remove the random initialization statements (based on initial begin)

Out of 4 combinations of these 2 flags, only 3 are meaningful
- `!useInitAsPreset & !disableRandomization` => default behavior
- `useInitAsPreset & disableRandomization` => Emit proper Preset
- `disableRandomization` => deactivate randomization

This API is an advanced transform API, with local effect (only one register is concerned by each annotations) and it has in fact little to do with the newer `--emission-options=` available from the command line since  #2396 
This command-line API sets **default values** for the above mentioned flags.
While it makes sense for disabling randomization at circuit scale (kind of cosmetic), it certainly doesn't for using init as preset at this scale.
Indeed,  except for small designs,  various reset trees must be preserved and only preset-annotated resets shall result in `RegInit` using init as preset: this is precisely how ` firrtl.annotations.PresetAnnotation(reset.toTarget)` works behind the scene.
 
> So I guess currently only useInitAsPreset is dubious and should be removed.

While your conclusion about `useInitAsPreset` makes sense from a command-line point of view (where it should never have ended up alongside `disableRegisterRandomization`), it is in fact very useful as `EmissionOptions` 😃 

Hope this helps to understand the big picture, even if I'm not sure anyone will read this long comment 😅

PR #2477, opened 2022-01-27
Dev/jean.bruant/mergify/bp/1.4.x/pr 2468

johnsbrew:
supersede and close #2475
as I couldn't push directly on the repo 
git push --set-upstream berkeley mergify/bp/1.4.x/pr-2468
ERROR: Permission to chipsalliance/firrtl.git denied to johnsbrew.


jackkoenig:
Thank you for the backport fix! I've pushed to #2475. Generally the best way to fix the backport is to open a PR against it, but this also works, thanks!

PR #2504, opened 2022-04-05
Use circuit name hashCode for circuit anno hashCode

jackkoenig:
It is unclear if having more than one circuit annotation even works. If
it does, it is implausible that they would have the same circuit name
since that would correspond to a namespace collision.

Looking at flamegraphs, FirrtlCircuitAnnotation.hashCode always comes up. Creating a HashMap with different Circuits that nevertheless have the same name is extremely uncommon so collisions are not a concern. Include productPrefix so that this doesn't collide with other types that use a similar strategy and hash the same String.

This speeds up anything using Stage/Phase with FIRRTL Circuits which includes Chisel

### Contributor Checklist

#### Type of Improvement

  - performance improvement    

#### API Impact

No impact

No impact

#### Desired Merge Strategy

   - Squash

#### Release Notes

Improve performance of `FirrtlCircuitAnnotation.hashCode`. This speeds up all uses of Stage/Phase.

- [ ] Did you review?


ekiwi (approved):
Seems reasonable :ship:

PR #2508, opened 2022-04-07
Make MemConf's MemPort serialization deterministic

chick:
Problem: MemConf serialization of MemPorts was not deterministic
and the ordering seems to have changed as we move projects to 2.13
Downstream project can be adversely affected by changes in ordering
This changes specifies as specific ordering that should be compatible with
the historically one.

### Contributor Checklist

#### Type of Improvement

Code cleanup

#### API Impact

This change should preserve existing behavior that was changing as we continue to move to 2.13
Without this change downstream users for MemConf output could break

Verilog output should not be altered

#### Desired Merge Strategy

Squash and Merge

#### Release Notes

The order of MemPort fields in serialized MemConf is now guaranteed to be in the following order:
- mrw, mwrite, rw, write, read

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/passes/memlib/MemConf.scala
+++ b/src/main/scala/firrtl/passes/memlib/MemConf.scala
@@ -3,7 +3,7 @@
 package firrtl.passes
 package memlib
 
-sealed abstract class MemPort(val name: String) { override def toString = name }
+sealed abstract class MemPort(val name: String) extends Product with Serializable { override def toString = name }
 
 case object ReadPort extends MemPort("read")
 case object WritePort extends MemPort("write")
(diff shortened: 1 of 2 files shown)


mwachs5 (commented):
Non-determinism and hardware generation don't mix, I agree in principle with this change have not done a deep dive on the implementation

src/main/scala/firrtl/passes/memlib/MemConf.scala
```diff
@@ -13,7 +13,22 @@ case object MaskedReadWritePort extends MemPort("mrw")
...
+      MaskedWritePort,
+      ReadWritePort,
+      WritePort,
+      ReadPort
+    )
+    .zipWithIndex
+    .toMap
+
+  // uses orderedPorts when sorting MemPorts
+  implicit def ordering: Ordering[MemPort] = Ordering.by(e => orderedPorts(e))
```
jackkoenig, line 31:
```suggestion
  // This is the order that ports will render in MemConf.portsStr
  val ordered: Seq[MemPort] = Seq(
      MaskedReadWritePort,
      MaskedWritePort,
      ReadWritePort,
      WritePort,
      ReadPort
    )
  
   val all: Set[MemPort] = ordered.toSet

  // uses orderedPorts when sorting MemPorts
  implicit def ordering: Ordering[MemPort] = {
    val orderedPorts = ordered.zipWithIndex.toMap
    Ordering.by(e => orderedPorts(e))
  }
```

The goals of my suggestion are:
1) Keep it dry, might as well derive `all` from the ordered list
2) If we're adding a public API, I think just having the `Seq` (which is ordered) makes more sense than `Map[_, Int]`

jackkoenig (commented):
Generally looks good but I have a couple of suggestions

jackkoenig (approved):

PR #2520, opened 2022-04-21
Fix optimization of register with reset but invalid connection

### LINKED ISSUE #2516: Incorrect optimization of a register with a reset
### Checklist

- [ ] Did you describe your environment?

```
circuit Test:
  module Test:
    input clock : Clock
    input reset : UInt<1>
    input p: UInt<1>
    output o : UInt<8>
    
    reg r : UInt<8>, clock with :
      reset => (p, UInt<8>(3))
    r is invalid

    o <= r
```
 outputs
 ```
 module Test(
  input        clock,
  input        reset,
  input        p,
  output [7:0] o
);
  assign o = 8'h0;
endmodule
```

I expected the `r is invalid` to be equivalent to `r <= UInt<8>(0)`:

```
module Test(
  input        clock,
  input        reset,
  input        p,
  output [7:0] o
);
  reg [7:0] r;
  assign o = r;
  always @(posedge clock) begin
    if (p) begin
      r <= 8'h3;
    end else begin
      r <= 8'h0;
    end
  end
endmodule
```

### Your environment

- Chisel Verions: 
- OS: 
- Verilator version: 

### External Information

@seldridge has indicated that the InferResets pass is missing a case to handle `IsInvalid`:
https://github.com/chipsalliance/firrtl/blob/master/src/main/scala/firrtl/transforms/RemoveReset.scala#L56

[This](https://github.com/chipsalliance/firrtl/blob/master/src/main/scala/firrtl/transforms/RemoveReset.scala#L77) match is going to move the reset value into the resets map. However, the pass never comes back to it because there is no case to match for `IsInvalid`.

### PR DESCRIPTION
**Pull request by jackkoenig**
Fixes #2516

Previously,
```
reg r : UInt<8>, clock with :
  reset => (p, UInt<8>(3))
r is invalid
```
would compile to:
```
reg r : UInt<8>, clock
r <= UInt<8>(0)
```
now it compiles to:
```
reg r : UInt<8>, clock
r <= UInt<8>(3)
```

### Contributor Checklist

#### Type of Improvement

 - bug fix        

#### API Impact

This is a 

No impact

#### Desired Merge Strategy

 - Squash

#### Release Notes

Fix lowering of synchronously reset registers that have an invalid connection.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/transforms/RemoveReset.scala
+++ b/src/main/scala/firrtl/transforms/RemoveReset.scala
@@ -93,6 +93,9 @@ object RemoveReset extends Transform with DependencyAPIMigration {
           // addUpdate(info, Mux(reset, tv, fv, mux_type_and_widths(tv, fv)), Seq.empty)
           val infox = MultiInfo(reset.info, reset.info, info)
           Connect(infox, ref, expr)
+        case IsInvalid(_, ref @ WRef(rname, _, RegKind, _)) if resets.contains(rname) =>
+          val Reset(_, init, info) = resets(rname)
+          Connect(info, ref, init)
         case other => other.map(onStmt)
       }
     }
--- a/src/main/scala/firrtl/transforms/RemoveReset.scala
+++ b/src/main/scala/firrtl/transforms/RemoveReset.scala
@@ -52,6 +52,7 @@ object RemoveReset extends Transform with DependencyAPIMigration {
     val resets = mutable.HashMap.empty[String, Reset]
     val asyncResets = mutable.HashMap.empty[String, Reset]
     val invalids = computeInvalids(m)
+    lazy val namespace = Namespace(m)
     def onStmt(stmt: Statement): Statement = {
       stmt match {
         case reg @ DefRegister(_, name, _, _, reset, init) if isPreset(name) =>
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by ekiwi — APPROVED** [review]
:shipit:

**Comment by ekiwi** [comment]
Needs some `scalafmtAll` though

**Review by seldridge — CHANGES_REQUESTED** [review]
(I spoke to Jack about this separately.)

I don't think the additional invalid interpretation here is correct or self-consistent with existing SFC behavior. The end-to-end compile of this should result in a mux with zero, not optimization to the reset value. My reasoning is that this pass is narrowly responsible for moving the reset mux in front of the register. Any further optimization should not happen and is introducing yet-another-interpretation of invalid that has to be special cased.

Put differently, this creates a reset `mux` not a reset `when`. I would expect the pass to generate or be equivalent to the following:

```scala
    reg r : UInt<8>, clock
    wire inv: UInt<8>
    inv is invalid
    r <= mux(p, UInt<8>(3), inv)
```

jackkoenig:
Changed it to what @seldridge said. The implementation is a bit uglier, but it is consistent with the behavior when the register has an async reset.

seldridge (approved):
LGTM

Weakening my previous statements: I would like to revisit the `mux(cond, _, invalid)` canonicalization in the future. It's annoying that this is not unified with `when` or `validif` formulations. Keeping clean separation of lowering (this pass) from optimization/folding will help us better tackle this fully in the future.

PR #2543, opened 2022-08-18
FIRRTL version support

zyedidia:
Now that FIRRTL supports version specifiers (https://github.com/chipsalliance/firrtl-spec/pull/30), the parser here should accept versions and the serializer should emit a version. The current version of FIRRTL is 1.1.0. This FIRRTL compiler will throw an error if used to compile FIRRTL with a version greater than 1.1.x (e.g., >= 1.2.0). The serializer currently emits FIRRTL version 1.1.0.

### Contributor Checklist

#### Type of Improvement

- new feature/API

#### API Impact

The serializer will now emit a FIRRTL version number, and the parser will accept one and throw an error if it is above 1.1.0.

No impact.
#### Desired Merge Strategy

Squash.
#### Release Notes

The serializer will now emit a FIRRTL version number, and the parser will accept one and throw an error if it is above 1.1.0.
- [ ] Did you review?

### DIFF
--- a/src/main/antlr4/FIRRTL.g4
+++ b/src/main/antlr4/FIRRTL.g4
@@ -36,7 +36,6 @@ version
   : 'FIRRTL' 'version' semver NEWLINE
   ;
 
-// Due to lexer problems, something like 1.1.0 is lexed as DoubleLit '.' UnsignedInt
 semver
   : SemVer .*?
   ;
--- a/src/main/antlr4/FIRRTL.g4
+++ b/src/main/antlr4/FIRRTL.g4
@@ -355,7 +355,7 @@ BinaryLit
   ;
 
 SemVer
-  : Digit+ '.' Digit+ '.' Digit+
+  : UnsignedInt '.' UnsignedInt '.' UnsignedInt
   ;
 
 DoubleLit
--- a/src/main/antlr4/FIRRTL.g4
+++ b/src/main/antlr4/FIRRTL.g4
@@ -38,7 +38,7 @@ version
 
 // Due to lexer problems, something like 1.1.0 is lexed as DoubleLit '.' UnsignedInt
 semver
-  : DoubleLit '.' UnsignedInt .*?
+  : SemVer .*?
   ;
 
 module
@@ -357,6 +357,10 @@ OctalLit
 
 BinaryLit
   : '"' 'b' ( '+' | '-' )? ( BinaryDigit )+ '"'
+  ;
+
+SemVer
+  : Digit+ '.' Digit+ '.' Digit+
   ;
 
 DoubleLit
--- a/src/main/antlr4/FIRRTL.g4
+++ b/src/main/antlr4/FIRRTL.g4
@@ -29,7 +29,16 @@ import firrtl.LexerHelper;
 
 // Does there have to be at least one module?
 circuit
-  : 'circuit' id ':' info? INDENT module* DEDENT EOF
+  : version? NEWLINE* 'circuit' id ':' info? INDENT module* DEDENT EOF
+  ;
+
+version
+  : 'FIRRTL' 'version' semver NEWLINE
+  ;
+
+// Due to lexer problems, something like 1.1.0 is lexed as DoubleLit '.' UnsignedInt
+semver
+  : DoubleLit '.' UnsignedInt .*?
   ;
 
 module
--- a/src/main/antlr4/FIRRTL.g4
+++ b/src/main/antlr4/FIRRTL.g4
@@ -33,7 +33,12 @@ circuit
   ;
 
 version
-  : 'FIRRTL' 'version' SemVer NEWLINE
+  : 'FIRRTL' 'version' semver NEWLINE
+  ;
+
+// Due to lexer problems, something like 1.1.0 is lexed as DoubleLit '.' UnsignedInt
+semver
+  : DoubleLit '.' UnsignedInt
   ;
 
 module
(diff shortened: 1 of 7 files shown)


src/main/scala/firrtl/ir/Serializer.scala
```diff
@@ -10,6 +10,8 @@ object Serializer {
   val NewLine = '\n'
   val Indent = "  "
 
+  val version = "1.1.0"
```
seldridge, line 13: This would probably be better described as a product and if this was defined globally so this can be used in `parser/Listener.scala`:

```scala
case class Version(major: Int, minor: Int, patch:Int) extends scala.math.Ordered[Version] {
  override def serialize: String = ???
  override def compare(that: Version): Bool = ???
}
```

This would simplify some things later, e.g., exiting if the input version is too new. 

Also, should this be defined in IR.scala?

src/test/scala/firrtlTests/ParserSpec.scala
```diff
@@ -92,6 +92,48 @@ class ParserSpec extends FirrtlFlatSpec {
...
+                  |circuit Test :
+                  |  module Test :
+                  |    input in : { 0 : { 0 : UInt<32>, flip 1 : UInt<32> } }
+                  |    input in2 : { 4 : { 23 : { foo : UInt<32>, bar : { flip 123 : UInt<32> } } } }
+                  |    in.0.1 <= in.0.0
+                  |    in2.4.23.bar.123 <= in2.4.23.foo
+      """.stripMargin
+    firrtl.Parser.parse(input)
+  }
+
```
mwachs5, line 122: should we have a test that 1.1.1 is OK?

src/main/antlr4/FIRRTL.g4
```diff
@@ -29,7 +29,16 @@ import firrtl.LexerHelper;
...
 circuit
-  : 'circuit' id ':' info? INDENT module* DEDENT EOF
+  : version? NEWLINE* 'circuit' id ':' info? INDENT module* DEDENT EOF
+  ;
+
+version
+  : 'FIRRTL' 'version' semver NEWLINE
+  ;
+
+// Due to lexer problems, something like 1.1.0 is lexed as DoubleLit '.' UnsignedInt
```
jackkoenig, line 39: I believe this comment is now untrue since you have the `SemVer` lex rule below.

src/main/antlr4/FIRRTL.g4
```diff
@@ -29,7 +29,16 @@ import firrtl.LexerHelper;
...
+semver
+  : SemVer .*?
```
jackkoenig, line 41: What is `.*?` intended to be matching? We should try to avoid non-greedy matching in general.

src/test/scala/firrtlTests/ParserSpec.scala
```diff
@@ -92,8 +92,8 @@ class ParserSpec extends FirrtlFlatSpec {
     ) ++ PrimOps.listing
   }
 
-  // FIRRTL version number
-  "Version" should "be parsed" in {
+  // ********** FIRRTL version number **********
```
mwachs5, line 95: should we also have an explicit test for "a file without a version number should be accepted"

src/main/antlr4/FIRRTL.g4
zyedidia: That is intended to match any additional version information that could be included in the semantic version (e.g., for 1.2.0-rc1).

src/main/scala/firrtl/ir/Serializer.scala
jackkoenig: I'm okay with it being in `IR.scala` but I'd like to avoid modifying anything in IR.scala if we can avoid it (to avoid backwards compatibility concerns with existing transforms).

src/main/antlr4/FIRRTL.g4
jackkoenig: Can you add a comment that mentions that? 

I suspect there are no performance concerns because I believe the `.` does not include newlines so in practice this won't lookahead very far, but can you do a sanity check that on a very large `.fir` file this change doesn't impact parsing speed?

mwachs5: FWIW the firrtl spec doesn't allow for any such modifiers

zyedidia: Okay, I see that in the grammar definition in the FIRRTL spec this is restricted to just X.Y.Z, so I'll remove this from here.

src/main/scala/firrtl/ir/Serializer.scala
zyedidia: I've made a `Version` class in `Serializer.scala` that is similar to this. Instead of having a `compare` method it has an `incompatible` method, since I think compare should check the patch version, but we want to be able to accept larger versions as long as only the patch is larger. Let me know if this needs more changes.

src/test/scala/firrtlTests/ParserSpec.scala
jackkoenig: Technically every other test in this file is testing that but I do agree that calling it out explicitly would be useful

jackkoenig (approved):
LGTM!

src/main/scala/firrtl/ir/Serializer.scala
jackkoenig: I think this is reasonable. If we need to do more work here later we can but this looks good enough to me.

ekiwi (approved):
:shipit:

PR #2545, opened 2022-08-24
implement sub-word assignments

ekiwi:
### Contributor Checklist

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/passes/ExpandWhens.scala
+++ b/src/main/scala/firrtl/passes/ExpandWhens.scala
@@ -66,20 +66,120 @@ object ExpandWhens extends Pass {
     */
   type Defaults = Seq[mutable.Map[WrappedExpression, Expression]]
 
+  private def getVoid(tpe: ir.Type): WIntVoid = WIntVoid(bitWidth(tpe))
+  private def doConnect(
+    netlist:  Netlist,
+    defaults: Defaults,
+    lhs:      ir.Expression,
+    rhs:      ir.Expression,
+    info:     Info
+  ): Unit =
+    lhs match {
+      case ref: ir.RefLikeExpression =>
+        netlist(ref) = InfoExpr(info, rhs)
+      case otherLhs =>
+        // normalize nested bits expressions before check
+        simplifyBits(otherLhs) match {
+          case ir.DoPrim(PrimOps.Bits, Seq(ref: ir.RefLikeExpression), Seq(hi, lo), _) =>
+            val refWidth = bitWidth(ref.tpe)
+            val highest = refWidth - 1
+            assert(hi < refWidth && lo >= 0)
+            // if we are assigning the whole range, things are simple
+            if (hi == highest && lo == 0) {
+              netlist(ref) = InfoExpr(info, rhs)
+            } else {
+              // perform a read modify write
+              val (prevInfo, prev) = unwrap(
+                netlist.getOrElse(ref, getDefault(ref, defaults).getOrElse(WIntVoid(refWidth)))
+              )
+              val msb = if (hi == highest) { simplifyBits(rhs) }
+              else {
+                Utils.cat(simplifyBits(prev, highest, hi + 1), simplifyBits(rhs))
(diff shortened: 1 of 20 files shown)


schoeberl:
Great that this comes back to Chisel.

src/test/scala/firrtlTests/SubWordAssignmentTests.scala
```diff
@@ -0,0 +1,342 @@
...
+        |""".stripMargin
+    val expected =
+      """circuit m :
+        |  module m :
+        |    input y : UInt<1>
+        |    output x : UInt<1>
+        |    x <= y
+        |""".stripMargin
+    check(src, expected)
+  }
```
zyedidia, line 341: Might also want a test for subword-assigning an SInt:

```
circuit test :
  module test :
    input c : UInt<2>
    output x : SInt<2>

    x[0] <= c[0]
    x[1] <= c[1]
```

Note that since the type of `x[...]` is always `UInt`, it must be connected to a `UInt` even though `x` is an `SInt`.

src/main/scala/firrtl/passes/ExpandWhens.scala
```diff
@@ -66,20 +66,120 @@ object ExpandWhens extends Pass {
...
+            throw new PassException(s"Invalid expression at the left hand side of an assignment: ${other.serialize}")
+        }
+    }
+  private def simplifyBits(e: ir.Expression, hi: BigInt, lo: BigInt): ir.Expression =
+    simplifyBits(Utils.bits(e, hi, lo))
+
+  /** performs special simplifications which are needed in the case of sub-word assignments to avoid false positives
+    *  in the connection check
+    */
+  private def simplifyBits(e: ir.Expression): ir.Expression = e match {
```
zyedidia, line 115: I'm not sure if this code already handles this, but I recently fixed a bug in the CIRCT implementation related to SInts so I thought I'd mention it here in case. First the connector must call `asSInt` if the bit-indexed value being assigned is an `SInt`.

Then the simplifications need to look through calls to `asSInt` (i.e., having a simplification that `bits(asSInt(x), ..., ...) -> bits(x, ..., ...)`).

src/main/scala/firrtl/passes/InferTypes.scala
```diff
@@ -97,6 +98,25 @@ object InferTypes extends Pass {
   }
 }
 
+/** Internal node used by the parser for the `[...]` and `[... : ...]` syntax.
+  *  CHIRRTL level type inference then converts this into either a `bits(..., ... , ...)`,
+  *  or [[ir.SubIndex]] depending on the inferred type of the inner expression.
+  */
+case class WSliceNode(expr: ir.Expression, hi: Int, lo: Int) extends ir.Expression {
```
zyedidia, line 105: Would it be better to define this as a node in `IR.scala` so that it can be more easily accessed from the Chisel converter?

ekiwi: No, this is an internal node used to hack around a short coming in the parser (that it does not do type inference). From Chisel you should know whether you want a bit-extract or a vector access since the types are available in the Chisel data structure.

src/main/scala/firrtl/passes/ExpandWhens.scala
ekiwi: I believe subword-assignment for SInt is broken in this PR right now. Thanks for pointing out how I could fix it.

src/test/scala/firrtlTests/SubWordAssignmentTests.scala
ekiwi: Yes. Thanks, I need to add this test (and fix the implementation to support it)

seldridge (commented):
I had one comment from a while back that I never actually filed. 🤦 

It is bad if we are handling vec of bools differently from UInt. Users can annoyingly start to rely on this behavior and it then becomes the de facto way that Chisel works.

PR #2549, opened 2022-09-12
Make the Parser handle errors more gracefully

jackkoenig:
Ever since introducing the Listener, the firrtl Parser now can hit errors in the code converting from concrete syntax to abstract syntax that may be due to syntax errors. These errors are essentially broken assumptions about the structure of the parsed code because there is an error. These errors are reported before the standard ANTLR syntax errors are aggregated and reported, and thus could result in less than elegant error messages (eg. NullPointerException). Now, the Parser will mask off such errors in the event of standard syntax errors caught by the ANTLR-generated parser.

This commit also cleans up some ParserSpec tests slightly to make the ScalaTest style more canonical.

### Contributor Checklist

#### Type of Improvement

 - bug fix    
#### API Impact

No impact

No impact

#### Desired Merge Strategy

- Squash

#### Release Notes

Improve error reporting for some syntax errors in parsing that previously were leading to crashes.

- [ ] Did you review?


src/main/scala/firrtl/Parser.scala
```diff
@@ -42,12 +44,26 @@ object Parser extends LazyLogging {
...
+        }
 
       val numSyntaxErrors = parser.getNumberOfSyntaxErrors
       if (numSyntaxErrors > 0) throw new SyntaxErrorsException(s"$numSyntaxErrors syntax error(s) detected")
 
+      // Any exceptions thrown by the Listener/Visitor should be due to syntax errors reported
+      // above, but we don't want to mask any bugs.
+      if (exception.nonEmpty) {
+        throw exception.get
+      }
```
seldridge, line 65: Nit: `if (foo) { throw bar }` is inconsistent with the style on line 59.

src/main/scala/firrtl/Parser.scala
```diff
@@ -42,12 +44,26 @@ object Parser extends LazyLogging {
       parser.getInterpreter.setPredictionMode(PredictionMode.SLL)
       parser.addParseListener(listener)
 
-      // Concrete Syntax Tree
-      parser.circuit
+      // Syntax errors may violate assumptions in the Listener and Visitor
+      // We need to handle these errors gracefully
```
seldridge, line 48: Nit: it would be good to use periods for sentences.

src/main/scala/firrtl/Parser.scala
seldridge: Is this code path ever expected to be hit or is this a complete fallback that "something went wrong"? This seems like the latter. If that is the case, it may be useful to add more information to this error that it looks like a real FIRRTL bug. Maybe this is already reported as such.

seldridge (approved):

src/main/scala/firrtl/Parser.scala
jackkoenig: > Nit: if (foo) { throw bar } is inconsistent with the style on line 59.

Fixed

> Is this code path ever expected to be hit or is this a complete fallback that "something went wrong"? This seems like the latter. If that is the case, it may be useful to add more information to this error that it looks like a real FIRRTL bug. Maybe this is already reported as such.

It is the latter. I've made the comment call that out more clearly and wrapped the issue in an "Internal Exception". Such wrapping was already occurting if you hit it from `FirrtlPhase`, but since `FirrtlPhase` is smart enough not to wrap twice, might as well wrap it here so that if you call the firrtl Parser directly, you still get the wrapping.

src/main/scala/firrtl/Utils.scala
```diff
@@ -165,7 +165,7 @@ object Utils extends LazyLogging {
     * @param message - possible string to emit,
     * @param exception - possible exception triggering the error.
     */
-  def throwInternalError(message: String = "", exception: Option[Exception] = None) = {
+  def throwInternalError(message: String = "", exception: Option[Throwable] = None) = {
```
jackkoenig, line 168: Note that this looks spooky and not binary compatible, but I tested it an it is binary compatible due to type erasure (since the real type here is `Option`). MiMa did not report any issues and I also manually checked in a project that compiled with firrtl 1.5.4 and ran with firrtl from this branch.

PR #2554, opened 2022-09-15
Add Serializer.lazily for buffered serialization

jackkoenig:
This is needed for emitting modules that serialize to Strings larger than 2 GiB (the maximum String size on the JVM). It includes micro-optimized logic for turning arbitrarily nested when scopes into Iterable[String].

I spent a lot of time benchmarking and micro-optimizing this code because it was pretty hard to make this version actually faster than the old version that just serialized to a single String. It turns out the JVM does a very very good job of handling large String serialization, so it was tricky to amortize the overhead of the additional Iterator objects (especially w.r.t to `whens`). Nevertheless, for the 264 MiB of .fir design I used for benchmarking, this is now faster than the current implementation.

This is not yet used, but it enables emission of very large circuits.

### Contributor Checklist

#### Type of Improvement

  - performance improvement  
  - new feature/API   

#### API Impact

- new API

No impact

#### Desired Merge Strategy

 - Squash

#### Release Notes

Add `Serializer.lazily` for buffered serialization of FIRRTL circuits.

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/ir/Serializer.scala
+++ b/src/main/scala/firrtl/ir/Serializer.scala
@@ -33,7 +33,7 @@ object Serializer {
       case n: Info        => s(n)(builder, indent)
       case n: StringLit   => s(n)(builder, indent)
       case n: Expression  => s(n)(builder, indent)
-      case n: Statement   => s(n)(builder, indent)
+      case n: Statement   => builder ++= lazily(n, indent).mkString
       case n: Width       => s(n)(builder, indent)
       case n: Orientation => s(n)(builder, indent)
       case n: Field       => s(n)(builder, indent)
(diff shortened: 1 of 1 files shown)


jackkoenig:
This buffered emission also lends itself really nice to multi-threading so we could consider using something like a [java.util.concurrent.ConcurrentLinkedQueue](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentLinkedQueue.html) to let 1 thread do the conversion to Strings and another to write out the file. I think the actual conversion to Strings is like 75-85% of the work so such multi-threading won't speed things up that much further, but it's cool that it's an option now.

azidar (approved):

PR #2591, opened 2023-01-13
Publish Dated SNAPSHOTs

jackkoenig:
The intent is that once these are being published, chisel3 can find the latest dated SNAPSHOT and then publish itself against it.

### Contributor Checklist

#### Type of Improvement

- release automation

#### API Impact

No impact

No impact

#### Desired Merge Strategy

  - Squash

#### Release Notes

Begin published dated SNAPSHOTs

- [ ] Did you review?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -13,9 +13,20 @@ lazy val isAtLeastScala213 = Def.setting {
   CrossVersion.partialVersion(scalaVersion.value).exists(_ >= (2, 13))
 }
 
+def currentVersion(major: String): String = {
+  val doDatedSnapshot = sys.props.get("datedsnapshot")
+  if (doDatedSnapshot.isDefined) {
+    val basicFormat = java.time.format.DateTimeFormatter.BASIC_ISO_DATE
+    val date = java.time.LocalDate.now().format(basicFormat)
+    s"$major-$date-SNAPSHOT"
+  } else {
+    s"$major-SNAPSHOT"
+  }
+}
+
 lazy val firrtlSettings = Seq(
   name := "firrtl",
-  version := "1.6-SNAPSHOT",
+  version := currentVersion("1.6"),
   addCompilerPlugin(scalafixSemanticdb),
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -13,9 +13,20 @@ lazy val isAtLeastScala213 = Def.setting {
   CrossVersion.partialVersion(scalaVersion.value).exists(_ >= (2, 13))
 }
 
+def currentVersion(major: String): String = {
+  val doDatedSnapshot = sys.props.get("datedsnapshot")
+  if (doDatedSnapshot.isDefined) {
+    val basicFormat = java.time.format.DateTimeFormatter.BASIC_ISO_DATE
+    val date = java.time.LocalDate.now().format(basicFormat)
+    s"$major-$date-SNAPSHOT"
```
jackkoenig, line 21: It would be a lot better if the date were derived from the commit hash so that we don't have old commits pushing over today's date. I also think the BuildInfo should include the commit hash so that users can check this info.

seldridge: Can you use `git describe --tag` or some variation of this to get `<latest-tag>-<new-commits>-g<git-hash>` instead of the date?

jackkoenig: If we include the commit hash, it will publish a SNAPSHOT for every commit whereas I was trying to just have 1 per day. To keep the number of snapshots down we could instead put it on a "nightly" trigger, but that makes downstream projects more annoying to do because you want to make sure they publish their nightly SNAPSHOT after and upstream dependencies do. Just sticking with a dated one makes that a little bit simpler.

seldridge: Every commit can be useful as you are known to be pinned to something obvious. Nightly is annoying because it's a question of when (have to check the cron job to see when nightly is defined). FWIW: here are the commit hash snapshots for `sifive/chisel-circt`: https://oss.sonatype.org/content/repositories/snapshots/com/sifive/chisel-circt_2.13/ This has been useful in the past.

PR #2594, opened 2023-01-19
Fix to support verilator tests in chisel3 that use MFC to generate unittests

chick:
Fix to support verilator tests in chisel3 that use MFC to generate code for `executeExpecting*` API

### Contributor Checklist

#### Type of Improvement

 - code refactoring

#### API Impact

Does not change the API structure. Slightly different behavior to support MFC in chisel3

- does not impact code generation

#### Desired Merge Strategy

#### Release Notes

Detects "assertion failed" messages in stderr in addition to existing checks on execution success or failure

- [ ] Did you review?

### DIFF
--- a/src/main/scala/firrtl/util/BackendCompilationUtilities.scala
+++ b/src/main/scala/firrtl/util/BackendCompilationUtilities.scala
@@ -188,7 +188,8 @@ object BackendCompilationUtilities extends LazyLogging {
           logger.info(line)
         },
         stderrLine => {
-          triggered = triggered || stderrLine.contains("Assertion failed") || (assertionMessageSupplied && stderrLine.contains(assertionMsg))
+          triggered = triggered || stderrLine
+            .contains("Assertion failed") || (assertionMessageSupplied && stderrLine.contains(assertionMsg))
           logger.warn(stderrLine)
         }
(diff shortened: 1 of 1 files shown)


src/main/scala/firrtl/util/BackendCompilationUtilities.scala
```diff
@@ -176,7 +187,11 @@ object BackendCompilationUtilities extends LazyLogging {
...
         },
-        logger.warn(_)
+        stderrLine => {
+          triggered = triggered || stderrLine
+            .contains("Assertion failed") || (assertionMessageSupplied && stderrLine.contains(assertionMsg))
+          logger.warn(stderrLine)
+        }
       )
     // Fail if a line contained an assertion or if we get a non-zero exit code
     //  or, we get a SIGABRT (assertion failure) and we didn't provide a specific assertion message
```
jackkoenig, line 197: Rather than hardcoding the assertion message, can you instead make this:
1. do the same that the stdout check does and allow the user to specify the assertion message
2. Make the return value below do the right thing even without providing Assertion failed? it's already looking for non-zero exit codes, are MFC tests that have assertion failures not returning non-zero?

PR #2595, opened 2023-01-25
[smem] fix read-under-write serialization

ekiwi:
Also adds some tests for the parser and
the serializer.

#### Type of Improvement

- bug fix 

#### API Impact

- smem now serializes correctly

- none

- [ ] Did you review?

### DIFF
--- a/src/test/scala/firrtlTests/ParserSpec.scala
+++ b/src/test/scala/firrtlTests/ParserSpec.scala
@@ -426,20 +426,7 @@ object SMemTestCircuit {
   def src(ruw: String): String =
     s"""circuit Example :
        |  module Example :
-       |    input clock : Clock
-       |    input reset : UInt<1>
-       |    input addr : UInt<3>
-       |    output data : UInt<8>
-       |
        |    smem mem : UInt<8> [8] $ruw@[main.scala 10:25]
-       |    wire _data_WIRE : UInt @[main.scala 11:20]
-       |    _data_WIRE is invalid @[main.scala 11:20]
-       |    when UInt<1>("h1") : @[main.scala 11:20]
-       |      _data_WIRE <= addr @[main.scala 11:20]
-       |      node _data_T = or(_data_WIRE, UInt<3>("h0")) @[main.scala 11:20]
-       |      node _data_T_1 = bits(_data_T, 2, 0) @[main.scala 11:20]
-       |      read mport data_MPORT = mem[_data_T_1], clock @[main.scala 11:20]
-       |    data <= data_MPORT @[main.scala 11:8]
        |""".stripMargin
 
   def findRuw(c: Circuit): ReadUnderWrite.Value = {


src/main/scala/firrtl/ir/Serializer.scala
```diff
@@ -302,7 +302,11 @@ object Serializer {
     // WIR
     case firrtl.CDefMemory(info, name, tpe, size, seq, readUnderWrite) =>
       if (seq) b ++= "smem " else b ++= "cmem "
-      b ++= name; b ++= " : "; s(tpe); b ++= " ["; b ++= size.toString(); b += ']'; s(info)
+      b ++= name; b ++= " : "; s(tpe); b ++= " ["; b ++= size.toString(); b += ']'
+      if (readUnderWrite != ReadUnderWrite.Undefined) {
+        b += ' '; b ++= readUnderWrite.toString
+      }
```
seldridge, line 308: It would be better to unconditionally serializing the read-under-write behavior than to have an implicit default. This only affects the MFC in terms of making this a breaking change.

src/test/scala/firrtlTests/ParserSpec.scala
```diff
@@ -405,6 +405,48 @@ class ParserSpec extends FirrtlFlatSpec {
...
+       |    smem mem : UInt<8> [8] $ruw@[main.scala 10:25]
+       |    wire _data_WIRE : UInt @[main.scala 11:20]
+       |    _data_WIRE is invalid @[main.scala 11:20]
+       |    when UInt<1>("h1") : @[main.scala 11:20]
+       |      _data_WIRE <= addr @[main.scala 11:20]
+       |      node _data_T = or(_data_WIRE, UInt<3>("h0")) @[main.scala 11:20]
+       |      node _data_T_1 = bits(_data_T, 2, 0) @[main.scala 11:20]
+       |      read mport data_MPORT = mem[_data_T_1], clock @[main.scala 11:20]
+       |    data <= data_MPORT @[main.scala 11:8]
+       |""".stripMargin
```
seldridge, line 443: Nit: This text is pretty long for what it is actually testing. This test will be more maintainable and understandable if it is shortened to:

```
circuit Example :
  module Example :
    smem mem : UInt<8>[8] $ruw @[main.scala 42:9000]
```

src/test/scala/firrtlTests/SerializerSpec.scala
```diff
@@ -121,4 +121,19 @@ class SerializerSpec extends AnyFlatSpec with Matchers {
     val serialized = Serializer.serialize(when, 1)
     serialized should be("  when cond :\n    skip\n")
   }
+
+  it should "serialize read-under-write behavior for smems correctly" in {
+    def parseSerializeParse(src: String): Circuit = Parser.parse(Parser.parse(src).serialize)
+    val undefined = parseSerializeParse(SMemTestCircuit.src(""))
+    assert(SMemTestCircuit.findRuw(undefined) == ReadUnderWrite.Undefined)
```
seldridge, line 128: Nit: using non-assert matchers may be more readable:

```scala
SMemTestCircuit.findRuw(undefined) should be (ReadUnderWrite.Undefined)
```

seldridge (commented):
Nice. 

One concern about the serialization format. I'd prefer to always serialize this.

src/main/scala/firrtl/ir/Serializer.scala
ekiwi: I can change it. The idea behind not serializing the default was that it would maintain backwards compatibility for most circuits (since RUW is rarely set).

src/test/scala/firrtlTests/SerializerSpec.scala
ekiwi: I disagree. I like to keep things simple and scalatest will still highlight the failure correctly when using plain asserts. The `should` or `must` matchers have some weird corner cases (I forgot what they were, but there was some trouble in the past that would have been avoided by sticking to the basics).

src/test/scala/firrtlTests/ParserSpec.scala
ekiwi: I will try out your simplification

src/main/scala/firrtl/ir/Serializer.scala
ekiwi: changed

src/test/scala/firrtlTests/ParserSpec.scala
ekiwi: works. changed

src/test/scala/firrtlTests/SerializerSpec.scala
ekiwi: Let me know if you insist on using the `should` and I will change it in order to get this change in.

seldridge: There's no problem with using them here and continuing to use them aligns with all the other tests in this file.

For corner cases, you may be thinking of string matching on `include` vs `contains`? The former looks for a substring match and the latter looks for an exact match.

seldridge (approved):
LGTM

src/test/scala/firrtlTests/SerializerSpec.scala
seldridge: Don't care. It's good to land as-is. 👍

PR #2633, opened 2023-04-25
remove protobuf support

ekiwi:
Remove protobuf support since it is a pain to support across systems and it does not see much use anyways.


ekiwi:
wrong repo

